PR #13, opened 2016-09-12
Adding a proposal to support Open Stack on top of SONiC

c-wichmann:
Looking at integration options for Open Stack and SONiC.

This initial proposal describes what we would like to achieve by the Open Stack integration and further updates will flush out the implementation details.


lihuay:
Close 6 year old design with no reviews.

PR #179, opened 2018-04-24
Introduction for SONiC system telemetry

jipanyang:
Signed-off-by: Jipan Yang <jipan.yang@alibaba-inc.com>

Introduction for SONiC telemetry via gRPC


jipanyang:
Corrupted by wrong forced push.

PR #187, opened 2018-05-10
SONiC warm reboot discussion

jipanyang:
The discussion for SONiC warm reboot support.


ipasha2018:
Clarifications/Comments:
1.	Warm reboot from the immediate previous version only.
2.	Kernel warm reboot cmd line option (similar to fast-reboot)
3.	Publish the SAI API to save the LibSai/LibASIC sate
4.	Document examples for the SAI API change use cases
a.	Attribute change
b.	Object change
i.	Remove the old object
ii.	Create a new object
iii.	Object dependency resolution

jipanyang:
@ipasha2018 Thanks for the comments! Will try to address them.

PR #202, opened 2018-06-01
Transceiver monitoring HLD

keboliu:
Transceiver monitoring HLD


doc/transceiver-monitor-hld.md
```diff
@@ -0,0 +1,140 @@
...
+    try:
+        for event in i.event_gen():
+            if event is not None:
+                (header, type_names, watch_path, filename) = event
+                _LOGGER.info("WD=(%d) MASK=(%d) COOKIE=(%d) LEN=(%d) MASK->NAMES=%s "
+                             "WATCH-PATH=[%s] FILENAME=[%s]",
+                             header.wd, header.mask, header.cookie, header.len, type_names,
+                             watch_path.decode('utf-8'), filename.decode('utf-8'))
+    finally:
+        i.remove_watch(b'/tmp')
```
andriymoroz: I understand this is just an example but looks like remove_watch does not match the add_watch
also /bsp/ is not available in pmon (it will run in pmon, right?) so it is better to use /sys/bus...

doc/transceiver-monitor-hld.md
```diff
@@ -0,0 +1,140 @@
...
+For the convenience of implementation and reduce the time consuming, need to enhance the `SfpUtilBase` class to provide functions to get `eeprom_if_dict` and `eeprom_dom_dict` separately, the intrested values are defined  in section 1.1.1 and 1.1.2, we can pick up these values from eeprom by calling the above new API with proper offset and number of bytes. 
+
+
+### 1.3 Transceiver plug in/out event ###
+
+Xcvrd need to be triggered by transceiver plug in/out event to refresh the transceiver info table.
+
+Transceiver plug in/out status can be derived from the content of sysfs file like `"/sys/bus/i2c/devices/2-0048/hwmon/hwmon7/qsfp10_status"`, if the content of the file is "good" represent SFP is present, conent change to "not_connected" means SFP plug out.
+
+To monitor the file change, can introduce python lib [inotify](https://pypi.org/project/inotify/), which can raise notification when target file change. Below is a sample for how to use inotify lib to monitor file change:
```
andriymoroz: here some details are missing. Who will generate events? where it will run? Is it part of Xcvrd or a separate daemon? If not part of Xcvrd - how event will be passed?

hui-ma: Sorry for the delay. I think I sent the comment back. I forget to hit the review button.

doc/transceiver-monitor-hld.md
keboliu: yes, we should mount /sys/bus to pmon, will revise the example.

doc/transceiver-monitor-hld.md
keboliu: I think each vendor have to implement their own API to fetch the watching path list and also the API to judge the status of the transceiver according to the file content. I'll define some API here. maybe the current CLI to show SFP presence is a good example here.

doc/transceiver-monitor-hld.md
```diff
@@ -0,0 +1,140 @@
...
+| 1.3.6.1.2.1.99.1.1.1.3.index | entPhySensorPrecision | Same as above | 4 |
+| 1.3.6.1.2.1.99.1.1.1.4.index | entPhySensorValue | Same as above | 7998 |
+| 1.3.6.1.2.1.47.1.1.1.1.2.index | entPhysicalDescr | Show interfaces alias | DOM RX Power Sensor for DOM RX Power Sensor for Ethernet29/1 |
+
+
+More detailed information about new table and new OIDs are described in [Sensor and Transceiver Info Monitoring Requirement](https://github.com/Azure/SONiC/blob/gh-pages/doc/OIDsforSensorandTransciver.MD#transceiver-requirements-entity-mib).
+
+### 2.2 New connection to STATE_DB ###
+
+To get the transceiver and dom sensor status, SNMP agent need to connect to STATE\_DB and fetch information from TRNASCEIVER_TABLE which will be updated by Xcvrd when this is status change.
```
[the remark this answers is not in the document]
keboliu: will add.

doc/transceiver-monitor-hld.md
hui-ma: Thanks! Sounds good to me.

doc/transceiver-monitor-hld.md
```diff
@@ -0,0 +1,151 @@
...
+
+### 2.1 MIB tables extension ###
+
+MIB table entPhysicalTable from [Entity MIB(RFC2737)](https://tools.ietf.org/html/rfc2737) need to be extended to support new OIDs.
+
+| OID | SNMP counter | Where to get the info in Sonic. | Example: |
+| --- | --- | --- | --- |
+| 1.3.6.1.2.1.47.1.1.1 | entPhysicalTable |   |   |
+| 1.3.6.1.2.1.47.1.1.1.1 | entPhysicalEntry |   |   |
+| 1.3.6.1.2.1.47.1.1.1.1.2. ifindex | entPhysicalDescr | Show interfaces alias | Xcvr for Ethernet29 |
```
hui-ma: The ifindex should be replaced with index, which is Ifindex * 1000, as changed in requirement

doc/transceiver-monitor-hld.md
```diff
@@ -0,0 +1,152 @@
...
+	voltage                 = FLOAT                                      ; voltage value
+	rx1power                = FLOAT                                      ; rx1 power in dbm
+	rx2power                = FLOAT                                      ; rx2 power in dbm
+	rx3power                = FLOAT                                      ; rx3 power in dbm
+	rx4power                = FLOAT                                      ; rx4 power in dbm
+	tx1bias                 = FLOAT                                      ; tx1 bias in mA
+	tx2bias                 = FLOAT                                      ; tx2 bias in mA
+	tx3bias                 = FLOAT                                      ; tx3 bias in mA
+	tx4bias                 = FLOAT                                      ; tx4 bias in mA
+
```
yongcanwang00: Is it possible to add TX power? Or TX bias is enough?

keboliu: @yongcanwang00 this is defined in the requirement doc, I would like to ask the requirement author @hui-ma to comment on this.

hui-ma: TX bias is from the output of "show interfaces transceiver eeprom" command, It is also the output of some other vendors' snmp output. I think they can be translated into TX power. I need to look for its equation. Do you prefer TX power in the output directly? 

    ChannelMonitorValues:
              RX1Power: -1.1936dBm
              RX2Power: -1.1793dBm
              RX3Power: -0.9388dBm
              RX4Power: -1.0729dBm
              TX1Bias: 4.0140mA
              TX2Bias: 4.0140mA
              TX3Bias: 4.0140mA
              TX4Bias: 4.0140mA
        ModuleMonitorValues :
                Temperature : 1.1111C
                Vcc : 0.0000Volts

yongcanwang00: Yes, TX power is useful.  It's helpful to have them both.

reinight:
If the transceiver is damaged, what will SfpUtil return when Xcrvd pull the info? Is it depends on the vendor's implementation and we don't care about it?

PR #205, opened 2018-06-04
[doc] Add sonic CLI design document.

wyc11112002:
Signed-off by: yongcan.wyc@alibaba-inc.com


liatgrozovik:
all documents should be in md format. any reason why you have it in doc format? in this way it is impossible to comment on specific line as we recently did with other design reviews lately.

liatgrozovik:
Your examples as part of '2.1.	How to reuse current CLI?' is for linux commands only. Can you please provide example also for current SONiC CLI commands which are not Linux based but were added as part of few features implementations. Some utilities added as part of sonic-utility git

liatgrozovik:
With this solution should we still use the sudo for configuration?

liatgrozovik:
According to the diagram provided in '2.	Major design' you have direct connection to BGP docker (Quagga or FRR), is it for show command only? if not, will you allow configuring it not via the config_db?

liatgrozovik:
Planning wise - will you include all available SONiC commands under this design implementation? will it be allowed to add more CLI commands in the current way? If so, i think we should not allow it as it will have 2 different ways for CLI commands and we should have only 1 way.

yongcanwang00:
Hi liatgrozovik,

Thank you for your comments.
Yes, I can add examples for current SONiC CLI commands. 
Only sudo user can enter to configure mode. 
You know configDB only support a few features of Quagga, we prefer to support all features on cli for debugging.  Any configuration changed by this CLI are not committed to save unless this is supported by configDB.
This CLI have include many available SONIC commands but not all. 
Currently, this CLI is just an option. I think different user prefers different cli style, and they can choose to provide which cli to their end user. We provide this CLI to end user as the only 1 in Alibaba. Many commands are just an wrapper on the available SONIC commands. But  available SONIC commands are not visible for end user. We are open if community wants to select one cli.

yongcanwang00:
Add md format document: SONIC_CLI.md

lihuay:
@yongcanwang00 @liatgrozovik : is this hld still valid or of interest?

PR #228, opened 2018-08-08
[xcvrd design] robust enhancement to xcvrd design

keboliu:
a robust enhancement design for xcvrd on mlnx platform

### DIFF
--- a/doc/transceiver-monitor-hld.md
+++ b/doc/transceiver-monitor-hld.md
@@ -108,9 +108,9 @@ mnlx-sfpd will populate error when:
 1. not able to get correct sfp change event from SDK
 2. mlnx-sfpd itself failed for some reason.
 
-mlnx-sfpd will periodically write a timestamp to STATE_DB to show the liveness. On above two failure cases mlnx-sfpd will stop update the timestamp and exit. Related info will be logged as error indication to user.  
+mlnx-sfpd will have a liveness indication mechanisim to let xcvrd know that it is working or not. mlnx-sfpd use STATE_DB to convey it's liveness status to the outside.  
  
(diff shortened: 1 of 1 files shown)


doc/transceiver-monitor-hld.md
```diff
@@ -95,6 +95,23 @@ Xcvrd will call this API to wait for the sfp plug in/out event, following exampl
                  
 It's possible that when received the plug in/out event, the transceiver eeprom is not ready for reading, so need to give another try if first reading failed. 
 
+#### 1.3.1 Transceiver plug in/out event implementation on mlnx platform ####
```
qiluo-msft: In above section, "default value 0 means never timeout" is not good. A blocking function with timeout zero normally means no wait at all.

doc/transceiver-monitor-hld.md
```diff
@@ -95,6 +95,23 @@ Xcvrd will call this API to wait for the sfp plug in/out event, following exampl
...
+When mlnx-sfpd get the event, it will populate it to STATE_DB. get_transceiver_change_event on mlnx platform will subscribe to STATE_DB and waiting for it.
+
+Since xcvrd does not talk to mlnx-sfpd directly, need to have some mechanism to notify xcvrd when mlnx-sfpd fail, so xcvrd can handle accordingly. The intermediate will still be STATE_DB.
+
+mnlx-sfpd will populate error when:
+
+1. not able to get correct sfp change event from SDK
+2. mlnx-sfpd itself failed for some reason.
+
+mlnx-sfpd will periodically write a timestamp to STATE_DB to show the liveness. On above two failure cases mlnx-sfpd will stop update the timestamp and exit. Related info will be logged as error indication to user.
```
qiluo-msft: Using timestamp is implementation details, not part of the design. I encourage you investigate another alternative implementation https://redis.io/commands/expire

keboliu: @qiluo-msft will investigate, thanks.

PR #231, opened 2018-08-17
Snmp configuation db schema proposal

MichelMoriniaux:
Proposal for a config DB SNMP configuration table schema

### DIFF
--- a/doc/snmp-schema-addition.md
+++ b/doc/snmp-schema-addition.md
@@ -42,22 +42,23 @@ In the end this is used to produce /etc/snmp/snmpd.conf
 }
 ```
 
-Where:  
-LOCATION_STRING:  String, defines the snmpLocation, default: ""  
-CONTACT_STRING:   String, defines the snmpContact, default: "" or the current hardcoded value for backwards compatibility.  
-COMMUNITY_STRING: String, defines the community string  
-ACL_STRING:       String, name of the ACL that controls access to SNMP daemon, default: "SNMP_ACL"  
-OID_STRING:       String. a dotted notation OID prefix  
-COMMAND_STRING:   String: command to be called when OID_STRING is requested
+Where:
+| Name | Type | meaning | default value |
+|------|------|---------|---------------|
+| LOCATION_STRING: | String | defines the snmpLocation | "" |
+| CONTACT_STRING: | String | defines the snmpContact | "" or the current hardcoded value for backwards compatibility. |
+| COMMUNITY_STRING: | String | defines the community string | |
+| ACL_STRING: | String | name of the ACL that controls access to SNMP daemon, default: "SNMP_ACL" | |
+| OID_STRING: | String | a dotted notation OID prefix | |
+| COMMAND_STRING: | String | command to be called when OID_STRING is requested | |
 
-Variables and dicts:  
-"v2c": we define a "v2c" tree to allow for future expansion for other versions of the SNMP protocol, this spec only defines for SNMP v2  
-       we could imagine the implementation of "v3" with the inclusion of users or references to central PAM methods  
-"type":  there are 2 possible types:  
-       "ro": read-only, the only implemented method at this time  
-       "rw": well you never know - here for completeness but unused in the code  
-"acl":   allows to override the default SNMP_ACL definition ( useful for management VRFs? )  
-"pass_trough": used for pass-trough definitions - currently the config only has one hard-coded pass-through defined for SysDescription
+Variables and dicts:
+| Name | description |
+|------|-------------|
+| "v2c": | we define a "v2c" tree to allow for future expansion for other versions of the SNMP protocol, this spec only defines for SNMP v2, we could imagine the implementation of "v3" with the inclusion of users or references to central PAM methods |
+| "type": | there are 2 possible types:   * "ro": read-only, the only implemented method at this time  *"rw": well you never know - here for completeness but unused in the code |
+| "acl": | allows to override the default SNMP_ACL definition ( useful for management VRFs? ) |
+| "pass_trough": | used for pass-trough definitions - currently the config only has one hard-coded pass-through defined for SysDescription |
 
 ### files needing modification for implementation ###
 
--- a/doc/snmp-schema-addition.md
+++ b/doc/snmp-schema-addition.md
@@ -1,0 +1,81 @@
+# SONiC SNMP TABLE schema proposal #
+
+## Scope of the change ##
+
+Currently SNMP configuration is managed from a mix of yaml files and DB based ACLs
+we propose to integrate the SMP configs into an SNMP table in the DB
+The following document proposes the database Schema in Json format and the list of required changes to the code
+
+## Current configurtions for SNMP ##
+sonic-buidimage and sonic-snmpagent depend on the following files:
+1. /etc/sonic/snmp.yaml
+    defines 2 variables:
+        snmp_rocommunity: 
+        snmp_location: 
+   this file is consumed in the snmp docker container by /usr/share/sonic/templates/snmpd.conf.j2 and /usr/bin/snmpd-config-updater
+2. /etc/sonic/config_db.json or /etc/sonic/minigraph.xml for ACL definitions 'SNMP_ACL'
+   this file is consumed by /usr/bin/snmpd-config-updater, /usr/share/sonic/templates/alias_map.j2, /usr/share/sonic/templates/sysDescription.j2
+
+in the end this is used to produce /etc/snmp/snmpd.conf
+
+## Limits and incentives to change ##
+- This goes against the principle of having the config DB be the central repository of configuration data
+- Imposes the provisioning of multiple static files
+- the snmpContact is hardcoded in the /usr/share/sonic/templates/snmpd.conf.j2 file
+- pass-trhoughs are hard-coded
+
+## Proposal: integrate the SNMP configuration into config DB ##
+### Proposed Schema ###
+
+"SNMP": {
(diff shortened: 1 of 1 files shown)


doc/snmp-schema-addition.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Proposal: integrate the SNMP configuration into config DB ##
+### Proposed Schema ###
+
+```
+"SNMP": {
+    "location": LOCATION_STRING,
+    "contact": CONTACT_STRING,
+    "v2c": {
+        COMMUNITY_STRING: {
```
qiluo-msft: >COMMUNITY_STRING [](start = 8, length = 16)

Is it true that COMMUNITY_STRING is not related to v2c?

doc/snmp-schema-addition.md
```diff
@@ -0,0 +1,83 @@
...
+            "type": "rw"|"ro",
+            "acl": ACL_STRING
```
qiluo-msft: >acl [](start = 13, length = 3)

Is it true that acl is not related to COMMUNITY_STRING?

doc/snmp-schema-addition.md
```diff
@@ -0,0 +1,83 @@
...
+- LOCATION_STRING:  String, defines the snmpLocation, default: "".  
+- CONTACT_STRING:   String, defines the snmpContact, default: "" or the current hardcoded value for backwards compatibility.  
+- COMMUNITY_STRING: String, defines the community string.  
+- ACL_STRING:       String, name of the ACL that controls access to SNMP daemon, default: "SNMP_ACL".  
+- OID_STRING:       String. a dotted notation OID prefix.  
+- COMMAND_STRING:   String: command to be called when OID_STRING is requested.
+
+New keys:
+- "v2c": we define a "v2c" tree to allow for future expansion for other versions of the SNMP protocol, this spec only defines for SNMP v2  
+         we could imagine the implementation of "v3" with the inclusion of users or references to central PAM methods.
```
qiluo-msft: >imagine [](start = 18, length = 7)

Is there any reference for the imagination?

doc/snmp-schema-addition.md
MichelMoriniaux: current behavior of /usr/bin/snmpd-config-updater and only this configuration method is to read the ACL table to link a rocommunity to a host
for eg.: this writes in the snmpd.conf file:
rocommunity mysnmpcomm 172.16.10.0/24
rocommunity mysnmpcomm 122.168.5.12

doc/snmp-schema-addition.md
MichelMoriniaux: I don't understand the comment
it is my understanding that in v1 and v2c the community is a 'password' and that in v3 this 'password' is replaced by user authentication+encryption key, so in this contect COMMUNITY_STRING should be part of v2c configuration only

doc/snmp-schema-addition.md
qiluo-msft: >ACL_STRING [](start = 19, length = 10)

Currently ConfigDB already contains "ACL_TABLE|SNMP_MGMT_ONLY", do you want to replace it?

doc/snmp-schema-addition.md
MichelMoriniaux: we could have for eg.:
"v3": {
    "key": ENCRYPTION_KEY,
    "pam": PAM_METHOD,
    "users": {
         USERNAME: PASSWORD,
         USERNAME: PASSWORD
    }
}

doc/snmp-schema-addition.md
qiluo-msft: I found "SNMPv3 also uses community strings, but allows for secure authentication and communication between SNMP manager and agent"
[https://en.wikipedia.org/wiki/Simple_Network_Management_Protocol](https://en.wikipedia.org/wiki/Simple_Network_Management_Protocol).
I am ok with the version string here.

---
In reply to: [211706220](https://github.com/Azure/SONiC/pull/231#discussion_r211706220) [](ancestors = 211706220)

doc/snmp-schema-addition.md
MichelMoriniaux: good point:
how about: if the "acl" key is present we override and if not present we used the default: "ACL_TABLE|SNMP_MGMT_ONLY"

MichelMoriniaux: makes sense, significantly simplifies config generation

qiluo-msft: Is it possible to refer to it, and keep the table contents in original place? This way your changes are minimized.

jipanyang:
It is good to move the configuration to configDB.  Eventually where are you going to store the schema data?  

A large part of SONiC schema could be found in swss-schema.md file though some of them are a little outdated.  My worry is that we may lose track of the schema soon if there is no well known central place to store and find them.  

If there is mechanism to automatically generate code from the schema, it would be great, that will ensure the schema is always up to date and match the code implementation, but I guess this will a lot of effort.

doc/snmp-schema-addition.md
jleveque: Qi is correct. the current SNMP ACL implementation disregards community strings, and works on a service-based level. We whitelist IPs for each service, so all whitelisted IPs for SNMP will apply regardless of the community string.

jleveque: As I mentioned above, the /usr/bin/snmpd-config-updater application currently only applies to Arista platforms, and I believe it will be removed in the coming months. Once this is done, all service ACLs (control plane ACLs) will be managed in the base image using the caclmgrd daemon. Therefore, there will be no need for ACL information inside the SNMP container. I believe you can simplify your design further and eliminate ACL information altogether.

All ACL information will continue to reside in the ACL_TABLE of ConfigDB.

MichelMoriniaux:
@jipanyang added requirement to the design proposal

doc/snmp-schema-addition.md
```diff
@@ -0,0 +1,87 @@
...
+        OID_STRING: COMMAND_STRING,
+    }
+}
+```
+
+Where:
+- LOCATION_STRING:  String, defines the snmpLocation, default: "".  
+- CONTACT_STRING:   String, defines the snmpContact, default: "" or the current hardcoded value for backwards compatibility.  
+- COMMUNITY_STRING: String, defines the community string.  
+- ACL_STRING:       String, name of the ACL that controls access to SNMP daemon, default: "SNMP_ACL".
```
jleveque: Since you have removed ACLs from your proposed schema, this line should also be removed.

doc/snmp-schema-addition.md
```diff
@@ -0,0 +1,87 @@
...
+- OID_STRING:       String. a dotted notation OID prefix.  
+- COMMAND_STRING:   String: command to be called when OID_STRING is requested.
+
+New keys:
+- "v2c": we define a "v2c" tree to allow for future expansion for other versions of the SNMP protocol, this spec only defines for SNMP v2  
+         we could imagine the implementation of "v3" with the inclusion of users or references to central PAM methods.  
+- "type":  Optional, if ommited defaults to 'ro', there are 2 possible values:  
+           "ro": read-only, the only implemented method at this time.  
+           "rw": well you never know - here for completeness but unused in the code.  
+- "acl":   allows to override the default SNMP_ACL definition ( useful for management VRFs? ).
```
jleveque: Since you have removed ACLs from your proposed schema, this line should also be removed.

doc/snmp-schema-addition.md
```diff
@@ -0,0 +1,87 @@
...
+
+The changes we propose are only additive to remain compatible with the current install base and the current way of doing things.
+
+In repo *sonic-buidlimage*:
+
+*dockers/docker-snmp-v2/snmpd.conf.j2*:  
+    verify the existence of the SNMP table in the datatbase and fork behavior if present, if not continue using old method.
+
+*dockers/docker-snmp-v2/snmpd-config-updater*:  
+    this file will be deprecated soon by cacl so no updates will be done
```
jleveque: For clarity:

"this file will be deprecated soon by **caclmgrd** so no updates will be done"

lguohan:
@MichelMoriniaux can you add backward compatibility into the design, so that we can make sure the existing way still works.

MichelMoriniaux:
definitely, I had this in the current commit but maybe it's not clear
enough:
*dockers/docker-snmp-v2/snmpd.conf.j2*:
verify the existence of the SNMP table in the datatbase and fork behavior
if present, if not continue using old method.

tientienle:
I don't see any value that use to store the snmp server ip address. Is there any reason for this or do you use it in anywhere else? 
For example, this is how I store it in config_db.json (or cfg data)
    "SNMPINFORM": {
        "192.168.0.1": {
            "community-string": "public"
        }
    },
    "SNMPTRAP2": {
        "192.168.0.1": {
            "community-string": "public"
        }
    },
    "SNMPTRAP3": {
        "1192.168.0.1": {
            "username": "spine1",
            "password": "password",
            "auth": "MD5", 
            "priv": "DES",
            "key":	"testingkey"
        }
    },
    "SNMPTRAP1": {
        "192.168.0.1": {
            "community-string": "public"
        }
    },
However, I still miss a few more values.

RichardWu-Hebut:
I tried to write the SNMP configuration according to your  proposal, and found that those SNMP configurations can not be automatically loaded from JSON file into DB. So, we need to handle this SNMP configuration specially?

MichelMoriniaux:
the proposal has not been implemented yet so there is no translation from
the json or redis DB to the sonic-snmpagent container configuration files

PR #242, opened 2018-09-04
VRF-HLD document

shine4chen:
VRF high level design for SONIC


doc/sonic-vrf-hld-v0.2.md
```diff
@@ -0,0 +1,412 @@
...
+key = ROUTE_TABLE:VRF_NAME:prefix ;
+
+nexthop = \*prefix, ;IP addresses separated “,” (empty indicates no gateway)
+
+intf = ifindex? PORT_TABLE.key ; zero or more separated by “,” (zero indicates
+no interface)
+
+blackhole = BIT ; Set to 1 if this route is a blackhole (or null0)
+
+Since global vrf name is null, global vrf key will becomes ROUTE_TABLE:prefix.
```
nikos-github: I suggest you use ROUTE_TABLE:DEFAULT:prefix or ROUTE_TABLE:GLOBAL:prefix

doc/sonic-vrf-hld-v0.2.md
```diff
@@ -0,0 +1,412 @@
...
+>   When query nexthop，keys now are (tableid, ipaddress)，tableid of nexthop
+>   can be acquired using nexthop interface.
+
+neighorch changes
+-----------------
+
+Adding the following logic:
+
+-   the Key of NextHop now is changed from only ipaddress to a pair of
+    (ipaddress, interface_name)
```
nikos-github: Should be (ipaddress, tableid) and not (ipaddress, interface_name). Expect to be doing a lot of conversions and lookups if you leave as (ipaddress, interface_name).

doc/sonic-vrf-hld-v0.2.md
```diff
@@ -0,0 +1,412 @@
...
+>   VRFOrch has three member routerCnt , neighCnt and rifCnt which record VRF
+>   related route number , neigh number and rif number.VRFOrch is added to
+>   RouteOrch , NeighOrch and intfsOrch member to update routeCnt , neighCnt and
+>   rifCnt.
+
+>   When  VRFOrch receives vrf-delete event VRF object won’t  be deleted until
+>   routerCnt  ,neighCnt and rifCnt  is decreased to zero.
+
+>   When device binds to specified VRF, the ip address of slave device will be
+>   removed and kernel will delete all neigh associated slave device.
```
nikos-github: I believe on enslavement, connected and local routes are automatically moved to the table associated with the VRF device. Is that not the case with the latest kernel? Can you explain how you plan to achieve the above removal/behavior you mention?

lguohan:
@jipanyang , I have re-formatted the doc, please comment.

doc/sonic-vrf-hld-v0.2.md
lguohan: can we point a route in vrf red to a nexthop (neighbor) in vrf blue? @prsunny , do we need to change the nexthop format in app db?

nikos-github: @lguohan Yes you can have routes in vrf red with nexthops in vrf blue.

prsunny: @nikos-github , yes it is possible but the format in APP_DB is same for route since every route has a nexthop IP with the interface name. Currently orchagent doesn't look for the interface which requires a fix.

shine4chen: @nikos-github When set neigh to SAI layer (ipaddress, interface_name) would be more comfortable. And relatedAPP_DB/ASIC_DB needn't be changed.

lguohan:
why close this PR?

doc/sonic-vrf-hld-v0.2.md
```diff
@@ -0,0 +1,343 @@
...
+The following is high level diagram of modules with VRF support.
+
+## The schema changes
+
+1. Adding VRF related configuration in config_db.json
+
+```
+"VRF": {
+    "VRF-blue": {
+        "fall_through":"1" //enable global fib lookup while vrf fib lookup missed
```
nikos-github: Please change fall_through to fallback. The term fall_through tends to be used for routemaps but for vrfs the industry established term is fallback.

There should be no restriction that the fallback be the global/main table even if that may be the only one we plan to support (and I don't see why we plan to support only global/main as fallback).

Your document needs to specify how the fallback is specified. I assume the table id will be used since linux doesn't have a name for the global/main table.

doc/sonic-vrf-hld-v0.2.md
```diff
@@ -0,0 +1,343 @@
...
+```
+
+app-intf-prefix-table is defined as the following:
+
+```
+;defines logical network interfaces with IP-prefix, an attachment to a PORT and
+list of 0 or more ip prefixes;
+
+;Status: stable
+key = INTF_TABLE:ifname:IPprefix ; an instance of this key will be repeated for
```
nikos-github: What is the advantage of having multiple keys?

Please provide the schema for this.

doc/sonic-vrf-hld-v0.2.md
shine4chen: @nikos-github Fallback term is better.I will revise it. The fallback global feature is very useful for specify VRF user access internet through global/main route which is defined by RFC4364. Some enterprise user still use this to access internet on vpn environment. Different vrf route leak is beyond this document scope.

doc/sonic-vrf-hld-v0.2.md
shine4chen: @nikos-github There is two reason why break up app-intf-table into app-intf-table and app-intf-prefix-table.
1. Multiple ip addresses can be configured on one interface.So I put interface common attribute into app-intf-table and keep ip-prefix specific attribute on app-intf-prefix-table.
2. Interface can be put to specific VRF before ip address is configured on it.

shine4chen: jason file looks like below:
    "Ethernet1":{
        "mtu":"1500”,
        "vrf":"vrf-red"
    },
    "Ethernet0|11.11.11.1/24": {},
    "Ethernet1|12.12.12.1/24": {},
    "Ethernet2|13.13.13.1/24": {},
    "Ethernet3|14.14.14.1/24": {},

shine4chen:
@lguohan @prsunny @jipanyang please help to approve this document if no more question.

shine4chen:
Nephos plan to submit this feature include warm-reboot support on 201906 branch. @lguohan @prsunny @jipanyang

prsunny:
Closing in favor of [this ](https://github.com/Azure/SONiC/commit/10d5d193bc3c12da7491732e107cb8a4970a4165)commit

PR #252, opened 2018-09-19
Create system-warmboot.md

lguohan:


nikos-github:
Will the asic reset during this warmboot?

lguohan:
@nikos-github , no

stepanblyschak:
Who will set WARM_RESTART_TABLE:system ? Is this a responsibility of warm reboot CLI script or user should do that via config warm_restart enable before executing warm reboot CLI?

doc/warm-reboot/system-warmboot.md
```diff
@@ -0,0 +1,46 @@
+# System-wide Warmboot
+
+# going down path
+
+- stop bgp docker 
+  - enable bgp graceful restart
+  - same as fast-reboot
+- stop teamd docker
+  - just kill
+  - same as fast-reboot
```
jipanyang: Is it exactly the same as fast reboot, or some extra state processing is needed for avoid process being killed in the middle of something?

doc/warm-reboot/system-warmboot.md
```diff
@@ -0,0 +1,46 @@
...
+  - implemented in database system service
+- start syncd docker
+  - implemented inside syncd docker
+  - recover SAI state from ```/host/warm-reboot/sai``` 
+  - the host interface will be also recovered.
+- start swss docker
+  - orchagent will wait till syncd has been started to do init view.
+  - will read from APP DB and do comparsion logic.
+- start teamd docker
+  - at the same time as swss docker. swss will not read teamd app db until it finishes the comparison logic.
```
jipanyang: Some reconciliation could be necessary if teamsyncd process is killed forcefully.  Ex. del request could be lost in teamsyncd before the data is put into appDB, due to process kill.  After restart, the state in teamd may become inconsistent with data in appDB.

pavel-shirshov:
```
stop teamd docker 
   just kill
   same as fast-reboot
```
We should  stop teamd gracefully to allow teamd to send last valid update to be sure we'll have 90 seconds reboot time available.

qiluo-msft:
Suggest refine the kernel argument as ```SONIC_BOOT_TYPE=[fast|warm]```. Benefits:
1. not possible to set both fast and warm
2. less conflict with vanilla linux kernel arguments

jipanyang:
Before dumping appDB (and other DB),  would it be better to call "systemctl stop swss" ?  There are processes other than orchagent,  some of them may continue writing appDB.

qiluo-msft:
Thanks @jipanyang for the suggestion. I updated the doc.

doc/warm-reboot/system-warmboot.md
```diff
@@ -0,0 +1,67 @@
...
+2. less conflict with vanilla linux kernel arguments
+3. all existing checker for ```*fast-reboot*``` still work
+
+Later if we improve the consistency ```SONIC_BOOT_TYPE=[fast|warm|cold]```, this way the production image upgrading process will be smooth (no disruptive change)
+
+## SAI expectations for warm shutdown
+- Application (e.g. SONiC) sets switch attribute SAI_SWITCH_ATTR_RESTART_WARM to true before calling remove_switch().
+  - Note that this attribute doesn't have to be set at switch_create() time. This is a dynamic decision, setting before calling remove_switch is sufficient.
+- Application sets profile attribute SAI_KEY_WARM_BOOT_WRITE_FILE to a valid path/filename where the SAI data will be saved during upcoming warm shutdown.
+  - Depending on the SAI implementation, this value might have been read by SAI at switch_create() time only. It is recommended to set this value before calling crete_switch().
```
pavel-shirshov: crete->create

doc/warm-reboot/system-warmboot.md
```diff
@@ -0,0 +1,67 @@
...
+In fast-reboot or warm-reboot, we will use kernel argument to indicate the whole system reboot type for the next boot up. The argument format is ```SONIC_BOOT_TYPE=[fast-reboot|warm|cold]```. Benefits:
+1. not possible to set both fast and warm
+2. less conflict with vanilla linux kernel arguments
+3. all existing checker for ```*fast-reboot*``` still work
+
+Later if we improve the consistency ```SONIC_BOOT_TYPE=[fast|warm|cold]```, this way the production image upgrading process will be smooth (no disruptive change)
+
+## SAI expectations for warm shutdown
+- Application (e.g. SONiC) sets switch attribute SAI_SWITCH_ATTR_RESTART_WARM to true before calling remove_switch().
+  - Note that this attribute doesn't have to be set at switch_create() time. This is a dynamic decision, setting before calling remove_switch is sufficient.
```
pavel-shirshov: `remove_switch` Please add () for regularity

doc/warm-reboot/system-warmboot.md
```diff
@@ -0,0 +1,67 @@
...
+- Application sets profile attribute SAI_KEY_WARM_BOOT_WRITE_FILE to a valid path/filename where the SAI data will be saved during upcoming warm shutdown.
+  - Depending on the SAI implementation, this value might have been read by SAI at switch_create() time only. It is recommended to set this value before calling crete_switch().
+
+# going up path
+
+- Use kernel argument ```warmboot``` to determine in warm starting mode
```
pavel-shirshov: warmboot -> warm ?

PR #274, opened 2018-10-26
[Design doc] View switching in ProducerStateTable

taoyl-ms:


qiluo-msft:
@jipanyang The motivation is from previous discussion https://github.com/Azure/sonic-swss/pull/599#issuecomment-416106825.
I believe the answer to your question is well written in "Problem Description" section in this PR.

jipanyang:
Could you be more specific?  In what scenarios it is needed,   neighsyncd, fpmsyncd, or any other application?

qiluo-msft:
It is a generic extension on ProducerStateTable, and could be used in neighsyncd, fpmsyncd, or any other application. This is an alternative solution to existing/ongoing implementation on warm starting feature, not to intend to fix anything wrong.

doc/warm-reboot/view_switch.md
```diff
@@ -0,0 +1,179 @@
+# View Switching in ProducerStateTable
+
+### Problem Description
+To support warm reboot, 
+ProducerStateTable needs to support a feature that producer can create a temporary view for a table, 
+into which write operation will not not be synced to consumers immediately.
```
zhenggen-xu: remove one "not"

jipanyang:
I don't see it being able to replace existing neighsyncd/fpmsyncd reconciliation funcitons.  It is missing stale data removal function, and the field processing capability required by fpmsyncd.

Again, I don't see the use case for this proposal for now.  Maybe you have something in mind.

doc/warm-reboot/view_switch.md
```diff
@@ -0,0 +1,179 @@
...
+However, the actual operations get passed to consumer should be optimized as best effort, 
+so if an object is present in both old view and new view, 
+it should not be `del()` and `set()` again.
+
+In order to archieve that, we are going to add two additional APIs: `create_temp_view()` and `apply_temp_view()`,
+as well as modifying existing `del()` and `set()` API.
+
+This design is based on the assumption that there is only one producer writing to a specific table. 
+If there are multiple producer writing to a same table, the changes that 'secondary' producer write to the table
+while 'main' producer was doing view switching might be lost.
```
zhenggen-xu: We probably only have one producer for a table for now. But is there any mechanism to enforce this constrain? or make some of the steps atomic to database?

doc/warm-reboot/view_switch.md
```diff
@@ -0,0 +1,179 @@
...
+| KeySet        | `ROUTE_TABLE_KEY_SET`      |
+| DelKeySet     | `ROUTE_TABLE_DEL_SET`      |
+| Channel       | `ROUTE_TABLE_CHANNEL`      |
+
+We are not going to create additional Redis object to support the temporary view. 
+Instead, an in-memory table `TableDump m_tempViewState` 
+that shares a similar structure with TableHash and StateHash will be created to maintain the target state of the temporary view.
+Additionally, a flag `bool m_tempViewActive` will indicates whether a temporary view is current being worked on or not.
+
+### Operation Pseudo Code
```
zhenggen-xu: Can we add a working flow of the API caller?  Producer calling flow and Consumer calling flow?

taoyl-ms:
If I understand what you mean by ""stale data removal" correctly, I think this design does support this functionality. When new view is applied all the data that exists only in old view will be deleted. 

Regarding "field processing capability" I'm not quite sure what you are referring to, could you give some more detailed description?

jipanyang:
https://github.com/Azure/sonic-swss/pull/602/files#diff-931e09d3d640a3f50a57b139ea40ce4cR305

I guess you guys are thinking of the view comparison logic in syncd and want to mimic that in orchagent.  Honestly,  I think that is a bad example, at least based on its current implementation with all the limitations and complexities.

doc/warm-reboot/view_switch.md
qiluo-msft:
```suggestion
while 'main' producer was doing view switching might be lost during `apply_temp_view()`.
```

doc/warm-reboot/view_switch.md
```diff
@@ -0,0 +1,179 @@
...
+}
+```
+
+#### `apply_temp_view()`
+
+To apply temporary view, there are several steps:
+1. Drop all pending operations by clearing KeySet, DelKeySet, and StateHash
+2. Dump content of current view
+3. Compare current view and target view, generate a minimal set of `set()` and `del()` operations
+4. Write to KeySet, DelKeySet, and StateHash accordingly, publish the change to the channel, and unset `m_tempViewActive` flag.
```
lguohan: what about missing field? can we just delete the field instead of delete the key and and the object?

doc/warm-reboot/view_switch.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+#### `pop()` in ConsumerStateTable
+As `apply_temp_view()` is translating view switching operation into a set of standard `set()` and `del()` operations,
+no change is needed for `pop()` and any other functions in ConsumerStateTable.
+
+
+### Q&A
+
+#### How about storing temporary view in DB instead?
```
lguohan: locking mechanism? should return fail when the second producer try to do view switching when there is already one going on.

doc/warm-reboot/view_switch.md
taoyl-ms: It's not supported even in the normal 'SET'/'DEL' operations currently. What's the scenario that new object will have a missing field, while we cannot just delete the object and create a new one?

PR #310, opened 2018-12-06
Rif counters design

mykolaf:
Open for discussion

Add new CLI commands to show L3 interfaces counters. L3 interface counters refers to router port as well as vlan interface.
The information should be: RX: OK, ERR in packets and bytes. TX: OK, ERR in packets and bytes

### DIFF
--- a/doc/RIF_counters.md
+++ b/doc/RIF_counters.md
@@ -226,12 +226,3 @@ Example virtual paths
 | COUNTERS_DB	| "RIF_COUNTERS/``<interface name>``/``<counter name>``" |	One counter on one interface |
 
 ### 4 Open questions
-
-#### 4.1 SNMP support?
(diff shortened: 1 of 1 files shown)


doc/RIF_counters.md
```diff
@@ -0,0 +1,261 @@
...
+Options:
+
+  -p, --period TEXT
+
+  --verbose          Enable verbose output
+
+  -?, -h, --help     Show this message and exit.
+```
+
+Note: the 'show interfaces counters' has '-c, --clear' option. From the CLI point of view, this should be available from 'sonic-clear' CLI. (There already is 'sonic-clear interfaces counters')
```
prsunny: As we discussed in meeting, this would need to be retained as it is being used in production scripts

PR #316, opened 2018-12-21
Vlan status enhancement

jihaix:
* New CLI command to set vlan's admin-status
  - config vlan adsmin-status <vid> <up|down>
* Make Vlanxx's oper_status be relavent to its admin_status and member ports' oper_status.

Signed-off-by: xiongjihai@baidu.com


jipanyang:
I think the cause of mismatch between vlan member oper status and vlan status is that only one Bridge is being used by all vlans. 

As long as there is one slave netdev(port or lag) of the Bridge is up operationally, the oper status of Bridge will be up, then Vlan is up  (oper status).

Probably we should create one dedicated bridge for each Vlan, then the oper status of VLAN will  automatically reflect the oper status of bridge which is controlled by vlan member status with current Linux behavior?

PR #328, opened 2019-01-28
Media based port settings in SONIC

dgsudharsan:
This pull request contains the high level document explaining the design of media based port settings in Sonic


doc/media-settings/Media-based-Port-settings.md
```diff
@@ -0,0 +1,313 @@
...
+
+              To solve the above problem, few SAI attributes required for serdes programming are listed in section below. These SAI attributes will be programmed during media insert event.
+
+Media settings file:
+====================
+
+              Each vendor need to define the media settings file under device/<vendor-name>/ <ONIE_PLATFORM_STRING>/ <HARDWARE_SKU>/media_settings.json. This file contains the list of optics and DAC media-based settings for each port in the device. Each media setting for a media type comprises of key-value pairs per lane that is expected to be programmed when an optics or DAC is used. Examples of key value pairs include pre-emphasis, idriver and ipredriver.
```
keboliu: can we have a way for some vendors to bypass this mechanism?  They may have other ways to tune the SFP.

doc/media-settings/Media-based-Port-settings.md
```diff
@@ -0,0 +1,313 @@
...
+}
+
+![](https://github.com/dgsudharsan/SONiC/blob/hld_media/doc/media-settings/event_flow.png)
+
+Flow:
+=====
+
+              When a media is detected in xcvrd daemon, it constructs the media key identifier string as discussed above and searches the media_settings.json file. On finding a matching media key entry the vendor key is constructed and if found the key value pairs are fetched. The xcvrd daemon notifies these key value pairs along with the alias port to portsorch. The portsorch task will convert the alias to SAI object Id and notifies the syncd by doing port attribute set of corresponding settings. Syncd invokes the SAI port set those attribute with specified values. The SAI implementation then programs the attribute into the hardware.
+
+              The notification of hardware profile from xcvrd to portsorch task will be done during initialization and during media detect event. This is not required during media removal event. Since the media settings are required only for the proper functioning of the optics or DAC, the handling can be restricted to media insert event alone and no action needs to be taken during media removal.
```
keboliu: during switch initialization, xcvrd will wait for all the ports configured then startup itself, so maybe need to find another way for the event detection in this phase.

doc/media-settings/Media-based-Port-settings.md
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+}
+
+![](https://github.com/dgsudharsan/SONiC/blob/hld_media/doc/media-settings/event_flow.png)
+
+Flow:
+=====
+
+              When a media is detected in xcvrd daemon, it constructs the media key identifier string as discussed above and searches the media_settings.json file. On finding a matching media key entry the vendor key is constructed and if found the key value pairs are fetched. The xcvrd daemon notifies these key value pairs along with the alias port to portsorch. The portsorch task will convert the alias to SAI object Id and notifies the syncd by doing port attribute set of corresponding settings. Syncd invokes the SAI port set those attribute with specified values. The SAI implementation then programs the attribute into the hardware.
```
keboliu: would you please elaborate on the interaction between the xcvrd and portsorch? like to way to send the notification, the notification message outline?

doc/media-settings/Media-based-Port-settings.md
```diff
@@ -0,0 +1,313 @@
...
+
+At the third level, the Vendor key is derived by concatenating vendor name and vendor part number (e.g. DELL-00-11-22).  If there is no exact match for vendor name -- vendor part number key, then the vendor key is reduced to vendor name alone (e.g. DELL) and the media type entry is looked up for such a key. If there is no match, then the default value listed under the media type is chosen. Below is an example for json file for a specific port. For the port Ethernet20 two specific media types (40GBASE-SR4 and 40GBASE-CR4-3M) and a 'Default' media type is defined. For 40GBASE-SR4 a DELL vendor name- vendor PN specific setting is listed along with a 'Default' value which is chosen for all other 40GBASE-SR4 optics apart from the listed key (DELL-00-11-22). For 40GBASE-CR4-3M media key a DELL vendor name alone key word is listed along with a 'Default' key word. This would imply that any 40GBASE-CR4-3M DAC manufactured by DELL will pick the values listed in the vendor name key whereas everything else will pick the default values.
+
+{
+
+    "PORT_OPTICS_SETTINGS": {
+
+        "Ethernet20": {
```
ben-gale: "Ethernet20": As discussed on the call, I think it would be good to allow port lists (e.g. "Ethernet1, Ethernet3, Ethernet5 etc") and port ranges (e.g. "Ethernet1-Ethernet48") here. A list can also include range entries. I think this may make the file much shorter and more usable/maintainable.

doc/media-settings/Media-based-Port-settings.md
```diff
@@ -0,0 +1,313 @@
...
+
+            "Default": {
+
+                "Default": {
+
+                    "preemphasis": [
+
+                         "0x1201",
```
ben-gale: Can you explain what these values represent please? They look like register contents, in which case, are these values similar/common in semantics and type across transceivers?

doc/media-settings/Media-based-Port-settings.md
```diff
@@ -0,0 +1,313 @@
...
+
+                     ],
+
+                     "ipredriver": [
+
+                        "0x1",
+
+                        "0x1"
+
+                     ]
```
ben-gale: Did you think about adding a FEC setting here (for copper transceivers)?

doc/media-settings/Media-based-Port-settings.md
ben-gale: As also discussed on the call, the file must also support Port Breakout scenarios - this may change the port references in the file. Please describe how will this be handled?

doc/media-settings/Media-based-Port-settings.md
dgsudharsan: Hi @keboliu . Yes. If the vendor does not define media_setting.json file this mechanism would not kick-in.

doc/media-settings/Media-based-Port-settings.md
dgsudharsan: We plan to have a one-shot task during bootup which will notify porsorch, the media settings for all ports based on media detected. Once xcvrd starts, it will start notifying the media settings during transceiver change events.

doc/media-settings/Media-based-Port-settings.md
dgsudharsan: Hi @ben-gale . Agreed. Will add this to the document

dgsudharsan: The document will be updated on breakout scenario.

doc/media-settings/Media-based-Port-settings.md
dgsudharsan: These are the best abstractions of relevant pre-emp, idriver and pre-driver registers that are used by vendors. These will be common in semantics across transceivers. ASIC vendors have APIs to program these. So, it's upto the libsaiASIC-VENDOR implementation to implement these to map to their SDKs

doc/media-settings/Media-based-Port-settings.md
dgsudharsan: The data will be sent based on notifications. xcvrd will have a NotificationProducer which will send notifications on transceiver detection. The message will contain the following key value pairs 1) Portname = 'logical Port name' on which transceiver detected.
2) Each key value pairs defined in the media_settings file corresponding to the media_type, vendor and port.
The portsorch on receiving this message will convert the logical port name to sai port id and map each key value pair to corresponding attribute?
An example of notification as seen from redis consumer
1) "message"
2) "MEDIANOTIFICATION"
3) "[\"media_change\",\"\",\"port\",\"Ethernet12\",\"ipredriver\",\"0x1\",\"preemphasis\",\"0x5678\",\"idriver\",\"0x1\"]"
1) "message"
2) "MEDIANOTIFICATION"
3) "[\"media_change\",\"\",\"port\",\"Ethernet20\",\"ipredriver\",\"0x1\",\"preemphasis\",\"0x1234\",\"idriver\",\"0x1\"]"
1) "message"
2) "MEDIANOTIFICATION"
3) "[\"media_change\",\"\",\"port\",\"Ethernet116\",\"ipredriver\",\"0x1\",\"preemphasis\",\"0x9abc\",\"idriver\",\"0x1\"]"

doc/media-settings/Media-based-Port-settings.md
dgsudharsan: FEC can already be provided in port_config.ini (e.g. wnc/x86_64-wnc_osw1800-r0/OSW1800-48x6q/port_config.ini). FEC settings should be coming from config DB through the configuration present in port_config.ini

lguohan:
as part of the development, please also add a checker program to make sure the media setting file submit by vendor is correct. here is an example. https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-device-data/tests/config_checker

dgsudharsan:
> as part of the development, please also add a checker program to make sure the media setting file submit by vendor is correct. here is an example. https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-device-data/tests/config_checker

Sure Guohan. Will add that during development

doc/media-settings/Media-based-Port-settings.md
MichelMoriniaux: correct me if I'm wrong but is not the FEC set in the INTERFACE section, at least that is how we do it for specific optics
    "Ethernet36": {
...
      "fec": "rs"
    },

dgsudharsan: > correct me if I'm wrong but is not the FEC set in the INTERFACE section, at least that is how we do it for specific optics
> "Ethernet36": {
> ...
> "fec": "rs"
> },

Hi. What i meant to convey here is there are multiple other mechanisms used to configure FEC and hence this specific mechanism need not have provision to configure FEC settings

doc/media-settings/Media-based-Port-settings.md
```diff
@@ -0,0 +1,281 @@
...
+
+              This mechanism is also very helpful in supporting new media types without upgrading the Operating system. If a new media type need to be supported the only change that needs to be done is modify media_settings.json to add the new media type.
+              
+![](https://github.com/dgsudharsan/SONiC/blob/hld_media/doc/media-settings/key_selection_flow.png)
+
+Breakout Scenario:
+==================
+
+              For breakout the particular media, if it has a global value is listed in the media_settings.json with appropriate lane values. If the values are specific to a port, then the entry is listed under the port. If the logical port is going to be created after breakout (E.g Ethernet0 is modified to Ethernet0,Ethernet1,Ethernet2,Ethernet3) those corresponding ports are listed in the ports section in media_settings.json. In the current breakout scenario, when port_config.ini is modified and a config reload is done, the flow will be similar to bootup sequence and thus all the media settings will be pushed from xcvrd to portsorch on restart during config apply.
```
lguohan: can you give detailed example. have difficult to understand.

lguohan:
can you format the doc? it is not well formatted.

doc/media-settings/Media-based-Port-settings.md
```diff
@@ -0,0 +1,281 @@
...
+
+              To solve the above problem, few SAI attributes required for serdes programming are listed in section below. These SAI attributes will be programmed during media insert event.
+
+Media settings file:
+====================
+
+              Each vendor need to define the media settings file under device/<vendor-name>/ <ONIE_PLATFORM_STRING>/ <HARDWARE_SKU>/media_settings.json. This file contains the list of optics and DAC media-based settings for each port in the device. Each media setting for a media type comprises of key-value pairs per lane that is expected to be programmed when an optics or DAC is used. Examples of key value pairs include pre-emphasis, idriver and ipredriver.
```
lguohan: since you are going to handle break out scenario, the media_settings should not be related to HARDWARE_SKU, it should only related to the true hardware platform.

doc/media-settings/Media-based-Port-settings.md
dgsudharsan: Hi @lguohan I have given an example of breakout. I have modified the design of media settings file slightly so that the users do not need to do anything additionally to handle breakout scenario

dgsudharsan:
> can you format the doc? it is not well formatted.

Hi @lguohan I had done some formatting now. Please let me know if this looks good

doc/media-settings/Media-based-Port-settings.md
dgsudharsan: Hi @lguohan I have modified the document to have the file under platform path rather than SKU path. I have also slightly changed design so as to handle breakout cases

Rama-Nemani:
The Pre-emphasis settings can be different, when the port operates in different speeds. How does the definition handle it?

dgsudharsan:
> The Pre-emphasis settings can be different, when the port operates in different speeds. How does the definition handle it?

Hi,
I assume pre-emphasis for different speed in-turn implies a different optic or DAC corresponding to the speed. Thus those are going to be different settings as such and would have different set of values listed. For example if a port can operate at 100G as well as 40G, there will be two sections like this. The QSFP28 implies a 100G DAC whereas QSFP implies a 40G DAC
```
            "QSFP28-40GBASE-CR4-1M":{
                "preemphasis": {
                    "lane0":"0x005678",
                    "lane1":"0x005678"
                },
                "idriver": {
                    "lane0":"0x1",
                    "lane1":"0x1"
                }
            }
            "QSFP-40GBASE-CR4-1M":{
                "preemphasis": {
                    "lane0":"0x001234",
                    "lane1":"0x001234"
                },
                "idriver": {
                    "lane0":"0x2",
                    "lane1":"0x2"
                }
            }
```

doc/media-settings/Media-based-Port-settings.md
satishkumarks: Can these abstractions be further be classified into specific TX_FIR setting values(like pre-tap, main, post_tap). Some vendor-types choose to have either 3-tap co-effs or more based on the modulation type(PAM4 or NRZ). The order of values in plain hex string would be misleading at times.

satishkumarks:
> For pre-emphasis, these abstractions be further be classified into specific TX_FIR setting values(like pre-tap, main, post_tap). Some vendor-types choose to have either 3-tap co-effs or more based on the different modulation type. The order of values in plain hex string would be misleading at times.

(This was supposed to be a review comment. Realised, it took as approval by mistake. Plz Ignore)

dgsudharsan:
> urther be classified into specific TX_FIR setting values(like pre-tap, main, post_tap). Some vendor-types choose to have either 3-t

Hi. Currently SAI doesn't have TX_FIR individual attributes. I think the Vendor SAI can handle decoding pre-emphasis values into TX_FIR setting values.

stevenlu99:
Actually there could cases that we can run different speed on same transceiver. e.g people can run 40G on a 100G optic. Seems we do not have such supported yet, but could we consider add in json schema?

dgsudharsan:
> Actually there could cases that we can run different speed on same transceiver. e.g people can run 40G on a 100G optic. Seems we do not have such supported yet, but could we consider add in json schema?

Hi,
Yes there will be cases where the media can be used at a lower speed but I am not sure if someone wants this use case. 
One more question i have in mind is if someone uses a 100G media at 40G rate will the pre-emphasis change? From my understanding pre-emphasis or any transceiver tuning parameter does not vary with the speed at which port operates. There might be different pre-emphasis settings for a 100G media and 40G media on same port, but if the 100G media operates at 40G speed (due to port speed) I think it still should be programmed with 100G pre-emphasis settings. Please let me know your thoughts on this.

stevenlu99:
The tuning value will change if speed changed even with same transceiver module.

stevenlu99:
More comments:
1. How about warm boot case? in case of warm boot, everything remaining same, but we should not invoke SAI function calls to set tunning value to avoid data path impact.
2. There seems a bug in xcvrd function get_media_settings_key() 
  there are some transceiver modules could be not properly programmed and we can not assume they have keys like "manufacturenamer", "modelname", 'cable_type', etc. basically need check key exiting before referring.
3. The "GLOBAL" settings new support range, which is pretty good and convenient. How about make it better and support mixed ranges and ports. e.g. 1-5, 7, 9-32, so that we can handle cases that some ports might need special tuning values?

dgsudharsan:
> The tuning value will change if speed changed even with same transceiver module.

I assume most optics/DAC do not support dual speeds. Even if few does, I assume this might not be a normal use case where the media is used at a lower than possible speed. Please let me know your views.

stevenlu99:
dgsudharsan, that's right, we do not see many cases that same optics/DAC support dual speed. We do not necessary have it support in this change. Just want make sure we have this fact considered and we can extend our schema if this case happens later.

dgsudharsan:
> More comments:
> 
> 1. How about warm boot case? in case of warm boot, everything remaining same, but we should not invoke SAI function calls to set tunning value to avoid data path impact.
> 2. There seems a bug in xcvrd function get_media_settings_key()
>    there are some transceiver modules could be not properly programmed and we can not assume they have keys like "manufacturenamer", "modelname", 'cable_type', etc. basically need check key exiting before referring.
> 3. The "GLOBAL" settings new support range, which is pretty good and convenient. How about make it better and support mixed ranges and ports. e.g. 1-5, 7, 9-32, so that we can handle cases that some ports might need special tuning values?

Hi @stevenlu99 
1) Thanks for the catch. I handled warmboot case and updated xcvrd pull request
2) I am not directly reading from eeprom but from a transceiver_info_dict API which i assume will fill up keys with empty values if they are not read(else even without my changes xcvrd would crash) Thus i am not doing checks there.
3) Yes it does support mixed port ranges currently. You can give range like you specified.

shihjeff:
@dgsudharsan - Sorry to bother you.  I want to add some functionality on your PR for media serdes attributes and wanted to make sure my idea is complete and don't break yours.

I want to add: SAI_PORT_ATTR_HW_PROFILE_ID to SONiC to support other vendors and not have to add new SAI attributes for the various ASIC types.

Since SONiC has quite a few repos just wanted to piggyback on yours and wanted to make sure I have all of them:

https://github.com/Azure/SONiC/pull/328
https://github.com/Azure/sonic-swss/pull/821/files
https://github.com/Azure/sonic-buildimage/pull/2713

Did these also already handle the port-breakout scenario (multiple logical ports per physical port)?

PR #332, opened 2019-02-09
[warm-reboot-version] warm reboot versioning document & design

yxieca:
- Document the requirement of warm reboot versioning.
- Design solution for post warm reboot versioning.

Signed-off-by: Ying Xie <ying.xie@microsoft.com>

### DIFF
--- a/doc/warm-reboot/warm_reboot_versioning.md
+++ b/doc/warm-reboot/warm_reboot_versioning.md
@@ -1,0 +1,48 @@
+# Nomenclature
+## From version:
+- the version being upgraded out from. 
+- the version of the SONiC image running on the device before upgrading.
+
+
+## To version:
+- the version upgraded into.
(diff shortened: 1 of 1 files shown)


jipanyang:
To add more clarification:
<1> We could save the "show version" data of "FROM" image, which includes the versions of the whole SONiC images and each specific docker, to /host/warmboot at the beginning of sonic_installer install processing.

<2> At preStartAction or postStartAction of database container,  a script plugin mechanism should be designed.
 So based on the specific container "FROM" version,   a data schema transformation script should be provided by the "TO" version container if needed.

<3> The "TO" version of container/SONiC image is also responsible for maintaining the white list of supported "FROM" version.   This one applies to both pre-upgrade and post-upgrade.

This doesn't conflict with Database container versioning, but gives finer control of software warm upgrade.

PR #363, opened 2019-04-10
Add SONiC platform test plan

wangxin:
This test plan is to check the functionalities of platform related software components. These software components are for managing platform hardware, including FANs, thermal sensors, SFP, transceivers, pmon, etc.

The software components for managing platform hardware on Mellanox platform is the [hw-management package](https://github.com/Mellanox/hw-mgmt).

### DIFF
--- a/doc/pmon/sonic_platform_test_plan.md
+++ b/doc/pmon/sonic_platform_test_plan.md
@@ -593,36 +593,36 @@ Because these scripts need to be upstreamed, community definitely will have lots
 
 ## Scripts to be implemented in phase1
 
-Test Case | Script | Common for all vendors?
-----------|--------|-----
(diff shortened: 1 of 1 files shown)


liat-grozovik:
1. Please arrange the document so it will be clear what can be vendor specific and what is generic. Vendor specific can be test case but also a in the test case itself.
2. The test should be topology unaware and should read the connectivity so once connectivity matters it will check only against it
3. Toggle PSU or anything else that will be part of the automatic test code should be coded in a way it can be different from one lag/setup to another. need to consider this action by default as not supported and if there is an interface in the lab for such.
4. The test code should be based on pytest. The design part should updated following that.

wangxin:
@lguohan @jleveque I've updated the platform test plan document per review comments. Could you please check the updated version?

PR #386, opened 2019-05-17
SONiC PVST feature HLD

sandeep-kulambi:
This document provides high level design details for SONiC PVST feature


doc/stp/SONiC_PVST_HLD.md
```diff
@@ -0,0 +1,700 @@
...
+    bpdu_guard          = "yes" / "no"                          ; port disabled due to bpdu-guard
+
+### STP_PORT_STATE_TABLE
+    ;Stores STP port state per instance
+    ;Status: work in progress
+    key                 = STP_PORT_STATE:ifname:instance        ; ifname and the STP instance
+    state               = 1DIGIT                                ; 0-listen, 1-learn, 2-block, 3-forward, 4-disabled 
+
+
+### STP_VLAN_INSTANCE_TABLE
```
lguohan: does it make sense to add this as an attribute in the existing vlan table?

doc/stp/SONiC_PVST_HLD.md
```diff
@@ -0,0 +1,700 @@
...
+
+### 3.4.1 BPDU trap mechanism
+SONiC uses copp configuration file 00-copp.config.json for configuring the trap group, ids, cpu priority queue and a policer. This functionality has been extended for supporting STP and PVST BPDU trap to CPU.
+
+## 3.5 SAI
+STP SAI interface APIs are already defined and is available at below location -
+
+https://github.com/opencomputeproject/SAI/blob/master/inc/saistp.h
+
+## 3.6 CLI
```
lguohan: can we have a section to describe how to enable and disable this feature? looks like this is a part that most community concern is coming from.

doc/stp/SONiC_PVST_HLD.md
```diff
@@ -0,0 +1,700 @@
...
+ 4. Support Openconfig yang model - with extensions for supporting PVST 
+ 5. Support REST APIs for config and operational data
+
+## 1.3 Scalability Requirements
+16k port-vlan instances with max 255 STP instances.
+The scaling limit might differ depending on the platform and the CPU used, which needs to be determined based on testing.
+
+## 1.4 Warm Boot Requirements
+
+Warm boot will not be supported
```
lguohan: can you describe more on what exact do you mean by warm boot will not be supported?

doc/stp/SONiC_PVST_HLD.md
```diff
@@ -0,0 +1,700 @@
...
+
+STP container will host STPMgr and STPd process.
+
+STPMgr will handle all the STP configurations and VLAN configurations via config DB and state DB respectively.
+
+STPd process will handle all the protocol functionality and has following interactions
+
+ * Linux kernel for packet tx/rx and netlink events
+ * STPMgr interaction for configuration handling
+ * STPSync is part of STPd handling all the STP operational data updates to APP DB
```
lguohan: it looks like the kernel stp state is not programmed at all, there will be inconsistency between the asic and config db. is there going to be any case that such inconsistency will break?

doc/stp/SONiC_PVST_HLD.md
```diff
@@ -0,0 +1,700 @@
...
+
+Note: As Port-channel port ids are dynamically allocated it might result in same port channel getting different value post reboot. To ensure a predictable convergence where port-channel port id is used as tie breaker, user needs to configure appropriate port priority value.
+
+### 3.4.1 BPDU trap mechanism
+SONiC uses copp configuration file 00-copp.config.json for configuring the trap group, ids, cpu priority queue and a policer. This functionality has been extended for supporting STP and PVST BPDU trap to CPU.
+
+## 3.5 SAI
+STP SAI interface APIs are already defined and is available at below location -
+
+https://github.com/opencomputeproject/SAI/blob/master/inc/saistp.h
```
lguohan: you also need to implement the STP trap id, can you list that?

doc/stp/SONiC_PVST_HLD.md
sandeep-kulambi: Yeah this can be moved to existing vlan table. We felt keeping the tables separate will allow us to delete the STP table when the feature is disabled, whereas keeping with vlan table the entry will still be present but needs to be populated with invalid value. Let me know your thoughts, we can change accordingly.

doc/stp/SONiC_PVST_HLD.md
sandeep-kulambi: 3.6.2.1.1 Global spanning-tree mode - This section describes how to enable or disable this feature, we can change the description to "Enabling/disabling of PVST feature" to make it clear.

doc/stp/SONiC_PVST_HLD.md
sandeep-kulambi: Warm reboot is not supported in this release. User is expected to do cold reboot when PVST is running so that topology will reconverge and traffic will be redirected via alternate paths.

If PVST is enabled and user tries to perform warm reboot a warning can be displayed indicating PVST doesn't support warm reboot.

doc/stp/SONiC_PVST_HLD.md
sandeep-kulambi: Thanks for bringing this point !! We did some investigation on this and with the vlan aware bridge model being used in SONiC we don’t have a way to program the STP state, it’s possible in case of linux traditional bridge where a netdevice is created for each vlan, port combination by updating the netdevice state. 

Below are few alternatives to achieve the STP blocking functionality in kernel to keep it in sync with HW state, request your feedback on this -
1) When STP state is blocking remove the VLAN membership on the port from the vlan aware bridge, this will ensure the traffic is not processed and L2 control packets like STP can still be trapped from kernel to user space as they rely on PF_PACKET socket. 

2) Use linux tc (traffic control) filtering mechanism where we could drop specific VLAN packets on an interface. This may not be efficient compared to 1) as we may need to program more number of filters on each device(or interface) on per VLAN basis.

doc/stp/SONiC_PVST_HLD.md
sandeep-kulambi: Sure will update the doc

doc/stp/SONiC_PVST_HLD.md
sandeep-kulambi: Keeping the table separate

doc/stp/SONiC_PVST_HLD.md
sandeep-kulambi: Document updated

doc/stp/SONiC_PVST_HLD.md
sandeep-kulambi: Document updated

doc/stp/SONiC_PVST_HLD.md
sandeep-kulambi: Updated the document

doc/stp/SONiC_PVST_HLD.md
sandeep-kulambi: Updated the document

doc/stp/SONiC_PVST_HLD.md
```diff
@@ -0,0 +1,736 @@
...
+REST APIs is not supported in this release
+
+# 4 Flow Diagrams
+
+## 4.1 Global spanning-tree mode
+
+![STP](images/Global_PVST_enable.png "Figure 2: Global PVST mode enable")
+
+__Figure 2: Global PVST mode enable__
+
```
rck-innovium: There might be config_db entries that have added VLANs to STP instances but this would not have taken effect until Global PVST is enable.  So when GloblaPVST is enabled, this flow should be extended to add VLANs to the STP instances.

sandeep-kulambi: Just to clarify, global pvst must be enabled for any other spanning tree config to be accepted. In this case (add VLANs to STP instance flow) is pretty much same as described in "Figure 4 STP VLAN enable".

NarendraBandarupalli:
can this be merged if no more comments.

chenkelly:
Hi @sandeep-kulambi 
According to https://github.com/Azure/sonic-swss-common/pull/305, DB schema use "PORT" instead of "INTF". Do you plan to update the SONiC_PVST_HLD.md file? 
Thanks very much.

sandeep-kulambi:
Thanks !! I have updated the document here.

https://github.com/Azure/SONiC/pull/527/

zhangyanzhao:
not all code PRs are merged, move to backlog @adyeung

PR #390, opened 2019-05-23
NAT feature HLD

kirankella:
This document describes high level design details for NAT feature in SONiC.

Signed-off-by: Kiran Kella <kiran.kella@broadcom.com>


doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1399 @@
...
+#### 3.3.1.4 Clear command
+When the administrator issues the clear command the conntrack table is flushed, that in turn results in deleting the entries in the APP_DB by the Natsyncd daemon. 
+
+### 3.3.2 Natsync daemon
+
+NatSyncd listens to the conntrack netlink notification events from the kernel for the creation, update and removal events of the connections in the conntrack table.
+Once the notifications for SNAT and DNAT entries are received, they are pushed into the NAT_TABLE/NAPT_TABLE/NAPT_TWICE_TABLE in the APP_DB to be consumed by the NatOrch Agent. More details in the section 3.4.
+
+### 3.3.3 NatOrch Agent
+NAT feature is enabled globally in the hardware by default by NatOrch Agent during startup.
```
rlhui: We should not enable the feature globally by default, as NAT will not be needed in all cases.  Would we please have a global NAT enable/disable command?

kirankella: Sure. We will add the knob and disable it by default.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1422 @@
...
+
+Dynamically created NAT/NAPT entries are timed out after a period of inactivity in the hardware. The inactivity timeout periods are independently configurable for TCP and UDP NAPT entries.
+The inactivity timeout period is configurable for the basic NAT entries.
+
+The dynamic port allocation from the pool is done in a random manner.
+When the pool is exhausted, new incoming connections are no longer NAT'ted. In such scenario, the new incoming traffic are reported as NAT miss and dropped in the kernel. Only when the inactive entries are released are the new incoming connections dynamically mapped to the freed up ports.
+
+#### 2.2.3.1 Full Cone NAT
+When the first connection from a particular internal IP + L4 port is dynamically SNAT mapped to an external IP + L4 port, the SNAT and the corresponding DNAT entries are added for that mapping in the hardware.
+Subsequently any new connections from the same internal IP + L4 port to new destination end points use the same mapping. As well, traffic sent by any external host to this external IP + L4 port is NAT'ted back to the same internal IP + L4 port. This behavior model is referred to as Full Cone NAT.
```
lguohan: kernel connection track is 5-tuple, kernel can use the same source port and IP for two different connections. for example, 
```
(before nat)                       -> (after nat)
1.1.1.1:5555 -> 3.3.3.3:80 -> 10.10.10.10:10000 -> 3.3.3.3:80
1.1.1.2:5556 ->  4.4.4.4:80 -> 10.10.10.10:10000 -> 4.4.4.4:80
```
The above setup is totally fine for kernel since the connection tracking is 5-tuple, however, it is a problem for the asic as asic only use 3 tuple for nat.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1422 @@
...
+The allowed range of configurable zone values is 0 to 3.
+
+### 2.2.5 Twice NAT and NAPT
+Twice NAT or Double NAT is a NAT variation where both the Source IP and the Destination IP addresses are modified as a packet crosses the address zones. It is typically used in  the communication between networks with overlapping private addresses.
+
+The configuration for Twice NAT/NAPT is achieved in 2 ways:
+- Putting two Static NAT/NAPT entries in the same Group ('twice_nat_id' value).
+- Putting a Dynamic NAT/NAPT Binding and a Static type NAT/NAPT entry in the same Group ('twice_nat_id' value).
+
+When a host matching a dynamic NAT pool binding sends traffic to host with a matching DNAT Static NAT/NAPT entry in the same 'twice_nat_id' group, a bi-directional Twice NAT/NAPT entry is created for the traffic flow.
```
lguohan: does linux support twice nat scenarios, what linux configuration do we need.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1422 @@
...
+
+### 3.3.2 Natsync daemon
+
+NatSyncd listens to the conntrack netlink notification events from the kernel for the creation, update and removal events of the connections in the conntrack table.
+Once the notifications for SNAT and DNAT entries are received, they are pushed into the NAT_TABLE/NAPT_TABLE/NAPT_TWICE_TABLE in the APP_DB to be consumed by the NatOrch Agent. More details in the section 3.4.
+
+### 3.3.3 NatOrch Agent
+NAT feature is disabled by default.
+When the NAT feature is enabled, NatOrch enables the feature and the NAT miss trap actions in the hardware.
+The SNAT or DNAT miss packets are rate limited to CPU @ 600pps.
```
lguohan: this limits the connection rate to 300pps, given that you will trap two syn into kernel, this needs to be configurable.

lguohan: also, we nat miss packet drops counters to show how many connections we have dropped in cli.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1422 @@
...
+    SAI_NAT_SNAT_DISCARDS
+    SAI_NAT_DNAT_TRANSLATION_NEEDED
+    SAI_NAT_SNAT_TRANSLATION_NEEDED
+    SAI_NAT_DNAT_TRANSLATIONS
+    SAI_NAT_SNAT_TRANSLATIONS
+
+```
+The following new counters are available per entry:
+```
+The counters in the COUNTERS_DB are updated every 10 seconds. The key for the entry in the COUNTERS_DB is same as the key in the APP_DB.
```
lguohan: potentially, there are lots counters to pull, can you look into the flex counter design to see how can we reduce the counter pulling overhead?

kannankvs:
How about the support for NAT ALGs ?

kannankvs:
Regarding the reply for Guohan's question about 5-tuple, I think that the issue cannot be solved until ASICs/SAI supports 5 tuple.  At the same time, I think that kernel does not reuse the source port number until the unique port numbers are exhausted. If user can configure enough port range, I think that it should work without having the necessity for reusing the source port number.  A quick prototype with just 2 or 3 port numbers and few TCP connections may help to get the actual implementation of this Debian kernel. 
NOTE: Am unable to reply under/against Guohan's reply.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1422 @@
...
+   - Monitors the translation activity status every sampling period for each of the NAT/NAPT entries in the hardware. Sampling period is chosen to be a optimal value of 60 seconds. If the NAPT entry is inactive for the duration of the NAT timeout period, the entry is removed. Static NAT/NAPT entries are not monitored for the inactivity. They have to be unconfigured explicitly to be removed from the hardware.
+
+### 3.3.3.1 Interaction with NeighOrch
+NatOrch registers as observer with the NeighOrch so that the DNAT entry and the corresponding SNAT entry can be pushed to ASIC_DB by NatOrch, only if the corresponding inside zone neighbor's entry is created in NeighOrch.
+When the Neighbor entry is deleted in NeighOrch, the corresponding DNAT and SNAT entries are removed from the hardware.
+ 
+### 3.3.3.2 Aging inactive entries
+Since the active data traffic for the TCP/UDP connections are not flowing through the kernel, the timing out of the connection tracking entries is the responsibility of the application.
```
madhupalu: What is the design to make kernel vs ASIC conntrack NAT entries are in sync w.r.t conntrack entries timeout and refresh?

doc/nat/nat_design_spec.md
madhupalu: How to support rate limit? let's say user wanted to rate limit 1K flows.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1422 @@
...
+the following iptable rules for udp, tcp, icmp protocol types are added as SNAT rules in the nat table:
+```
+iptables -t nat -A POSTROUTING -p tcp -s 20.0.1.0/24 -j RETURN
+iptables -t nat -A POSTROUTING -p udp -s 20.0.1.0/24 -j RETURN
+iptables -t nat -A POSTROUTING -p icmp -s 20.0.1.0/24 -j RETURN
+
+iptables -t nat -A POSTROUTING -p tcp -s 20.0.0.0/16 -j SNAT -o Ethernet15 --to-source 65.55.42.1:1024-65535 --random
+iptables -t nat -A POSTROUTING -p udp -s 20.0.0.0/16 -j SNAT -o Ethernet15 --to-source 65.55.42.1:1024-65535 --random
+iptables -t nat -A POSTROUTING -p icmp -s 20.0.0.0/16 -j SNAT -o Ethernet15 --to-source 65.55.42.1:1024-65535 --random
+```
```
madhupalu: Is the current design supports ALGs (bump-in-the-wire) on SONiC?

doc/nat/nat_design_spec.md
kirankella: In the discussion here, let's refer only to the single NAT case.

I am correcting the example cited above:
(before nat)                       -> (after nat)
1.1.1.1:5555 <-> 3.3.3.3:80 -> 10.10.10.10:10000 <-> 3.3.3.3:80
1.1.1.1:5555 <-> 4.4.4.4:80 -> 10.10.10.10:10020 <-> 4.4.4.4:80

In kernel we will have an entry created like this for every flow, which is tracked as a 5-tuple (Proto, SIP, SPORT, DIP, DPORT). You are right. Kernel does track the flow matching the 5-tuple fields. 

In this 5-tuple entry, if the source IP + port are translated, kernel sets the flag SNAT for that conntrack entry. If only destination IP + port are translated, kernel sets the DNAT flag. If all the fields are changed, both flags are set.

The NAT application will only pick up the 3-tuple value of interest from the kernel (from the SNAT'ted 5-tuple) and programs them in the ASIC. The DIP and DPORT are don't care for us in case of SNAT.

Say the 5-tuple conntrack entry created in the kernel looks like
1.1.1.1:5555 <-> 3.3.3.3:80 -> 10.10.10.10:10000 <-> 3.3.3.3:80

Corresponding to the above entry, the NAT application adds two 3-tuple entries in ASIC that are created corresponding to this entry looks like below (for both directions, including the implicit DNAT):
[Proto = TCP, SIP = 1.1.1.1, SPORT =5555]  => [SIP = 10.10.10.10, SPORT = 10000] (SNAT)
[Proto = TCP, DIP = 10.10.10.10, DPORT = 10000] => [DIP = 1.1.1.1, DPORT = 5555] (implicit DNAT)

In the context of Full Cone NAT that is desired, we want to ensure any new connection from the same internal IP + port (1.1.1.1:5555) to a different destination (4.4.4.4:80) to also use the same ASIC entries created above.
Once the above 3-tuple entry is created in the ASIC, the subsequent connections are not NAT missed and are just translated using the same ASIC entries. Another 5-tuple won't be created for the new destination 4.4.4.4:80.

But there can be a narrow window where the new connection to the destination 4.4.4.4:80 hit the CPU (via NAT miss) when that traffic came before the above ASIC entry is added to the hardware for the first connection.
In that case, the kernel creates a new 5-tuple [1.1.1.1:5555 <-> 4.4.4.4:80 -> 10.10.10.10:10020 <-> 4.4.4.4:80]. We cannot push this entry to hardware as that disrupts the Full Cone NAT behavior.
The NAT application (which ensures 3-tuple logic) notices that a different SNAT'ted 5-tuple is reported for the same Source IP+Port. To ensure the Full Cone NAT behavior, the NAT application deletes the newly reported conntrack 5-tuple for the same SIP+Port from the kernel.

doc/nat/nat_design_spec.md
kirankella: No additional configuration is needed to support the twice NAT.

In the kernel, when the packet is traversing from the ingress to the egress:
1. It creates a conntrack entry if it doesn't exist
2. It first hits the PREROUTING rules in iptables, does the DNAT and updates the conntrack entry with the translated DIP+DPORT if matched
3. L3 routes the packet
4. Next it hits the POSTROUTING rules in iptables, does the SNAT and updates the conntrack entry with the translated SIP+SPORT if matched

We get netlink notifications in the NAT application whenever the conntrack entry is modified. After step 4, we see that the conntrack 5-tuple is modified with both the SNAT and DNAT actions. It implies the Twice NAT happened.

doc/nat/nat_design_spec.md
kirankella: > this limits the connection rate to 300pps, given that you will trap two syn into kernel, this needs to be configurable.

The rate limit for nat trap group should be configurable via copp config. In the lines of 

        "COPP_TABLE:trap.group.nat": {
            "trap_ids": "snat_miss, dnat_miss",
            "trap_action":"trap",
            "trap_priority":"4",
            "queue": "4",
            "meter_type":"packets",
            "mode":"sr_tcm",
            "cir":"600",
            "cbs":"600",
            "red_action":"drop"
        }

> also, we nat miss packet drops counters to show how many connections we have dropped in cli.
We should be able to retrieve the policer stats for the above trap group.

doc/nat/nat_design_spec.md
kirankella: Sure, will have a look.

doc/nat/nat_design_spec.md
kannankvs: @kirankella : Trying to explain more since I have same doubts.
I agree with Guohan on this and I replied to him in this page. Please check the same.
In your reply, you modified the example given by Guohan, which is incorrect.
C1: 1.1.1.1:5555 <-> 3.3.3.3:80 -> 10.10.10.10:10000 <-> 3.3.3.3:80
C1 is the entry created for first connection.

Let us assume that C1 TCP connection is successful, SNAT & DNAT programming are completed and traffic is going on fine.

Later, if 1.1.1.1 initiated a new TCP connection to 4.4.4.4:80, it will use a different source port number 5556 as Guohan mentioned. 
1.1.1.2:5556 <-> 4.4.4.4:80 -> 10.10.10.10:10000 <-> 4.4.4.4:80
Now, we will get a NAT miss, packet will go to linux kernel and kernel MAY use the same translated source port number 10000 to use since the 4-tuple is unique for the translated connection (10.10.10.10:10000 <-> 4.4.4.4:80 is different from 10.10.10.10:10000 <-> 3.3.3.3:80). Theoretically this is possible.   If I am correct, it is not mandatory for linux kernel to use 10001 or 10020. 
If this happens, we will have issue in creating DNAT table. That's an issue.
Please read my comment below where I tried to answer Guohan. You can verify the same and provide your reply accordingly.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1422 @@
...
+5.3 Ability to enable the logging at different severity levels of the NAT module.
+
+## 1.2 Configuration and Management Requirements
+Configuration of the NAT feature can be done via:
+
+- JSON config input
+- Incremental CLI
+
+## 1.3 Scalability Requirements
+Ability to support the scale of up to 40K Bi-directional NAT entries, but this is SAI capability dependent.
```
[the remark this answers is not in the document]
kirankella: We are looking into using the CRM to fetch the ASIC resource limits using SAI.
And not pushing the entries to hardware when the limit is reached.

doc/nat/nat_design_spec.md
kirankella: Through Copp configuration. Please refer to earlier response above.

doc/nat/nat_design_spec.md
kirankella: @kannankvs 
We verified during prototyping that with an iptables POSTROUTING SNAT rule with a port range given, we see that it is translating to a new port number (and not reusing the port number).
Like you said, if it translates to the same port number, we cannot ensure the reverse mapping in the ASIC unless it is also 5-tuple based.

doc/nat/nat_design_spec.md
kirankella: NAT application periodically queries the activity of the NAT flows every 30 seconds.
For every active NAT flow, the corresponding conntrack entry's timeout value is reset in the kernel (say set the timeout to a bigger value 3600 secs).
For the inactive NAT flows (that are inactive for the configurable timeout periods of nat-tcp-timeout, nat-udp-timeout), the corresponding connctrack entry is deleted from the kernel by the NatOrch. The delete notification is picked by NatSync and removes the entry from APP_DB and further down it gets deleted from the ASIC.

doc/nat/nat_design_spec.md
kannankvs: @kerankella  - Can you try one more prototype by giving only one port number in port range and check if linux reuses the port number if the destination IP is different for the 2nd connection? This is to check when linux might reuse the same source port number. There could be some implementations where port number is never reused.

kirankella:
@kannankvs @madhupalu  
> How about the support for NAT ALGs ?

For ALG support, it needs to be handled in the slow path (in the linux kernel) and cannot be handled in hardware. And that requires the corresponding ALG kernel modules for the protocols (like SIP, DNS,...).
This is added as one of the future features to be supported in NAT.

Currently only the ICMP error reply packets are trapped to CPU to be handled by the ALG logic for ICMP packets in the kernel.

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
lguohan: I am not sure crm has been enhanced to include nat table. Also, if you can query the nat table capacity, we can first query the capacity and use the capacity to check on application side.

doc/nat/nat_design_spec.md
lguohan: linux can rewind the source port, if there is long live the connection, i believe use same src port for different connection is possible in linux. unless kernel has option to prevent source port reuse, I think my case will happen.

rlhui: @kirankella, can we check for this case in the code and delete this second connection in kernel and also have alerts, and document that this case (port overloading) is not supported.  We should also investigate and confirm in what scenario this will happen.  Thanks.

kirankella: @lguohan @kannankvs  Verified with smaller port range sample to hit the scenario. Kernel is reusing the source port number.
@rlhui Deleting the second connection in the kernel would not help here, since the kernel will keep on trying to reuse the source port whenever the traffic keeps coming from different source sent to different destination. Using --random option spreads the port numbers in the range but doesn't prevent the reuse.
Looks like we need to make changes in the kernel in the port allocation area. I am investigating if there is any other option to address this.

kirankella: @lguohan @kannankvs @rlhui
Currently I believe the better solution seems to be to use the '--random-fully' option while adding the rules in the kernel, that randomizes the spread much better than '--random' to reduce the probability of overlap/reuse of the source ports. And if any overlap is there, we will delete those connections from the kernel and log those errors like @rlhui said. And those connections may succeed without overlap after multiple retries (by the connection initiating host) due to full randomization.

I will document the same.

On an average, with a pool port range of 180, we see the random spread is assigning nearly 120 unique source ports, and 60 connections are assigned overlapping ports.
And with a port range of 1024, we see 800 uniquely assigned source ports and others are assigned overlapping ports.
Extrapolating this, with a pool size of (64511 = port range of 1024-65535), we should have better distribution of connections (with less overlap) if the number of connections are of the order of 40000.

In Linux conntrack, the current behavior is like this.
-----------
Different Source -> Different Destinations - reuses the translated Source Port (Will address this as mentioned above).
Different Source -> Same Destination - does not reuse the translated Source Port.
Same Source -> Different Destinations - reuses the same translated Source Port (Full Cone NAT).
-----------

rlhui: @kirankella, thanks a lot for the detailed experiments and info.  For first case "Different Source -> Different Destinations - reuses the translated Source Port", it will not function correctly I believe.  Do we know why Linux is doing that?  For 3rd case, we can keep the behavior for full cone nat.

lguohan: what about this one? https://github.com/Chion82/netfilter-full-cone-nat

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1446 @@
...
+
+# 2 Functionality
+## 2.1 Target Deployment Use Cases
+NAT feature is targeted for DC and Enterprise deployments.
+
+In DC deployments, the ToR switch can enable NAT feature to translate the traffic from/to the attached Servers running in the private network and the public domain.
+![DC deployment use case](images/nat_dc_deployment_usecase.png)
+
```
lguohan: i feel we need to explain a little bit more on the DC deployment use case here. Can you add these three packet flows in the diagram. Then, the design can centered around how to support these three scenarios.

1. to and from internet, we do the DNAT and SNAT.
2. within DC, we do not need to do the NAT. Can use private IP for communication, save ASIC resource.
3. within the TOR, if server needs to talks to each other. In some scenarios, we need to DNAT and SNAT via the TOR.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1446 @@
...
+Before the L3 forwarding is done in the Kernel, the iptables rules in PREROUTING chain are applied. If there are any DNAT rules that the packets match against, they translate the DIP and L4 port to the IP and L4 port as applicable. In this process, the DNAT state and DIP of the connection entry in the connection tracking table are updated.
+
+After the L3 forwarding is done and before the packet is about to be sent out, the rules in the POSTROUTING chain are applied which do the SNAT (translate the SIP + L4 port to the public IP + L4 port). In this process, the SNAT state and SIP of the connection entry in the connection tracking table are updated.
+
+As long as the hardware entry is not installed, the NAT translation and forwarding is done in the Linux kernel.
+
+### 3.4.2 Connection tracking
+Connection tracking module in the kernel creates the connection entries and maintains their states as and when the packet traverses the forwarding path in the kernel. It keeps track of all the connections created in the system. IPtables module consults the connections tuples tracked in the system during NAT process and updates the connections.
```
lguohan: I feel we need to first describe the asic limitation, 3-tuple matching instead of 5-tuple matching for nat. The limitation creates mismatch between the asic nat model and linux traditional conntrack model. Then, our design should address such mismatch.

doc/nat/nat_design_spec.md
lguohan: for --random-fully option, the collision ratio is a little bit high. it means that once we reach 40000, many connections will not be setup without a few retries. I feel we need a better solution here.

lguohan:
since zone concept is not available in the kernel, how do represent the zone in the ip table, or do we need that?

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1446 @@
...
+    "packet_action": "forward"
+
+NAT_BINDINGS|nat1    
+    "access_list": "10"
+    "nat_pool": "pool1
+    "nat_type": "snat"
+
+PORT|Ethernet15
+    "nat_zone": 1
+
```
rlhui: For dynamic source NAT, the public IP to translate to may not be a physical interface address.  It can be a loopback interface IP.  Please support this.  In the sai pipeline as well as in the spec here, it's stated that outgoing zone-id is from the next-hop interface from the L3 route lookup of the destination.  How to derive the zone-id when the public IP is a loopback IP interface address?  Thanks.

kirankella: Using loopback IP as public IP makes sense. It will be supported. Will update the spec to mention about it.
The outgoing zone-id is not affected by the Source Public IP used. The next-hop interface for the L3 route lookup to an external destination would be physical interface facing outbound.

kirankella:
> since zone concept is not available in the kernel, how do represent the zone in the ip table, or do we need that?

We are not mapping to any equivalent zone attribute in the kernel, since iptables do not take such an attribute. Since SNAT MISS packets come up to CPU only if zone crossing happens, the zone filtering check is kind of taken care of in the hardware, and only the translation happens in the iptables based on the ACL matching subnets.

doc/nat/nat_design_spec.md
kirankella: > what about this one? https://github.com/Chion82/netfilter-full-cone-nat since the asic only support full cone nat, three tuple for connection tracking instead of five tuple, then in the kernel we should do the same full cone nat in the kernel. however, the patch says it only applies to udp packets.

@lguohan Right, we need a solution that applies to both udp and tcp.
Once the hardware entry for the first traffic outbound SNAPT (and implicit DNAPT) are added,  we need not have to do the full cone nat for the incoming traffic in the kernel, as the hardware translates (fullcone) the incoming traffic from new external sources (falling in the same 3-tuple). So, currently it is not a hard requirement to implement full cone nat in the kernel for the inbound traffic.

doc/nat/nat_design_spec.md
kirankella: > this limits the connection rate to 300pps, given that you will trap two syn into kernel, this needs to be configurable.

@lguohan Just an update. Verified that the Linux kernel sets the TCP connection tracking entry's state to ASSURED only after the 3 way handshake.
We in the application are accepting TCP connections only after the ASSURED flag set.
Documentation elsewhere mentioned just the bi-directional packets sets the connection to ASSURED. So, we were expecting the 2 SYNs will set the TCP state to ASSURED.
But the l4 protocol tracker (TCP layer here) is setting the state to ASSURED only after the 3-way handshake. So, the 2 SYNs + ACK are trapped to CPU, before the entry is added to the hardware.
I am updating the HLD accordingly.

doc/nat/nat_design_spec.md
kirankella: > @kirankella, thanks a lot for the detailed experiments and info. For first case "Different Source -> Different Destinations - reuses the translated Source Port", it will not function correctly I believe. Do we know why Linux is doing that? For 3rd case, we can keep the behavior for full cone nat.

@rlhui Linux was by default trying to optimize on the translated port to pick up from the port range. Since it tracks by 5-tuple uniqueness, it is reusing the translated Source Port from the port range. It will work correctly in the Linux Kernel for any NAT'ted traffic being forwarded back and forth in the kernel. Since we have the limitation in the hardware to match with, we are using the random option to minimize the reuse of the translated port.

rlhui:
@kirenlella, please add this behavior to the HLD.  Also the NAT overall feature behavior is really tightly coupled with NAT SAI pipeline, in all cases that if resulting behavior does not exactly match what's mentioned in the SONiC spec, due to differences between SW (SONiC) and HW (logically represented by the SAI pipeline), can we please mention it in HLD?  Thanks,

kirankella:
> @kirenlella, please add this behavior to the HLD. Also the NAT overall feature behavior is really tightly coupled with NAT SAI pipeline, in all cases that if resulting behavior does not exactly match what's mentioned in the SONiC spec, due to differences between SW (SONiC) and HW (logically represented by the SAI pipeline), can we please mention it in HLD? Thanks,


Done. Updated the section 3.4.

doc/nat/nat_design_spec.md
kirankella: Done.

doc/nat/nat_design_spec.md
kirankella: Done

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1517 @@
...
+### 2.2.4 NAT zones
+NAT zones refer to different network domains between which the NAT translation happens when the packet crosses between them. NAT zones are also referred to as NAT realms.
+NAT zones are created by configuring zone-id per L3 interface. The L3 interface referred to for NAT purposes can be an Ethernet, VLAN or PortChannel or Loopback interface that are configured with IP address(es).
+
+In this document, the interface that is towards the private networks (private realm) on the NAT router is referred to as an inside interface.
+And the interface that is towards the public network (public realm) on the NAT router is referred to as outside interface.
+By default, L3 interface is in NAT zone 0 which we refer to as an inside interface.
+
+NAT/NAPT is performed when packets matching the configured NAT ACLs cross between different zones.
+The source zone of a packet is determined by the zone of the interface on which the packet came on. And the destination zone of the packet is determined by the zone of the L3 next-hop interface from the L3 route lookup of the destination.
```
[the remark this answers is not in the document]
kirankella: The egress interface would also be an L3 interface right? So, user would have configured zone value on that outbound L3 interface (which happens to be egress interface that has the nexthop via that interface).

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1517 @@
...
+    "nat_ip": {{ip_address}} (or) {{ip_addr_start}}-{{ip_addr_end}} 
+    "nat_port": {{start_l4_port}}-{{end_l4_port}}  (OPTIONAL)
+```
+
+#### 3.2.1.5 ACL to Pool Bindings
+```
+NAT_BINDINGS|{{binding-name}}
+    "nat_pool": {{pool-name}}
+    "access_list": {{access-list-name}} (OPTIONAL) 
+    "nat_type": {{snat-or-dnat}}		(OPTIONAL)
```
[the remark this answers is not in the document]
kirankella: NAT bindings are used for dynamic NAT/NAPT. Having the NAT type as SNAT means, the Source addresses are dynamically translated when the bindings are applied. We currently support only SNAT type for dynamic pool bindings.
In future, we also can add support for dynamically doing DNAT (use case being FQDN service with session distribution at the firewall https://docs.paloaltonetworks.com/pan-os/8-1/pan-os-admin/networking/nat/source-nat-and-destination-nat/destination-nat.html)..

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1517 @@
...
+
+ACL rules are added with 'do-not-nat' action for the hosts for which the NAT is to be avoided, and with 'forward' action for the hosts for which the NAT is to be performed.
+
+The action 'do-not-nat' tells the hardware to skip doing NAT zone checks and NAT processing for the packet, and instead the packet is L3 forwarded. For the rest of the permitted packets, the NAT processing is done in the hardware.
+If a matching SNAT entry or DNAT entry does not exist in the hardware, the packets are trapped to CPU as NAT miss packets to be processed by the kernel software forwarding path.
+
+Corresponding to the ACL, the equivalent deny and permit rules are added in the iptables so that the intended NAT miss packets are dynamically SNAT'ted in the kernel.
+
+When the ACL that is bound to the NAT pool is deleted, the corresponding iptables SNAT rules are deleted in the kernel.
+When the ACL rules are created or modified, the corresponding iptables SNAT rules are updated in the kernel.
```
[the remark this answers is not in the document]
kirankella: No. AclOrch is only responsible for configuring the ACL configuring (do-not-nat) in the hardware.
NatMgr daemon listens on the ACL configuration changes and only those ACL config that is used in NAT Pool/Bindings are acted upon by the NatMgr for updating the iptable rules in the kernel.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1517 @@
...
+Multiple ACLs can be bound to a single NAT pool. 1 ACL can be bound at most to a single NAT pool.
+
+If there is no matching ACL configured, it is treated as 'implicit permit all' and hence the corresponding NAT pool in the binding is applied on all the traffic.
+
+Following match actions of the ACL are handled to match against the traffic for dynamic SNAT mapping:
```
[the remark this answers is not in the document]
kirankella: They are enforced using different source/destination ip and port options in the iptable rules that do the SNAT translation mapping for the new NAT miss traffic.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1517 @@
...
+And the interface that is towards the public network (public realm) on the NAT router is referred to as outside interface.
+By default, L3 interface is in NAT zone 0 which we refer to as an inside interface.
+
+NAT/NAPT is performed when packets matching the configured NAT ACLs cross between different zones.
+The source zone of a packet is determined by the zone of the interface on which the packet came on. And the destination zone of the packet is determined by the zone of the L3 next-hop interface from the L3 route lookup of the destination.
+
+Currently only 2 zones are supported, which correspond to the inside interfaces and the outside interfaces.
+
+Any inbound traffic ingressing on the outside interface that is L3 forwarded on to the inside interface, is configured by user via Static NAT/NAPT entries to be DNAT translated.
+Any outbound traffic ingressing on the inside interface is configured to be dynamically SNAT translated.
```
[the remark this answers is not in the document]
kirankella: In the above statements, we are mentioning the typical use case where the DNAT is configured to be done and the SNAT is configured to be done (in which directions usually).
And it is SAI pipeline specific that the DNAT_POOL match should happen first for the DIP, before NAT happens while crossing the zones. Hence it is not listed in this HLD.

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: Ok. I wanted to know where the out zone is configured exactly because it makes a difference for the other applications. We don't want to mix it with the next hops which are configured by other applications.

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: Let me rephrase my question. In the config example, the rule looks like this:
ACL_RULE|10|1
    "priority": "20", 
    "src_ip": "20.0.1.0/24",
    "packet_action": "do_not_nat"

How are you going to ensure that only those fields from the 5-tuple are going to be used as match criteria, and not, for example, an ethertype or any other field? Also, what about VRF? Do you want to include it as a match field?

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: In the ACL there is only a do-not-nat action proposed. So how is the do-not-nat going to distinguish between nat_type for which it should disable translation? E. g. for this configuration:
```
ACL_TABLE|10
    "stage": "INGRESS", 
    "type": "L3", 
    "policy_desc": "nat-acl", 
    "ports": "Vlan2000"

ACL_RULE|10|1
    "priority": "20", 
    "src_ip": "20.0.1.0/24",
    "packet_action": "do_not_nat"

ACL_RULE|10|2
    "priority": "10", 
    "src_ip": "20.0.0.0/16", 
    "packet_action": "forward"

NAT_BINDINGS|nat1    
    "access_list": "10"
    "nat_pool": "pool1
    "nat_type": "snat"
```
How do we tell HW to disable only SNAT for packets with SRC_IP "20.0.1.0/24" and to not disable DNAT for the same packet?

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: OK

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
kirankella: ACL action do-not-nat disables any NAT that can happen on the packet.
And the nat-type attribute in the bindings config is only of use for the kernel iptables to act upon (whether to apply or not apply snat or dnat) on the NAT miss traffic to the CPU.
And like mentioned above, only snat type is supported in the bindings.

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
kirankella: The iptable rules will check only for the fields passed as arguments for the SNAT action (in the above example, we only pass the source ip option) and will not check other fields in the packet. Did I get your question correctly? 
I don't see VRF match as is that is taken by iptables rules. But we achieve the similar result by matching against the outgoing interface entry (that is a result of the lookup in the particular VRF that the packet is routed in).

Just for info - the 5-tuple conntrack entry is created for the first packet in the flow that matches the iptables rules. And for the rest of the packets in the same flow, the iptables rules are not applied and the already created conntrack entry is referred to.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1517 @@
...
+iptables -t nat -A POSTROUTING -p tcp -j SNAT -o Ethernet28 --to-source 65.55.42.1:1024-65535 --random-fully
+iptables -t nat -A POSTROUTING -p tcp -j SNAT -o Vlan100 --to-source 65.55.42.1:1024-65535 --random-fully
+```
+
+### 3.4.2 Connection tracking
+Connection tracking module in the kernel creates the connection entries and maintains their states as and when the packet traverses the forwarding path in the kernel. It keeps track of all the connections created in the system. IPtables module consults the connections tuples tracked in the system during NAT process and updates the connections.
+
+Connections when added, deleted or updated in the connection tracking system are notified via netlink interface to interested listeners (natsyncd in our case).
+
+#### 3.4.2.1 Handling NAT model mismatch between the ASIC and the Kernel
```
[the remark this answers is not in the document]
kirankella: From the configuration perspective (of the NAT feature) of most of the vendors I see, create the 3-tuple mappings. Believe it translates to most of the hardware designs (3-tuples). Also, using 5-tuples in the hardware might take up more hardware resources in the cases where we just need 1 entry (Full Cone NAT case).

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: So in the example above, if a packet with the SRC IP from 20.0.1.0/24 and DST IP from the DNAT_POOL, won't be translated and/or trapped to CPU because of that ACL with do-not-nat action.
Is it desired behavior? If the application controls SNAT and DNAT separately, why not do the same in HW?

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: It is clear that iptables can only check for that fields, but schema does not restrict the user from using other fields. So how, for example, this rule will be treated where I also put a dscp value?

```
ACL_RULE|10|1
"priority": "20",
"src_ip": "20.0.1.0/24",
"dscp": "50";
"packet_action": "do_not_nat"
```

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: I agree with the motivation to reduce resource usage.
Does this retry routine always converge? Is it possible to have a configuration where we're close to filling up the NAT table and the number of retries will significantly affect the latency of TCP handshake?

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
kirankella: Actually the iptables rules can be matched against other options too (like dscp, ecn, length, tos, ttl ..).
Just that we need to parse those options also from the ACL config and add them to the rules from NatMgr.
NatMgr needs to be extended to handle those options as well as needed. Currently NatMgr is handling only the fields mentioned above in the HLD.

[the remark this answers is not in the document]
marian-pritsak: OK

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
kirankella: I believe this is again SAI pipeline specific. Zone is configured only on L3 interface in the hardware. And the egress nexthop in the hardware picks the zone attribute from the nexthop L3 interface it points to.

doc/nat/nat_design_spec.md
rlhui: Hi Kiran, when there is a port-reuse but sources are different, it'll not function correctly. Please continue to find a solution so that this will not happen.

kirankella: The changes for 3-tuple uniqueness and fullcone nat in the kernel conntrack are being tested. Will update the HLD once the changes are ready.

doc/nat/nat_design_spec.md
kirankella: Will move to flex counters once we scale up the NAT entries beyond 1K in the near future.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1517 @@
...
+1. Since we do not want to use the NAT table space in the hardware to be filled up with any uni-directional SYN flood traffic.
+2. So that the kernel does not timeout the unidirectional SYN-SENT connection state entries early (if only SYN packet is software forwarded in the kernel and the SYN+ACK/ACK packets are hardware forwarded).
+3. To have the conntrack TCP entries and the hardware entries in sync.
+
+The conntrack entry notified by the kernel has the SIP and DIP fields in both directions of the flow. If only SIP or DIP is modified in any direction, it is a case of Single NAT/NAPT. If both the SIP and DIP are modified in any direction, it is a case of Twice NAT/NAPT.
+
+The conntrack netlink DESTROY events result in the deletion of the NAT entry from the APP_DB. The DESTROY events are received on the timeouts of the TCP connections in the connection tracking table.
+
+The TCP FIN flagged packets are not trapped to CPU. Hence the NAT entries for the closed TCP connections are not removed immediately from the hardware. They are timed out eventually based on the translation inactivity and removed.
```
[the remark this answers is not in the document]
kirankella: Using the same DNAPT entry, lots of TCP flows (new connections) are translated in the hardware (based on the fullcone nat model in the hardware). For the very first flow detected, the hardware entry is added. And the follow up flows are not in the kernel. Hence trapping FIN/RST packets to the kernel may result in getting them dropped in the kernel.
We can configure lower TCP NAT timeout in order to get the inactive sessions timed out early in the hardware.

doc/nat/nat_design_spec.md
kirankella: Done.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1550 @@
...
+sai_attribute_t nat_zone_counter_attr[10];
+
+nat_zone_counter_attr[0].id = SAI_NAT_ZONE_COUNTER_ATTR_NAT_TYPE;
+nat_zone_counter _attr[0].value = SAI_NAT_TYPE_SOURCE_NAT;
+
+nat_zone_counter_attr[1].id = SAI_NAT_ZONE_COUNTER_ATTR_ZONE_ID;
+nat_zone_counter_attr[1].value.u32 = 1;
+
+nat_zone_counter_attr[2].id = SAI_NAT_ZONE_COUNTER_ATTR_ENABLE_TRANSLATION_NEEDED;
+nat_zone_co
```
stepanblyschak: Could you please describe a bit more reconciliation in natsyncd?
Is the reconciliation done after a timer expires like in fpmsyncd warm restart case?
Based on what do you mark an entry as stale?
Do we need reconciliation here? What if we just restore and then dynamic NAT entries will be removed if no traffic hits them.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1550 @@
...
+0d3b86e300f9        docker-lldp-sv2:latest            "/usr/bin/supervisord"   2 minutes ago        Up About a minute                       lldp
+f65865e90fd1        docker-platform-monitor:latest    "/usr/bin/supervisord"   2 minutes ago        Up About a minute                       pmon
+4374ebb6e0e1        docker-teamd:latest               "/usr/bin/supervisord"   2 minutes ago        Up About a minute                       teamd
+6244613730fb        docker-fpm-frr:latest             "/bin/sh -c '/usr/biâ¦"   2 minutes ago        Up About a minute                       bgp
+9ba9d3e63426        docker-database:latest            "/usr/bin/supervisord"   2 minutes ago        Up 2 minutes                            database
+```
+## 3.6 SAI
+Table shown below represents the SAI attributes which shall be used for NAT.
+
+NOTE: NAT SAI Spec is not finalized at the time this spec is written, and is subject to change. The PR for the NAT SAI Spec is @ [NAT SAI Spec PR](https://github.com/opencomputeproject/SAI/pull/937)
```
stepanblyschak: Seems that NAT SAI spec PR is merged, could you please update the design accordingly?

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1550 @@
...
+    "local_ip": {{ip_address}}
+    "local_port" :{{l4_port}}
+    "nat_type"   :{{snat-or-dnat}}      (OPTIONAL)
+    "twice_nat_id" :{{twice-nat-id}}    (OPTIONAL)
+```
+
+```
+This config tells to do:
+- If the "nat_type" is 'dnat':
+  - DNAT translation of the DIP/DPORT in the IP packet from 'global_ip' address and 'global_port' to 'local_ip' address and 'local_l4_port' when the packet crosses the zones.
```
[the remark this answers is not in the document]
kirankella: This is meant to say, DNAT happens from the configured values of global to local ip addresses, when the matching DNAT entry exists in the hardware tables. Crossing the zones statement is no longer required here. Shall remove the same. Since zones can still be equal in case of hair-pin nat'ting case.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1550 @@
...
+[SIP=1.0.0.2, SPORT=120 ==> DIP=2.0.0.3, DPORT=300] is SNAT'ted as [SIP=65.55.45.1, SPORT=600 ==> DIP=2.0.0.3, DPORT=300]
+since the translated 5-tuples are unique in each case.
+
+For the first traffic flow above, the 3-tuple entry is added in the hardware to
+[SIP=1.0.0.1, SPORT=100] SNAT to [SIP=65.55.45.1, SPORT=600]
+
+The second traffic flow [SIP=1.0.0.2, SPORT=120] cannot be added in the hardware to translate to the same IP/PORT [SIP=65.55.45.1, SPORT=600], since the reverse traffic flows cannot be uniquely translated to the original Source endpoints.
+
+This mismatch in the NAT models between the ASIC and the Kernel is addressed by:
+- Changes in the Linux kernel to do 3-tuple unique translation and full cone NAT functionality in the outbound (SNAT) direction.
```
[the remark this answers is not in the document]
kirankella: Patch file is added in the sonic-linux-kernel submodule for the kernel changes. Shall be submitting a PR shortly for these changes.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1550 @@
...
+    "nat_type"   : "dnat"
+```
+```
+/* Step 1: Create a Source NAPT Entry object:
+ * ---------------------------------------- */
+sai_attribute_t nat_entry_attr[10];
+nat_entry_t snat_entry;
+
+nat_entry_attr[0].id = SAI_NAT_ENTRY_ATTR_NAT_TYPE;
+nat_entry_attr[0].value = SAI_NAT_TYPE_SOURCE_NAT;
```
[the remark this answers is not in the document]
kirankella: The example is correct. The nat_entry_attr is populated with the translated (global ip) values, and the snat_entry.data (which has the key) is populated with the local_ip.

doc/nat/nat_design_spec.md
kirankella: Yes, reconciliation starts after the timer expires.
As explained in section 6 in this document:
The warmRestartAssist class is used to cache the data read from APP_DB.
Now the conntrack entries are read and pushed to warmRestartAssist class.
After the timer expires, the reconciliation is done by warmRestartAssist which removes those entries the stale entries from the APP_DB. Ideally all the entries should be in sync. In case for reason if a connrtack entry failed to be restored in the kernel, that entry is marked as stale and removed from app-db during reconciliation.

doc/nat/nat_design_spec.md
kirankella: Sure.

doc/nat/nat_design_spec.md
```diff
@@ -1520,31 +1529,31 @@ The Unit test case one-liners are as below:
...
-| 50 | Verify the tracing of natmgrd at different log levels.                                                                                                                                                                                 |
-| 51 | Verify the tracing of natorch at different log levels.                                                                                                                                                                                |
-| 52 | Execute the debug dump commands for dumping the internal operational data/state of NatOrch and NatMgrd.                                                                                                                               |
-| 53 | Verify NAT happens on ICMP traffic like ping, traceroute traffic.                                                                                                                               |
-| 54 | Verify NAT happens on TCP traffic like ssh, telnet and UDP traffic like TFTP.                                                                                                                               |
```
rlhui: Could we please add and verify test cases when static NAT entries are deleted, as well as the ACLs are changed for both static and dynamic NAT?

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1559 @@
...
+STATIC_NAPT|65.55.42.1|TCP|1024 
+    "local_ip": 20.0.0.1
+    "local_port" :6000 
+    "nat_type": "dnat"
+
+INTERFACE|Ethernet15|65.55.42.1/24
+...
+
+INTERFACE|Ethernet15
+    "nat_zone": 1
```
rlhui: From this example, it seems configuring "nat_zone" at an interface is the way to enable this static DNAT on the interface., is it correct?  If there is another ethernet interface configured with a different nat_zone id, it'll not use DNAT entry, is it correct?  What if we want one more DNAT entry to be bound to ethernet15, how to do so?  Do we support two different static DNAT entries on two different interfaces with different zone-ids?  It is not clear to me how we can clearly associate which static NAT entries to which interface.

kirankella: The nat_zone mentioned here is used to push the zone attribute to the hardware on the L3 RIF (and only hardware uses it for the zone-id differences when doing the translation).

For binding the interface to the static dnat entry, the L3 IP (65.55.42.1) of the nat entry is matched against any L3 interface IP in the system. [This matching is needed for the application to add an entry to the hardware].
We can add any number of static napt entries with the same L3 IP (and with different ports) that will be matched to the same interface that has that IP (Ethernet15 here).
For example, in addition to the above static dnapt entry, we can add another static dnapt entry bound to the same interface.
STATIC_NAPT|65.55.42.1|TCP|1025 
    "local_ip": 30.0.0.1
    "local_port" :8000 
    "nat_type": "dnat"

Yes, we can add different static dnat entries matching different interfaces with same or different zone-ids.
For example, we have another L3 interface (Ethernet16) with IP 65.55.43.1, we can add another static dnapt entry
STATIC_NAPT|65.55.43.1|TCP|1024 
    "local_ip": 40.0.0.1
    "local_port" :6000 
    "nat_type": "dnat"

where 
INTERFACE|Ethernet16|65.55.43.1/24
...

and zone add can be 
INTERFACE|Ethernet16
    "nat_zone": 1

or
INTERFACE|Ethernet16
    "nat_zone": 2

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1550 @@
...
+
+After the L3 forwarding is done and before the packet is about to be sent out, the rules in the POSTROUTING chain are applied which do the SNAT (translate the SIP + L4 port to the public IP + L4 port). In this process, the SNAT state and SIP of the connection entry in the connection tracking table are updated.
+
+As long as the hardware entry is not installed, the NAT translation and forwarding is done in the Linux kernel using the iptables and conntrack entries.
+
+#### 3.4.1.1 No NAT Zones in the Kernel
+Similar to the zone per L3 interface that is programmed in the hardware, there is no concept of NAT zone on the interface in the Kernel. As a result, no zone checks can be done in the kernel when doing NAT. Matching is done against the outbound interface for the POSTROUTING/SNAT traffic and matching against the inbound interface for the PREROUTING/DNAT traffic. The outgoing interface parameter should be passed to the iptables rules to ensure that the NAT is done for the software routed traffic in the kernel only when the packet is sent on the interface corresponding to the outside zone. Without the interface parameter, any software routed packet can get NAT'ted.
+
```
rlhui: Would be good to improve the clarify of this sentence " The outgoing interface parameter should be passed to the iptables rules to ensure that the NAT is done for the software routed traffic in the kernel only when the packet is sent on the interface corresponding to the outside zone."  - what parameter is being passed.  Here it seems you're talking about packets received from downlinks belonging to one zone, to an uplink belonging to a different zone-id.  What's being checked to ensure "packet is sent on the interface corresponding to the outside zone"?

doc/nat/nat_design_spec.md
rlhui: Would we please add the debug commands?

rlhui: Did we handle the case that  previous L3 Route/neighbor/nexthop entries were gone after restarting?  What's the behavior for NAT in this case?

doc/nat/nat_design_spec.md
rlhui: Did you mean the specific DNAT matching and action happens only if the original DIP matches the IP interface of the packets' incoming interface?  In example above, the DNAT action to translate from public IP 65.55.42.1|TCP|1024, will be applied to Ethernet15 only and not any other Ethernet interface, is it correct?  Or it is not interface specific?  E.g. if the same packet is received at another interface Ethernet17 with nat_zone 1 configured, but Ethernet17 has a different IP address configured, what will be the behavior?

doc/nat/nat_design_spec.md
kirankella: Current design is as below:
For normal interfaces (non loopback interfaces) that are used as Public IP for NAT, interface parameter is passed when programming the iptables so that if the packets are going out on that interface, iptables apply the dynamic nat pool.
for eg., iptables -t nat -A POSTROUTING -j SNAT ... -o <Ethernetxx> --to-source <ip+port pool>

For loopback interface whose IP is used as NAT public IP, the packets can go on any interface that are in the same zone as the loopback interface. So in this case, interface parameter is not passed to the iptables, and hence it is handled differently. The iptables mangles table rules set the MARK value to the zone value when a packet is going out on an interface. And in the iptables nat table rules for loopback, we match on the MARK value that is matching the loopback zone value.

Sure, will update this section for clarity.

If we want the same behavior on normal interface too (to not just match only on the outgoing interface) as in the case of the loopback interface, we will be restricted to configuring only 1 pool for 1 outside zone.

doc/nat/nat_design_spec.md
rlhui: @kirankella, I don't see a test case in the hld for nat pkt trapping rate.  Could we please add? Thanks.

doc/nat/nat_design_spec.md
kirankella: The DNAT entries are added to the hardware when the translated dip address is either resolved via a neighbor/nexthop entry or via a L3 Route. If unresolved, the DNAT entry is deleted from the hardware.
So, after restart, the DNAT entries are added only after the translated DIP is resolved. This can happen if the packet came up and software forwarded resulting in the ARP entry for connected internal hosts.

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
rlhui: @kirankella, please confirm if we're querying the capacity as suggested by Guohan?  Thanks.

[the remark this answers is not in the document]
kirankella: @rlhui Currently the application limits it to 1K. 
We are implementing the SAI attributes to query the capacities (SAI_SWITCH_ATTR_AVAILABLE_SNAT_ENTRY, SAI_SWITCH_ATTR_AVAILABLE_DNAT_ENTRY...)

kirankella:
> "When ACL rule is changed from 'forward' action to 'do_not_nat' action, the matching traffic flows corresponding to the NAT entries that were created before due to the 'forward' action continue to be translated till the NAT entries are timed out."
> 
> why could we not delete the NAT entries in HW upon this? If traffic is still there, there would be no way to not do NAT even though CLI is changed.

@rlhui It requires to find out what conntrack entries were created because of which ACLs (iptables in the kernel). To identify such conntrack entries, we need to kind of run the whole ACL rules matching logic in the NAT application. 
Once we are able to identify such matched conntrack entries for any ACL, we can delete such entries in the kernel, if the ACL action was changed from 'forward' to 'do_not_nat'. This can be tracked as a future item.

Moreover, apart from the DIP, DPORT, SIP, SPORT fields (that are available in the conntrack entry), we cannot match the conntrack entry against the other ACL matching rules if configured by the user (like length, DSCP, TOS,...).

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
rlhui: @marian-pritsak , in NAT SAI pipeline, SNAT and DNAT are also controlled separately. Please raise again if this should be discussed further.

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
rlhui: @kirankella, the corresponding section in doc seems not yet updated.

doc/nat/nat_design_spec.md
rlhui: @kirankella, wondering if you get a chance to verify/update the hld any change here with latest SAI spec?  Thanks.

doc/nat/nat_design_spec.md
rlhui: And what's the behavior when the configured DNAT global IP is matched to a loopback interface IP only?  thanks.

doc/nat/nat_design_spec.md
rlhui: @kirankella, I was referring to warm boot case where the DNAT entries were already in hardware.  
With warm reboot, a DNAT entry was already created prior to rebooting.  If next-hop was gone after rebooting, then the DNAT entries in HW or all SW tables (e.g. APP_DB?) should be deleted.  Wanted to double check if this case was tested.

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: OK

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: OK

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: OK

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: @rlhui the DO_NOT_NAT is global in the SAI pipeline. Once you disabled the NAT, it is disabled for all the types - SNAT< DNAT, DOUBLE_NAT. I don't get why we should do it differently in SONiC.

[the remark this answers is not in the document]
rlhui: @marian-pritsak , you're right that if a pkt matches DO_NOT_NAT condition in the SAI pipeline, then DNAT or DOUBLE_NAT is bypassed too.  In SONiC, currently it seems acl binding including DO_NOT_NAT is supported with SNAT Only.

@kirankella, can you please confirm?  Is there a way to bypass NAT for some pkts e.g. some known protocols for static DNAT?

doc/nat/nat_design_spec.md
kirankella: Test for pkt rate limiting is already present in the latest HLD. Yes, the rate limit configured in the Copp trap table is tested.

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
kirankella: This section is already updated for the DNAT in the latest HLD. Crossing the zones statement is valid for SNAT. It is not valid for DNAT.

doc/nat/nat_design_spec.md
kirankella: Done.

doc/nat/nat_design_spec.md
kirankella: > Did you mean the specific DNAT matching and action happens only if the original DIP matches the IP interface of the packets' incoming interface? In example above, the DNAT action to translate from public IP 65.55.42.1|TCP|1024, will be applied to Ethernet15 only and not any other Ethernet interface, is it correct? Or it is not interface specific? E.g. if the same packet is received at another interface Ethernet17 with nat_zone 1 configured, but Ethernet17 has a different IP address configured, what will be the behavior?

@rlhui Actually, matching against any interface in the same subnet is a criteria for application's purpose only to add the matching iptables rules in the kernel, and to push the entry to the hardware.
Once the entry is added in the hardware the behavior as per SAI pipeline, will not match on the packet's incoming interface. DNAT will happen for the packet ingressing with that original IP when received on any interface as long as DNAT criteria is met.

doc/nat/nat_design_spec.md
kirankella: Done.

doc/nat/nat_design_spec.md
rlhui: Thanks.  I see that the test case " Modify the ACL rules and verify that the iptable rules are updated in the kernel. " is added.  It seems corresponding conntrack or hardware rules may not be cleaned up after ACL rules change?  If so, please mark it as the current limitation.

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
stepanblyschak: Ok.

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
stepanblyschak: I see

doc/nat/nat_design_spec.md
stepanblyschak: According to warm reboot design the table is 'WARM_RESTART_TABLE', Why do we need a separate table 'NAT_RESTORE_TABLE'?

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
kirankella: @marian-pritsak @rlhui What you said is right. In SONiC, currently acl binding including DO_NOT_NAT is supported with SNAT only (ideally other vendors config apply ACLs to the SNAT dynamic pool allocation). So on those lines the iptables added apply the do-not-nat checks to exclude the some hosts during SNAT. Also, the ACL (with do-not-nat) is typically applied on all the inside zone ports where the outbound traffic is being SNAT'ted. That way do-not-nat is not applied on the outside zone ports in the hardware.

But as per the SAI pipeline, the DO_NOT_NAT condition is like a global condition i see, but that be can be achieved only if this ACL rule is applied on all the ports (both the inside zone and the outside zone ports).
Currently for the static DNAT (you are not referring to NAPT right?), all the traffic is DNAT'ted in the hardware for all protocols by the single DNAT entry. To bypass DNAT for some packets, we have to use the ACL DO_NOT_NAT rule and apply the ACL on those outside zone ports. So that they will be regularly forwarded in the hardware.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1550 @@
...
+ACL_TABLE|10
+    "stage": "INGRESS", 
+    "type": "L3", 
+    "policy_desc": "nat-acl", 
+    "ports": "Vlan2000"
+
+ACL_RULE|10|1
+    "priority": "20", 
+    "src_ip": "20.0.1.0/24",
+    "packet_action": "do_not_nat"
```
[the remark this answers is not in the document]
kirankella: @stepanblyschak Packet actions are mutually exclusive for a particular match criteria, so as far as the ACL schema is concerned the packet actions are mutually exclusive. So "do-not-nat" action is mutually exclusive of l3-forward or redirect or other actions. For eg., Packet is either l3-forwarded or not-natted or redirected...
That's the reason "packet_action" attribute in the ACL schema is re-used to configure do-not-nat value for a match.

doc/nat/nat_design_spec.md
kirankella: Mentioned as a limitation in section 3.3.1.1

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
ritalhui: @kirankella, thanks.  For each dynamic SNATP entry going outbound, there's at least one reverse flow, that needs the DNATP back to the source.  If DO_NOT_NAT was applied to the outbound flow, same would apply to the corresponding inbound flow.  Can this be supported?
For DO_NOT_NAT for static DNAT or DNATP, it could be a future item.

doc/nat/nat_design_spec.md
```diff
@@ -815,10 +818,19 @@ NAT_BINDINGS|nat1
     "nat_type": "snat"
 ```
 
-The following iptables rules are added in the kernel for each protocol type traffic for SNAT purpose:
+The iptables 'mangle' table rules are added on each L3 interface (other than loopback) to set the MARK value.
+```
+iptables -t mangle -A PREROUTING -i Ethernet20 -j MARK --set-mark 2
+iptables -t mangle -A PREROUTING -i Vlan100 -j MARK --set-mark 2
+iptables -t mangle -A POSTROUTING -o Ethernet20 -j MARK --set-mark 2
+iptables -t mangle -A POSTROUTING -o Vlan100 -j MARK --set-mark 2
```
rlhui: Should it be Ethernet28 here?

kirankella: Typo, thanks for the correction.

kirankella:
>>> For each dynamic SNATP entry going outbound, there's at least one reverse flow, that needs the DNATP back to the source. If DO_NOT_NAT was applied to the outbound flow, same would apply to the corresponding inbound flow. Can this be supported?
For DO_NOT_NAT for static DNAT or DNATP, it could be a future item.

@rlhui Yes it is supported, since there won't be any matching DNAT entries in the hardware for the reverse traffic.

doc/nat/nat_design_spec.md
kirankella: This is on similar lines as used by the neighsync application. The purpose of RESTORE_TABLE is different, it is for the state exchange between the restore_nat_entries script and the natsync application that is waiting for the conntrack entries to be restored in the kernel by the script. Following the kernel restore, the natsync application gets the dump of the conntrack entries from the kernel and then uses the warmrestartassist helper class to complete the reconciliation and set the reconciliation state in the WARM_RESTART_TABLE.

doc/nat/nat_design_spec.md
```diff
@@ -0,0 +1,1399 @@
...
+ACL_TABLE|10
+    "stage": "INGRESS", 
+    "type": "L3", 
+    "policy_desc": "nat-acl", 
+    "ports": "Vlan2000"
+
+ACL_RULE|10|1
+    "priority": "20", 
+    "src_ip": "20.0.1.0/24",
+    "packet_action": "do_not_nat"
```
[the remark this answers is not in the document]
kirankella: This acl action "SAI_ACL_ENTRY_ATTR_ACTION_NO_NAT" (defined in SAI) is also passed by AclOrch like for other actions to SAI.
...
attr.id = SAI_ACL_ENTRY_ATTR_ACTION_NO_NAT;
attr.value.booldata = true;
...
sai_acl_api->create_acl_entry(..)

doc/nat/nat_design_spec.md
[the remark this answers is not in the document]
marian-pritsak: OK

doc/nat/nat_design_spec.md
rlhui: @kirenkella, 
I see the following can happen with DNAT:
If the outside/public IP address to use is a local interface IP address (e.g. an uplink interface IP), in SONiC NAT spec, inbound traffic must be arriving at this interface, only then corresponding rule is added to iptables in kernel and hardware DNAT entry programmed.  Otherwise, it’ll not do so.
But in SAI pipeline, such check is not present, so long there is a “DNAT Pool Prefix Lookup” table hit which indicates the DIP needs a DNAT and if DNAT lookup is a miss, pkts will be trapped to CPU.
So if inbound traffic carrying this public IP is being received at a different uplink - an interface which does not own that public IP, no hardware DNAT entry is programmed but pkts will keep coming to CPU.

Will you please document this in the HLD so that those who use interface IP as public IP is aware of this and may avoid the situation.

kirankella: @rlhui, similar to loopback interfaces, the physical interface IP if used as public IP, can also be handled. Packet destined to interface IP received on any interface in the same zone as the interface that has that IP, is DNAT'ted if trapped to CPU. Section 3.4.1.1 is updated with the same.

doc/nat/nat_design_spec.md
kirankella: @rlhui If the nexthop is not resolved after warm reboot, the DNAT entry is not added by orchagent in the h/w. The corresponding conntrack entry and the APP_DB entry are timed out and deleted.

PR #391, opened 2019-05-23
Error handling framework initial draft

sivamukka:
This document describes high level design details for Error Handling framework in SONiC.

Signed-off-by: Siva Mukka siva.mukka@broadcom.com


doc/error-handling/error_handling_design_spec.md
```diff
@@ -0,0 +1,429 @@
...
+
+In this document, the term '**object**' refers to an entry in the table OR a specific attribute in the table.
+
+# 1 Requirements Overview
+## 1.1 Functional Requirements
+
+The requirements for error handling framework are:
+
+1.1.1 Provide registration/de-registration mechanism for applications to enable/disable error notifications on a specific table. More than one application can register for notifications on a given table.
+
```
madhupalu: Can framework supports applications to register for notifications at attribute level?

doc/error-handling/error_handling_design_spec.md
```diff
@@ -0,0 +1,429 @@
...
+For some objects, a notification may be needed even when the object is successfully programmed (SAI API reports SUCCESS). In such scenarios, the framework does not store the object in ERROR_DB (to reduce memory usage), but does notify the registered applications using ERROR_DB notification channel.
+
+Defining a separate error database has the following advantages over extending existing APP_DB tables for error notifications:
+
+- Applications do not register as consumer & producer for same table, which avoids redundant notifications.
+- Flexible, not tied to any other table format.
+- Extensible to all types of errors in the system, not restricted to APP_DB definitions.
+- Efficient, as notifications are limited to failures in the DB.
+- Notification for delete failures can be supported even when corresponding objects are deleted from APP_DB.
+
```
madhupalu: I understand the new DB approach, however here is my thought, why can't we have a single DB (APP DB) and separated by namespaces? ex: configured vs applied/error in the same table so that it could be easy to maintain one table.

doc/error-handling/error_handling_design_spec.md
```diff
@@ -0,0 +1,429 @@
...
+![Framework high level design](images/error_handling_overview.png)
+
+The following is the sequence of events involved in reporting a failure notification from Syncd process to application:
+
+1. Syncd reports errors using a single notification channel to OrchAgent. Using a single notification channel ensures that order of the notifications is retained.
+2. After receiving the notification from Syncd, OrchAgent:
+   - Translates it from SAI data types to ERROR_DB data types
+   - Adds an entry in to error database. If the entry already exists, the corresponding failure code is updated.
+   - Publishes the notifications to respective error listeners.
+3. Error listener waits for the incoming notifications, filters them and invokes the application callback.
```
madhupalu: Can you describe how does Error listener filters the notifications ? what is the criteria supported? please add the use case?

doc/error-handling/error_handling_design_spec.md
```diff
@@ -0,0 +1,429 @@
...
+
+The following table describes how framework handles some of the notifications:
+
+| Previous Notification | Current Notification | Framework Action                                             |
+| --------------------- | -------------------- | ------------------------------------------------------------ |
+| Create failure        | Update failure       | Update the entry in the database and notify the registered applications |
+| Create failure        | Delete failure       | Remove the entry from database and notify the registered applications |
+| Create failure        | Update success       | Remove the entry from the database and notify the registered applications |
+| Create success        | Delete failure       | Add the entry to the database and notify the registered applications |
+| Delete failure        | Create success       | Remove the entry from the database and notify the registered applications |
```
madhupalu: Does the applications get out of order notifications from feedback loop? How to handle in the case of it? Ex: User does create/delete/create and do you expect the error feedback come in order?

doc/error-handling/error_handling_design_spec.md
santoshdoke: Notifications can only be enabled per-table. Failure status code is reported per object. For example, port table has multiple objects like MTU, admin state. Notifications can be enabled at Port table level, but not on MTU failures specifically.

doc/error-handling/error_handling_design_spec.md
santoshdoke: Can you please provide more details here? Are you suggesting that ERROR tables can be stored in APP_DB, and there is no need to create ERROR_DB? We want to avoid modifying the existing ROUTE_TABLE schema in APP_DB - error handling can optionally be disabled and retain the current behavior.

doc/error-handling/error_handling_design_spec.md
santoshdoke: Sure, we will add more details on this.

doc/error-handling/error_handling_design_spec.md
santoshdoke: Order of notifications will be preserved - because changes to APP_DB and ASIC_DB maintain the sequence. In case the same object fails multiple times, we need a unique transaction id to associate the operation and failure. To address this, we are looking at adding unique ID to each APP_DB operation and reporting the ID back as part of failure notification.

PR #398, opened 2019-06-10
Documents: Debug Framework Design Spec

anilsadineni:
* A framework that allows components to register and dump running snapshots of component internals using dump routines.
* Handle assert conditions to collect more info.

Signed-off-by: Anil Sadineni anil.sadineni@broadcom.com


doc/debug_framework_design_spec.md
```diff
@@ -0,0 +1,394 @@
...
+
+**Component Responsibilities**
+
+Components implement dump routines following below guidelines
+- Implement a function of type DumpInfoFunc.  
+  `typedef void (*DumpInfoFunc)(std::string, KeyOpFieldsValuesTuple);`
+- Within the function, dump the information into a string buffer and use a macro `SWSS_DEBUG_PRINT`.  Macro will further process the information based on intended framework action. Wrap the function that uses multiple invocations of `SWSS_DEBUG_PRINT` with `SWSS_DEBUG_PRINT_BEGIN` and `SWSS_DEBUG_PRINT_END`.
+- Handle KeyOpFieldsValuesTuple as argument. Arguments like dumpType and pass thru arguments that instruct the level of info dump ( short summary/specific info/filtered output/full info) should be handled by components. These arguments are opaque to framework. Additional optional arguments like output location and post-actions are for the framework consumption.  
+- Interpreting arguments is at discretion of components.  Components may choose to dump same info for all trigger types ignoring input arguments.
+- Dump routines registered with framework should be thread safe. Necessary synchronization needs to be ensured.
```
lguohan: orchagent is single thread application, it has no lock. how are you going to make the dump routine to be thread safe?

doc/debug_framework_design_spec.md
```diff
@@ -0,0 +1,394 @@
...
+The below activity is delegated to the component
+- Optional creation of a swss::Selectable object. 
+- Create a subscriber instance to a table/channel in APPL_DB in RedisDB for receiving triggers.
+- Optionally create a thread to wake on triggers.
+- Invoke component specific dump routine on receipt of a trigger.
+- Update Redis DB to indicate callback action completion.
+
+
+**Implementation brief**  
+Framework will be implemented in C++ as a class in SWSS namespace.
```
lguohan: how to open source project which are not written in C++, for example dhcp relay which is written in C?

doc/debug_framework_design_spec.md
```diff
@@ -0,0 +1,394 @@
...
+
+## 3.3 CLI
+
+### 3.3.1 Show Commands
+
+|Syntax                         | Description                                                                   |
+|-------------------------------|-------------------------------------------------------------------------------|
+|show debug all                 | This command will invoke dump routines for all components with default action |
+|show debug < component >       | This command will invoke dump routine for specific component                  |
+|show debug actions < options > | This command will display the configured framework actions                    |
```
lguohan: how do we know all the component that has such debug capability? is there a command for that?

doc/debug_framework_design_spec.md
```diff
@@ -0,0 +1,394 @@
...
+### 3.3.1 Show Commands
+
+|Syntax                         | Description                                                                   |
+|-------------------------------|-------------------------------------------------------------------------------|
+|show debug all                 | This command will invoke dump routines for all components with default action |
+|show debug < component >       | This command will invoke dump routine for specific component                  |
+|show debug actions < options > | This command will display the configured framework actions                    |  
+
+### 3.3.2 Debug/Error Interface counters
+show interfaces pktdrop is added to display debug/error counters for all interfaces.
```
lguohan: how is this different from the portstat command line currently we have?

doc/debug_framework_design_spec.md
anilsadineni: component owner trying to debug his component will enter component name so that the trigger to dump will be specific to the component specified.  If the admin is trying to collect debug information and has no knowledge of registered components he may consider show debug all.   

If its an engineer but not component owner who is trying to figure out the registered components he can query Redis DB for the registered component list using redis-cli and then proceed with the trigger.

doc/debug_framework_design_spec.md
anilsadineni: Good question.  For python programs SWIG will help us.  Check pyext folder for interface file definition. 

For C,  yes we need to write wrapper functions.

doc/debug_framework_design_spec.md
anilsadineni: If any application is not thread safe we recommend to use option #2 to link with framework.  In our pilot implementation for dump routines of  we chose option #2 for orchagent

doc/debug_framework_design_spec.md
anilsadineni: The initial thought was to fetch more errors and counters from the asic but this is not a framework item and very asic dependent hence dropped it.  Shall remove in next version.

PR #406, opened 2019-06-17
PDK PDDF HLD PR request

fk410167:
This document describes the Platform Driver Development Framework (PDDF) which optimises platform driver and SONiC plugin development using a data-driven model.


doc/platform/brcm_pdk_pddf.md
```diff
@@ -0,0 +1,1170 @@
...
+| PDK                      | Platform Development Kit              |
+| SAI                      | Switch Abstraction Interface |
+| PSU                      | Power Supply Unit |
+| I2C                      | Inter-integrated Circuit communication protocol |
+| SysFS                    | Virtual File System provided by the Linux Kernel |
+
+
+## 1 Requirements Overview
+SONiC OS is portable across different network devices with supported ASIC via Switch Abstraction Interface (SAI). These devices primarily differ in the way various device specific hardware components are accessed, and thus require custom device drivers and python plugins. Each platform vendor implements these custom device drivers and plugins. The feature requirement is to support a SONiC platform driver development framework to enable rapid development of custom device drivers and plugins.
+
```
keboliu: Currently SONiC is the process to replace platform plugin with a newly designed platform APIs,  the target is to switch to  new platform API in 201908 release, so should also considering to support platform API for the coming transform.  The design doc of new platform API is here:  https://github.com/Azure/SONiC/pull/285  the new platform base class is defined here: https://github.com/Azure/sonic-platform-common/tree/master/sonic_platform_base

fk410167: Hi Kebo Liu,

Yes, we are aware of the new platform API in August, 2019 release. We plan to support new platform base class in PDDF.

-Fuzail

stepanblyschak:
@fk410167 assuming platform vendor will use its own drivers and not using generic drivers with JSON descriptor files approach, how can vendor use generic plugin/platform API implementation by providing sysfs path or symlink to sysfs and attributes list?
From the picture in https://github.com/Azure/SONiC/blob/6d59a095d6bb9dcb3ecb18e8dacec0ce44e886ad/doc/platform/brcm_pdk_pddf.md#33-generic-plugin-design it is assumed sysfs path is auto generated based on json description.

fk410167:
> @fk410167 assuming platform vendor will use its own drivers and not using generic drivers with JSON descriptor files approach, how can vendor use generic plugin/platform API implementation by providing sysfs path or symlink to sysfs and attributes list?
> From the picture in https://github.com/Azure/SONiC/blob/6d59a095d6bb9dcb3ecb18e8dacec0ce44e886ad/doc/platform/brcm_pdk_pddf.md#33-generic-plugin-design it is assumed sysfs path is auto generated based on json description.

Hi stepanblyschak,
The SysFS path derivation is solely based on I2C topology info and attribute name in JSON file.  In the scenario you mentioned, the vendor can still use the generic plugins/CLI utils given,

1. User has defined the attribute lists for the non-pddf drivers in JSON descriptor file. 
2. The topology data, parent_bus, virt_bus, parent device, dev_addr, dev_type etc are provided correctly in JSON as per your custom driver implementation.
3. Copy the generic PDDF plugins to your platform plugins folder. 

-Fuzail

doc/platform/brcm_pdk_pddf.md
jleveque: @fk410167: Do you have plans to update this document to reflect the new platform API? Vendors are currently in the process of migrating over to the new API. At this point in time, this document should reflect the new API rather than the old plugins.

fk410167: Yes we have already started to move the whole framework to use 2.0 platform APIs. We will modify the HLD accordingly.

PR #408, opened 2019-06-19
VRF HLD version 1.1

preetham-singh:
Updating VRF HLD to create version 1.1


jeffreyzfzeng:
In the VRF diagram, in each VRF, site-1 and site-2 are repeated. Should they be site-3 and site-4? Or it means a multi-homing configuration? Please clarify.

doc/sonic-vrf-hld-v1.0.md
```diff
@@ -81,10 +86,48 @@ The fallback feature which defined by RFC4364 is very useful for specified VRF u
...
+## Functionality
+### Target Deployment Use Cases
+Virtual Routing and Forwarding is used to achieve network virtualization and traffic separation over on a shared network infrastructure in Enterprise and DC delpoyments.
+
+![Deployment use case](https://github.com/preetham-singh/SONiC-1/blob/master/images/vrf_hld/Multi-VRF_Deployment.png "Figure 1: Multi VRF Deployment use case")
+__Figure 1: Multi VRF Deployment use case__
+
+Above figure depicts typical customer deployment where multiple Customer facing devices are connected to Provider edge routers.
+With such deployment Provider Edge routers associate each input interface with one VRF instance.
+Note that multiple input interfaces can be associated to a VRF instance. This input interface can be Physical interface, Port-channel or Ve interface.
```
ben-gale: Suggest saying "SVI interface" instead of "Ve interface" - more widely recognized term

doc/sonic-vrf-hld-v1.0.md
```diff
@@ -81,10 +86,48 @@ The fallback feature which defined by RFC4364 is very useful for specified VRF u
...
+
+### Functional Description
+
+Multi-VRF is the ability to maintain multiple "Virtual Routing and Forwarding"(VRF) tables on the same Provider Edge router.
```
ben-gale: Pls define the terms "Provider Edge Router" and "Customer Edge Router", and/or show them in figure 1.

doc/sonic-vrf-hld-v1.0.md
```diff
@@ -81,10 +86,48 @@ The fallback feature which defined by RFC4364 is very useful for specified VRF u
...
+Multi-VRF uses multiple instances of a routing protocol such as BGP to exchange routing information among peer provider edge routers.
```
ben-gale: What does "instance" mean in this context? Is it really multiple instances of the BGP Daemon in FRR? I doubt it. I suspect there's a single instance of BGP running, but it's VRF-aware, so it knows which interfaces/sessions map to which VRFs, and can process the RD field and RT attribute.

doc/sonic-vrf-hld-v1.0.md
```diff
@@ -81,10 +86,48 @@ The fallback feature which defined by RFC4364 is very useful for specified VRF u
...
+The Multi-VRF capable Provider Edge router maps an input customer interface to a unique VRF instance. Provider Edge router maintains unique VRF table for each VRF instance on that Provider Edge router.
+
+Multi-VRF routers communicate with one another by exchanging route information in the VRF table with the neighboring Provider Edge router.
+This exchange of information among the Provider Edge routers is done using routing protocol like BGP.
+Customers connect to Provider Edge routers in the network using Customer Edge routers as shown in Figure 1. 
+
+Due to this overlapping address spaces can be maintained among the different VRF instances.
+
+FRR's BGP implementation is capable of running multiple autonomous systems (AS) at once. Each configured AS is associated with a VRF instance. The  SONiC VRF implementation will enable this capability of FRR BGP in SONiC.
```
ben-gale: Please describe the use case for this. It seems like this would be to put different VRFs under different administrative control, which seems like an unlikely use case for SONiC, no?

doc/sonic-vrf-hld-v1.0.md
```diff
@@ -81,10 +86,48 @@ The fallback feature which defined by RFC4364 is very useful for specified VRF u
...
+
+#### VRF route leak support:
+
+VRF route leak is a case where route and nexthops are in different VRF.
+
+VRF route leak can be achieved via both Static or Dynamic approach.
+
+In Static VRF route leak, FRR cli can be used where user can specify nexthop IP along with nexthop VRF, where the nexthop is reachable.
```
ben-gale: Pls say "CLI" not "cli", and do so consistently throughout.

ben-gale: Should "where the nexthop is reachable" say "where the nexthop is reachable through a different VRF"?

doc/sonic-vrf-hld-v1.0.md
```diff
@@ -81,10 +86,48 @@ The fallback feature which defined by RFC4364 is very useful for specified VRF u
...
+
+In Dynamic VRF route leak, Route maps can be used to import routes between VRFs. Prefix lists are used to match route prefixes.
```
ben-gale: Please explain how Prefix lists are used. Does this mean that, if a prefix list if defined for a given VRF, then only prefixes in that list are imported?

doc/sonic-vrf-hld-v1.0.md
```diff
@@ -81,10 +86,48 @@ The fallback feature which defined by RFC4364 is very useful for specified VRF u
...
+Leaked routes will be deleted when source route is deleted.
+
+Note: Since static leak route requires nexthop vrf name, VRF route leak from user-defined-vrf to Global vrf will not be supported.
```
ben-gale: Use "VRF" not "vrf" - use consistently throughout

ben-gale: What does this mean use-case wise, and why this is OK. It seems like it means that a configured VRF can use a nexthop in the global VRF, but not the other way around. This seems like it would hit the main use case, where common services (e.g. internet) are defined in the global table and available to user VRFs. Please clarify in the doc.

doc/sonic-vrf-hld-v1.0.md
```diff
@@ -545,22 +588,19 @@ $ config vrf add <vrf_name>
 $ config vrf del <vrf_name>
 
 //bind an interface to a VRF
-$ config interface <interface_name> vrf bind <vrf_name>
+$ config interface bind <interface_name> <vrf_name>
```
prsunny: My suggestion is to have it as  `config interface vrf add <interface_name> <vrf_name>`. When user just give "`config interface`" the options are listed as `ip, pfc` etc and "`bind`" may not be good choice whereas "`vrf`" can be self explanatory.

lguohan:
remove the version number in the file name.

nikos-github:
Regarding the comment I made in the community meeting last Tuesday on changing config_db.json when it comes to interfaces belonging in the global table and also mentioned in here https://github.com/Azure/SONiC/blob/master/doc/sonic-vrf-hld-v1.0.md

```"Ethernet2":{}, // it means this interface belongs to global vrf. It is necessary even user doesnt use vrf.```

File: doc/sonic-vrf-hld-v1.0.md [hunk 9]
```diff
@@ -238,7 +277,7 @@ Add vrf-binding information in config_db.json file.
...
+Logically IP address configuration must be processed after interface binding to vrf is processed. In intfmgrd/intfOrch process intf-bind-vrf event must be handled before IP address event. So interface-name entry in config_db.json is necessary even though user doesn't use VRF feature. e.g. `"Ethernet2":{}` in the above example configuration. For version upgrade compatibility we need to add a script, this script will convert old config_db.json to new config_db.json at bootup automatically, then the new config_db.json would contain the interface-name entry for interfaces associated in the global VRF table.
```
nikos-github: @preetham-singh @ben-gale @prsunny Regarding the comment I made in the community meeting last Tuesday on changing config_db.json when it comes to interfaces belonging in the global table and also mentioned in here https://github.com/Azure/SONiC/blob/master/doc/sonic-vrf-hld-v1.0.md

"Ethernet2":{}, // it means this interface belongs to global vrf. It is necessary even user doesnt use vrf.

In order to preserve backwards compatibility with existing config_db.json files and deployments where the config_db.json is checked for deviations, when saving the contents of config db from redis back into config_db.json, default entries like the the entry above can be removed. Basically it shouldn't be a requirement for the user to either provide that in their config_db.json or have this saved back.

Such approach is being followed already by some BGP cmds I introduced and implemented.

doc/sonic-vrf-hld-v1.0.md
preetham-singh: This statement no longer holds good, hence removed same.

doc/sonic-vrf-hld-v1.0.md
preetham-singh: Updated CLI in HLD

doc/sonic-vrf-hld-v1.0.md
prsunny: Based on discussion on this topic in the Sonic user group, it was agreed upon that in certain cases where the feature demands it, it is expected to have a `config_db` change as long as there is a provision to migrate the old `config_db.json` to the new image version to ensure backward compatibility. This means, after a save of the migrated `config_db`, the saved json file could be different from the old one and the expectation is that the monitoring application must accommodate to the new schema defined in the migrated image version.

ben-gale: Prince - what about the reverse direction (i.e. downgrade)? I assume the user is expected to manually 'repair' their configuration? 

Is there really no way that this could be handled differently?

prsunny: For the downgrade, the user has to manually repair. In general, it is not an automated support. W.r.t VRF, if you are asking whether it could handled differently, yes it is possible but with the complexities of supporting move feature. Moreover, some of the code to support the config_db approach like the db_migrator, interface changes etc were already in-place for VNet feature. That's why we want to take this path.

PR #409, opened 2019-06-20
add test plan for SONIC vrf feature #392

shine4chen:
It includes ansible test cases and virtual switch test cases.

Signed-off-by: shine.chen <shine.chen@nephosinc.com>


preetham-singh:
Can you please add test cases for below features in VRF context:
1. Test Link Local IPv6 address in VRF, ping, forwarding & BGP neighborship.
2. In Loopback interface test, check loopback interface IP can be used as a source in ping/ping6, BGP/BGP+ update-source and router-id for the VRF.
3. BGP multihop peer over IGP(static) which has ECMP in VRF. Failover of a nhop in a ECMP should not impact BGP-TCP connection in kernel. This will ensure control plane ECMP in VRF.

prsunny:
At a higher level, the test coverage looks good to me. Consider adding few cases (routes, nexthops) to check CRM resources with VRF. You can check the values before and after config. Also suggest to combine test cases wherever possible instead of separate test cases for each scenario. Scalability can be given as a configurable option.

shine4chen:
@preetham-singh Thanks for your suggestion.
1. In the existing sonic there are some issues about local-link address. So I suggest we can add local-link address test case after fixing these issues.
2. Yes, it is the default behavior of FRR BGP template session. If no explicit configured BGP session always chooses the first loop-back device address on the specified vrf as its route-id and chooses the second loop-back device address as update-source. We can add a test case to cover loop-back address as ping/ping6 source.
3. Control plane tcp connection is handled by linux kernel either in global or specified vrf. I think it may beyond vrf test scope.

shine4chen:
@prsunny  agree, we can rerun some crm ansible test case in vrf environment.

dawnbeauty:
@prsunny @shine4chen Currently we do not add testcases about CRM with VRF.  Since i dont think there is any particularly difference of CRM usage between global and VRF. If CRM ansible testcases passed (which means CRM works as normal without VRF), then it also means CRM works with VRF.

vboykox:
@prsunny @shine4chen @lguohan  Hi
> Depending on the test cases PTF will verify the packet is arrived or dropped. For vrf 
> "src_mac" option test, PTF will analyze ip packet dst_mac after L3 forwarding 
> through vrf and do L3 forwarding only when ip packet's dst_mac is matched with 
> configed vrf "src_mac".

It looks that this contradicts the current SONiC behaviour - RIFs' src mac do not
get updated on VRF's src mac change.
Please see this issue:
https://github.com/Azure/sonic-swss/issues/1833

So what behaviour for VRF/RIF src mac update is expected here by SONiC design?
SAI does not impose a requirement to "auto-update" RIF's src macs on VRF change.
I guess either the auto-update should be handled by SONiC, or testplan and
related test should be changed appropriately to not count on it.

Thanks in advance

vboykox:
> @prsunny @shine4chen @lguohan Hi
> 
> > Depending on the test cases PTF will verify the packet is arrived or dropped. For vrf
> > "src_mac" option test, PTF will analyze ip packet dst_mac after L3 forwarding
> > through vrf and do L3 forwarding only when ip packet's dst_mac is matched with
> > configed vrf "src_mac".
> 
> It looks that this contradicts the current SONiC behaviour - RIFs' src mac do not get updated on VRF's src mac change. Please see this issue: [Azure/sonic-swss#1833](https://github.com/Azure/sonic-swss/issues/1833)
> 
> So what behaviour for VRF/RIF src mac update is expected here by SONiC design? SAI does not impose a requirement to "auto-update" RIF's src macs on VRF change. I guess either the auto-update should be handled by SONiC, or testplan and related test should be changed appropriately to not count on it.
> 
> Thanks in advance

@prsunny kindly reminder

shine4chen:
@vboykox I prefer to revise the related test case.

PR #420, opened 2019-07-10
Sub port interface high level design

wendani:
Signed-off-by: Wenda Ni <wenni@microsoft.com>


wendani:
@jipanyang

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,475 @@
...
+| 0.1 | 07/01/2019  | Wenda Ni           | Initial version                   |
+
+# Scope
+A sub port interface is a logical interface that can be created on a physical port or a port channel.
+A sub port interface serves as an interface to either a .1D bridge or a VRF, but not both.
+This design focuses on the use case of creating a sub port interface on a physical port or a port channel and using it as a router interface to a VRF.
+
+![](https://github.com/wendani/SONiC/blob/sub_port_master/images/sub_interface_hld/sub_intf_rif.png)
+
+Multiple L3 sub port interfaces, each characterized by a VLAN id in the 802.1q tag, can be created on a physical port or a port channel.
```
qiluo-msft: previous `.1D`, now `802.1q`. Could you clarify? This line is difficult to understand.

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,475 @@
...
+
+mtu of a sub port interface should be no greater than that of the parent physical port or port channel. In the case field "mtu" is absent in the config_db.json file, a sub port interface inherits the mtu from its parent physical port or port channel.
+
+Example configuration:
+```
+"DOT1Q_INTERFACE": {
+    "Ethernet64.10": {
+        "mtu" : "9100"
+    },
+    "Ethernet64.10|192.168.0.1/21": {}
```
qiluo-msft: Is it better to add ipv6 into the example?

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,475 @@
...
+DOT1Q_INTERFACE|Ethernet64.10
+    "mtu" : "9100"
+
+DOT1Q_INTERFACE|Ethernet64.10|192.168.0.1/21
+    "NULL" : "NULL"
+```
+
+## 2.2 APPL_DB
+```
+INTF_TABLE:{{ port_name }}.{{ vlan_id }}
```
qiluo-msft: Are you reusing an old table? If yes, is it possible that old schema and new schema coexist in one switch?

qiluo-msft: If coexist, is it possible to normalize the schema?

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,475 @@
...
+------------------  -------  -----  ------  ------  -------  -------------------
+     Ethernet64.10     100G   9100      10      up       up  dot1q-encapsulation
+```
+
+# 5 Warm reboot support
+There is no special runtime state that needs to be kept for sub port interfaces.
+This said, current warm reboot infrastructure shall support sub port interfaces naturally without the need for additional extension.
+This is confirmed by preliminary trials on a Mellanox device.
+
+# 6 Uinit test
```
qiluo-msft: Unit test?

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,475 @@
...
+| Verify that sub port interface configuration is removed from APPL_DB INTF_TABLE by Intfmgrd            |
+| Verify that sub port interface state ok is removed from STATE_DB by Intfmgrd                           |
+
+# 7 Scalability
+Scalability is ASIC-dependent.
+We enforce a minimum scalability requirement on the number of sub port interfaces that shall be supported on a SONiC switch.
+
+| Name                                                              | Scaling value             |
+|-------------------------------------------------------------------|---------------------------|
+| Number of sub port interfaces per phyical port or port channel    | 250                       |
```
qiluo-msft: How do you choose this number? Is 256 more natural? How about other vendors?

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,475 @@
...
+| Number of sub port interfaces per switch                          | 750                       |
```
qiluo-msft: Why limit this? We have fixed number of port, and max `Number of sub port interfaces per phyical port or port channel` above, so we can get it by multiplication.

doc/sonic-sub-port-intf-hld.md
qiluo-msft: phyical -> physical?

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,475 @@
...
+},
+
+"VLAN": {
+    "Vlan10": {
+        "vlanid" : "10"
+    }
+},
+```
+
+### 2.1.2 CONFIG_DB
```
qiluo-msft: Is there an concept of port alias in this usage? For example Ethernet64.10 -> etp16.10

doc/sonic-sub-port-intf-hld.md
wendani: Sub port interfaces may not be uniform across physical port and/or port channels. This is a number referred from Vendor manual. https://github.com/Azure/SONiC/pull/420#discussion_r303186368

doc/sonic-sub-port-intf-hld.md
wendani: This is the number referred from Vendor manual. https://github.com/Azure/SONiC/pull/420#discussion_r303186270

doc/sonic-sub-port-intf-hld.md
wendani: The tag is called 802.1q tag. Just a name to refer to the 32-bit fields in the Ethernet frame.

doc/sonic-sub-port-intf-hld.md
wendani: It will co-exist. There is no need to normalization as both follows the naming convention.

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,483 @@
...
+### 2.5.1 Host sub port interfaces
+
+We use iproute2 package to manage host sub port interfaces.
+Specifically, we use `ip link add link <parent_port_name> name <subif_name> type vlan id <vlan_id>` to create a host sub port interface.
+This command implies the dependancy that a parent host interface must be created before the creation of a host sub port interface.
+At creation, a host sub port interface is always set to admin status up.
+
+Example:
+```
+ip link add link Ethernet64 name Ethernet64.10 type vlan id 10
```
tieguoevan: admin status of Ethernet64.10 is coupled with Ethernet64.
If Ethernet64.10 and Ethernet64 are both shutdown, startup Ethernet64 will also startup Ethernet64.10.
Is it better as follows:
```
ip link add br-Ethernet64 up type bridge
ip link add link br-Ethernet64 name Ethernet64.10 type vlan id 10
```

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,483 @@
...
+We enforce a minimum scalability requirement on the number of sub port interfaces that shall be supported on a SONiC switch.
+
+| Name                                                              | Scaling value             |
+|-------------------------------------------------------------------|---------------------------|
+| Number of sub port interfaces per physical port or port channel   | 250                       |
+| Number of sub port interfaces per switch                          | 750                       |
+
+# 8 Port channel renaming
+Linux has the limitation of 16 characters on an interface name.
+For sub port interface use cases on port channels, we need to redesign the current naming convention for port channels (PortChannelXXXX, 15 characters) to take shorter names (such as, PoXXXX, 6 characters).
```
tieguoevan: Physical port such as Ethernet128.1024 will exceed limitation of 16 characters in Linux kernel. It is better to take physical port into account for supporting 128-ports device.

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,483 @@
...
+fpmsyncd should receive route add/del updates on sub port interfaces from zebra over TCP socket port 2620. These updates are received in the format of netlink messages.
+neighsyncd should receive neigh add/del netlink messages on sub port interfaces from its subscription to the neighbor event notification group RTNLGRP_NEIGH.
+
+Internally, a sub port interface is represented as a Port object to be perceived seamlessly by NeighOrch and RouteOrch to create neighor and route entries, respectively, on it.
+
+
+# 3 Event flow diagrams
+## 3.1 Sub port interface creation
+![](https://github.com/wendani/SONiC/blob/sub_port_master/images/sub_interface_hld/sub_intf_creation_flow.png)
+
```
rck-innovium: In the section title 'SCOPE', it is mentioned that each sub port interface is considered to stay in a separate bridge domain and is useful. Given this, Is vlan-10 object creation by intfOrch necessary? If yes, it needs to be mentioned how 

1)  vlan-10 will be re-used when another support with vlan-10 is created 

2) when will vlan-10 be deleted.

In the current call sequence in Sec 3.3, vlan-10 object is lingering around after the subport has been deleted.

I think this complication arises from the way SAI_ROUTER_INTERFACE_ATTR_VLAN_ID is defined in the SAI spec. Since the VLANs mentioned in the support configuration have only port namespace, we really don't need a global VLAN object and just a uint16_t type should be sufficient.

    /**
     * @brief Associated Vlan
     *
     * @type sai_object_id_t <<<<<<<<
     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
     * @objects SAI_OBJECT_TYPE_VLAN <<<<<<<
     * @condition SAI_ROUTER_INTERFACE_ATTR_TYPE == SAI_ROUTER_INTERFACE_TYPE_VLAN or SAI_ROUTER_INTERFACE_ATTR_TYPE == SAI_ROUTER_INTERFACE_TYPE_SUB_PORT
     */
    SAI_ROUTER_INTERFACE_ATTR_VLAN_ID,

doc/sonic-sub-port-intf-hld.md
prsunny: I think it is ok to be coupled with Ethernet64 considering how it is configured in H/W. This is a sub-interface of Ethernet64 with a vlan tag and with above commands it is not related to Ethernet64. Not sure how the packet processing happens if we have above configuration and packet comes on Ethernet64 tagged with vlan 10.

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,493 @@
...
+## 3.2 Sub port interface runtime admin status change
```
prsunny: The diagram is showing `mtu `change but the heading states `admin `status change. You may want to modify the diagram.

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,493 @@
...
+![](https://github.com/wendani/SONiC/blob/sub_port_master/images/sub_interface_hld/sub_intf_set_mtu_flow.png)
+
+## 3.3 Sub port interface removal
+![](https://github.com/wendani/SONiC/blob/sub_port_master/images/sub_interface_hld/sub_intf_removal_flow.png)
+
+# 4 CLIs
+## 4.1 Config commands
```
prsunny: IMO, this can be under the current `config interface` CLI adding an option `subinterface` instead of a new set of command.

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,493 @@
...
+```
+Usage: config subinterface add [OPTIONS] <sub_port_interface_name>
+```
+```
+Usage: config subinterface del [OPTIONS] <sub_port_interface_name>
+```
+
+Once a sub port interface is added, `add` and `del` ip address commands are supported on it.
+```
+Usage: config subinterface ip [OPTIONS] COMMAND [ARGS]...
```
prsunny: Please use existing CLI to configure IP address for an interface. It also takes care of any default-VRF handling. Suggest to have one CLI to configure ip address for an interface.

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,493 @@
...
+                 / [ *4( h16 ":" ) h16 ] "::"              ls32
+                 / [ *5( h16 ":" ) h16 ] "::"              h16
+                 / [ *6( h16 ":" ) h16 ] "::"
+h16             = 1*4HEXDIG
+ls32            = ( h16 ":" h16 ) / IPv4address
+```
+
+Example:
+```
+SUB_INTERFACE|Ethernet64.10
```
jipanyang: Would it make more sense to rename the table to SUBPORT?

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,493 @@
...
+
+Once the host sub port interfaces are properly set, route and neighbor subsystems should function properly on sub port interfaces.
+fpmsyncd should receive route add/del updates on sub port interfaces from zebra over TCP socket port 2620. These updates are received in the format of netlink messages.
+neighsyncd should receive neigh add/del netlink messages on sub port interfaces from its subscription to the neighbor event notification group RTNLGRP_NEIGH.
+
+Internally, a sub port interface is represented as a Port object to be perceived seamlessly by NeighOrch and RouteOrch to create neighor and route entries, respectively, on it.
+
+
+# 3 Event flow diagrams
+## 3.1 Sub port interface creation
```
jipanyang: Table name in the diagrams seems outdated, dot1q interface?

doc/sonic-sub-port-intf-hld.md
```diff
@@ -105,13 +105,13 @@ For APPL_DB and STATE_DB, we do not introduce new tables for sub port interfaces
...
 ```
-A key in the DOT1Q_INTERFACE table is the name of a sub port, which consists of two sections delimited by a "." (symbol dot). The section before the dot is the name of the parent physical port or port channel. The section after the dot is the dot1q encapsulation vlan id.
+A key in the SUB_INTERFACE table is the name of a sub port, which consists of two sections delimited by a "." (symbol dot). The section before the dot is the name of the parent physical port or port channel. The section after the dot is the dot1q encapsulation vlan id.
 
 mtu of a sub port interface should be no greater than that of the parent physical port or port channel. In the case field "mtu" is absent in the config_db.json file, a sub port interface inherits the mtu from its parent physical port or port channel.
 
 Example configuration:
 ```
-"DOT1Q_INTERFACE": {
+"SUB_INTERFACE": {
```
lguohan: better to be dot1q_subinterface  as we could have other typwa of subinterface

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,499 @@
...
+        "admin_status" : "{{ adminstatus }}"
+    },
+    "{{ port_name }}.{{ vlan_id }}|{{ ip_prefix }}": {}
+},
+
+"VLAN": {
+    "Vlan{{ vlan_id }}": {
+        "vlanid" : "{{ vlan_id }}"
+    }
+},
```
lguohan: I think we should not create clan since the clan id in the subject interface is just a tag.  it does not need vlan object to be created

doc/sonic-sub-port-intf-hld.md
lguohan: support is more generic than sub interface.  subport cab be non l3 mode which is out of the scope of this proposal

doc/sonic-sub-port-intf-hld.md
venkatmahalingam: Yes @lguohan, but I got the reply from @wendani in the WG call that VLAN object creation is required for some reason, @wendani if not required, please remove the vlan-id configuration for sub-port feature.

doc/sonic-sub-port-intf-hld.md
venkatmahalingam: How about VLAN_SUB_INTERFACE? do we need DOT1Q for some reason?

lguohan: sounds good

doc/sonic-sub-port-intf-hld.md
venkatmahalingam: Since we have an explicit admin_status config for sub_intf, it is better to not to inherit the admin status from parent intf, if user has not configured the admin_status, we should use the default admin_status and when the packet comes on parent intf with VLAN 10 and the corresponding sub_intf created for VLAN 10 is admin down, kernel is expected to drop the packet which is expected, but having admin_status on sub_intf is of no significance, I think, we can always inherit admin_status from parent intf as Prince mentioned.

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,499 @@
...
+
+This design does NOT cover the case of creating and using sub port as a bridge port to a .1D bridge shown in Fig. 2.
+
+![](https://github.com/wendani/SONiC/blob/sub_port_master/images/sub_interface_hld/sub_intf_bridge_port.png "Fig. 2: Sub port bridge port (out of scope of this design")
+__Fig. 2: Sub port bridge port__
+
+# Acronyms
+| Acronym                  | Description                                |
+|--------------------------|--------------------------------------------|
+| VRF                      | Virtual route forward                      |
```
venkatmahalingam: Please replace "Virtual route forward" with "Virtual Routing and Forwarding"

doc/sonic-sub-port-intf-hld.md
```diff
@@ -0,0 +1,499 @@
...
+
+Manage the life cycle of a sub port interface created on a physical port or a port channel and used as a router interface to a VRF:
+* Creation with the specified dot1q vlan id encapsulation and optional mtu
+* Runtime mtu change
+* Removal
+
+A sub port interface shall support the following features:
+* L3 forwarding (both unicast and multicast)
+* BGP
+* ARP
```
venkatmahalingam: ARP and neighbors (IPv6)??

doc/sonic-sub-port-intf-hld.md
wendani: This is what the current SAI specification requires, and how vendor SAI is implemented if already supporting sub port interface. https://github.com/Mellanox/SAI-Implementation/blob/master/mlnx_sai/src/mlnx_sai_rif.c#L636

What Guohan suggests is to change the SAI spec before we can make any update accordingly here.

lguohan: the sai spec is not correct, can you change that?  @wendani

doc/sonic-sub-port-intf-hld.md
wendani: Not in CONFIG_DB. But in CLI, we support interface config using port alias, and the internally CLI will map port alias to default port name to set the CONFIG_DB.

wendani:
🚢

wendani:
Yes. sub-interface <Port/PortChannel> can coexist with Vlan interface at same time. You can stitch Fig. 3 and Fig. 1 together. Looking right side from VRF0 in Fig. 3, what you get is Fig. 1. https://github.com/Azure/SONiC/pull/420#commitcomment-35249048

PR #422, opened 2019-07-12
Mclag ansible test plan

shine4chen:
This PR is Mclag test plan


chenkelly:
Hi
We install teamd on ptf docker, but cannot execute it.
Has anyone experienced this? 
How to fix?
Please advise us your suggestion.
Thanks very much.

root@9a5d2e8482f3:/home# teamd 
This program is not intended to be run as root. Failed to create team device. teamd_init() failed. Failed: Operation not permitted

PR #424, opened 2019-07-16
bgp error handling changes for review on master branch

sudhanshukumar22:
This document gives an overview of how to handle hardware install related errors in BGP. Based on community feedback, I will add an updated document with the following features:
1. Add a knob for disabling and enabling error handling. If the error-handling is disabled, FRR to FPM communication will work as before. The hardware-related errors will be dropped in fpmsyncd and will not be sent to Zebra,
2. When the error handling feature is enabled, we should wait in BGP for installing the route in hardware. Once the route is successfully programmed in hardware, we will advertise the route to its peers.


doc/bgp_error_handling/BGP_Route_Error_Handling_Arlo.md
```diff
@@ -0,0 +1,241 @@
...
+The Orchagent reads the route from APP_DB, creates new resources like nexthop or nexthop group Id and installs the route in ASIC_DB. The syncd triggers the appropriate SAI API and route is installed in hardware. The CRM manages the count of critical resources allocated by orchagent through SAI API.
+Due to resource allocations failures in hardware, SAI API calls can fail and these failures should be notified to Zebra and BGP.
+On learning the prefix, BGP can immediately advertise the prefix to its neighbors. However, if the error-handling feature is enabled, BGP waits for success notification from hardware before advertising the same to its peers. If the hardware returns error, the routes are not advertised to the peers. 
+
+## 1.1 Functional Requirements
+
+ 
+
+ 1. BGP should not advertise the routes which have failed to be installed in hardware.
+ 1. BGP should mark the routes which are not installed in hardware as  "FIB-install pending" routes in its RIB-IN table.
```
dev-aviz: what is the reconciliation on process/daemon crashes/restarts? how to reconcile the routes? Please list down the scenarios and the observations?

doc/bgp_error_handling/BGP_Route_Error_Handling_Arlo.md
```diff
@@ -0,0 +1,241 @@
...
+## 2.1 Target Deployment Use Cases
+
+## 2.2 Functional Description
+Refer to section 1.1
+
+# 3 Design
+## 3.1 Overview
+On enabling the error-handling feature, fpmsyncd subscribes to the changes in the ERROR_ROUTE_TABLE entries. Whenever the error status in ERROR_ROUTE_TABLE is updated, fpmsyncd is notified. It then sends a message to Zebra's routing table to take appropriate action.
+Zebra should lookup the route and mark it as not installed in hardware. It should create a route-netlink message to withdraw this state in kernel. Also, it sends message to the source protocol for the route (BGP). 
+BGP marks the route as not installed in hardware and does not advertise the route to its peers. It should mark the route in RIB-IN as not installed in hardware and remove it from RIB-OUT list, if any.
```
dev-aviz: Can this feature turn-off on demand ? is yes, can this affect the system stability?

doc/bgp_error_handling/BGP_Route_Error_Handling_Arlo.md
```diff
@@ -0,0 +1,241 @@
...
+
+# 6 Warm Boot Support
+
+# 7 Scalability
+
+# 8 Unit Test
+
+The UT testcases are as follows:
+
+|**Test-Case ID**|**Description**|**Status**|**Comments**|
```
dev-aviz: What is the scope of the QuickTests? Is it covered only happy paths alone? do you have numbers with non-happy path scenarios?

doc/bgp_error_handling/BGP_Route_Error_Handling_Arlo.md
sudhanshukumar22: There are 2 scenarios here. One is warm-reboot case and another is unplanned reboot (like docker restart due to cold reboot). 
During warm reboot, fpmsyncd supports syncing of existing routes in APP_DB with newly learnt routes.  After warm reboot of BGP docker, BGP sends newly learnt routes to fpmsyncd. Since warm reboot is enabled, fpmsyncd will mark the old routes and send only the newly learnt routes to APP_DB. In our case, since we do not store any error route in APP_DB, so we 
don't need to do anything here.
For e.g., before warm reboot, suppose, we had 5 routes sent by BGP to fpmsynd, out of them., 5th route failed to be installed in hardware. Zebra will delete the 5th route from kernel, APP_DB, ASIC_DB and ERROR_DB. After warm reboot, when fpmsyncd receives the 5 routes again from BGP, it will send add for the 5th route again (since 5th route was never present in APP_DB prior to warm reboot).
During unplanned reboot, for the same 5 routes, suppose, fpmsyncd crashed before processing the add failure for the 5th route. Now, all the 5 routes are present in APP_DB and ASIC_DB. Since 
the 5th route failed to be installed in hardware, it is present in ERROR_DB. When docker comes up again, if warm-reboot feature is not enabled, all 5 routes will be pushed by fpmsyncd.
If warm-reboot feature for BGP is enabled, fpmsynd will reconcile the 5 routes with those present in APP_DB. In this case, it will receive the notification for the 5th route failure from ERROR_DB. Or, fpmsyncd, during
warm-reboot should check if any add failure entry is present in ERROR_DB and should process it by sending delete to the APP_DB.

doc/bgp_error_handling/BGP_Route_Error_Handling_Arlo.md
sudhanshukumar22: This feature can be turned off on demand. But it can affect the system stability. When the config was turned on, there may be some routes in BGP, for which, it is waiting for update from hardware. Suddenly, the feature is turned off. In this case, we will unsubscribe from ERROR_DB and will no longer receive any notifications from hardware. Hence, some of the routes may not receive any notification from hardware. 
It is recommended to restart the BGP docker when the config state is changed to disable from enable. By default, this config is disabled. If the config is changed from disable to enable, we do not need to restart the docker. But the feature will be affecting only those routes which will be learnt after enabling the feature.

doc/bgp_error_handling/BGP_Route_Error_Handling_Arlo.md
sudhanshukumar22: Yes, it covers the happy paths alone. I am still testing the feature using quicktest and will cover the non-happy paths also. From the BGP perspective, for routes that have failed to be installed in hardware, the route will not be sent to its peers and traffic will not be received at the receivers.

prsunny:
Currently orchagent crashes when there is a table full error, i.e when SAI returns SAI_STATUS_TABLE_FULL. With the error-handling, are you planning to change this behavior when the number of routes exceed the HW limits?

sudhanshukumar22:
Hi,
When error-handling is enabled, the system will not crash if we try to create an entry in hardware and it returns SAI_STATUS_TABLE_FULL. In this case, we will mark the route as not installed in hardware and create an entry in ERROR_ROUTE_TABLE.

PR #428, opened 2019-07-19
[everflow] Update everflow test plan to cover egress mirroring

wangxin:
This document is an updated version of the existing everflow test plan: https://github.com/Azure/SONiC/wiki/Everflow-test-plan

Purpose of this change is to cover the everflow egress mirroring capability added in https://github.com/Azure/SONiC/pull/411. Some other enhancement is also included in this change.

Major changes:
1. Extend the testing to cover everflow egress mirroring
2. Add a new set of ACL rules for IPv6 testing
3. Add ACL rule to match "IN_PORTS"

### DIFF
--- a/doc/acl/Everflow-test-plan.md
+++ b/doc/acl/Everflow-test-plan.md
@@ -464,14 +464,10 @@ There will be j2 template files for generating ACL tables and ACL rules configur
 
 Different sets of configuration files will be generated for different test scenarios:
 * IPv4
-  * ACL table: MIRROR, implicit ingress; ACL rules, MIRROR_INGRESS_ACTION
-  * ACL table: MIRROR, implicit ingress; ACL rules, MIRROR_EGRESS_ACTION
   * ACL table: MIRROR, ingress; ACL rules, MIRROR_INGRESS_ACTION
   * ACL table: MIRROR, ingress; ACL rules, MIRROR_EGRESS_ACTION
   * ACL table: MIRROR, egress; ACL rules, MIRROR_EGRESS_ACTION
 * IPv6
-  * ACL table: MIRRORV6, implicit ingress; ACL rules, MIRROR_INGRESS_ACTION
-  * ACL table: MIRRORV6, implicit ingress; ACL rules, MIRROR_EGRESS_ACTION
   * ACL table: MIRRORV6, ingress; ACL rules, MIRROR_INGRESS_ACTION
   * ACL table: MIRRORV6, ingress; ACL rules, MIRROR_EGRESS_ACTION
   * ACL table: MIRRORV6, egress; ACL rules, MIRROR_EGRESS_ACTION
@@ -481,20 +477,6 @@ Different sets of configuration files will be generated for different test scena
 For each test scenario, an ACL table configuration need to be created and loaded to DUT. Command `sonic-cfggen` can be used to load the ACL table configuration into config_db. Command syntax:  `sonic-cfggen -j <acl_table_configuration_filename> --write-to-db`.
 
 Example ACL table configuration files to be generated for each scenario:
-* ACL table: MIRROR, implicit ingress
-```
-{
-    "ACL_TABLE": {
-        "EF_INGRESS1": {
-            "policy_desc": "EVERFLOW implicit ingress",
-            "ports": [
-                "Ethernet100", "Ethernet104", "Ethernet92", "Ethernet96", "Ethernet84", "Ethernet88", "Ethernet76", "Ethernet80", "Ethernet108", "Ethernet112", "Ethernet64", "Ethernet60", "Ethernet52", "Ethernet48", "Ethernet44", "Ethernet40", "Ethernet36", "Ethernet120", "Ethernet116", "Ethernet56", "Ethernet124", "Ethernet72", "Ethernet68", "Ethernet24", "Ethernet20", "Ethernet16", "Ethernet12", "Ethernet8", "Ethernet4", "Ethernet0", "Ethernet32", "Ethernet28"
-            ],
-            "type": "MIRROR"
-        }
-    }
-}
-```
 * ACL table: MIRROR, ingress
 ```
 {
@@ -521,20 +503,6 @@ Example ACL table configuration files to be generated for each scenario:
             ],
             "type": "MIRROR",
             "stage": "egress"
-        }
-    }
-}
-```
-* ACL table: MIRRORV6, implicit ingress
-```
-{
-    "ACL_TABLE": {
-        "EFV6_INGRESS1": {
-            "policy_desc": "EVERFLOW IPv6 implicit ingress",
-            "ports": [
-                "Ethernet100", "Ethernet104", "Ethernet92", "Ethernet96", "Ethernet84", "Ethernet88", "Ethernet76", "Ethernet80", "Ethernet108", "Ethernet112", "Ethernet64", "Ethernet60", "Ethernet52", "Ethernet48", "Ethernet44", "Ethernet40", "Ethernet36", "Ethernet120", "Ethernet116", "Ethernet56", "Ethernet124", "Ethernet72", "Ethernet68", "Ethernet24", "Ethernet20", "Ethernet16", "Ethernet12", "Ethernet8", "Ethernet4", "Ethernet0", "Ethernet32", "Ethernet28"
-            ],
-            "type": "MIRRORV6"
         }
     }
 }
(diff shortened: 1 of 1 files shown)


doc/acl/Everflow-test-plan.md
```diff
@@ -0,0 +1,1157 @@
...
+      - [ACL rule for matching "IN_PORTS"](#ACL-rule-for-matching-IN_PORTS)
+      - [IPv6 everflow](#IPv6-everflow)
+    - [How to extend the testing](#How-to-extend-the-testing)
+  - [Test configurations](#Test-configurations)
+    - [ACL table configurations](#ACL-table-configurations)
+  - [Related **DUT** CLI commands](#Related-DUT-CLI-commands)
+    - [`sonic-cfggen`](#sonic-cfggen)
+    - [`acl-loader`](#acl-loader)
+    - [`aclshow`](#aclshow)
+    - [`config mirror_session`](#config-mirror_session)
```
stcheng: could you add config acl table/rule session? I would like to have some CLI supports for table/rule creation/removal

doc/acl/Everflow-test-plan.md
```diff
@@ -0,0 +1,1157 @@
...
+  - [Test case \#3 - Remove longer prefix route.](#Test-case-3---Remove-longer-prefix-route)
+    - [Test objective](#Test-objective-2)
+    - [Test steps](#Test-steps-2)
+  - [Test case \#4 - Change neighbor MAC address.](#Test-case-4---Change-neighbor-MAC-address)
+    - [Test objective](#Test-objective-3)
+    - [Test steps](#Test-steps-3)
+  - [Test case \#5 - Resolved ECMP route.](#Test-case-5---Resolved-ECMP-route)
+    - [Test objective](#Test-objective-4)
+    - [Test steps](#Test-steps-4)
+  - [Test case \#7 - ECMP route change (remove next hop not used by session).](#Test-case-7---ECMP-route-change-remove-next-hop-not-used-by-session)
```
stcheng: Please add test case \#6 as well

doc/acl/Everflow-test-plan.md
```diff
@@ -0,0 +1,1157 @@
...
+- [Overview](#Overview)
+  - [Scope](#Scope)
+  - [Summary of the existing everflow test plan](#Summary-of-the-existing-everflow-test-plan)
+  - [Extend the test plan to cover both ingress and egress mirroring](#Extend-the-test-plan-to-cover-both-ingress-and-egress-mirroring)
+    - [What new enhancements need to be covered?](#What-new-enhancements-need-to-be-covered)
+      - [Egress ACL table](#Egress-ACL-table)
+      - [Egress mirroring](#Egress-mirroring)
+    - [Some existing areas not covered by the existing scripts](#Some-existing-areas-not-covered-by-the-existing-scripts)
+      - [ACL rule for matching "IN_PORTS"](#ACL-rule-for-matching-IN_PORTS)
+      - [IPv6 everflow](#IPv6-everflow)
```
stcheng: Could you add ICMP code/type test as well?

doc/acl/Everflow-test-plan.md
```diff
@@ -0,0 +1,1157 @@
...
+Combining these two enhancements, there are 4 scenarios for everflow.
+1. ACL table of type `MIRROR` have `stage` field ignored or set to "ingress". Action type of ACL_RULE is `MIRROR_ACTION` or `MIRROR_INGRESS_ACTION`.
+2. ACL table of type `MIRROR` have `stage` field ignored or set to "ingress". Action type of ACL_RULE is `MIRROR_EGRESS_ACTION`.
+3. ACL table of type `MIRROR` have `stage` field set to "egress". Action type of ACL_RULE is `MIRROR_ACTION` or `MIRROR_INGRESS_ACTION`.
+4. ACL table of type `MIRROR` have `stage` field set to "egress". Action type of ACL_RULE is `MIRROR_EGRESS_ACTION`.
+
+Expected behaviors for the combinations:
+
+- | ACL table stage: ingress | ACL table stage: egress
+-|-|-
```
stcheng: could you fix the table format?

doc/acl/Everflow-test-plan.md
```diff
@@ -0,0 +1,1157 @@
...
+
+Each test case will be additionally validated by the loganalyzer utility.
+
+Each test case will add dynamic Everflow ACL rules at the beginning and remove them at the end.
+
+Each test case will run traffic for persistent and dynamic Everflow ACL rules.
+
+Each test case will analyze Everflow packet header and payload (if mirrored packet is equal to original). In case of egress mirroring, verify that TTL of the mirrored packet in GRE tunnel is decremented comparing with the injected packet.
+
+### Test case \#1 - Resolved route
```
stcheng: I notice that some of the tests are pretty similar to each other; is it possible to merge some of the tests into single ones so that we could have a simplified list of tests.

doc/acl/Everflow-test-plan.md
wangxin: Sure, added in the new commit.

doc/acl/Everflow-test-plan.md
wangxin: The original missing case 6 was added. Because I combined the similar test cases, in the last version, new ID of the original case 6 is case 4.

doc/acl/Everflow-test-plan.md
wangxin: Sure, fixed.

doc/acl/Everflow-test-plan.md
wangxin: Agree. I have merged some of the test cases. Now we totally have 4 test cases.

doc/acl/Everflow-test-plan.md
wangxin: @stcheng Does SONiC supports matching ICMP type/code? If yes, could you please point me to relevant document/design/code? I need to know how to add ACL rules for matching ICMP type/code. Thanks!

stcheng: hello, yes https://github.com/Azure/sonic-swss/blob/cadc5bd8542d8691f3ce39f4d2569f786a558ff7/orchagent/aclorch.cpp#L343 although `acl-loader` does not support icmp code/type, if you use configuration database, you could still program it.

wangxin: Thanks for the info! I have updated the test plan to cover ACL rules matching ICMP type and code.

wangxin:
In addition to covering ACL rules matching ICMP type and rule, I also added two more scenarios to be covered in the last commit:
* Mirrored IPv4 packets forwarded to mirror destination in IPv6 tunnel.
* Mirrored IPv6 packets forwarded to mirror destination in IPv4 tunnel.

doc/acl/Everflow-test-plan.md
```diff
@@ -0,0 +1,1213 @@
...
+
+##### ACL rule for matching "IN_PORTS"
+Now the SONiC ACL rules support matching "IN_PORTS". New ACL rule for matching "IN_PORTS" need to be added and covered.
+
+#### ACL rule for matching ICMP type and code
+The SONiC ACL rules also support matching ICMP type and code. New ACL rules for matching ICMP type and code need to be added and covered. The acl-loader utility does not support ICMP type and code yet. The sonic-cfggen tool will be used for directly loading such ACL rules to config_db.
+
+##### IPv6 everflow
+The existing scripts only covered IPv6. IPv6 is also supported by SONiC now. The scripts need to be extended to cover IPv6 everflow too. To cover IPv6:
+* Everflow ACL table of type "MIRRORV6" needs to be defined and loaded during testing.
```
rck-innovium: Are both the MIRROR and MIRRORv6 ACL Tables created together?

Some devices would be able to fit both IPv4 and IPv6 rules in the same ACL table; in this scenario creating two ACL tables for mirroring would result in wasting the h/w resources. Is there a way we can define based on the platform whether to have separate mirror ACL tables for v4 and v6?

wangxin: @rck-innovium For simplicity, the MIRROR and MIRRORv6 ACL tables can be created together in configuration/setup stage of the testing. 

It's for testing everflow, as long as h/w resources are enough, it is not a issue. Why would we need to save h/w resources when we have enough of them? It's not for production.

If on platforms which hardware resources are not enough to do the testing, the script can create ACL table for testing each scenario. Then cleanup the ACL rules and tables after testing the scenario is completed.

PR #434, opened 2019-07-30
Configurable drop counters HLD

daall:
This document describes the high level design details for configurable packet drop counters in SONiC.

Signed-off-by: Danny Allen daall@microsoft.com


yxieca:
It might make sense to specifically separate SAI requirement v.s. SONiC requirement. Though most of the requirements are SONiC requirements.

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -0,0 +1,268 @@
...
+### 3.2.2 PACKET_DROP_COUNTER Table
+Example:
+```
+{
+    "PACKET_DROP_COUNTER": {
+        "LEGAL_RX_DROPS": {
+            "counter": "DEBUG_0",
+            "type": "ingress",
+            "reasons": [
+                SMAC_EQUALS_DMAC,
```
prsunny: Do you want to have reasons as a list here? How is it planned to handle add/delete to this after the initial config?

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -0,0 +1,268 @@
...
+                INGRESS_VLAN_FILTER
+            ],
+            "desc": "Legal RX pipeline drops"
+        },
+        "LEGAL_TX_DROPS": {
```
prsunny: `LEGAL `or `LEGIT `is a better name for this?

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -0,0 +1,268 @@
...
+
+## 3.4 SWSS
+Portorch should be extended to support a variable number of SAI_PORT_STAT_IN/OUT_DROP_REASON counters.
+
+Debugcountsorch should be implemented to handle debug counter creation and configuration.
+
+## 3.5 CLI
+The CLI tool will provide the following functionality:
+* Show drop counts: ```show drops```
+* Clear drop counters: ```clear drops```
```
prsunny: I think this must be `sonic-clear`

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -0,0 +1,268 @@
...
+(4) Syncd subscribes to Flex Counter DB and sets up flex counters. Flex counters periodically query ASIC counters and publishes data to Counters DB.
+
+(5) CLI uses counters DB to satisfy CLI requests.
+
+(6) (not shown) CLI uses State DB to display hardware capabilities (e.g. how many counters are available, supported drop reasons, etc.)
+
+# 5 Unit Tests
+A separate test plan will be uploaded and reviewed by the community.
+
+# 6 Open Questions
```
prsunny: Can you add a section for `warm-boot` and how it is handled?

doc/drop_counters/drop_counters_HLD.md
daall: I don't think it makes too much of a difference since this is just an example, but I can update it if people feel like a different name would be more clear.

doc/drop_counters/drop_counters_HLD.md
daall: To address adding/deleting after the initial config there's now a separate table for drop reasons, in the same vein as ACL and Portchannel support.

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -0,0 +1,288 @@
...
+## 3.4 SWSS
+Debugcountsorch should be implemented to handle debug counter creation and configuration.
+
+### 3.4.1 SAI APIs
+This orchestrator will interact with the following SAI APIs (all via sai_debug_counter_api_t):
+* `sai_create_debug_counter_fn` to create/configure new drop counters.
+* `sai_remove_debug_counter_fn` to delete/free up drop counters that are no longer being used.
+* `sai_get_debug_counter_attribute_fn` to gather information about counters that have been configured (e.g. index, drop reasons, etc.).
+* `sai_set_debug_counter_attribute_fn` to re-configure drop reasons for counters that have already been created.
+
```
rck-innovium: This list should also include sai_query_attribute_enum_values_capability()

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -100,12 +106,157 @@ Users must be able to use all counters and drop reasons provided by the underlyi
...
+```
+
+### 3.1.5 Configuring counters from the CLI
+```
+$ config drops init --counter="DEBUG_3" --alias="EXAMPLE" --type="SWITCH_EGRESS" --desc="example" --reasons=["L2_ANY", "L3_ANY"]
+Initializing DEBUG_3 as TX_LEGIT...
+
+Counter   Alias     Type           Reasons  Description
+-------   --------  -------------  -------  -----------
+DEBUG_3   TX_LEGIT  SWITCH_EGRESS  L2_ANY   Legitimate switch-level TX pipeline drops
```
ben-gale: These values don't match the given CLI example (e.g. alias, desc) - confusing.

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -142,141 +293,39 @@ Example:
 }
 ```
 
-## 3.2 State DB
+## 3.3 State DB
```
ben-gale: Can we see the table definition please (like you have for CONFIG_DB)?

doc/drop_counters/drop_counters_HLD.md
daall: Thanks for catching this! It's fixed now. :)

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -0,0 +1,353 @@
...
+(3) The debug counts orchagent publishes counter configurations to Flex Counter DB.
+
+(4) Syncd subscribes to Flex Counter DB and sets up flex counters. Flex counters periodically query ASIC counters and publishes data to Counters DB.
+
+(5) CLI uses counters DB to satisfy CLI requests.
+
+(6) (not shown) CLI uses State DB to display hardware capabilities (e.g. how many counters are available, supported drop reasons, etc.)
+
+# 5 Warm Reboot Support
+At this stage, debug counters will be deleted prior to warm reboot and re-installed when orchagent starts back up. This is intended to simplify upgrade behavior and conserve hardware resources during the warm reboot.
```
dev-aviz: Do you preserve the drop counters values on warm reboots?

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -0,0 +1,353 @@
...
+PORT_EGRESS:
+    L2_ANY
+    L3_ANY
+    A_CUSTOM_REASON
+
+```
+
+### 3.1.2 Displaying current counter configuration
+```
+$ show drops config
```
dev-aviz: can it be possible to logical/aggregate grouping of drop counters?

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -0,0 +1,353 @@
...
+    6. Users can clear counters
+
+## 2.2 Configuration and Management Requirements
+Configuration of the drop counters can be done via:
+* config_db.json 
+* minigraph.xml
+* CLI
+
+## 2.3 Scalability Requirements
+Users must be able to use all counters and drop reasons provided by the underlying hardware.
```
dev-aviz: ex: when we delete drop counters will not delete the other hanrdware counters? hope these counters are not overlap with other hardware counters?

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -0,0 +1,353 @@
...
+## 2.4 Supported Debug Counters
+* PORT_INGRESS_DROPS: port-level ingress drop counters
+* PORT_EGRESS_DROPS: port-level egress drop counters
+* SWITCH_INGRESS_DROPS: switch-level ingress drop counters
+* SWITCH_EGRESS_DROPS: switch-level egress drop counters
+
+# 3 Design
+
+## 3.1 CLI (and usage example)
+The CLI tool will provide the following functionality:
```
dev-aviz: is there any default configurations for drop counters? 
do you have any template to confiugre the debug counters?

doc/drop_counters/drop_counters_HLD.md
daall: Correct. These counters are separate from the existing hardware counters and their behavior (e.g. deleting, clearing, etc.) will not affect the existing counters. I've added a note.

doc/drop_counters/drop_counters_HLD.md
daall: I've added the caveat mentioned in the meeting yesterday. We will only remove the debug counters during warm reboot on platforms where the additional hardware resources are necessary.

doc/drop_counters/drop_counters_HLD.md
daall: We decided during the meeting yesterday that this is out of the scope of the current design. This is a good idea though and I think we'll address it during the discussions about the management framework or potentially as a future extension to this feature.

doc/drop_counters/drop_counters_HLD.md
```diff
@@ -0,0 +1,383 @@
...
+
+### 3.3.1 DEBUG_COUNTER_CAPABILITIES Table
+Example:
+```
+{
+    "DEBUG_COUNTER_CAPABILITIES": {
+        "SWITCH_INGRESS_DROPS": {
+            "total": 3,
+            "used":  1,
+            "reasons": [L2_ANY, L3_ANY, SMAC_EQUALS_DMAC]
```
dev-aviz: What is the scope of L2_ANY drop reason? is this specific hardware platform? what are the SONiC platforms supported this?

daall: L2_ANY refers to any drop reason coming from layer 2, so for example if a frame has the same source mac and destination mac, if a frame is larger than the MTU, etc. Any packet dropped due to problems at layer 2 will be counted under L2_ANY.

L2_ANY is defined in the SAI so in theory it should not be platform specific. That being said, I don't think any of the platforms support this drop reason yet except for maybe Mellanox (let me get back to you on that!)

PR #435, opened 2019-08-01
SONIC VRF HLD 1.2

shine4chen:
Nephos cooperate with brcm for this version.
- update on loopback device
- enhance frr template for vrf and static route configuration
- decrease vrf-name len to 15 for kernel limitation

@preetham-singh

### DIFF
--- a/doc/sonic-vrf-hld.md
+++ b/doc/sonic-vrf-hld.md
@@ -144,7 +144,7 @@ FRR's BGP implementation is capable of running multiple autonomous systems (AS) 
 
 #### VRF route leak support
 
-VRF route leak is a case where route and nexthops are in different VRF.
+Generally VRF route leak is a case where route and nexthops are in different VRF. VRF route leak for directly connected destinations in another VRF is also supported.
 
 VRF route leak can be achieved via both Static or Dynamic approach.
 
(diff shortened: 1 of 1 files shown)


shine4chen:
@prsunny

preetham-singh:
> Route leak support:
> Looks like we have not captured the case "route leaking for directly connected destinations (not thru nexthops) in another VRF", don't we have this use-case?

Connected route leak feature was already supported as part of VRF HLD 1.1.
We missed to list this as part of VRF HLD 1.1. 
@prsunny / @shine4chen 
Perhaps we can update route leak section?

shine4chen:
@venkatmahalingam good point, I will change it.

> VRF_NAME = 1*16VCHAR ;
> 
> This should be 1*15VCHAR ;, please change it.
> 
> root@sonic:/# ip link add Vrf-12345678910 type vrf table 10
> root@sonic:/# ip link add Vrf-123456789101 type vrf table 11
> Error: argument "Vrf-123456789101" is wrong: "name" too long
> 
> root@sonic:/#

@venkatmahalingam good point , thanks. I will change it.

shine4chen:
> > Route leak support:
> > Looks like we have not captured the case "route leaking for directly connected destinations (not thru nexthops) in another VRF", don't we have this use-case?
> 
> Connected route leak feature was already supported as part of VRF HLD 1.1.
> We missed to list this as part of VRF HLD 1.1.
> @prsunny / @shine4chen
> Perhaps we can update route leak section?

@preetham-singh yes , I will update it.

shine4chen:
@prsunny update HLD for acl-redirect schema backward compatibility

PR #436, opened 2019-08-09
Management Framework HLD PR

anand-kumar-subramanian:
Management Framework HLD PR


doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2226 @@
...
+
+    1. Certificate-based authentication
+    2. User/password based authentication
+    3. Role based authorization
+
+* Ease of use for developer workflow
+
+    1. Specify data model and auto-generate as much as possible from there
+
+* Must support Validation and Error Handling - data model, platform capability/scale, dynamic resources
```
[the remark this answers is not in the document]
dutta-partha: @dev-aviz Yes it supports, please refer to section "3.2.2.8.2.3 Platform specific validation".

doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2226 @@
...
+
+#### 1.2.1 Basic Approach
+
+* Management framework takes comprehensive approach catering:
+	* Standard based YANG Models and custom YANG
+	* Open API spec
+	* Industry standard CLI
+	* Config Validation
+* REST server, gNMI server, App module and Translib - all in Go
+* Translation by using the Translib Library and application specific modules
```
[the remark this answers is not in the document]
dutta-partha: @dev-aviz translib is a framework library whereas app modules uses the translib framework and implements feature specific translation logic.

doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2029 @@
...
+2. Run time components
+
+#### 3.2.1 Build time components
+
+Following are the build time components of the management framework
+
+1. Pyang compiler (for YANG to OpenAPI conversion)
+2. Swagger generator
+3. YGOT generator
+4. pyang compiler (for YANG to YIN conversion)
```
daall: Are these components compatible with Automake and/or do they support an incremental build approach? More generally, if I add a new object or tweak the definition of an object slightly does it require a complete re-build of these components or just the dependencies?

doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2029 @@
...
+
+Framework provides swagger-codegen generated Python client SDK. Developers can generate client SDK code in other programming languages from the OpenAPI definitions as required.
+
+Client applications can use swagger generated client SDK or any other REST client tool to communicate with the REST server.
+
+##### 3.2.2.3 gNMI Client
+
+SONiC Teleletry service provides the gNMI server, while the client must be provided by the user.
+
+GNMI client developed by JipanYANG.(github.com/jipanYANG/gnxi/gnmi_get, github.com/jipanYANG/gnxi/gnmi_set)
```
daall: Can you include a table or list somewhere with all the different libraries this feature depends on?

doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2029 @@
...
+Details are in [DB Schema](#3_2_2_4_14-db-schema) section.
+
+HTTPS certificates are managed similar to that of existing gNMI Telemetry program.
+Server key, certificate and CA certificate file paths are configured in CONFIG_DB
+DEVICE_METATDATA table entry. Same certificate is used by both gNMI Telemetry and
+REST server.
+
+###### 3.2.2.4.2 Translib linking
+
+REST server will statically link with Translib. For each REST request, the srever
```
daall:
```suggestion
REST server will statically link with Translib. For each REST request, the server
```

doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2029 @@
...
+returning application errors. In case of CVL constraint violation the REST server will pick
+the error message from the YANG "error-message" statement of CVL schema YANG.
+
+Table below lists possible error conditions with response status and data returned by REST server.
+
+Method  | Error condition          | Status | error-type  | error-tag        | error-message
+--------|--------------------------|--------|-------------|------------------|----------------------
+*any*   | Incorrect request data      | 400 | protocol    | invalid-value    |
+*write* | Bad content-type            | 415 | protocol    | invalid-value    | Unsupported content-type
+*write* | OpenAPI schema validation fails | 400 | protocol| invalid-value    | Content not as per schema
```
daall: I'm assuming these error messages are coming from swagger - is this all the detail they provide? Can we get more context other than "the schema was wrong?" I can see this being tricky to debug for more complicated schemas.

doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2029 @@
...
+###### 3.2.2.7.1 Components
+
+Transformer consists of the following components and data:
+* **Transformer Spec:** a collection of translation hints
+* **Spec Builder:** loads YANG and annotation files to dynamically build YANG schema tree and Transformer Spec.
+* **Transformer Core:** perform main transformer tasks, i.e. encode/decode YGOT, traverse the payload, lookup Transformer spec, call Transformer methods, construct the results, error reporting etc.
+* **Built-in Default Transformer method:** perform static translation
+* **Overloaded Transformer methods:** callback functions invoked by Transformer core to perform complex translation with developer supplied translation logic
+* **Ouput framer:** aggregate the translated pieces returned from default and overloaded methods to construct the output payload
+* **Method o
```
daall: Does this refer just to the application code or do the generated code fragments also provide syslog messages that we can trace?

doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2029 @@
...
+
+Management framework is a SONiC application which is responsible for providing various common North Bound Interfaces (NBIs) for the purposes of managing configuration and status on SONiC switches. The application manages coordination of NBI’s to provide a coherent way to validate, apply and show configuration.
+
+### 1.1 Requirements
+
+* Must provide support for:
+
+    1. Standard [YANG](https://tools.ietf.org/html/rfc7950) models (e.g. OpenConfig, IETF, IEEE)
+    2. Custom YANG models ([SONiC YANG](https://github.com/Azure/SONiC/blob/master/doc/mgmt/SONiC_YANG_Model_Guidelines.md))
+    3. Industry-standard CLI
```
pavel-shirshov: Can you please define "Industry-standard CLI"?

doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2029 @@
...
+## 1 Feature Overview
+
+Management framework is a SONiC application which is responsible for providing various common North Bound Interfaces (NBIs) for the purposes of managing configuration and status on SONiC switches. The application manages coordination of NBI’s to provide a coherent way to validate, apply and show configuration.
+
+### 1.1 Requirements
+
+* Must provide support for:
+
+    1. Standard [YANG](https://tools.ietf.org/html/rfc7950) models (e.g. OpenConfig, IETF, IEEE)
+    2. Custom YANG models ([SONiC YANG](https://github.com/Azure/SONiC/blob/master/doc/mgmt/SONiC_YANG_Model_Guidelines.md))
```
pavel-shirshov: All of them must be supported? What if SONiC doesn't have support for all the models?

doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2029 @@
...
+    3. Industry-standard CLI
+
+* Must provide support for [OpenAPI spec](https://swagger.io/specification/) to generate REST server side code
+* Must provide support for NBIs such as:
+
+    1. CLI
+    2. gNMI
+    3. REST/RESTCONF
```
pavel-shirshov: What about netconf which is considered as standard transport? https://tools.ietf.org/html/rfc6241

dutta-partha: NETCONF is another configuration management protocol which mainly uses SSH as transport. This is considered as management application in context of management framework and  can be supported like RESTCONF. NETCONF will call translib APIs to manage its datastore (running, candidate etc.).

doc/mgmt/Management Framework.md
dutta-partha: If SONiC does not have support, those models won't be implemented. Also if SONiC has partial support for a model, other non-supported objects will be marked as "not-supported" through deviation.

doc/mgmt/Management Framework.md
dutta-partha: "Industry-standard CLI" refers to mainly CISCO style CLI.

doc/mgmt/Management Framework.md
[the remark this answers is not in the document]
kwangsuk: @daall By default, the built-in (default) transformer method performs 1:1 translation from NBI model, e.g. openconfig model, to Redis ABNF using underlying SONiC model.
e.g. if the request lands at the Transformer, the default method maps the container/list name to the TABLE, leaf/leaf-list name/value to FIELD/value unless the external transformer callback overrides the default method. The callbacks can be defined with YANG extensions, “xxx-transformer” on the YANG path, e.g. key-transformer {callback} to translate the YANG list key to Redis Keys. The method overloader will be used by Transformer core to look up the specific callback to override the default method to perform user-supplied code based translation.

kwangsuk: @daall Method overloader is a part of transformer infra code to override the default transformer and perform data translation with user-supplied callbacks. Yes, the translation processes are syslog-ed for trace.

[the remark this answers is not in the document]
daall: Thank you for the explanation! :) 

I'm not sure why github put this comment here though, I was asking about line 1864: "Describe key scaling factor and considerations." It shows up under files but not on the conversation page for some reason.

doc/mgmt/Management Framework.md
pavel-shirshov: Can we put it in the document explicitly?

doc/mgmt/Management Framework.md
pavel-shirshov: So we create another model based on standard model.
Do we need another non-standard model?

doc/mgmt/Management Framework.md
pavel-shirshov: Sorry. I don't get it. Are you going to support NETCONF or not?

anand-kumar-subramanian: Currently we do not support NETCONF, we only support REST. But extending this architecture to support NETCONF should be easy.

doc/mgmt/Management Framework.md
anand-kumar-subramanian: With the SONiC yang model, the conversion is fairly straightforward as it is 1-1 with the ABNF schema in Redis. Because of this many in the work group wanted support for SONiC yang. With this architecture we will support both the models.

doc/mgmt/Management Framework.md
anand-kumar-subramanian: Industry-standard CLI is not a custom construct and it is widely used in the industry.

pavel-shirshov: I think it's better to have good defined spec. Do we try to avoid Cisco in this document?

doc/mgmt/Management Framework.md
pavel-shirshov: So we should remove standard model here, because SONiC doesn't support it. Right?

doc/mgmt/Management Framework.md
anand-kumar-subramanian: Industry standard CLI reference[https://www-01.ibm.com/support/docview.wss?uid=isg3T7000651&aid=1](url)

doc/mgmt/Management Framework.md
anand-kumar-subramanian: SONiC supports both. We have equal number in the work group asking support for standard models as well. So our architecture supports both.

pavel-shirshov: I really want to get the idea.
I don't know why, but I'm pretty sure that SONiC doesn't have support for all settings for the standard model. And I also sure that the standard models doesn't have support for all SONiC settings. It's a reason why you want to use SONiC models (otherwise why do we need them? Let's use the standard?). But as soon as we want support the standard models, we must change them to make compatible with our requirements making them non-standard.
So why do want to support standard models, making them non-standard?

doc/mgmt/Management Framework.md
pavel-shirshov: The link you provided describes CLI available on some IBM equipment. And it's name ISCLI, I'm not sure how is this compatible with Cisco.
I think it's better to change wording from "industry standard CLI" to something "Cisco-like CLI". Because there is no such thing like a standard for CLI in the industry.

doc/mgmt/Management Framework.md
anand-kumar-subramanian: For a given feature we may have support for both Standard models as well as SONiC models. When standard models does not have support, we will provide standard like yang models. This is to support users who want only standard models/standard like models and users who want SONiC only models.

doc/mgmt/Management Framework.md
anand-kumar-subramanian: Updated the md file to have cisco like cli

doc/mgmt/Management Framework.md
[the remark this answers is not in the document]
anand-kumar-subramanian: Updated the doc

doc/mgmt/Management Framework.md
sachinholla: OpenAPI schema validation is applicable for REST APIs defined directly via an OpenAPI spec. We have very basic schema validation support here. It is possible to provide more information, but not planned for now. Payload validation for yang defined RESTCONF APIs will be done through YGOT bindings and it provides detailed error info.

doc/mgmt/Management Framework.md
sachinholla: All code generators and compilers are integrated with SONiC make infrastructure. Incremental build is supported.

dutta-partha: In general the components are having individual Makefiles manually written where dependencies are captured and they support incremental build approach.

doc/mgmt/Management Framework.md
dutta-partha: Are you asking about dependent library list for telemetry only or management framework in general ?

daall: Sorry I should have specified, management framework in general.

doc/mgmt/Management Framework.md
```diff
@@ -0,0 +1,2029 @@
...
+
+###### 3.2.2.7.1 Components
+
+Transformer consists of the following components and data:
+* **Transformer Spec:** a collection of translation hints
+* **Spec Builder:** loads YANG and annotation files to dynamically build YANG schema tree and Transformer Spec.
+* **Transformer Core:** perform main transformer tasks, i.e. encode/decode YGOT, traverse the payload, lookup Transformer spec, call Transformer methods, construct the results, error reporting etc.
+* **Built-in Default Transformer method:** perform static translation
+* **Overloaded Transformer methods:** callback functions invoked by Transformer core to perform complex translation with developer supplied translation logic
+* **Ouput framer:** aggregate the translated pieces returned from default and overloaded methods to construct the output payl
```
daall: Could you be a little more specific? How does this framework scale in terms of the number of models/APIs? Do you anticipate any significant impact in terms of latency, memory usage, etc? Is this impact static or is it somehow tied to the schema or the number of APIs or some other metric?

kwangsuk: All the translation metadata – transformer callbacks or simple hints like field-name change etc. -  defined as YANG extensions across YANG modules are loaded and stored in the in-memory Transformer Spec. As the number of YANG modules requiring translation increases, the size of Transformer Spec grows and consumes more memory. During runtime, if the request URI represents a resource which can span multiple augmented modules with a deep depth, the number of external transformer callbacks involved in data translations between YANG and Redis ABNF schema could increase to incur process/method call overhead and consume more resources. In general, the overall latency and resource consumption is mainly dependent on the number of translation points based on the request URI/payload.

doc/mgmt/Management Framework.md
anand-kumar-subramanian: Added an Appendix A section at the end and listed all the open source tools used in the Management framework

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,584 @@
+# RADIUS Management User Authentication
```
renukamanavalan: Can you please add this file in a new PR?

This file being newly added is very likely to be missed by reviewers. Let us commit the reviewed changes and then take this in next pull.

a-barboza: ok. this was added to master of project-arlo and got tagged along with other mgmt-framework SONiC pull request. Will raise as a separate PR from a different branch

PR #437, opened 2019-08-14
EVPN VXLAN HLD.

srj102:
Initial Version of the EVPN VXLAN HLD.


doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+
+### 4.1.3 SAI Overview
+
+To support this feature, SAI will be extended as below:
+
+- **sai_fdb_entry_type_t** - will be enhanced to have a new enum. 
+  - SAI_FDB_ENTRY_TYPE_STATIC_MACMOVE - This will be used when a MAC should not be aged out but a MAC move should be allowed. An example is remote MAC learnt by EVPN procedures which can be subjected to a MAC move from remote to local.
+- **sai_tunnel_attr_t** - It will be enhanced to have one new attribute
+  - SAI_TUNNEL_ATTR_ENCAP_DEST_IP - New tunnel attribute will be added to specify the tunnel destination IP address. This will be used to support warm reboot. Currently only the SIP is part of the tunnel attributes which does not allow for unique set of attributes when there are multiple tunnels with same SIP and different DIPs. The unique set of attributes is required for reconcilation during a warm reboot.  bridgeport associated with a tunnel also requires the tunnel id to be reconcilable post warm restart. Hence, it is required as part of the sai_tunnel_attr_t list.
```
marian-pritsak: VxLAN is P2MP tunnel. The DST IP is part of fdb encap entry or a tunnel route. What is the reason for creating multiple tunnels if we can implement P2MP already?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+### 4.1.2 Container
+
+No new container is added. The changes are added to existing containers such as FRR, SwSS and Syncd. The details of the changes will be discussed in the Design section below.
+
+### 4.1.3 SAI Overview
+
+To support this feature, SAI will be extended as below:
+
+- **sai_fdb_entry_type_t** - will be enhanced to have a new enum. 
+  - SAI_FDB_ENTRY_TYPE_STATIC_MACMOVE - This will be used when a MAC should not be aged out but a MAC move should be allowed. An example is remote MAC learnt by EVPN procedures which can be subjected to a MAC move from remote to local.
```
marian-pritsak: How is the behavior different from SAI_FDB_ENTRY_TYPE_STATIC?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+- **sai_tunnel_attr_t** - It will be enhanced to have one new attribute
+  - SAI_TUNNEL_ATTR_ENCAP_DEST_IP - New tunnel attribute will be added to specify the tunnel destination IP address. This will be used to support warm reboot. Currently only the SIP is part of the tunnel attributes which does not allow for unique set of attributes when there are multiple tunnels with same SIP and different DIPs. The unique set of attributes is required for reconcilation during a warm reboot.  bridgeport associated with a tunnel also requires the tunnel id to be reconcilable post warm restart. Hence, it is required as part of the sai_tunnel_attr_t list.
+- **sai_vlan_attr_t** - It will be enhanced to have one new attribute
+  - SAI_VLAN_ATTR_NEIGH_SUPP_MODE - By default this value will be FALSE. Reason for this new attribute is due to recent changes in ARP behaviour. The default arp behavior in copp file is changed from trap to copy(as part of commit -5e4b71d4). But for ARP Suppression, we need Pkt to get trapped instead of copy to avoid flooding of packets over vxlan tunnel.So a differentiated tcam rule for these arp suppressed vlans will redirect the ARP pkts to cpu instead of copy.
```
marian-pritsak: By defining this kind of attribute, the ARP flooding will be disabled for locally connected hosts on that VLAN, which seems not the best solution. I think it is possible to control flooding per port. Instead of disabling flood completely, it should be restricted to the local ports only.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+```
+;New table
+;Stores Neigh Suppress configuration per VLAN
+
+key             = NEIGH_SUPPRESS_VLAN:"Vlan"vlanid       ; VLAN with prefix "NEIGH_SUPPRESS"
+suppress_neigh  = "ON" / "OFF" ; Default "OFF"
+
+```
+
+**VRF_TABLE**
```
marian-pritsak: There is already a L3 VxLAN implementation already in SONiC called VNET. Why not integrate with it?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: As mentioned these should NOT be aged out but could be subjected to a MACMOVE. We don't
want to delete and add this.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: To handle the case of ingress replication for BUM we need P2P tunnels to be created and a bridge port
associated with them. These P2P tunnels are discovered in the EVPN case. To handle warm reboot  we would need the DEST IP for these P2P tunnels.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+3. Support configuration of L3VNI association with VRF. 
+
+
+
+FRR version 7.0.1, and later, is assumed here. FRR configuration is assumed in "split" configuration mode.
+BGP EVPN configurations in FRR are referred wherever required in this document. However, this document does not serve as reference guide for the FRR configurations.
+
+
+
+## 2.3 Scalability Requirements
```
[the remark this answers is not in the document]
srj102: These are part of an overall requirements section. Scale is controlled by hardware as well as 
FRR capabilities for the control plane.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
kananthak: The Local connected hosts will always get the Arp Requests flooding. 
When Arp Suppression is turned on that Vlan, the new trap action in hw ensures that hw doesn't flood the packet over tunnel for that specified vlan.
When the Kernel IP Stack receives the packet, it floods the arp requests to local connected hosts even when it  acts as proxy for tunnel neighbour. 
Here we try to only stop the Hw flooding of Arp pkts over tunnel as the copp file is recently modified from trap to copy.
If we have had the  old copp file (before the commit mentioned above), the SAI change is not required. 
So here the flooding of arp pkts to local connected hosts happens by IP stack, instead of hw.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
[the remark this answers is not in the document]
marian-pritsak: I'm not sure how relevant those scale requirements are because I don't understand how they affect the design. Can you please provide the justification for each of them where the number was derived from and what implication it has on a design if any?

What about CRM? Any work expected to be done in this area?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
hasan-brcm: EVPN L3VNI design here is already integrated with VXLAN design/implementation done as part of VNET. Most of the work is being leveraged. However, since VNET is not VRF, and VRF support was added later into SONiC, it makes more sense to provide L3VNI solution with VRF.
VNET will continue to co-exist with VRF-L3VNI.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
marian-pritsak: Now I understand. Thanks

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
marian-pritsak: Agree. Isn't the host interface a better place for defining such attribute? It's all about traps, and not the VLAN, VLAN has nothing to do with traps.

kananthak: Still this is per vlan attribute. For example , Vlan-x can be a configured with arp suppression enabled.  But Vlan-y can be configured with arp suppression disabled.
Hence we have made this attribute configurable per vlan.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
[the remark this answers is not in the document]
srj102: On the Scaling - Agreed that different platforms can support different numbers.  If the suggestion is to remove the numbers and mention that this will be specified in the platform specific documentation then this can be incorporated. 

However I do see scalability numbers put in as part of the 
https://github.com/Azure/SONiC/blob/master/doc/vxlan/Vxlan_hld.md
 
On CRM - no work is planned. The CRM depends on SAI Switch attributes to be defined for VXLAN which is not defined presently.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
kananthak: Since the cpu is going to act as ARP & ND proxy for the configured vlans, the cpu load might be little extra when compared to normal ARP/ND programming. Hence on top of the sai vlan attribute, we are also going to introduce the two new HOSTIF trap types in copp file. This will help us to set adjust the cir/cbs values when required.  
Lets us name it as"
i) SAI_HOSTIF_TRAP_TYPE_ARP_SUPPRESS
ii)SAI_HOSTIF_TRAP_TYPE_ND_SUPPRESS

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+
+EVPN VXLAN feature supports L2 forwarding and L3 routing over VXLAN, dynamic creation and deletion of VXLAN tunnels, dynamic extension and removal of VLANs over the discovered tunnels based on BGP EVPN procedures described in RFC 7432 and 8365.
+
+#### L2 forwarding over VXLAN
+
+Traffic received from the access ports are L2 forwarded over VXLAN tunnels if the DMAC lookup result does not match the switch MAC. The frame format is as described in RFC 7348. The VNI is based on the configured global VLAN-VNI map. BUM traffic is ingress replicated to all the tunnels  which are part of the VLAN. The DIP of the BUM packets is the IP address of the remote VTEP. L2 unicast traffic is forwarded to the tunnel to which the MAC points to. 
+Traffic received from VXLAN tunnels are mapped to a VLAN based on the received VNI and the VNI-VLAN lookup. The received traffic can be L2 unicast or BUM depending on the inner MAC lookup. Traffic received from VXLAN tunnels are never forwarded onto another VXLAN tunnels. 
+
+![L2 forwarding over VXLAN ](images/pktflowL2.PNG "Figure 1 : L2 Flow")
+__Figure 1: L2 forwarding Use Case__
```
prsunny: The figure is not a use case diagram. May be rename it to packet flow.

srj102: will do.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+
+
+## 3.2 Functional Description
+
+EVPN VXLAN feature supports L2 forwarding and L3 routing over VXLAN, dynamic creation and deletion of VXLAN tunnels, dynamic extension and removal of VLANs over the discovered tunnels based on BGP EVPN procedures described in RFC 7432 and 8365.
+
+#### L2 forwarding over VXLAN
+
+Traffic received from the access ports are L2 forwarded over VXLAN tunnels if the DMAC lookup result does not match the switch MAC. The frame format is as described in RFC 7348. The VNI is based on the configured global VLAN-VNI map. BUM traffic is ingress replicated to all the tunnels  which are part of the VLAN. The DIP of the BUM packets is the IP address of the remote VTEP. L2 unicast traffic is forwarded to the tunnel to which the MAC points to. 
+Traffic received from VXLAN tunnels are mapped to a VLAN based on the received VNI and the VNI-VLAN lookup. The received traffic can be L2 unicast or BUM depending on the inner MAC lookup. Traffic received from VXLAN tunnels are never forwarded onto another VXLAN tunnels.
```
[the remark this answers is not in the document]
srj102: will do.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+## 4.2 DB Changes
+
+At a high level below are some of the interactions between relevant components and the DB involved for EVPN L2 VXLAN support in SONiC architecture.
+
+![EVPN L2 VXLAN Arch](images/arch.PNG "Figure : Arch")
+__Figure 4: EVPN L2 VXLAN DBs__
+
+### 4.2.1 CONFIG_DB changes
+
+**EVPN_NVO_TABLE**
```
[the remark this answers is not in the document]
srj102: Only 1 for now.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+
+```
+;Existing table
+;defines virtual routing forward table. Updated to stores VNI associated with VRF
+;
+;Status: stable
+
+key = VRF_TABLE:VRF_NAME ;
+fallback = "true"/"false"
+vni = 1*8DIGIT ; VNI associated with VRF
```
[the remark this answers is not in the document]
srj102: yes

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+```
+;Stores a list of routes
+;Status: Mandatory
+
+key = ROUTE_TABLE:VRF_NAME:prefix ;
+nexthop = prefix,              ; IP addresses separated ',' (empty indicates no gateway). May indicate VXLAN endpoint if vni_label is non zero.
+intf = ifindex? PORT_TABLE.key ; zero or more separated by ',' (zero indicates no interface)
+vni_label = VRF.vni            ; New Field. zero or more separated by ',' (empty value for non-vxlan next-hops). May carry MPLS label in future.
+router_mac = mac_address       ; New Field. zero or more remote router MAC address separated by ',' (empty value for non-vxlan next-hops)
+blackhole = BIT                ; Set to 1 if this route is a blackhole (or null0)
```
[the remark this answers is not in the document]
tapashdas: Generic.  vni_label can be vni or mpls label in future

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+Schema:
+
+```
+;Stores a list of routes
+;Status: Mandatory
+
+key = ROUTE_TABLE:VRF_NAME:prefix ;
+nexthop = prefix,              ; IP addresses separated ',' (empty indicates no gateway). May indicate VXLAN endpoint if vni_label is non zero.
+intf = ifindex? PORT_TABLE.key ; zero or more separated by ',' (zero indicates no interface)
+vni_label = VRF.vni            ; New Field. zero or more separated by ',' (empty value for non-vxlan next-hops). May carry MPLS label in future.
```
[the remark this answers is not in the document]
tapashdas: New fields added are part of Nexthop tuple {Nexthop IP, VNI, RMAC}. In case of Overlay ECMP, we will have Overlay Nexthops seperated by ',' . Nexthop IP and RMAC will be different. Encap VNI can be same or different.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+- sai_fdb_entry_add called with the following attributes
+  - SAI_FDB_ENTRY_ATTR_BRIDGE_PORT_ID with the bridge port id corresponding to the tunnel.
+  - SAI_FDB_ENTRY_ATTR_ENDPOINT_IP with the tunnel DIP.
+  - SAI_FDB_ENTRY_ATTR_TYPE with type SAI_FDB_ENTRY_TYPE_STATIC_MACMOVE (corresponding to type dynamic in VXLAN_FDB_TABLE)  or SAI_FDB_ENTRY_TYPE_STATIC (corresponding to type static in VXLAN_FDB_TABLE)  
+
+![Remote MAC Handling](images/remotemac.PNG "Figure : Remote MAC handling")
+__Figure 10: Remote MAC handling__
+
+
+#### 4.3.7.3 MAC Move handling
```
prvattem: Is there a limit on how many times a mac can be moved between 2 VTEPs?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+
+
+
+## 2.2 Configuration and Management Requirements
+
+Configuration and Management for this feature will be based the proposed management framework for SONiC.
+
+1. Support configuration of Source IP address of the VTEP. 
+2. Support configuration of a global VLAN-VNI map.
+3. Support configuration of L3VNI association with VRF.
```
prvattem: Is it worth mentioning that, L3VNI can only be associated with non-default VRF (Symmetric IRB - case)

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+
+
+
+## 2.3 Scalability Requirements
+
+1. Total Local VXLAN terminations (source VTEP) - 1.
+2. Total Remote VTEPs (VXLAN destination tunnels) - 512.
+3. Total L2 VNI per switch- 4K.
+4. Total VNI per tunnel - 4K.
+5. Total EVPN participating VRF per switch - 512
```
prvattem: If all the VRFs support Symmetric IRB, those many extra VNIs(L3) will be consumed.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+![EVPN IP Prefix route handling](images/type5route.PNG "Figure 11: EVPN_Type5_Route")
+__Figure 12: EVPN_Type5_Route__
+
+#### VRF Route Table with Tunnel Nexthop
+
+The connected prefixes bound to VNET go into VNET_ROUTE_TABLE, and remote prefixes are programmed in VNET_ROUTE_TUNNEL_TABLE.
+
+In contrast to the above, VRF routes are programmed in ROUTE_TABLE:{{vrf_name}} table including local (connected, static) and remote (BGP, IGP) VRF-lite prefixes.
+
+In order to program VRF routes imported from EVPN Type-5 prefix routes, <endpoint, router-MAC, and VNI> values are required along with the prefix in the route entry.
```
prvattem: Please describe a little more about how router-mac is derived.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+This feature adds the following enhancements.
+
+- EVPN Type 2, 3 routes support.
+- VLAN extension over VXLAN tunnels.
+- Flooding of L2 BUM (Broadcast, Unknown Unicast, and Multicast) traffic in and out of VXLAN tunnel.
+- Forwarding of L2 unicast traffic in and out of VXLAN tunnel.
+- EVPN Type 5 route support based on VRF construct.
+- Routing of L3 (IPv4 and IPv6) traffic in and out of the VXLAN tunnel.
+- Overlay ECMP support.
+- EVPN ARP and ND suppression.
```
prvattem: Is distributed Anycast Gateway support, not a requirement?
What would be the default Gateway for a host in a given subnet?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
prvattem: Can this be renamed to SAI_FDB_ENTRY_TYPE_STICKY ?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+
+Consumer: VxlanMgr
+
+Description: This is a singleton configuration which holds the VTEP source IP used for BGP-EVPN discovered tunnels. 
+
+Schema:
+
+```
+;New table
+;holds the VTEP source IP used for BGP-EVPN discovered tunnels
```
prvattem: Is the description misleading here? We dont store the VTEP source IP here.. rather the Name of the Tunnel, which has the IP associated with.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+Schema:
+
+```
+;Existing table
+;defines virtual routing forward table. Updated to stores VNI associated with VRF
+;
+;Status: stable
+
+key = VRF_TABLE:VRF_NAME ;
+fallback = "true"/"false"             ; default false
```
prvattem: What is intended behavior with fallback option?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+
+Consumer: VxlanOrch
+
+Description: This specifies the remote VNI to be extended over a tunnel. This is populated by fdbsyncd on receipt of a IMET route from the remote side. 
+
+Schema:
+
+```
+; New table
+; specifies the VLAN to be extended over a tunnel
```
prvattem: This table is confusing. For a given local VLAN the VNI is mapped. If an IMET route is received with a matching VNI, we can build a mapping of VLAN,VNI to list of remote VTEPs.  
But the current schema allows a combination of <V1,RIP1>, <V2,RIP1> map to same VNI.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+Schema:
+
+```
+; New table
+; Defines schema for VXLAN State
+key             = VXLAN_TUNNEL:name    ; tunnel name
+; field         = value
+SRC_IP          = ipv4                 ; tunnel sip
+DST_IP          = ipv4                 ; tunnel dip 
+tnl_src         = "CLI"/"EVPN"
```
prvattem: the option "CLI" is for static VXLAN case? There is a mention of not supporting static VXLAN in this document.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: The term sticky is specific to EVPN whereas the SAI definitions tend to be fairly generic and more reflective of the device functionality. Hence the term STATIC_MACMOVE was given. May be we could propose STATIC_WITH_MACMOVE. However the naming convention is finally to be decided by the SAI community.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: By default, 5 MAC Moves in a 180 sec period according to RFC 7432 sec 15.1. Will update on FRR behavior.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: Will correct.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: This is to handle the case of downstream assigned VNI in DCI cases in future. The VNI in the table specifies the VNI carried in the IMET route. This VNI will be used during encap. The remote will advertise unique VNI values per VLAN. In the DCI case the remote VNI will be different from the local VNI for a VLAN whereas in a global VNI case it will be the same.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: Yes for the static case or could be extended for orchestrator configured tunnels like NSX. This is purely informational. With the current feature only EVPN will be filled in. Static tunnels configured through VXLAN_TUNNEL table with a DIP could fill this with the tnl_src field as static.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
tapashdas: Yes. Will do.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
tapashdas: This is existing field in VRF Table. Please refer Following VRF HLD for detail.
https://github.com/Azure/SONiC/blob/master/doc/vrf/sonic-vrf-hld.md

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
tapashdas: distributed Anycast Gateway will be supported.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+
+### 4.1.3 SAI Overview
+
+To support this feature, SAI will be extended as below:
+
+- **sai_fdb_entry_type_t** - will be enhanced to have a new enum. 
+  - SAI_FDB_ENTRY_TYPE_STATIC_MACMOVE - This will be used when a MAC should not be aged out but a MAC move should be allowed. An example is remote MAC learnt by EVPN procedures which can be subjected to a MAC move from remote to local.
+- **sai_tunnel_attr_t** - It will be enhanced to have one new attribute
+  - SAI_TUNNEL_ATTR_ENCAP_DEST_IP - New tunnel attribute will be added to specify the tunnel destination IP address. This will be used to support warm reboot. Currently only the SIP is part of the tunnel attributes which does not allow for unique set of attributes when there are multiple tunnels with same SIP and different DIPs. The unique set of attributes is required for reconcilation during a warm reboot.  bridgeport associated with a tunnel also requires the tunnel id to be reconcilable post warm restart. Hence, it is required as part of the sai_tunnel_attr_t list.
+- **sai_vlan_attr_t** - It will be enhanced to have one new attribute
```
prvattem: We should also have the ability to enable the ARP suppression system wide.
Higher the number of VLANs, higher will be the configuration.

kananthak: Same review comment is given to us during EVPN work group meeting.
We have accepted to implement this enhancement.
Few hardware platforms which cannot scale will continue to use the individual per vlan setting.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+    
+EVPN_NVO_TABLE|{{nvo_name}}
+    "source_vtep" : {{source_vtep_name}}
+```
+
+
+
+Following configurations are required for enabling EVPN L3 service on SONiC device:
+
+- Configure L3VNI-VLAN mapping for the given NVO.
```
prvattem: This probably is the MAC-VRF VNI, which is different from L3VNI.
VLAN<--> MAC-VRF VNI  and IP-VRF <-->L3VNI.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1162 @@
...
+   +=========+=======+
+   | Vlan100 |   100 |
+   +---------+-------+
+   | Vlan101 |   101 |
+   +---------+-------+
+   Total count : 2
+
+
+3. show vxlan vrfvnimap 
+   - Displays all the VRF VNI mappings.
```
prvattem: Please provide the sample o/p for this?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1162 @@
...
+
+5. show vxlan remote_mac <remoteip/all>  <vlanid/all> 
+   - lists all the MACs learnt from the specified remote ip or all the remotes for the specified/all vlans. (APP DB view) 
+   - VLAN, MAC, RemoteVTEP,  VNI,  Type are the columns.
+
+   show vxlan remote_mac all
+   +---------+-------------------+--------------+-------+--------+
+   | VLAN    | MAC               | RemoteVTEP   |   VNI | Type   |
+   +=========+===================+==============+=======+========+
+   | Vlan101 | 00:00:00:00:00:01 | 4.4.4.4      |  1001 | static |
```
prvattem: Here the remote MAC type "static" conveys a meaning that, the learnt MAC is static on the Remote VTEP. This needs to be changed??

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1162 @@
...
+   +=========+===================+==============+=======+========+
+   | Vlan101 | 00:00:00:00:00:02 | 3.3.3.3      |  1001 | static |
+   +---------+-------------------+--------------+-------+--------+
+   | Vlan101 | 00:00:00:00:00:99 | 3.3.3.3      |  1001 | static |
+   +---------+-------------------+--------------+-------+--------+
+   Total count : 2
+
+
+6. show vxlan remote_vni <remoteip/all> 
+   - lists all the VLANs learnt from the specified remote ip or all the remotes.  (APP DB view)
```
prvattem: This needs to be rephrased to say "lists all the VNIs learnt from" ?? VLAN is local and mapped to a VNI

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1235 @@
...
+- fdbsyncd adds the new remote MAC entry to the APP_VXLAN_FDB_TABLE.
+- fdborch looks up internal cache to see if an entry exists. If it exists with a local port as the destination then the entry is deleted and the new entry populated.
+- If the local MAC is configured as static then the received remote MAC will not be programmed and instead will be marked as pending. This will be programmed when the local static MAC is removed. 
+
+##### Remote to Remote
+
+MAC Move from remote tunnel-1 to remote tunnel-2 is handled as follows. 
+
+- On identifying a MAC Move, BGP deletes the old entry from APP_VXLAN_FDB_TABLE and adds the new entry pointing to the tunnel-2.
+- fdborch processing is same as for a regular remote MAC add/delete.
```
gord1306, line 861: is there any section for the remote MAC delete flow.  I only see the new Remote MAC flow

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
tapashdas: yes. One VNI per VRF.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1023 @@
...
+
+
+#### Symmetric and Asymmetric IRB
+
+In the asymmetric IRB model, the inter-subnet routing functionality is performed by the ingress VTEP with the packet after the routing action being VXLAN bridged to the destination VTEP.
+
+The egress VTEP removes the VXLAN header and forwards the packet onto the local Layer 2 domain based on the VNI to VLAN mapping. The same thing happens on the return path, with the packet routed by the ingress router and then overlay switched by the egress router. The overlay routing action occurs on a different router on the reverse path vs. the forward path, hence the term asymmetric IRB.
+
+![Asymmetric IRB Use Case](images/pktflowAsymmIRB.PNG "Figure 2 : Asymm IRB Flow")
+__Figure 2: Asymmetric IRB Use Case__
```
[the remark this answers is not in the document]
tapashdas: done

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
tapashdas: Yes.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
tapashdas: Interface MAC of IRB router interface is used as router-mac.
L3 VNI will be associated with a Vlan interface (IRB router interface), Vlan interface MAC of that associated IRB interface is used as router-mac.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
tapashdas: updated the document.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+
+Description: This is a singleton configuration which holds the tunnel name specifying the VTEP source IP used for BGP-EVPN discovered tunnels. 
+
+Schema:
+
+```
+;New table
+;holds the tunnel name specifying the VTEP source IP used for BGP-EVPN discovered tunnels
+
+key = EVPN_NVO_TABLE:nvo_name
```
gord1306, line 309: In sonic-swss-commin [PR339](https://github.com/Azure/sonic-swss-common/pull/339/), this table is named as VXLAN_EVPN_NVO

srj102: Will update the HLD to match the PR 339.

ben-gale:
As discussed 6/23, can we approve/merge this now please?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+
+There is no special handling by SwSS. 
+
+Local to Local MAC moves will be identified by BGP and no MAC route will be re-originated. 
+
+##### Remote to Local
+
+For MAC move cases from remote to local, fdbsyncd removes the 
+remote entry from VXLAN_FDB_TABLE. The fdborch should not delete the 
+MAC from the internal cache and the hardware.
```
gord1306, line 836: For the remote to local case, why the MAC can not delete from the internal cache and the hardware.  Does this mean the MAC address will exist both in remote and local in the hardware?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+
+In addition there is a need to display the operstatus of the tunnel.  The operstatus is reflective of the dataplane forwarding status of the tunnel. The operstatus can toggle depending on the dataplane state. 
+
+A new table STATE_VXLAN_TUNNEL_TABLE is used for the above purposes.
+
+A new field "tunnel source" - EVPN or CLI will be added to indicate the source of tunnel creation. 
+
+The tunnel source (EVPN, CLI), SIP, DIP can be filled in during creation time. 
+
+The operstatus of a tunnel  will be updated based on notification from SAI through a notification channel similar to port oper status.
```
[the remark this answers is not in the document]
srj102: This is not an extra SAI enhancement. The existing port oper status notification is to be used with the bridgeport id being that of the p2p tunnel.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: 1. The remote to local is initiated by a learn/move  event from the SAI/hardware. 
2. The internal cache would be updated. 
3. The fdbsyncd knows about this when the STATE DB is updated with the MAC which is moved to the local.
4. The VXLAN_FDB_TABLE entry will have to be removed by the fdbsyncd. 
5. After the above step fdborch gets a notification to remove the MAC, the fdborch should identify that this is a MAC which has already been moved and should not delete.  In a normal remote MAC removal, the MAC would have been removed by fdborch from its internal cache.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+For MAC move cases from remote to local, fdbsyncd removes the 
+remote entry from VXLAN_FDB_TABLE. The fdborch should not delete the 
+MAC from the internal cache and the hardware. 
+
+##### Local to Remote
+
+MAC move from local to remote tunnel is handled as follows.
+
+- fdbsyncd adds the new remote MAC entry to the APP_VXLAN_FDB_TABLE.
+- fdborch looks up internal cache to see if an entry exists. If it exists with a local port as the destination then the entry is deleted and the new entry populated.
```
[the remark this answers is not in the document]
srj102: Yes this is removed from the state db as well.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
gord1306: I am still not clear why the remote MAC address can not remove from the hardware. And when the remote MAC address is removed from the hardware. Did you mean the fdborch will keep the remote MAC address in its cache until an event (maybe timeout event?) , Thanks

srj102: The remote MAC is programmed as not ageable but moveable. So as part of step 1 the device would have re-learnt the MAC over a local port. At this point itself the MAC is moved in the hardware to a local port. The indication is sent to the fdborch and fdbsyncd to update the necessary software state.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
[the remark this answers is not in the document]
gord1306: which PR contains this oper_up status operations? thanks.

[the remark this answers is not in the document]
srj102: sonic swss PR 1264.
updateDbTunnelOperStatus is called from the portsorch.cpp.

However I notice that getPort(sai_object_id_t id, Port &port) has to be enhanced to also support TUNNEL.
This function is the one which derives the tunnel id from the notified bridge port id.

There was a dependency on PR#885 around this area as the getPort function was reorganised to use a 
portOidToName structure.  Earlier comments in PR 1264 pointed out this dependency and it was commented out
to let PR 1264 be independent of 885. 

I will add the separate TUNNEL handling in getPort to derive the tunnel id.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
prvattem: What happens in-case of MCLAG use case? How is the router-mac derived?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+The corresponding CONFIG_DB entries are as follows. 
+
+
+```
+VXLAN_TUNNEL_TABLE|{{source_vtep_name}}
+    "src_ip" : {{ipv4_address}}
+    
+VXLAN_EVPN_NVO_TABLE|{{nvo_name}}
+    "source_vtep" : {{source_vtep_name}}
+```
```
gord1306, line 558: These 2 tables' are different with the sonic-swss-common schema.h [link](https://github.com/Azure/sonic-swss-common/pull/339)

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+- VNI VLAN map cannot be deleted if there is corresponding VRF VNI map.
+
+```
+
+## 6 Serviceability and Debug
+
+The existing logging mechanisms shall be used. The show commands and the redis table
+dumps can be used as debug aids.
+
+## 7 Warm Reboot Support
```
prsunny, line 1205: This is an enhancement to existing warmboot sequence and may need to be taken with warmboot forum. I'll check and get back on any feedback.

prsunny:
@marian-pritsak , @prvattem , @gord1306 , could you please sign-off.  This is pending merge.

gord1306:
hi, @prsunny  I can not see any sign-off options in my comments.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+There is no change to fdborch wrt local MAC handling. 
+
+Currently Local MACs learnt by the fdborch are written to the STATE_DB. 
+
+Fdbsyncd listens to the above table and notifies Linux FDB which notifies to Zebra, Zebra to BGP and BGP originates the MAC route. 
+
+The above is applicable for locally added static MACs also. 
+
+![Local MAC Handling](images/localmac.PNG "Figure : Local MAC handling")
+__Figure 9: Local MAC handling__
```
gord1306, line 806: For the right side in the local mac handling flow chart, whether the fdbsyncd directly notify to zebra?? or it should write the fdb entry to kernel and then the zebra receive the kernel notification for the new local mac event

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+dumps can be used as debug aids.
+
+## 7 Warm Reboot Support
+
+Warm reboot will support below requirements
+
+1. System Warm reboot for EVPN 
+   The goal of the system warm reboot is to achieve system software reload without impacting the data plane and graceful restart of the control plane (BGP).
+
+![System Warm reboot and reconciliation sequence](images/warmrebootseq.PNG "Figure 16: System Warm reboot and reconciliation sequence")
```
gord1306, line 1212: There is a question about the 3rd last step - learn local MAC from kernel and re-dist to peers. How these MAC entries be stored during the warm-reboot? Whether these MAC entries do no need to sync from ASIC DB? Because there is no ASIC DB in this flow chart

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+
+__Figure 16: System Warm reboot and reconciliation sequence__
+
+During the system warm-reboot restoration/reconciliation, the fdbsyncd is dependent on Intf, Vlan, Vrf, Vxlan Mgrs to complete their data restoration to kernel. Before fdbsyncd can program the MACs to the kernel it needs to wait for these subsystems data replay to kernel. Without this the kernel programming for fdbsyncd will fail.  Subsequently if the FDB/VTEP info is not available in kernel/zebra for BGP to distribute to its peers, it will converge with premature info causing data-loss. To solve this dependency , we use a replay-done notification by each subsystem. The notification is passed between the subsystems by using the existing  WARM_RESTART_TABLE. A new state "replayed" in the data restoration flow  ( "state" = "replayed") is used by each subsystem to indicate it has replayed its data to kernel.
```
qiluo-msft, line 1216: >A new state "replayed" [](start = 666, length = 22)

The state enum becomes  more complex. Could you draw a state transition diagram and describe it?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+
+The dependent subsystems use the WARM_RESTART_TABLE table to wait before attempting to program their info to the kernel.   fdbsyncd replays its info to kernel only after the data is replayed for Intf, Vlan, Vxlan and Vrf Mgrs.  BGP is configured/started only when FDB has replayed all its data to kernel. 
+
+There is also a requirement that the app_db entries are not modified by *syncd subsystems until the orchagent has reconciled its data after warm-reboot. To achieve this dependency, the *syncd applications which write data to the app_db wait till orchagent reports "reconciled" status in the WARM_RESTART_TABLE.
```
qiluo-msft, line 1220: >There is also a requirement that the app_db entries are not modified by *syncd subsystems until the orchagent has reconciled its data after warm-reboot [](start = 0, length = 151)

I think there is no such requirement. Could you check?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+
+fpmsyncd and fdbsyncd also need to wait for some time for the BGP protocol to converge and populate the routing info from peers, before they can reconcile their respective app_db entries. Hence they start a fixed 120 seconds timer once BGP is started. Once this timer expires and orchagent is reconciled, they start the app_db reconciliation process.
+
+The below detailed sequence diagram explains the interaction during the restore, replay and reconcile phase.
+
+![System Warm reboot and reconciliation sequence](images/warmrebootseq2.PNG "Figure 17: System Warm reboot and reconciliation sequence detail")
+
+__Figure 17: Warm reboot replay sequence__
+
+To achieve this dependency new states "replayed", "reconciled"  and "disabled" are added to the WARM_RESTART_TABLE  table in STATE_DB.  The value "disabled" is set when the subsystems perform a cold restart.
```
qiluo-msft, line 1230: >"disabled" [](start = 68, length = 10)

Why "disabled" is added to achieve this dependency?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
rpjain-nw: This mechanism has been put in place in this design to ensure that the orchagent/syncd/SAI restores the s/w states, h/w states to "pre warmboot states".
The "delta" changes during the warmboot process exchanged from all the remote EVPN peers are replayed later as an update to ensure that the syncd/SAI updates to the latest available state.
In this process, the following things happen in parallel.
After all the process like vlan/vxlan/intf/vrf has replayed the existing config and reconciled with kernel/app_db, the FRR/BGP starts the process of communicating with the remote peers to send/receive the latest underlay/overlay mac/arp/route information.
During the same time, the orchagent/syncd/SAI, tries to restore the "pre-warmboot" state based on the app/asic DB.
Once both the above process are finished (one may finish before the other), if there is any delta changes, they are replayed and reprogram the hardware.
With the above design approach, there is a better control and deterministic behavior.
Please let us know if you have any further questions.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+# 1 Introduction and Scope
+
+This document describes the Functionality and High level design of the EVPN VXLAN feature.
+
+The following are the benefits of an EVPN approach for VXLAN.
+
+ - Standards based and Interoperable solution.
+ - Auto discovery of remote VTEPs, Auto provisioning of tunnels and VLANs over VXLAN tunnels.
+ - Support for L2 and L3 VPN services.
+ - Allows for dual homing support.
```
heidinet2007, line 102: But dual homing support is "outside the scope of this document" ?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+ - Control plane MAC learning and ARP suppression leads to reduced flooding over an EVPN network.
+ - Eases planning and configuration when supporting downstream assigned VNI for DCI usecases.
+ - Aids in VM mobility.
+
+In the current implementation, there is support for L3 forwarding over VXLAN and EVPN Type 5 route support based on VNET constructs.
```
heidinet2007, line 107: This VNET based EVPN type 5 only works with some ASIC SAI, but not all, right?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+The numbers specified here serve as a general guideline. The actual scale numbers are dependent on the platform.
+
+## 2.4 Warm Boot Requirements
+
+Warm reboot is intended to be supported for the following cases:
+
+- Planned system warm reboot. 
+- Planned SwSS warm reboot.
+- Planned FRR warm reboot.
+
```
heidinet2007, line 193: Do you see any VxLAN specific issue to support unplanned FRR warm reboot?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+- [Support for L2VXLAN](https://github.com/opencomputeproject/SAI/pull/1025)
+- [Support for ARP/ND Suppression](https://github.com/opencomputeproject/SAI/pull/1056)
+
+## 4.2 DB Changes
+
+At a high level below are some of the interactions between relevant components and the DB involved for EVPN L2 VXLAN support in SONiC architecture.
+
+![EVPN L2 VXLAN Arch](images/arch.PNG "Figure : Arch")
+__Figure 4: EVPN L2 VXLAN DBs__
+
```
heidinet2007, line 292: There should be an arrow from zebra to kernel for remote MAC (type2) ?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+```
+;New table
+;holds the tunnel name specifying the VTEP source IP used for BGP-EVPN discovered tunnels
+
+key = VXLAN_EVPN_NVO|nvo_name
+                          ; nvo or VTEP name as a string
+; field = value
+source_vtep = tunnel_name ; refers to the tunnel object created with SIP only.
+```
+
```
heidinet2007, line 314: Can we have an example of config_db, similar to https://github.com/prsunny/SONiC/blob/f85f2b178743343b2c26d0a3e21392bdaae9745f/doc/vxlan/Vxlan_hld.md#4-example-configuration

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+Schema:
+
+```
+;Stores a list of routes
+;Status: Mandatory
+
+key = ROUTE_TABLE:VRF_NAME:prefix ;
+nexthop = prefix,              ; IP addresses separated ',' (empty indicates no gateway). May indicate VXLAN endpoint if vni_label is non zero.
+intf = ifindex? PORT_TABLE.key ; zero or more separated by ',' (zero indicates no interface)
+vni_label = VRF.vni            ; New Field. zero or more separated by ',' (empty value for non-vxlan next-hops). May carry MPLS label in future.
```
heidinet2007, line 459: Does the "VRF.vni"  have to match the l3vni configured for "VRF_NAME"?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+
+
+MAC addresses learnt in the forwarding plane are populated by fdborch in the FDB_TABLE in STATE_DB. The Linux FDB learns only those MAC addresses which are CPU bound (e.g. from ARP request/response). In order to provide EVPN L2 service, all of the MAC addresses learnt in the forwarding plane need to be advertised.
+
+In order to achieve above, new process fdbsyncd is added in SwSS. Fdbsyncd would subscribe to FDB entries for a given VLAN once VLAN-VNI mapping is available. Fdbsyncd will maintain the VLAN subscription and send FDB events to Linux FDB.
+
+![EVPN MAC route handling](images/l2macflowcombined.PNG )
+__Figure 8: EVPN MAC route handling__
+
+
```
heidinet2007, line 783: Any performance concern here to use kernel netlink message to parse all MAC?  It's well known that when the number of routes in network becomes large,  the number of kernel notification messages becomes a bottleneck and would be unreliable, and kernel expecting applications to handle that.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1352 @@
...
+ bgp graceful-restart
+ bgp graceful-restart preserve-fw-state
+ bgp graceful-restart stalepath-time 600  ( default 720)
+ bgp graceful-restart restart-time 300    ( default 240)
+```
+
+
+
+2. BGP Docker Warm reboot for EVPN
+   The goal of the BGP Docker warm reboot is to achieve BGP docker restart without impacting the dataplane. This is achieved with the help of graceful restart of the control plane (BGP).  As soon as BGP docker restarts, zebra re-learns the EVPN entries from kernel and marks them as stale. BGP protocol then re-learns the entries from the BGP peers and marks them for reconcile in zebra.  At the end of retention timer zebra  sweeps stale entries from the kernel. The remaining reconciliation sequence is same as the system warm-reboot, except that in this case only BGP docker and related tables are involved and data is already available in the DB and kernel.
```
heidinet2007, line 1268: Why " zebra re-learns the EVPN entries from kernel"?  appDB has all the routes, no need for Zebra to read them from kernel.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: That was an introductory para listing the benefits of EVPN in general and not specific to what is being implemented. The subsequent paragraphs mention that this is not in the scope of the implementation.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
tapashdas: Yes. That's right.  This is the VNI associated with VRF (L3 VNI).

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: This statement was from a SONiC perspective.  Compliance to SAI and Integration with SONiC per ASIC is outside the scope of this document.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: Will give an example.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: will update

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
srj102: Will update.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
nkelapur: will add more info there

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
nkelapur: The Local MACs will be present in the Local FDB table in StateDB during and after warm-reboot. During warm-reboot reconciliation, these Local entries are reprogrammed into the kernel.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
nkelapur: Disabled denotes cold restart.  When the system is restarted with cold-reboot, the warm-reboot status will be set to "disabled" for all the modules. Hereafter when only one docker ex. BGP goes for warm-reboot, it can read the status as "reconciled" or "disabled" for the other docker modules on which it depends and just continue processing.  For this reason the state "disabled" is added to imply that BGP in this case need not wait for reconcile of the other module

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
nkelapur: In normal processing, zebra learns the Local EVPN routes from kernel and gives the info to BGP, which then sends these EVPN routes to the BGP EVPN peers. After wam-reboot, the local routes in BGP/Zebra are lost and hence zebra needs to learn them again from kernel so that it can give that info to BGP. BGP will then re-send these to its EVPN peers so that the BGP-GR process can successfully happen for its peers.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
nkelapur: No, we don't see any issue with VXLAN for unplanned FRR warm-reboot. VXLAN listens to APP-DB and until FRR is able to populate/reconcile APP-DB entries correctly, there should be no impact to VXLAN

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
heidinet2007: thanks for the clarification

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
kishorekunal01: We have not seeing much issue with 40K MAC handled from Kernel, We have tested on AS7326(TD3).

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
heidinet2007: So this "relearn from Zebra" is for local routes only? not including routes advertised by its peers?

prsunny: WG discussion: 
"-R" retains the routes but if this option is not present, kernel cleans up the routes when zebra dies. Suggest to add an option for retaining vs cleaning. If its without "-R" option. This would require fdbsyncd to be aware of bgp docker warmrestart. In this release, we can go with the default without "-R" option

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1434 @@
...
+VLAN|Vlan{{vlan_id}}
+    "vlanid" : {{vlan_id}}
+
+VXLAN_TUNNEL_MAP|{{source_vtep_name}}|{{tunnel_map_name}}
+    "vni" : {{vni_id}}
+    "vlan" : {{vlan_id}}
+
+VRF|{{vrf_name}}
+    "l3vni" : {{vni_id}}
+
```
heidinet2007, line 588: Should the field be "vni", instead of "l3vni", as described above?
key = VRF_TABLE:VRF_NAME ;
fallback = "true"/"false"
vni = 1*8DIGIT ; VNI associated with VRF

tapashdas: yes. will update the document.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
```diff
@@ -0,0 +1,1434 @@
...
+        "vtep1|map_50_Vlan5": {
+            "vlan": "Vlan5",
+            "vni": "50"
+        },
+        "vtep1|map_60_Vlan6": {
+            "vlan": "Vlan6",
+            "vni": "60"
+        }
+    }
+
```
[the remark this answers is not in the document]
tapashdas: We need to configure to determine the Vlan (Router) interface associated with L3 VNI.
Based on VXLAN_TUNNEL_MAP we bring the Vlan interface Operationally up. 
Zebra advertise Prefix route only if Vlan interface associated with L3 VNI is Operationally Up.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
heidinet2007: Can both VNET based tunnel and VRF based tunnel be created simultaneously on a switch?  If not, any warning message when a user does so (mostly by mistake, say when migrating from VNET to VRF)?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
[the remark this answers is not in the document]
heidinet2007: If a router interface is already associated with a VRF, but not a vlan, does it still need to be added into VXLAN_TUNNEL_MAP?

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
tapashdas: VNET co-exist with VRF-L3VNI. Currently no such checks.

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
prsunny: WG discussion: (10/23):
1. If there are no l2, then there is no additional wait time. The wait is only for the local address to be populated. 
2. Wait time:  There is a fixed wait time of 240sec, @nikhil can you confirm on this? Default BGP time is 120sec
3. User to ensure bgp graceful restart time and neighbor recovery timing are aligned

prsunny:
WG request for timing sequence during warmboot for the following scenarios:
1. No EVPN configuration
2. EVPN configuration with only Layer 3
Broadcom shall provide these numbers using the optimized image

nkelapur:
> WG request for timing sequence during warmboot for the following scenarios:
> 
> 1. No EVPN configuration
> 2. EVPN configuration with only Layer 3
>    Broadcom shall provide these numbers using the optimized image

Time required for BGP/fpmsynd to start replay/reconciliation
1) 500 L2 VNIs :  50 seconds
2) 500 L3 VNIs : 240 seconds
3) 500 L3 and 500 L2 VNIs : 245 seconds

Without EVPN config
4) 1K ipv4 routes, 200  Vrfs : 47 seconds
Without dependency wait for BGP/fpmsyncd
5) 1K ipv4 routes, 200  Vrfs : 28 seconds

Please note that the these tests were done of broadcom build which uses netlink interface for kernel programming. The time required for L3VNIs is more and is attributed to over all more config ( 500 Vrf , 1000 Vlans and corresponding BGP instances and the corresponding increase in reconciliation of Vlanmgr, VrfMgr, orchagent etc)

doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
nkelapur: @prsunny   Please find my reply below
1. Yes, no additional wait time when no L2 entries.
2. There is no fixed wait time. Just for fdbsyncd the reconcile time is 120 seconds and yes for BGP GR the default reconnect time is set to 120 seconds ( basically not change there from existing behaviour)
3. Yes, for EVPN config, to account for the dependency wait, GR timers need to be increased accordingly

prsunny (approved):

PR #448, opened 2019-08-22
Sonic yang guide

praveen-li:
This document lists the guidelines, which will be used to write Yang Modules for Sonic. These Yang Modules will be primarily based on ABNF Schema of Sonic.


doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+```
+container acl-entries {
+    description "Access list entries container";
+    list acl-entry {
+        key "sequence-id";
+```
+
+
+
+### 7.  All must, when, pattern and enumeration constraints will come from .h files (or code). If code has the possibility to have unknown behavior with some config, then we will put a constraint in yang models objects. (This will always be Ongoing effort).
```
dutta-partha: I would prefer to say "must, when, leafref, pattern and enumeration constraints can be derived from .h files (or code logic)." 
Developer can put any additional constraint to stop invalid configuration (for which there might not be any code logic)

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+                        type inet:ipv4-prefix; <<<<
+                    }
+
+                    leaf DST_IP {
+                        type inet:ipv4-prefix;
+                    }
+```
+
+
+## All above steps are manual right now, in future few of them can be automated.
```
dutta-partha: This needs to be clarified. Are we saying that ABNF to YANG schema translation will be automated to some extend ?

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Revision ###
+
+ | Rev |     Date    |       Author       | Change Description                |
+ |:---:|:-----------:|:------------------:|-----------------------------------|
+ | 1.0 | 22 Aug 2019             |      Praveen Chaudhary      | Initial version                   |
+
+
+
+## This document lists the guidelines, which will be used to write Yang Modules for Sonic. These Yang Modules will be primarily based on ABNF Schema of Sonic.
```
dutta-partha: May be we should just mention that RFC 7950 should be base reference point for writing SONiC YANG model.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+        ......
+    ......
+            leaf IP_TYPE {
+                type sonic-head:ip_type;
+            }
+    }
+}
+``` 
+
+### 4.  if an object is used as a key in Redis, it should be a key in yang model, irrespective of need. Exception must be recorded in Yang Model.
```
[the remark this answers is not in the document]
renukamanavalan: Where & when do we see a scope for exception?

If none, we could skip the "exception ...".

In future, if there ever would be a need, we can revisit.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+
+### 5.  It is best to categorized Yang Modules based on a networking components. For Example, it is good to have separate modules for VLAN,  ACL, PORT and IP-ADDRESSES etc.
+```
+sonic-acl.yang
+sonic-interface.yang
+sonic-port.yang
+sonic-vlan.yang
+```
+
+### 6. When it comes about the choice between objects, it is good to stick to standard models.
```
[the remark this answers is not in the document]
renukamanavalan: I don't understand point # 6.

redis/ABNF schema is very clear and single hierarchy. The CVL is sticking to ABNF. 
You follow the single model of describing ABNF in YANG, irrespective of table.
I can see some grouping of related tables in one YANG module. Other than that, I don't see a scope for choice.

Please help me understand.

doc/Yang Model Guidelines.md
renukamanavalan: I don't get it either. This is a one time task. In future, when we add/update/remove, we work on the YANG, which is a 1:1 representation of ABNF schema.

The CVL YANG is such that it is just a representation of ABNF schema, to the extent, it should be feasible to write a simple tool to get ABNF schema from YANG. Other way round may not be possible, as YANG could have constraints laid out along with some metadata to meet mgmt FW requirements.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+                    leaf SRC_IP {
+                        type inet:ipv4-prefix; <<<<
+                    }
+
+                    leaf DST_IP {
+                        type inet:ipv4-prefix;
+                    }
+```
+
+
```
dutta-partha: @praveen-li We need to have guidelines how to model a repeating key : 
Example :
"BUFFER_PG": {
    "Ethernet0,Ethernet1,Ethernet2|0": {
        "profile": "[BUFFER_PROFILE|ingress_lossy_profile]"
    },

dutta-partha: @praveen-li We need to have guideline how to model mapping tables .

Example : 
        "TC_TO_QUEUE_MAP": {
                "AZURE": {
                        "1": "1",
                        "0": "0",
                        "3": "3",
                        "4": "4"
                }
        }

Please refer to "How to write CVL/SONiC Northbound YANG" section in Mgmt Framwork HLD PR https://github.com/Azure/SONiC/pull/436/files

dutta-partha: @praveen-li There should be guideline covering how to write read-only (config false) nodes, rpc and notification schema nodes.

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
dutta-partha: Is it possible to get permission to modify the MD file itself so that we can directly modify it ?

doc/Yang Model Guidelines.md
dutta-partha: Yes, we will bring some guidelines from the section "How to write CVL/SONiC Northbound YANG".

dutta-partha: So, should we throw error if repeating key is there?

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
dutta-partha: @praveen-li When we give the interface table data to 'libyang' by skipping ip-prefix key, it will throw syntax as all keys should be provided.

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
praveen-li: Yes Sharing is done already. 
https://drive.google.com/drive/folders/1FL1_cjxlKxMzU0K5UPz5RDsdM8YWTpYR
You should be able to edit it in any online editor.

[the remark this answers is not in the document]
praveen-li: Address in V2.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+### 1. Each primary section of Config DB (i.e a dictionary in config_db.json) for Example, VLAN, VLAN_MEMBER, INTERFACE  in config DB will be mapped to a container in Yang model.
+
+Example:
+
+#### ABNF
+
+```
+"INTERFACE": {
```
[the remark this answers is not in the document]
praveen-li: From Schema file on github, I see only APP DB entries not Config DB entries. So, Let me know If you can contribute here. Else ABNF Data instances are good enough.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+Example:
+Leaf names are same PACKET_ACTION, IP_TYPE and PRIORITY.
+
+#### ABNF
+```
+        "NO-NSW-xxx|xxx": {
+            "PACKET_ACTION": "FORWARD",
+            "IP_TYPE": "IPv6ANY",
+            "PRIORITY": "9xxxx20",
+            .....
```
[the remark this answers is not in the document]
praveen-li: This is conventional way on Programing books, to show that few more lines may exist here, But in current context, these line are irrelevant. 

Let me know if any better way to show it.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+####  Yang
+--
+```
+container INTERFACE {
+~~
+~~
+}
+```
+
+### 2. Each leaf in yang module should have same name as corresponding key in Config DB.
```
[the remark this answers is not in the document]
praveen-li: Data instance ideally must have same key as ABNF Schema. But I feel, there is no ABNF schema for many tables in config DB in below file
https://github.com/Azure/sonic-swss/blob/master/doc/swss-schema.md

here, when I refer key, it is for leafs, for example:
"scope": "global", [scope is key here for 'global' value]

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+            "PRIORITY": "9xxxx20",
+            .....
+            ....
+        },
+```
+
+####  Yang
+```
+            leaf PACKET_ACTION {
+                type xxxx;
```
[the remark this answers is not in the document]
praveen-li: I do not want any interpretation of type here, the point is about LEAF name. That`s why type is xxx. I think, we must either put ...... or xxx, but actual type value is open for interpretation.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
...
+            }
+            leaf PRIORITY {
+                type uint32 {
+                    range xxxx;
+                }
+            }
+```
+
+
+### 3. Hierarchy of the an objects in yang models will be same as for all the objects at same hierarchy in Redis. If any exception is created then it must be recorded properly.
```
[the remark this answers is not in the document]
praveen-li: I would prefer to use "Data Node/Object Hierarchy " instead of "Hierarchy" ....AGREE.

I strongly believe we also must have a top level container (along with container for every list) for every model...AGREE NOW. 
Let`s add this in Guideline please. Since we want the name of the container should be same as module, please add that as well. Thanks.

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
praveen-li: Right now, we have 2 scenarios, 1.) ip-prefix as key and 2.) VRF configuration where only interface is key.
I feel, this scenario is tricky to handle. 
One option is: We can make if-name + ip-prefix + vrf name as key, Where while translation to YANG (or with default option in YANG), we can populate non specified field as empty, that way key will be unique.

For Example:
"Ethernet1"|""|"Vrf-red"
"Ethernet2"|""|""
"Ethernet0|"11.11.11.1/24"|"",
"Ethernet0|12.12.12.1/24"|"",
"Ethernet3"|""|"" 
All 2 unique and all valid as per: https://github.com/Azure/SONiC/blob/master/doc/sonic-vrf-hld.md#the-schema-changes.

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
praveen-li: I feel, We can remove this point now, Since we have specified individual points for table names, Keys and for all leafs. What you guys think?

doc/Yang Model Guidelines.md
praveen-li: AGREE.

doc/Yang Model Guidelines.md
praveen-li: As per last comment, this is resolved already, I guess.

dutta-partha: @praveen-li I am not sure if we can mark this as resolved unless community confirms the same.

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
dutta-partha: @praveen-li Can you explain "with default option in YANG" a bit. Default value in YANG key is ignored as per RFC 7950. What should be the empty values be when we translate ABNF to YANG ?
One option is to have "default 0.0.0.0/0" in YANG as we have been discussing. But user must provide the default value also in REST API call, otherwise RESTCONF server rejects such request since it always expects all key values to be specified. So, if users sends  "0.0.0.0/0" as ip-prefix (may be in YANG 'description' we should mention that user must send default value for creating empty key), we know that it is default value in YANG and hence empty key item should be created in Redis DB.

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
dutta-partha: I agree.

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
renukamanavalan: I go for 0.0.0.0/0 for IP addresses. I wonder what we could for in places, where it is empty string.

Please suggest.

One option could be to have a reserved word indicating empty value. e.g. "None"
Can we confirm a way ?

doc/Yang Model Guidelines.md
renukamanavalan: AGREE

doc/Yang Model Guidelines.md
renukamanavalan: Can we please drop this comments as we all agree it is manual for now ?
Any future enhancement ideas, let us hold back, until we bake the first cut.

doc/Yang Model Guidelines.md
renukamanavalan: I am fine with Nested list.

Do we all agree?

renukamanavalan: Agree.

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
praveen-li: I checked that Libyang allows empty keys.

For Example:
            leaf ip-prefix {
                mandatory true;
                type union {
                        type inet:ip-prefix;
                        type string {
                                pattern "";
                       }
                }
           }

and Data:
			{
	        	"interface": "Ethernet8",
	        	"ip-prefix": "",
				"scope": "global",
				"family": "IPv4"
        	},

is acceptable. 
But since default keys are ignored, we need to find a way to auto-populate this.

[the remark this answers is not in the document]
renukamanavalan: Can you please elaborate on 'where we should/required-to auto-populate'?

doc/Yang Model Guidelines.md
praveen-li: This will be added by Partha with another PR, since it is already recorded Mgmt Framwork HLD PR.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,253 @@
+#   YANG MODEL GUIDELINES
```
dutta-partha: Please use SONiC YANG model instead of YANG model throughout.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,248 @@
+#   SONiC YANG MODEL GUIDELINES
+
+### Rev 1.0 ###
+
+### Revision ###
+
+ | Rev |     Date    |       Author       | Change Description                |
+ |:---:|:-----------:|:------------------:|-----------------------------------|
+ | 1.0 | 22 Aug 2019             |      Praveen Chaudhary      | Initial version                    |
```
renukamanavalan: We could leave at 1.0 as this is yet to be committed.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,248 @@
...
+ | 1.1 | 29 Aug 2019             |      Praveen Chaudhary      | Addressed review comments.
+
+
+
+###  This document lists the guidelines, which will be used to write YANG Modules for SONiC. These YANG Modules will be primarily based on or represent the ABNF Schema of SONiC, and the syntax of YANG models must follow RFC 7950.
+
+### These YANG models will be used to verify the configuration for SONiC switches, so a library which supports validation of configuration on SONiC Switch must use these YANG Models. List of such Libraries are: 1.) Configuration Validation Library. (CVL)
```
renukamanavalan: Can we also add that the same YANG file is used at NB too, with possibly additional metadata as required ?

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,248 @@
...
+            "PRIORITY": "0"
+        }
+}
+```
+####  YANG
+In YANG, "Family" of VLAN_INTERFACE and "IP_TYPE" of ACL_RULE is at same level.
+```
+container VLAN_INTERFACE {
+        description "VLAN_INTERFACE part of config_db.json";
+        list xxx {
```
renukamanavalan: Here we need to have a convention for list name. If I recall, from last meeting, we were in favor of something like <container name>_LIST, which in this case VLAN_INTERFACE_LIST

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,248 @@
...
+    ......
+            leaf family {
+                type sonic-head:ip-family;
+            }
+        }
+}
+
+container ACL_RULE {
+        description "ACL_RULE part of config_db.json";
+        list xxx {
```
renukamanavalan: Same here.  Follow convention for list name.

doc/Yang Model Guidelines.md
```diff
@@ -0,0 +1,248 @@
...
+        "Vlan100|Ethernet0": {<<<< KEYS
+            "tagging_mode": "untagged"
+        }
+   }
+```
+#### YANG
+```
+container VLAN_MEMBER {
+    description "VLAN_MEMBER part of config_db.json";
+        list xxx {
```
renukamanavalan: Same here. List name

doc/Yang Model Guidelines.md
```diff
@@ -220,13 +214,13 @@ leaf L4_DST_PORT_RANGE {
         }
 }
 ```
-### 8. Comment all must, when and patterns conditions.
+### 7. Comment all must, when and patterns conditions.
```
dutta-partha: Comment all must, when and patterns conditions using 'description' keyword.

doc/Yang Model Guidelines.md
```diff
@@ -250,4 +244,5 @@ Example:
 ```
 
 
-## All above steps are manual right now, in future few of them can be automated.
+## All above steps are manual right now, in future few of them can be automated. Below tasks are automated already:
```
dutta-partha: I believe this is not a guideline step and should be moved to a different section.

doc/Yang Model Guidelines.md
```diff
@@ -237,7 +231,7 @@ leaf family {
 
 
 
-### 9.  Use IETF data types for leaf type first if applicable (RFC 6021) . Declare new type (say Sonic types) only if IETF type is not applicable. All Sonic Types must be part of same header type or common yang model.
+### 8.  Use IETF data types for leaf type first if applicable (RFC 6021) . Declare new type (say SONiC types) only if IETF type is not applicable. All SONiC Types must be part of same header type or common YANG model.
```
dutta-partha: "Declare new type in common YANG model (say SONiC types) only if IETF type is not applicable."

We can remove "All SONiC types must be part of same header type or common YANG model."

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
dutta-partha: @praveen-li We don't have a control if REST API client chooses to skip the key and then REST Server will reject the request before we auto populate it inside management framework.

doc/Yang Model Guidelines.md
```diff
@@ -39,7 +42,7 @@ container INTERFACE {
 }
 ```
 
-### 2. Each leaf in yang module should have same name as corresponding key in Config DB.
+### 2. Each leaf in YANG module should have same name as corresponding key in Config DB.
```
dutta-partha: Word 'key' here is confusing. Please rephrase the sentence accordingly.

doc/Yang Model Guidelines.md
```diff
@@ -71,10 +74,10 @@ Leaf names are same PACKET_ACTION, IP_TYPE and PRIORITY.
 ```
 
 
-### 3. Hierarchy of the an objects in yang models will be same as for all the objects at same hierarchy in Redis. If any exception is created then it must be recorded properly.
-For Example: 
+### 3. Data Node/Object Hierarchy of the an objects in YANG models will be same as for all the objects at same hierarchy in Redis. If any exception is created then it must be recorded properly.
```
dutta-partha: The step does not tell how to record exception. Will the exception be understood by processing library (e.g. CVL). Please provide example of exception and how to record the same.

doc/Yang Model Guidelines.md
```diff
@@ -116,11 +119,11 @@ container ACL_RULE {
             }
     }
 }
-``` 
+```
 
-### 4.  if an object is used as a key in Redis, it should be a key in yang model, irrespective of need. Exception must be recorded in Yang Model.
+### 4.  if an object is used as a key in Redis, it should be a key in YANG model. Exception must be recorded in YANG Model.
```
dutta-partha: ###4. If

Also provide steps how to record such exception with an example.

doc/Yang Model Guidelines.md
```diff
@@ -140,27 +143,18 @@ container VLAN_MEMBER {
 }
 ```
 
-### 5.  It is best to categorized Yang Modules based on a networking components. For Example, it is good to have separate modules for VLAN,  ACL, PORT and IP-ADDRESSES etc.
+### 5.  It is best to categorized YANG Modules based on a networking components. For Example, it is good to have separate modules for VLAN,  ACL, PORT and IP-ADDRESSES etc.
```
dutta-partha: Please change to ... "It is best to categorize YANG Modules based on a networking components. For example ......."

doc/Yang Model Guidelines.md
```diff
@@ -140,27 +143,18 @@ container VLAN_MEMBER {
...
-container acl-entries {
-    description "Access list entries container";
-    list acl-entry {
-        key "sequence-id";
-```
-
 
 
-### 7.  All must, when, pattern and enumeration constraints will come from .h files (or code). If code has the possibility to have unknown behavior with some config, then we will put a constraint in yang models objects. (This will always be Ongoing effort).
+### 6.  All must, when, pattern and enumeration constraints can be derived from .h files (or code). If code has the possibility to have unknown behavior with some config, then we should put a constraint in YANG models objects. Also, developr can put any additional constraint to stop invalid configuration.
```
dutta-partha: I believe we need to explain "unknown behavior with some config" a bit here, with an example. This will help developer to think on appropriate constraint in this context.

"Developer"...

doc/Yang Model Guidelines.md
```diff
@@ -116,11 +119,11 @@ container ACL_RULE {
...
     }
 }
-``` 
+```
 
-### 4.  if an object is used as a key in Redis, it should be a key in yang model, irrespective of need. Exception must be recorded in Yang Model.
+### 4.  if an object is used as a key in Redis, it should be a key in YANG model. Exception must be recorded in YANG Model.
 
-Example: VLAN_MEMBER dictionary in ABNF.json has both vlan-id and Port part of the key. So yang model should have the same keys.
+Example: VLAN_MEMBER dictionary in ABNF.json has both vlan-id and Port part of the key. So YANG model should have the same keys.
```
dutta-partha: ABNF.json or config_db.json ?

doc/Yang Model Guidelines.md
dutta-partha: What if a new feature is developed, developer may first with writing YANG model and there won't be any .h file or code. Please clarify this in the step.

doc/Yang Model Guidelines.md
```diff
@@ -140,27 +143,18 @@ container VLAN_MEMBER {
...
-    description "Access list entries container";
-    list acl-entry {
-        key "sequence-id";
-```
-
 
 
-### 7.  All must, when, pattern and enumeration constraints will come from .h files (or code). If code has the possibility to have unknown behavior with some config, then we will put a constraint in yang models objects. (This will always be Ongoing effort).
+### 6.  All must, when, pattern and enumeration constraints can be derived from .h files (or code). If code has the possibility to have unknown behavior with some config, then we should put a constraint in YANG models objects. Also, developr can put any additional constraint to stop invalid configuration.
+(This will always be Ongoing effort).
```
dutta-partha: I wonder why we need even to mention "(This will always be Ongoing effort)." If new constraint has to be added, instead of modifying the code, developer will anyway add the same in YANG model as part of maintenance or bug fixing.

doc/Yang Model Guidelines.md
```diff
@@ -1,20 +1,23 @@
-#   YANG MODEL GUIDELINES
+#   SONiC YANG MODEL GUIDELINES
```
dutta-partha: Please also rename the file "SONiC YANG Model Guidelines.md"

doc/Yang Model Guidelines.md
```diff
@@ -1,20 +1,23 @@
...
 
 ### Rev 1.0 ###
 
 ### Revision ###
 
  | Rev |     Date    |       Author       | Change Description                |
  |:---:|:-----------:|:------------------:|-----------------------------------|
- | 1.0 | 22 Aug 2019             |      Praveen Chaudhary      | Initial version                   |
+ | 1.0 | 22 Aug 2019             |      Praveen Chaudhary      | Initial version                    |
```
dutta-partha: We have not released "1.0" major version yet. Should we change to 0.1, 0.2 etc.

doc/Yang Model Guidelines.md
```diff
@@ -1,20 +1,23 @@
...
+ | 1.1 | 29 Aug 2019             |      Praveen Chaudhary      | Addressed review comments.
 
 
 
-## This document lists the guidelines, which will be used to write Yang Modules for Sonic. These Yang Modules will be primarily based on ABNF Schema of Sonic. 
+###  This document lists the guidelines, which will be used to write YANG Modules for SONiC. These YANG Modules will be primarily based on or represent the ABNF Schema of SONiC, and the syntax of YANG models must follow RFC 7950.
```
dutta-partha: We should also  clearly state where user should start from - ABNF schema and/or ABNF JSON i.e. Redis JSON data in config_db.json.

doc/Yang Model Guidelines.md
renukamanavalan: I would like to drop these, as this YANG is where we put the schema  and ***the primary source*** for all other requirements. Like CVL YANG is generated from this using a build time tool. The header files would be generated off this.

I really don't see scope for this YANG getting auto generated, which would imply one another source for schema, which is none in my knowledge.

doc/Yang Model Guidelines.md
renukamanavalan: Agree.

doc/Yang Model Guidelines.md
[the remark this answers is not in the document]
renukamanavalan: @dutta-partha  -- Please share your suggestions on optional-key handling.

doc/Yang Model Guidelines.md
renukamanavalan: We could drop this line "_If code has the possibility to have unknown behavior with some config, then we should put a constraint in YANG models objects._ ", as this implies a ***missed*** validation in code.  This is part of reading code to understand constraints, some explicitly present and some implicit (like unknown behavior)

doc/Yang Model Guidelines.md
renukamanavalan: I guess Praveen refer, [Azure/SONiC/Wiki/Configuration](https://github.com/Azure/SONiC/wiki/Configuration#vlan).

doc/Yang Model Guidelines.md
renukamanavalan: Agree. This is true, for the first cut of each YANG module, as we learn schema from multiple sources that include header, code, .... Going ahead, the YANG will drive/generate the header files.

doc/Yang Model Guidelines.md
renukamanavalan: Agree.

doc/Yang Model Guidelines.md
renukamanavalan: Agree. 
As this is not committed yet, you may not increment for adding comments. 
It can be 0.1
Subsequent PRs can increment.

praveen-li:
>>>What if a new feature is developed, developer may first start with writing YANG model and there won't be any .h file or code. Please clarify this in the step.

Ideally, we should not allow config for new feature if back-hand code does not exist. Else back-hand will crash. So code come first and then YANG model. (.h) files can be autogenerated, but back-hand code should be present to handle new config data.

praveen-li:
>>>>We could drop this line "If code has the possibility to have unknown behavior with some config, then we should put a constraint in YANG models objects. ", as this implies a missed validation in code. This is part of reading code to understand constraints, some explicitly present and some implicit (like unknown behavior)

It may not be necessary a missed validation code, For example we do not want lots of code to verify a pattern of a string in back-hand. So it is better to handle it with YANG validation. 
From Code, if we know that additional constraints must be added in YANG then we must add them proactively. Thanks.

praveen-li:
>We can remove "All SONiC types must be part of same header type or common YANG model."

I think all Sonic type definition must be in single file so that importing them is easy. Also, there will be one place to look for. Thanks.

praveen-li:
> Should we comment all must, when and patterns conditions using 'description' keyword ?

In YANG, description mainly describes the purpose of the node. So we may use a comment. 
But if we have a way to split out description part till user level (Say KLISH CLI level) in case of the wrong config, then It is beneficial. Because user does not have to decode must statement.

renukamanavalan:
@Praveen Chaudhary<mailto:pchaudhary@linkedin.com>,

Ideally when a new feature, brings in new schema, the objects created per new schema are unknown, hence ignored by backend, so we normally start with schema first.
This is sometimes become even a *requirement* to checkin schema *only* first, when that feature needs work from multiple teams.
With schema build/Make generate headers and we write code using it.

In cases, where we are amending a feature, like dropping/replacing/re-defining an attribute, we might need to do, schema changes + code changes in one commit. Even here, while the dev is coding, he updates schema first, run the Make to get the updated header(s) and then update code as needed.

In normal work flow, schema come first and the rest follow.

Thank you,
Renuka

renukamanavalan:
Agree. In fact, with this FW handling validation, any validation done by back end code would become redundant, hence can be dropped over time.

renukamanavalan:
Agree. Let us go with this model for *now*

praveen-li:
In that case, I feel, I can add below line:

"For new features, constraints may be derived from the low-level design document or from the requirement document of the feature."

renukamanavalan:
@Praveen Chaudhary<mailto:pchaudhary@linkedin.com>: Agree. Thank you!

doc/Yang Model Guidelines.md
dutta-partha: We agreed to have a top level container also e.g. sonic-vlan. I will add in next PR on top of this.

doc/SONiC_YANG_Model_Guidelines.md
```diff
@@ -0,0 +1,270 @@
...
+```
+container VLAN_MEMBER {
+    description "VLAN_MEMBER part of config_db.json";
+        list ..... {
+            key "vlanid port";<<<< KEYS
+       }
+}
+```
+
+### 5.  It is best to categorize YANG Modules based on a networking components. For Example, it is good to have separate modules for VLAN,  ACL, PORT and IP-ADDRESSES etc.
```
dev-aviz: Do we plan to model RPC's and Notifications? 
Can we put guidelines around rpc's and notifications?

dutta-partha: Please check latest PR https://github.com/Azure/SONiC/pull/464, step 23 and step 24.

PR #465, opened 2019-09-11
Functional spec to get CPU/memory utilization from process and docker 

pra-moh:
adding requirement document to collect process and docker CPU/memory utilization data via telemetry agent.


doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+"START"  
+"00:39"  
+"TIME"  
+"0:01"  
+"COMMAND"  
+"containerd-shim -namespace moby -workdir /var/lib/containerd/io.containerd.runtime.v1.linux/moby/07983d8d914904ac8054af2be0aa6aa70a8325700aa2588f7424ece3fbfe648c -address /run/containerd/containerd.sock -containerd-binary /usr/bin/containerd -runtime-root /var/run/docker/runtime-runc"  
+
+###### Docker stats
+
+$ docker stats --no-stream -a
```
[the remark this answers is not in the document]
lguohan: no, we are not.

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+For 1st part, Daemon code will be added under sonic-buildimage/files/image_config.  A Daemon will start when OS starts and upload Process and docker stats data every 2 mins to state-DB. Details of CLI and state-DB given below. 
+
+##### Part 2
+From state-DB data need to be available via telemetry agent
+
+##### CLI output  and corresponding structure in state-DB for process and docket stats
+
+###### Process stats
+
+$ ps aux
```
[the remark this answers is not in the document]
lguohan: joe is right. I think I have given the exact the command line we need for this purpose?

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+"?"  
+"STAT"  
+"SI"  
+"START"  
+"00:39"  
+"TIME"  
+"0:01"  
+"COMMAND"  
+"containerd-shim -namespace moby -workdir /var/lib/containerd/io.containerd.runtime.v1.linux/moby/07983d8d914904ac8054af2be0aa6aa70a8325700aa2588f7424ece3fbfe648c -address /run/containerd/containerd.sock -containerd-binary /usr/bin/containerd -runtime-root /var/run/docker/runtime-runc"  
+
```
lguohan: we need PID, PPID in the state db for process stats

lguohan: I suggest to add DOCKER in the process stats. If cannot find the docker for a process, then the value is NA

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+"snmp"  
+"CPU%"  
+"3.93"  
+"MEM USAGE"  
+"41.56MiB"  
+"LIMIT"  
+"7.784GiB"  
+"MEM%"  
+"0.52"  
+"NET I"
```
lguohan: NET I -> NET_IN

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+"3.93"  
+"MEM USAGE"  
+"41.56MiB"  
+"LIMIT"  
+"7.784GiB"  
+"MEM%"  
+"0.52"  
+"NET I"  
+"0B"  
+"NET O"
```
lguohan: -> NET_OUT

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+above output will be stored inside state-DB as follows:
+
+DockerStats|4dc60c743348    
+"NAME"  
+"snmp"  
+"CPU%"  
+"3.93"  
+"MEM USAGE"  
+"41.56MiB"  
+"LIMIT"
```
lguohan: MEM_LIMIT

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+|07983d8d9149 |   syncd |             41.89% |             291MiB / 7.784GiB  |   3.65%   |           0B / 0B |      83MB / 406kB |      33       |
+
+above output will be stored inside state-DB as follows:
+
+DockerStats|4dc60c743348    
+"NAME"  
+"snmp"  
+"CPU%"  
+"3.93"  
+"MEM USAGE"
```
lguohan: MEM

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+"41.56MiB"  
+"LIMIT"  
+"7.784GiB"  
+"MEM%"  
+"0.52"  
+"NET I"  
+"0B"  
+"NET O"  
+"0B"  
+"BLOCKI"
```
lguohan: BLOCK_IN

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+"7.784GiB"  
+"MEM%"  
+"0.52"  
+"NET I"  
+"0B"  
+"NET O"  
+"0B"  
+"BLOCKI"  
+"31.5MiB"  
+"BLOCKO"
```
lguohan: BLOCK_OUT

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+$ docker stats --no-stream -a
+
+|CONTAINER ID |     NAME|              CPU % |             MEM USAGE / LIMIT|   MEM % |           NET I/O|             BLOCK I/O |          PIDS|
+|-----------------|----------|-------------------|--------------------------------|-----------|-----------------|----------------------|-------------|
+|4dc60c743348 |    snmp |              3.93%  |            41.56MiB / 7.784GiB| 0.52%    |           0B / 0B |    31.5MB / 81.9kB|     7          |
+|07983d8d9149 |   syncd |             41.89% |             291MiB / 7.784GiB  |   3.65%   |           0B / 0B |      83MB / 406kB |      33       |
+
+above output will be stored inside state-DB as follows:
+
+DockerStats|4dc60c743348
```
lguohan: container id can change, the agent need to remove non existing container ids from the state db.

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+|root    | 4276 | 0.0    | 0.0     | 108816    | 5552 |?      |    Sl     |00:39 |0:01 | containerd-shim -namespace moby -workdir /var/lib/containerd/io.containerd.runtime.v1.linux/moby/07983d8d914904ac8054af2be0aa6aa70a8325700aa2588f7424ece3fbfe648c -address /run/containerd/containerd.sock -containerd-binary /usr/bin/containerd -runtime-root /var/run/docker/runtime-runc|
+|root    | 6601 |  0.0   |  0.0    |108816     | 5516 |?      |    Sl     |00:42 |0:01 |containerd-shim -namespace moby -workdir /var/lib/containerd/io.containerd.runtime.v1.linux/moby/4dc60c74334813d6c833d967b1196d1783b90bff0488aa0c35d544db66dc8a81 -address /run/containerd/containerd.sock -containerd-binary /usr/bin/containerd -runtime-root /var/run/docker/runtime-runc|
+
+above output will be stored inside state-DB as follows:
+
+ProcessStats|4276
```
lguohan: process id can change, the program needs to remove non existent process id from the state db.

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+"CPU"  
+"0.0"  
+"MEM"
```
lguohan: to be consistent with dockerstats, this should be MEM%

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,84 @@
...
+"TTY"  
+"?"  
+"STAT"  
+"SI"  
+"START"  
+"00:39"  
+"TIME"  
+"0:01"  
+"COMMAND"  
+"containerd-shim -namespace moby -workdir /var/lib/containerd/io.containerd.runtime.v1.linux/moby/07983d8d914904ac8054af2be0aa6aa70a8325700aa2588f7424ece3fbfe648c -address /run/containerd/containerd.sock -containerd-binary /usr/bin/containerd -runtime-root /var/run/docker/runtime-runc"
```
lguohan: need USER column

doc/system-telemetry/process-docker-stats.md
lguohan: in the worst case, there might be too many processes. It is better to have some limit, i am think should be limit to 1024 largest mem consumption processes and 1024 largest cpu consumption processes.

doc/system-telemetry/process-docker-stats.md
pra-moh: What way do you suggest? One way can be setting key expiry  time in state_db?

jleveque: If the daemon doesn't crash, then after it scrapes the latest data and before it uploads it to the DB, it can do a diff of the current DB contents and delete entries which no longer exist.

However, if the daemon were to ever crash, stale data could be present in the DB, and would be sent out via streaming telemetry. Maybe for this case it would be good to include a timestamp stating when the data was collected. This way any services dependent on this data can decide whether or not to use the data based on the age of the timestamp.

doc/system-telemetry/process-docker-stats.md
pra-moh: We are planning to do correlation between them at driver side so data going to kusto will be correlated

doc/system-telemetry/process-docker-stats.md
pra-moh: will add expiry at key level

jleveque: I think a timestamp is better than expiry. Consumer can determine what to do based on age of timestamp.

doc/system-telemetry/process-docker-stats.md
lguohan: the monitor process should be restarted if it crashes, otherwise we will not have process data updated in the database. for the timestamp, maybe we do not need to add for every entry since they are all updated at the same time.

jleveque: Agree it *should* be restarted, but just in case...

Also agree that one timestamp is sufficient.

doc/system-telemetry/process-docker-stats.md
pra-moh: Discussed with Joe. Adding  new entry in state DB as "Process_Stats|LastUpdateTime" and "Docker_Stats|LastUpdateTime"

doc/system-telemetry/process-docker-stats.md
pra-moh: Daemon on start will delete all entries from state db and then update. This way non existent process id won't persist in db. Adding new entry as "Process_stats|LastUpdateTime" to track timestamp. Added this in functional spec part1.

doc/system-telemetry/process-docker-stats.md
pra-moh: Agree. I'll look for ways to get top 1024 processes only.

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,95 @@
...
+Details of CLI and state-DB given below. 
+
+##### Part 2
+Verify that from state-DB data is available via telemetry agent
+
+##### CLI output  and corresponding structure in state-DB for process and docker stats
+
+###### Process stats
+
+$ ps -eo uid,pid,ppid,%mem,%cpu,stime,tty,time,cmd
```
jleveque: Please place all console-based input/output in code blocks (three backticks) for better formatting

doc/system-telemetry/process-docker-stats.md
jleveque: The concept here is to add a timestamp to the results. This way the consumer can decide whether or not the data is stale. Every time the daemon updates the DB, it will first clear the table of all old data, thus we need not worry about container IDs changing.

doc/system-telemetry/process-docker-stats.md
jleveque: I think it makes more sense for this daemon to be simple, and purely collect the raw data and push to the DB. Let the consumer do the correlation.

doc/system-telemetry/process-docker-stats.md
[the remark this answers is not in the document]
pra-moh: yes, used command given by Joe is same given earlier.

doc/system-telemetry/process-docker-stats.md
pra-moh: Updated

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,109 @@
...
+    0     1     0  0.1  0.0 Oct15 ?        00:00:09 /sbin/init
+    0     2     0  0.0  0.0 Oct15 ?        00:00:00 [kthreadd]
+    0     3     2  0.0  0.0 Oct15 ?        00:00:01 [ksoftirqd/0]
+    0     5     2  0.0  0.0 Oct15 ?        00:00:00 [kworker/0:0H]
+
+```
+above output will be stored inside state-DB as follows for largest 1024 CPU consumption processes:  
+
+```
+ProcessStats|4276
```
jleveque: In SONiC Redis DB, convention is for all table names to be `ALL_CAPS_SPACE_SEPARATED`. Please adjust accordingly.

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,109 @@
...
+"TIME"  
+"00:00:09"  
+"CMD"  
+"/sbin/init"  
+
+```
+Along with data new entry for timestamp will be updated in state_db:  
+
+```
+Process_Stats|LastUpdateTime
```
jleveque: Table name does not match the one in your above example. Also, in SONiC Redis DB, convention is for all table names to be `ALL_CAPS_SPACE_SEPARATED`. Please adjust accordingly.

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,109 @@
...
+CONTAINER ID        NAME                CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
+bf7e49f494ee        grpc                0.00%               0B / 0B               0.00%               0B / 0B             0B / 0B             0
+209c6e6116c6        syncd               10.82%              286MiB / 3.844GiB     7.26%               0B / 0B             0B / 639kB          32
+8a97fafdbd60        dhcp_relay          0.02%               12.15MiB / 3.844GiB   0.31%               0B / 0B             0B / 65.5kB         5
+
+```
+above output will be stored inside state-DB as follows:
+
+```
+DockerStats|bf7e49f494ee
```
jleveque: In SONiC Redis DB, convention is for all table names to be `ALL_CAPS_SPACE_SEPARATED`. Please adjust accordingly.

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,109 @@
...
+"0B"  
+"BLOCK_OUT"  
+"0B"  
+"PIDS"  
+"0"  
+```
+Along with data new entry for timestamp will be updated in state_db:  
+
+```
+Docker_Stats|LastUpdateTime
```
jleveque: In SONiC Redis DB, convention is for all table names to be `ALL_CAPS_SPACE_SEPARATED`. Please adjust accordingly.

pra-moh: updated

doc/system-telemetry/process-docker-stats.md
pra-moh: updated

doc/system-telemetry/process-docker-stats.md
pra-moh: updated

doc/system-telemetry/process-docker-stats.md
pra-moh: updated

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,109 @@
...
+"NET_IN"  
+"0B"  
+"NET_OUT"  
+"0B"  
+"BLOCK_IN"  
+"0B"  
+"BLOCK_OUT"  
+"0B"  
+"PIDS"  
+"0"
```
jleveque: The container you chose to create this example isn't a very good example case, as everything is 0. Can you remove grpc from your output above and use one of the other two containers here?

pra-moh: updated

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,109 @@
...
+PROCESS_STATS|LastUpdateTime  
+```
+
+###### Docker stats
+
+```
+$ docker stats --no-stream -a
+
+CONTAINER ID        NAME                CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
+bf7e49f494ee        grpc                0.00%               0B / 0B               0.00%               0B / 0B             0B / 0B             0
```
jleveque: Can you remove this 'grpc' container? It is not a standard SONiC container.

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,109 @@
...
+
+```
+above output will be stored inside state-DB as follows:
+
+```
+DOCKER_STATS|209c6e6116c6     
+"NAME"  
+"syncd"  
+"CPU%"  
+"10.82%"
```
jleveque: I don't think we should store the percent sign with the value. The key states that the value is a percentage.

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,109 @@
...
+"NAME"  
+"syncd"  
+"CPU%"  
+"10.82%"  
+"MEM"  
+"286MiB"  
+"MEM_LIMIT"  
+"3.844GiB"  
+"MEM%"  
+"7.26%"
```
jleveque: I don't think we should store the percent sign with the value. The key states that the value is a percentage.

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,109 @@
...
+above output will be stored inside state-DB as follows:
+
+```
+DOCKER_STATS|209c6e6116c6     
+"NAME"  
+"syncd"  
+"CPU%"  
+"10.82%"  
+"MEM"  
+"286MiB"
```
jleveque: I wonder whether we should store the units here like this, or if we should convert all values to a standard, like bytes and store that instead. Storing units here means that all consumers may have to implement similar logic to convert. And if something ever changes, all consumers would have to reflect that change.

This applies to "MEM", "MEM_LIMIT", "NET_IN", "NET_OUT", "BLOCK_IN", and "BLOCK_OUT".

pra-moh: I think exposing raw data as-is from the device is preferable so that if something changes then the consumer can understand and make the change. Thoughts?

doc/system-telemetry/process-docker-stats.md
pra-moh: removed

doc/system-telemetry/process-docker-stats.md
pra-moh: removed %

doc/system-telemetry/process-docker-stats.md
pra-moh: removed

doc/system-telemetry/process-docker-stats.md
pra-moh: updated as discussed

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,108 @@
...
+```
+above output will be stored inside state-DB as follows:
+
+```
+DOCKER_STATS|209c6e6116c6     
+"NAME"  
+"syncd"  
+"CPU%"  
+"10.82"  
+"MEM"
```
jleveque: Now that we've removed units from the value and standardized the value unit to bytes, we should probably add the unit in the key. E.g., `MEM_BYTES`

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,108 @@
...
+
+```
+DOCKER_STATS|209c6e6116c6     
+"NAME"  
+"syncd"  
+"CPU%"  
+"10.82"  
+"MEM"  
+"299892736"  
+"MEM_LIMIT"
```
jleveque: Now that we've removed units from the value and standardized the value unit to bytes, we should probably add the unit in the key. E.g., `MEM_LIMIT_BYTES`

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,108 @@
...
+"syncd"  
+"CPU%"  
+"10.82"  
+"MEM"  
+"299892736"  
+"MEM_LIMIT"  
+"4127463571.456"  
+"MEM%"  
+"7.26"  
+"NET_IN"
```
jleveque: Now that we've removed units from the value and standardized the value unit to bytes, we should probably add the unit in the key. E.g., `NET_IN_BYTES`

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,108 @@
...
+"10.82"  
+"MEM"  
+"299892736"  
+"MEM_LIMIT"  
+"4127463571.456"  
+"MEM%"  
+"7.26"  
+"NET_IN"  
+"0"  
+"NET_OUT"
```
jleveque: Now that we've removed units from the value and standardized the value unit to bytes, we should probably add the unit in the key. E.g., `NET_OUT_BYTES`

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,108 @@
...
+"299892736"  
+"MEM_LIMIT"  
+"4127463571.456"  
+"MEM%"  
+"7.26"  
+"NET_IN"  
+"0"  
+"NET_OUT"  
+"0"  
+"BLOCK_IN"
```
jleveque: Now that we've removed units from the value and standardized the value unit to bytes, we should probably add the unit in the key. E.g., `BLOCK_IN_BYTES`

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,108 @@
...
+"4127463571.456"  
+"MEM%"  
+"7.26"  
+"NET_IN"  
+"0"  
+"NET_OUT"  
+"0"  
+"BLOCK_IN"  
+"0"  
+"BLOCK_OUT"
```
jleveque: Now that we've removed units from the value and standardized the value unit to bytes, we should probably add the unit in the key. E.g., `BLOCK_OUT_BYTES`

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,108 @@
...
+```
+DOCKER_STATS|209c6e6116c6     
+"NAME"  
+"syncd"  
+"CPU%"  
+"10.82"  
+"MEM"  
+"299892736"  
+"MEM_LIMIT"  
+"4127463571.456"
```
jleveque: Fractional bytes make no sense. This is due to the value being converted to GiB for display and back to bytes; data was lost. Lets just round all byte values down to the nearest integer value.

hui-ma: +1

doc/system-telemetry/process-docker-stats.md
```diff
@@ -0,0 +1,108 @@
+# Process and docker stats availability via telemetry agent
+
+## Revision
+
+| Rev | Date     | Author      | Change Description |
+|:---:|:--------:|:-----------:|--------------------|
+| 0.1 | 09/12/19 | Pradnya Mohite | Initial version    |
```
hui-ma: please add an new revision number for the date of today. Then mentioned in the code that it is against when revision number. For further comments, you can address in the future code change not the existing one.

pra-moh: @hui lets discuss this offline

jleveque: I think it's fine leaving this as-is, because this is technically still the initial version , because the PR has not yet been merged.

PR #466, opened 2019-09-13
SONiC Open Config Platform Model Support HLD

fk410167:
This document describes the support for open config platform models in SONiC, based on the old platform APIs. This support is based on the enhanced 1.0 platform model APIs, and follows the highlighted design in 
https://github.com/Azure/SONiC/blob/master/doc/pmon/pmon-enhancement-design.md


fk410167:
We planned to modify the document including the platform 2.0 APIs and re-submit the PR.

PR #475, opened 2019-09-30
VRRP HLD

dks19:
Hi all,
    This is high level design for VRRP feature in SONiC. Currently, VRRPv2 along with interface tracking is supported. Please review and provide feedback/comments.

Thanks,
Dilip


doc/SONiC_VRRP_HLD.md
```diff
@@ -0,0 +1,582 @@
...
+
+
+# 1 Introduction and Scope
+
+This document describes the Functionality and High level design of the VRRP feature.
+
+The Virtual Router Redundancy Protocol (VRRP) functionality is designed to eliminate the single point of
+failure inherent in the static default routed environment. VRRP specifies an election protocol that
+dynamically assigns responsibility of gateway router to a VRRP instance on one of the routers on a LAN. The VRRP instance controlling the IP address(es) associated with a virtual router is called the Master, and routes the traffic. The election process provides dynamic fail-over in the forwarding responsibility should the Master become unavailable. Any of the virtual router's IP addresses on a LAN can then be used as the default first hop router by end-hosts. The advantage gained from using VRRP is a higher availability default path without requiring configuration of dynamic routing or router discovery protocols on every end-host.
+
```
dev-aviz: It would be good to list the out the possible use cases for the sonic user to enable to this feature? Can this work on data center MLAG kind of deployments ?

doc/SONiC_VRRP_HLD.md
```diff
@@ -0,0 +1,582 @@
...
+6. Support configurable priority for VRRP instance
+
+7. Support configuration and management of various VRRP parameters
+
+    
+
+Following requirements are beyond scope of this release. 
+
+1. VRRPv3 (IPv6) support
```
dev-aviz: what is preventing not to support vrrp3? 
How is it different from FRR VRRP? Since FRR is part of SONiC stack, can't we stick to FRR VRRP (one stack)?

doc/SONiC_VRRP_HLD.md
```diff
@@ -0,0 +1,582 @@
...
+
+VRRP control packets have IP protocol type as 112 (reserved for VRRP), and are sent to VRRP multicast address 224.0.0.18. Source MAC in VRRP control packets is virtual MAC address and source IP is interface IP.                                                               
+
+### 3.2.5    ARP Request Handling 
+
+Only master responds to the ARP requests for virtual IP address. In ARP replies sent by master, the source MAC in Ethernet header and ARP payload is virtual MAC address.
+
+### 3.2.6   Uplink Interface Tracking
+
+Interfaces other than the VRRP instance interface can be tracked for up/down events. When interface-tracking is enabled in the VRRP instance configuration, the tracked interface's operational status will be monitored. When a interface operational down event is detected on a tracked-interface, the track-priority/weight is subtracted from the current router’s priority value. Similarly, when interface operational up event is detected on the tracked-interface, the track-priority/weight is added to the router’s current priority value.
```
dev-aviz: How does the uplink tracking works? for instance let's say there are more than 8 uplink interfaces how do we does it effects on mastership?

doc/SONiC_VRRP_HLD.md
```diff
@@ -0,0 +1,582 @@
...
+```
+
+
+## 6 Serviceability and Debug
+
+The existing logging mechanisms shall be used. Proposed debug framework shall be used for internal state dump.
+
+## 7 Warm Reboot Support
+
+Currently, warm-reboot is not supported for VRRP. That is, warm-reboot will simply restart the VRRP docker without VRRP storing any data for warm restart.
```
lguohan: we should prevent warm reboot if there is vrrp configuraiton in place.

doc/SONiC_VRRP_HLD.md
```diff
@@ -0,0 +1,582 @@
...
+### vrrpsyncd
+
+Listens to MACVLAN interface programming in kernel. Status of MACVLAN interface determines Master/Backup state of VRRP instances. VRRP_Table in APP_DB will be programmed with interface name and VIP for Master instances.
+
+- For every IP add to MACVLAN interface, adds interface name and Virtual IP in APP_DB in VRRP_Table
+- For every IP delete from MACVLAN interface, deletes interface name and Virtual IP in APP_DB from VRRP_Table
+
+### vrrporchd
+
+- Listens to VRRP_Table in APP_DB and adds virtual MAC entries in ASIC_DB for Master instances
```
lguohan: it is unclear. can you explain what does vrrporch do?

doc/SONiC_VRRP_HLD.md
```diff
@@ -0,0 +1,582 @@
...
+MAC address allocated to it. This document currently only details the mapping to networks using the IEEE
+802 48-bit MAC address. The virtual router MAC address is used as the source in all periodic VRRP
+messages sent by the Master router to enable bridge learning in an extended LAN.
+A virtual router is defined by its virtual router identifier (VRID) and a set of IP addresses. A VRRP router
+may associate a virtual router with its real addresses on an interface, and may also be configured with
+additional virtual router mappings and priority for virtual routers it is willing to backup. The mapping
+between VRID and addresses must be coordinated among all VRRP routers on a LAN. However, there is
+no restriction against reusing a VRID with a different address mapping on different LANs. The scope of
+each virtual router is restricted to a single LAN.
+To minimize network traffic, only the Master for each virtual router sends periodic VRRP Advertisement
```
lguohan: break paragraph

doc/SONiC_VRRP_HLD.md
```diff
@@ -0,0 +1,582 @@
...
+
+## 6 Serviceability and Debug
+
+The existing logging mechanisms shall be used. Proposed debug framework shall be used for internal state dump.
+
+## 7 Warm Reboot Support
+
+Currently, warm-reboot is not supported for VRRP. That is, warm-reboot will simply restart the VRRP docker without VRRP storing any data for warm restart. 
+
+## 8 Unit Test cases
```
lguohan: we need detailed integration test plan

doc/SONiC_VRRP_HLD.md
```diff
@@ -0,0 +1,582 @@
...
+
+Following virtual MAC address is used by the protocol (as per RFC 5798)
+
+IPv4 case: **00-00-5e-00-01-{vrid}**
+
+where, vrid is user configured 1-byte virtual router identifier. VRID has interface scope; that is, VRID has to be unique among the VRRP instances on an interface. However, same VRID can be used for two  or more virtual router instances across different interfaces.
+
+### 3.2.4    VRRP Advertisement Frame
+
+VRRP control packets have IP protocol type as 112 (reserved for VRRP), and are sent to VRRP multicast address 224.0.0.18. Source MAC in VRRP control packets is virtual MAC address and source IP is interface IP.
```
lguohan: we need sai trap id for this. do we have it defined in the SAI, if yes, then add this as a sai requirement.

doc/SONiC_VRRP_HLD.md
```diff
@@ -0,0 +1,582 @@
...
+    vip                   ; virtual IP address in a.b.c.d/32 format
+    type                  ; IP address type string. only IPv4 supported currently
+    
+; field = value
+vmac = virtual_mac_address ; Virtual MAC address associated with VRRP instance
+```
+
+Example:- 
+
+**Key**: VRRP_TABLE:Vlan1000:[40.10.8.101/32:ipv4]
```
lguohan: why not use interface table? why can we extend intforch? why do we need a separate vrrporch?

doc/SONiC_VRRP_HLD.md
```diff
@@ -0,0 +1,582 @@
...
+
+- Reads the configuration from keepalived.conf file.
+- Listens to kernel link and IP events.
+- Runs VRRP protocol state machine, 
+  - For VRRP instances state transition to Master programs a MACVLAN interface with VIP as interface IP in kernel
+  - For VRRP instances state transition to non-Master deletes MACVLAN interface, if programed, from kernel.
+
+### vrrpsyncd
+
+Listens to MACVLAN interface programming in kernel. Status of MACVLAN interface determines Master/Backup state of VRRP instances. VRRP_Table in APP_DB will be programmed with interface name and VIP for Master instances.
```
lguohan: macvlan interface can be used for other use cases, are we assuming all macvlan interface are created by keepalived?

disaster123:
Does this one support active/active setups?

dks19:
This implementation of VRRP doesn't support Active-Active mode.

disaster123:
@dks19 any chance to implement active active or vrr? I‘m willing to pay for it.

doc/SONiC_VRRP_HLD.md
vvbrcm: The use cases are networks needing backup of logical gateway. The network could be like described below or for a similar use case,
- 3-tier access layer, distribution layer & core layer network, where the redundancy for gateway is needed in distribution layer.
- Layer 3 CLOS network, where the redundancy is needed for the first hop gateways.


Generally static anycast gateway is used in MLAG deployment, but yes VRRP could also be used in MLAG deployment.

doc/SONiC_VRRP_HLD.md
vvbrcm: Currently VRRPv3 (IPv6) support is added.

Since keepalived VRRP is robust, feature rich and well maintained in linux community, keepalived VRRP was chosen over FRR VRRP.

doc/SONiC_VRRP_HLD.md
vvbrcm: Ok, will split it.

doc/SONiC_VRRP_HLD.md
vvbrcm: The already defined SAI trap IDs viz., SAI_HOSTIF_TRAP_TYPE_VRRP & SAI_HOSTIF_TRAP_TYPE_VRRPV6 will be used to trap the VRRP packets.

doc/SONiC_VRRP_HLD.md
vvbrcm: Uplink tracking affects the overall priority for VRRP router. If the tracked interface is in 'UP' state, then the VRRP priority for the router is increased by the weight associated with the tracked interface.

Currently only 8 interfaces can be tracked per VRRP instance. Please let us know if there is a real use case of tracking more than 8 interfaces per VRRP instance.

Note multiple VRRP instances could be configured on an VRRP interface.

doc/SONiC_VRRP_HLD.md
vvbrcm: Intforch was not used to separate the use case. In VRRP multiple my IPs and my MACs (virtual RIF) needs to be programmed and it could so happen that there could be multiple sources (different VRRP instances on the same interface), generating same my IP.

As intforch cannot handle it as is, creating vrrporch module to handle the functionality separately.

doc/SONiC_VRRP_HLD.md
vvbrcm: No, only MACVLANs whose interface name starts with 'vrrp' in kernel is associated with VRRP. keepalived adds MACVLANs with name starting with 'vrrp' substring.

doc/SONiC_VRRP_HLD.md
vvbrcm: vrrpsyncd listen to kernel and for each VRRP master instances, vrrpsyncd will add interface name, VIP & VMAC to APP_DB. 
vrrporch will inturn listen to APP_DB and for entry in VRRP_TABLE, program the VIP as my IP and the VMAC as my MAC(virtual RIF) in ASIC_DB.

doc/SONiC_VRRP_HLD.md
vvbrcm: In the new design VRRP master's will relinquish the mastership of VRRP instances (by sending priority 0 keepalives, so that the standby can take over the mastership instantaneously) before going to warm-reboot. During warm-reboot the neighboring routers will be master. Post warm-reboot the rebooted router could claim back the mastership.

This way the forwarding is kept functional (and not changing) during the process of warm-reboot.

doc/SONiC_VRRP_HLD.md
vvbrcm: Ok.

jefftant:
Please note that we (IETF RTGWG chairs) have just last called draft-ietf-rtgwg-vrrp-rfc5798bis, upon becoming RFC it will absolete RFC5798, please use it as a base.

Cheers,
Jeff

adyeung:
Converge VRRP proposal with https://github.com/sonic-net/SONiC/pull/1446

PR #489, opened 2019-10-12
Thermal control design

keboliu:
Add a design doc for SONiC thermal control design.
Thermal control is intended to add a thermal control daemon to the system which will:

1.  to keep the switch at the proper temperature by using cooling devices, e.g., fan.

2. Thermal control daemon needs to monitor the temperature of devices (CPU, ASIC, optical modules, etc) and the running status of fan. It stores temperature values fetched from sensors and the thermal device running status to the DB, to make these data available to CLI and SNMP or other apps which interested.


thermal-control-design.md
```diff
@@ -0,0 +1,261 @@
...
+- FAN absence action, suspend the algorithm or not, fan speed value to set;
+
+- PSU absence action, suspend the algorithm or not, fan speed value to set.
+
+Below is an example for the policy configuration:
+
+	{
+	    "thermal_control_algorithm": {
+		"run_at_boot_up": true,
+		"fan_speed_when_suspend": 10000
```
jleveque: For these fan speed values, maybe we should use a percentage from 0-100% instead of RPMs? Vendors can define the max fan speed in RPMs in the platform API, and then the target RPMs can be calculated based on the percentage?

Question to platform vendors: Is there any reason to prefer target speeds in RPMs instead of percentages?

keboliu: revised.

keboliu: @jleveque should we wait for some more inputs from other vendors or we can move forward?

jleveque: @padmanarayana: Have you had time to review this PR? Please let us know if you do not have any comments. Thanks!

keboliu: @padmanarayana any comments on this design?

thermal-control-design.md
```diff
@@ -0,0 +1,272 @@
...
+## 1. Overview
+
+The purpose of Thermal Control is to keep the switch at a proper temperature by using cooling devices, e.g., fan.
+Thermal control daemon need to monitor the temperature of devices (CPU, ASIC, optical modules, etc) and the running status of fan. It store temperature values fetched from sensors and thermal device running status to the DB, to make these data available to CLI and SNMP or other apps which interested. 
+
+Thermal control also enforce some environment related polices to help the thermal control algorithm to adjust the switch temperature.     
+
+## 2. Thermal device monitoring
+
+Thermal monitoring function will retrieve the switch device temperatures via platform APIs, platform APIs not only provide the value of the temperature, but also provide the threshold value. The thermal object status can be deduced by comparing the current temperature value against the threshold. If it above the high threshold or under the low threshold, alarm shall be raised.
```
padmanarayana: Is the thermal/PSU/FAN monitoring daemon going to run as a separate entity than the thermal control daemon described in 3.1 ? If so, please indicate if the polling frequency of the monitoring daemon can be configurable ?

It would very much be desirable to run the monitoring/poller daemon separately (especially in BMC based units) so the thermals/PSU/FAN status may be used for CLI/diagnostics/telemetry purposes (since BMC would handle the thermal control).

thermal-control-design.md
```diff
@@ -0,0 +1,272 @@
...
+If fan broken or become up present, log shall be generated：
+
+	Fan removed warning: Fan 1 was removed from the system, potential overheat hazard!
+	Fan removed warning cleared: Fan 1 was inserted.
+
+## 3. Thermal control management
+
+Adjust cooling device according to the current temperature can be very vendor specific and some vendors already have their own implementation. In below Appendix chapter describes a Mellanox implementation. But handle the cooling device according to some predefined policies can be generic, this is part of what Thermal control management will do.
+
+This cooling device control function can be disabled if the vendor have their own implementation in the kernel or somewhere else.
```
padmanarayana: By default, thermal control may be disabled. Platforms (typically lacking BMC) may enable this as necessary.

thermal-control-design.md
```diff
@@ -0,0 +1,272 @@
...
+    timestamp               = STRING                         ; timestamp for the temperature fetched
+    high_threshold          = FLOAT                          ; temperature high threshold
+    critical_high_threshold = FLOAT                          ; temperature critical high threshold
+    low_threshold           = FLOAT                          ; temperature low threshold
+    critical_low_threshold  = FLOAT                          ; temperature critical low threshold
+    warning_status          = BOOLEAN                        ; temperature warning status
+
+These devices shall be included to the temperature monitor list but not limited to: CPU core, CPU pack, ASIC, PSU, Optical Modules, etc.
+
+TEMPERATURE_INFO Table key object_name convention can be "device_name + index"  or device_name if there is no index, like "cpu_core_0", "asic", "psu_2".  Appendix 1 listed all the thermal sensors that supported on Mellanox platform.
```
padmanarayana: Hope this convention is flexible since it may be possible to retrieve/save multiple values for a sensor (say current value and max value recorded since last read). So, the object name may be descriptive like sensor_1_current and sensor_1_max.

thermal-control-design.md
```diff
@@ -0,0 +1,272 @@
...
+
+	  Show platform-specific hardware info
+
+	Options:
+	  -?, -h, --help  Show this message and exit.
+
+	Commands:
+	  fanstatus  Show fan status information
+	  mlnx       Mellanox platform specific configuration...
+	  psustatus  Show PSU status information
```
padmanarayana: psustatus may be enhanced to dump PSU FAN values as well..

thermal-control-design.md
```diff
@@ -0,0 +1,272 @@
...
+### 3.1 Thermal control management flow
+
+It will be a routing function to check whether the policies was hit an the fan speed need to adjust, and also run vendor specific thermal control algorithm.
+
+
+Below policies are examples that can be applied:
+
+- Set PWM to full speed if one of PS units is not present 
+
+- Set PWM to full speed if one of FAN drawers is not present or one of tachometers is broken present
```
padmanarayana: I guess this is the case where the FAN is apparently stuck at a non 0 low tacho reading (minimal threshold below which it is deemed faulty). You may want to add it to the policy configuration example.

thermal-control-design.md
```diff
@@ -0,0 +1,272 @@
...
+    critical_high_threshold = FLOAT                          ; temperature critical high threshold
+    low_threshold           = FLOAT                          ; temperature low threshold
+    critical_low_threshold  = FLOAT                          ; temperature critical low threshold
+    warning_status          = BOOLEAN                        ; temperature warning status
+
+These devices shall be included to the temperature monitor list but not limited to: CPU core, CPU pack, ASIC, PSU, Optical Modules, etc.
+
+TEMPERATURE_INFO Table key object_name convention can be "device_name + index"  or device_name if there is no index, like "cpu_core_0", "asic", "psu_2".  Appendix 1 listed all the thermal sensors that supported on Mellanox platform.
+    
+### 2.2 Fan device monitoring
```
padmanarayana: It would be desirable to monitor both the FAN tray's FANs as well as the PSU's FANs.

thermal-control-design.md
keboliu: the monitor functions and thermal control part will be two separated threads of one daemon.  we will provide the flexibility for the vendors to decide not to run the thermal control part if they decide to cover this part somewhere else, in your case by the BMC, so this should be not an issue. 

for the polling interval,  it's not configurable from the CLI, because currently we don't have a framework to support this is kind of configuration for pmon.

thermal-control-design.md
keboliu: the vendor can disable it in the configuration file so it will be disabled on start, or if the vendor doesn't provide a configuration file, by default it will also be disabled.

thermal-control-design.md
keboliu: both of them will be monitored.

thermal-control-design.md
keboliu: for a thermal object it's composed by a current temp value, and several threshold values (min/max), the name is for the thermal objects, not for each value.   and also for the name will call the get_name API first, if not able to get a valid name then this convention applied.

thermal-control-design.md
keboliu: psu fan should be covered by fan cli, not sure whether we want to add PSU fan info here.

thermal-control-design.md
keboliu: these are only some examples, the framework will allow the vendors to add their own specific policies.

thermal-control-design.md
padmanarayana: Thanks for providing flexibility to not run thermal control part.
w.r.t polling interval, will it be configurable using a configDB entry if not CLI? since we may not want to enforce the same interval across all platforms..

thermal-control-design.md
padmanarayana: Thanks!

thermal-control-design.md
padmanarayana: ok

thermal-control-design.md
padmanarayana: From a modularity perspective, PSU FANs are added under PsuBase : so ideally, PSU FANs should be dumped as part of psustatus if at all.

thermal-control-design.md
padmanarayana: ok thanks

thermal-control-design.md
padmanarayana: ok thanks

thermal-control-design.md
```diff
@@ -0,0 +1,272 @@
...
+    ; field                 = value
+    temperature             = FLOAT                          ; current temperature value
+    timestamp               = STRING                         ; timestamp for the temperature fetched
+    high_threshold          = FLOAT                          ; temperature high threshold
+    critical_high_threshold = FLOAT                          ; temperature critical high threshold
+    low_threshold           = FLOAT                          ; temperature low threshold
+    critical_low_threshold  = FLOAT                          ; temperature critical low threshold
+    warning_status          = BOOLEAN                        ; temperature warning status
+
+These devices shall be included to the temperature monitor list but not limited to: CPU core, CPU pack, ASIC, PSU, Optical Modules, etc.
```
padmanarayana: Can you confirm if ASIC monitoring here refers to the external sensors placed in proximity to ASIC or ASIC internal diodes ? I was planning to add a poller (running in orchagent:SwitchOrch context) which will collect SAI_SWITCH_ATTR_MAX_TEMP, SAI_SWITCH_ATTR_TEMP_LIST, SAI_SWITCH_ATTR_AVERAGE_TEMP and dump to this TEMPERATURE_INFO table (since the other external thermal sensor values may be retrieved as part of standard Platform API calls instead of via SAI)...

thermal-control-design.md
keboliu: also not possible to config from config DB under curent pmon framework, can consider to add it as part of the thermal control daemon config file(we have a file to config the policies)

thermal-control-design.md
keboliu: thermal value is reading from the thermal API, in this layer we don't know exactly where the sensor is located, it could be various per vendors, let me know how these SAI temp defined(in terms which sensor it polling), I can check internally.

thermal-control-design.md
keboliu: @jleveque what's your view on show PSU fan info as part of PSU show cli? Since we have a PSU daemon, if go this way we may consider to extend psud to get it's fan info?

thermal-control-design.md
padmanarayana: Any way to configure this per device during init would be good.

thermal-control-design.md
padmanarayana: Understood. The sensors I was referring to are ASIC internal sensors. I get that currently the Platform APIs don't make SAI calls. So, it has to be a separate poller.

thermal-control-design.md
```diff
@@ -0,0 +1,261 @@
...
+	  temperature  Show fan status information
+	  
+out put of the new CLI
+
+	admin@sonic# show platform temperature
+	NAME    Temperature    High Threshold    Low Threshold    Critical High Threshold   Critical Low Threshold   Warning Status
+	-----  -------------  ----------------  ---------------  ------------------------- ------------------------ ----------------
+	CPU        85              110              -10                 120                        -20                    false
+	ASIC       75              100               0                  110                        -10                    false
+
```
srideepDell: For platforms which does thermal logic via BMC ( Ex monitor high/log and take action accordingly) in these cases do we  need to display CLI's values or can we just show as N/A ?

thermal-control-design.md
jleveque: I see how it could be helpful to display PSU fan info along with status, voltage, etc. I think it is worth exploring extending the PSU daemon.

liat-grozovik:
@jleveque please review the feedback as well as the changes following to your request.
If all is good we would like to merge this PR as code should be ready for review in the next few days.

PR #496, opened 2019-10-18
IP Multicast integration in SONiC - HLD

krishna-chilakapati:
This HLD describes the functionality and high level design of PIM Source Specific Multicast integration in SONiC by leveraging FRR 7.0 (or later) PIM and IGMP protocol suite.


krishna-chilakapati:
Shall submit a separate one.

PR #500, opened 2019-10-21
RADIUS Management User Authentication 0.6

a-barboza:


doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,583 @@
...
+    |  Success(7)     |                 |                    |
+    |<----------------|                 |                    |
+    |                 |                 |                    |
+
+```
+
+1. A user tries to login through the Console
+2. The PAM configuration causes an Access-Request to RADIUS server.
+3. RADIUS returns Access-Accept with Management-Privilege-Level (MPL)(Goto 4.),
+   or Access-Reject (Authentication Failed).
```
dev-aviz: Where does the Cached MPL (management-privilege-level)stored ? 
Can the MPL cache associate with TTL?
How about user login's on device reboots? does it expect login failed/success?

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,583 @@
...
+```
+; Key
+aaa_key              = "authentication"   ; AAA type
+; Attributes
+login                = LIST(1*32VCHAR)   ; pam modules for particular protocol, now only support login for (local, tacacs+, radius). Legal combinations are [ [local], [local, tacacs+ ], [local, radius], [tacacs+, local], [radius, local] ]
+failthrough          = "True" / "False"  ; failthrough mechanism for pam modules
+```
+
+Note(s):
+- protocol = [[tacacs+, local], [radius, local]] are the options
```
dev-aviz: Can the framework allow user change from tacacs+ to radius?

doc/aaa/radius_authentication.md
a-barboza: > Where does the Cached MPL (management-privilege-level)stored ? 

Cached MPL is stored under /var/run/radius. This is mentioned on line 271

> Can the MPL cache associate with TTL?

Yes, every time the user logs in, their cache MPL is updated.

> How about user login's on device reboots? does it expect login failed/success?

I didn't get the question here. Is the inquiry regarding cached MPL cached across device reboot. The cached contents are in /var/run/radius, and are updated when the user logs in.

doc/aaa/radius_authentication.md
a-barboza: > Can the framework allow user change from tacacs+ to radius?

We can configure using click, or REDIS the change from TACACS+ to RADIUS authentication. Later we can use the available Yang in the framework to change authentication from TACACS+ to RADIUS.

doc/aaa/radius_authentication.md
a-barboza: Updated the the location of the cache to /var/cache/radius. since that is recommended place for Application cache data

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,634 @@
...
+information obtained from the RADIUS server during authentication, to
+determine the user information that is needed to create a local user, at
+the time of first login. The information of this local user is returned
+during a getpwnam_r() library call.
+
+There is an option to map all users at a privilege level to one user as
+specified in the configuration file /etc/radius_nss.conf. This option is
+the many_to_one=y option. When this option is in effect, no local user is
+created at the time of login. Because of the mapping, not all features
+may apply, since the LOGNAME of the user, and the mapped uid may not match.
```
renukamanavalan, line 303: 1) Are we planning to use this feature ?
2) If yes, how
3) How does this affects home dir and files created by user, his uid/gid ...?

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,634 @@
...
+
+### AAA Table Schema
+
+```
+; Key
+aaa_key              = "authentication"   ; AAA type
+; Attributes
+login                = LIST(1*32VCHAR)   ; pam modules for particular protocol, now only support login for (local, tacacs+, radius). Legal combinations are [ [local], [local, tacacs+ ], [local, radius], [tacacs+, local], [radius, local] ]
+failthrough          = "True" / "False"  ; failthrough mechanism for pam modules
+debug                = "True" / "False"  ; debug logs for nss or pam modules
```
renukamanavalan, line 351: Which log file?

How is this flag going to behave, for tacacs & local auth ?

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,634 @@
...
+passkey              = 1*32VCHAR  ; shared secret (Valid chars: ASCII printable except SPACE, '#', and COMMA)
+auth_type            = "pap" / "chap" / "mschapv2"  ; method used for authenticating the communication message
+src_ip               = IPAddress  ;  source IP address (IPv4 or IPv6) for the
+                                  ;  outgoing UDP/IP dgram. This is being
+                                  ;  obsoleted in favor of the RADIUS_SERVER
+                                  ;  table src_intf attribute. If both src_ip
+                                  ;  and src_intf are specified a warning log
+                                  ;  can be given, and src_intf is preferred.
+                                  ;  Default is determined by routing stack.
+nas_ip               = IPAddress  ;  NAS-IP|IPV6-Address (Type 4|95) attribute
```
renukamanavalan, line 375: What is the added value/gain/pro in configuring nas-ip?
Can you please provide some semantics on the usage ?
Some of it could get translated to help hints for user in CLI

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,634 @@
...
+server_key           = NameOrIPAddress; RADIUS server's DNS name or IPv4|6 addr
+                                      ; * DNS name is Work In Progress(WIP) *
+                                      ; * in future phase                   *
+; Attributes
+auth_port            = 1*5DIGIT
+passkey              = 1*32VCHAR  ; per server shared secret (Valid chars: ASCII printable except SPACE, '#', and COMMA)
+auth_type            = "pap" / "chap" / "mschapv2"  ; method used for authenticating the communication message
+priority             = 1*2DIGIT  ; specify RADIUS server's priority
+timeout              = 1*2DIGIT
+retransmit           = 1*2DIGIT
```
renukamanavalan, line 406: Can you please provide some details on retransmit ?
Part of UDP config?

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,634 @@
...
+; Key
+server_key           = NameOrIPAddress; RADIUS server's DNS name or IPv4|6 addr
+                                      ; * DNS name is Work In Progress(WIP) *
+                                      ; * in future phase                   *
+; Attributes
+auth_port            = 1*5DIGIT
+passkey              = 1*32VCHAR  ; per server shared secret (Valid chars: ASCII printable except SPACE, '#', and COMMA)
+auth_type            = "pap" / "chap" / "mschapv2"  ; method used for authenticating the communication message
+priority             = 1*2DIGIT  ; specify RADIUS server's priority
+timeout              = 1*2DIGIT
```
renukamanavalan, line 405: This is the max time, spent to reach a server ?
Some semantics please

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,634 @@
...
+For TACACS+, the existing module click comands will continue to be available.
+The AAA commands are an existing click module command which will be extended
+for the RADIUS Management User Authentication feature.
+
+
+```
+sonic# show aaa
+sonic# show radius
+
+sonic(config)# [no] radius-server source-ip <IPAddress(IPv4 or IPv6)>
```
renukamanavalan, line 464: I am unable to understand "sonic(config)# [no]"

Also can you paste a sample o/p for the show command

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,634 @@
...
+4. Access-Accept: The MPL is cached.
+5. A getpwnam_r() call is made to obtain user information.
+6. If the RADIUS user is not found in the local /etc/passwd file, the NSS
+   RADIUS plugin uses the cached MPL information to return user information.
+7. The user has authenticated successfully.
+
+Note:
+1. If an application requests getpwnam_r() before any authentication is
+   done, the RADIUS NSS plugin can be configured to return a least privileged
+   (MPL 1) temporary user info. On subsequent sessions, the user info will
```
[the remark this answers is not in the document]
a-barboza: The TACACS+ client on SONiC uses the TACACS+ Authorization PDU to obtain User-Info. This works most of the time for TACACS+, but there are some TACACS+ servers which cannot be configured to return Authorization information before a successful authentication. For RADIUS, there is no such Authorization PDU.

The user information is configured using the Tables mentioned in the "User Privilege Table" subsection of "NSS" section below.

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,634 @@
...
+    |  Success(7)     |                 |                    |
+    |<----------------+                 |                    |
+    |                 |                 |                    |
+```
+
+1. A user tries to login through a SSH client.
+2. A getpwnam_r() call is made to obtain user information.
+3. If the RADIUS user is not found in the local /etc/passwd file, the NSS
+   RADIUS plugin uses the cached MPL information to return user information.
+   If no MPL is available, and the many_to_one=a is configured, a least
```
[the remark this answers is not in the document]
a-barboza: This is configured in the file as mentioned in the "NSS" section, 2nd paragraph. I'll add a pointer.
The primary advantage of using the many_to_one=a (anonymous) option  is that users do not have to be manually created beforehand by the admin. The primary downside is that unknown users (i.e. unknown before the first successful RADIUS login for that user) will appear as the least privileged user. I can add this information to the HLD as well.

The Management-Privilege-Level cache is only used by the RADIUS NSS plugin if the user is not present in the /etc/passwd file. The many-to_one=a (anonymous) option tells the RADIUS NSS plugin to use the configuration for the least privileged user in the User Privilege Table subsection, if there is no entry for the user in the Management-Privilege-Level cache.

doc/aaa/radius_authentication.md
a-barboza: 1. There is a corresponding (many_to_one=y)  option for TACACS+. I believe it would be used similarly. 
2. There are less use cases for this. It could be used, for example, if there is a requirement of creating only a fixed number of users (say one per privilege-level) on the SONiC device.
3. The user logging in will get the identity of the mapped user (Eg: If Jane's MPL was 15, and John's was 1, and the default table from section "User Privilege Table" were used, Jane would be operating as admin (uid=1000), and John would be operating as nobody (uid=65534). Thus all their files creations would be given the mapped identity of admin, and nobody. Similarly their home directories, by default without any other configuration, would be /home/admin and /nonexistent.)

doc/aaa/radius_authentication.md
a-barboza: Most PAM modules have a debug option which sends messages to syslog(3). This has a direct mapping to that option

```
OPTIONS
       debug
           Turns on debugging via syslog(3).
```

doc/aaa/radius_authentication.md
a-barboza: Every RADIUS PDU sent to the RADIUS server must have a NAS-IP|IPV6-Address or NAS-Identifier (or both) attribute. If the admin wants to 
(a) configure something other than the MGMT_INTERFACE table IP addresses
(b) reliably get one IP address (on a multi-homed/managementip device)

they could use this configuration.. Yes, I can add this to the HLD.

doc/aaa/radius_authentication.md
a-barboza: I will update as per the documentation of the pam_radius_auth module.

```
#  The timeout field controls how many seconds the module waits before
#  deciding that the server has failed to respond.
```

doc/aaa/radius_authentication.md
a-barboza: I will update as per the documentation of the pam_radius_auth module.

```
retry = #      - How many times to re-send a packet if there is no
                 response.  Once the retry count has been reached,
                 the module fails, and PAM continues to the next module.
```

doc/aaa/radius_authentication.md
a-barboza: The optional "no" prefix to most commands is a way to un-configure the corresponding configuration. I will add something like this to the HLD to give a better picture of the CLI.

```
S5(config)# radius-server nas-ip 10.59.143.170
S5(config)# radius-server host 10.59.143.220
S5(config)# do show radius-server
---------------------------------------------------------
RADIUS Global Configuration
---------------------------------------------------------
nas-ip-addr: 10.59.143.170
--------------------------------------------------------------------------------
HOST            AUTH-TYPE KEY       AUTH-PORT PRIORITY TIMEOUT RTSMT VRF   SI 
--------------------------------------------------------------------------------
10.59.143.220   -         -         1812      1        -       -     -     - 
S5(config)# no radius-server nas-ip 
S5(config)# do show radius-server
---------------------------------------------------------
RADIUS Global Configuration
---------------------------------------------------------
--------------------------------------------------------------------------------
HOST            AUTH-TYPE KEY       AUTH-PORT PRIORITY TIMEOUT RTSMT VRF   SI 
--------------------------------------------------------------------------------
10.59.143.220   -         -         1812      1        -       -     -     - 
S5(config)# 
```

doc/aaa/radius_authentication.md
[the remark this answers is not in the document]
renukamanavalan: `If the RADIUS user is not found in the local /etc/passwd file, the NSS RADIUS plugin uses the cached MPL information to return user information. If no MPL is available, and the many_to_one=a is configured, a least privileged (MPL 1) temporary user info is returned. Otherwise, the login fails`

When would be the case, "user not found in /etc/passwd", but MPL cache is found ?

doc/aaa/radius_authentication.md
```diff
@@ -0,0 +1,634 @@
...
+   privileged (MPL 1) temporary user info is returned. Otherwise, the
```
renukamanavalan, line 246: Does this mean, a Radius user has to login twice to get his  real privilege level/role, in the following two scenarios ?

a) First time login into SONiC switch (implying no entry in /etc/passwd)
b) When user's privilege level has changed at RADIUS server, since his last login into this SONiC switch?

Say user had higher privilege level. He logged in, say twice to ensure he has the right level.  And he logs out.
Say, later his privilege level at server goes down ( which by the way is standard practice)
Now the user logs into the device again.
This time he has entry in /etc/passwd & an entry in MPL cache too. Note: The cache will have the higher privilege level.
Questions:
a) Will he now get a session at higher privilege level ?
b) Is there any MPL cache timeout that would prevent the above ?
c) Even with MPL cache timed out, /etc/passwd would have the user for remote_user_su. How does this impact ?

Trying to understand, all the different behaviors to deal with when switching to Radius.

doc/aaa/radius_authentication.md
renukamanavalan: Thanks. So we are going to benefit from this flag, for both Tacacs & Radius. Pretty good.

doc/aaa/radius_authentication.md
renukamanavalan: This will be a very handy knob, to tune certain group of devices, differently. The knowledge in HLD doc will be very useful. Thanks

doc/aaa/radius_authentication.md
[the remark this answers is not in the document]
a-barboza: This would be the many_to_one=y (map users) case. In that case, the RADIUS user is not present on the system, thus the RADIUS NSS plugin would need to return mapped user information based on the MPL cached information. This is also discussed in the following comment. I work on adding this information as well.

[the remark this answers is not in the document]
renukamanavalan: Thanks

doc/aaa/radius_authentication.md
a-barboza: Thanks for the use-cases. These are all for the many_to_one=a (anonymous) use-case.

This case can be addressed by giving each user a distinct user id, and a distinct group id. Thus  the primary group will be distinct on the SONiC device for every RADIUS user. Any changes to the user's privilege level on the RADIUS server will not affect their primary group id. (I'm not in favor of re-privileging a user, since there are the issues of what can be done about existing resources that are already created for  that user. Another approach would be to re-create re-privileged users as a different username. However, having a distinct primary group id does seem to alleviate that situation.)

With these changes, the answers to the questions would be.
a) No, because after authentication the lower privilege level would be cached and used to start the user's session.
b) The cache would not be used after the login. Once the user's session is active the cache has no effect, since the /etc/passwd entry would have been created.
c) The /etc/passwd entry for the RADIUS user would be created with the RADIUS user's name, and the unique primary group created for that user. None of these have any bearing on the privileges assigned to that user. The privileges assigned to the user are based on the supplementary/secondary groups assigned to the user (User Privilege Table "secondary grps" column).

Yes, thanks for considering the different use-cases. I can make these changes to the HLD. Please advise?

doc/aaa/radius_authentication.md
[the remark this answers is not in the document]
a-barboza: Now, creating a temporary user is default, so there is no need of the "anonymous" option. The spec is updated.

doc/aaa/radius_authentication.md
a-barboza: Added additional nas-ip description to the RADIUS Table Schema in the form of a note.

doc/aaa/radius_authentication.md
a-barboza: Changes made. Summary of Changes section describes the major changes.

doc/aaa/radius_authentication.md
a-barboza: Added additional description for many_to_one=y case under NSS --> Mapping Users sub-section

PR #510, opened 2019-10-28
Kernel kdump support and management

olivier-singla:
In the event of a kernel crash, we need to gather as much information
as possible to understand and identify the root cause of the crash.
Currently, the kernel does not provide much information, which make
kernel crash investigation difficult and time consuming.

Fortunately, there is a way in the kernel to provide more information
in the case of a kernel crash. kdump is a feature of the Linux kernel
that creates crash dumps in the event of a kernel crash. This proposal
will add kernel kdump support.

An extension to the CLI utilities config and show is provided to
 configure and manage kdump:
 - enable / disable kdump functionality
 - configure kdump (how many kernel core files can be stored locally,
   how much memory is allocated for kdump)
 - view kernel crash logs


doc/kdump/SONiC-kdump.md
```diff
@@ -0,0 +1,425 @@
...
+#### config kdump <enable|disable> [-y] [-n]
+
+Enable or disable kernel core dump functionality.
+
+Administrative state of kdump is stored in ConfigDB.
+
+The variable *USE_KDUMP* in the file /etc/default/kdump-tools is set to 0 to disable kdump, and set to 1 to enable kdump.
+
+Since this command might require changing the kernel parameters to specify the amount of memory reserved for the capture kernel (the kernel parameters which are exported through */proc/cmdline*), a reboot is necessary. By default, the command displays a message showing that kdump functionality will be either enabled or disabled following the next reboot, unless the optional parameter *-n* has been provided. The optional parameter *-y* allows to automatically perform the reboot without asking for a confirmation. In this case, the command "config save -y" will be automatically performed prior to the reboot.
+
```
lguohan: user can choose to do warm reboot or fast reboot, recommend to do not integrate reboot into this config tool.

PR #518, opened 2019-11-08
[doc] Add rates and Utilization HLD 

mykolaf:
Add design proposal for 

calculating port and interface BPS, PPS rates and utilization using lua plugins(Flex counter infrastructure).

Aiming to resolve this issue: https://github.com/Azure/sonic-buildimage/issues/2738


mykolaf:
@lguohan could you please review the updated design and approve/merge if no further comments?

PR #523, opened 2019-11-20
Update configurable drop counters HLD

daall:
- Update HLD with changes from code review
- Add clarifications from discussion with Aviz
- Add acknowledgements and references

Signed-off-by: Danny Allen <daall@microsoft.com>


daall:
@dev-aviz I can't seem to add you as a reviewer but here are those changes we talked about!

PR #531, opened 2019-12-03
[doc] Create configurable drop counters test plan

daall:
Signed-off-by: Danny Allen <daall@microsoft.com>

### DIFF
--- a/doc/drop_counters/drop_counters_test_plan.md
+++ b/doc/drop_counters/drop_counters_test_plan.md
@@ -1,0 +1,584 @@
+# Configurable Packet Drop Counters Test Plan
+
+## Table of Contents
+- [1 Overview](#1-overview)
+    - [1.1 Scope](#11-scope)
+    - [1.2 Test Cases](#12-test-cases)
+    - [1.3 Related CLI Commands](#13-related-cli-commands)
+- [2 Testing Parameters](#2-testing-parameters)
+    - [2.1 Supported Topologies](#21-supported-topologies)
+    - [2.2 Testbed Constraints](#22-testbed-constraints)
+    - [2.3 IPv6 Support](#23-ipv6-support)
+- [3 Testing Flow](#3-testing-flow)
+    - [3.1 Setup](#31-setup)
+    - [3.2 Testing](#32-testing)
+    - [3.3 Teardown](#33-teardown)
+- [4 Test Cases](#4-test-cases)
+    - [4.1 Capabilities](#41-capabilities)
+    - [4.2 Individual Pipeline Drops](#42-individual-pipeline-drops)
+    - [4.3 Multiple Pipeline Drops](#43-multiple-pipeline-drops)
+    - [4.4 Multiple Counters](#44-multiple-counters)
+    - [4.5 Configuration](#45-configuration)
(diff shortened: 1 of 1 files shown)


doc/drop_counters/drop_counters_test_plan.md
```diff
@@ -0,0 +1,584 @@
...
+    version = 4
+    ttl = [auto]
+    proto = tcp
+    src = 169.254.0.0
+    dst = 2.2.2.2
+...
+```
+
+#### Additional Steps:
+Because counter B includes all the drop reasons, the total in counter B should be equal to the total number of drops. Counter A should only be equal to the number of SMAC_EQUALS_DMAC packets sent.
```
yvolynets-mlnx: Can you please clarify a little bit more about "total number of drops" in 
"the total in counter B should be equal to the total number of drops."

doc/drop_counters/drop_counters_test_plan.md
```diff
@@ -0,0 +1,584 @@
...
+    src = 169.254.0.0
+    dst = 2.2.2.2
+...
+```
+
+#### Additional Steps:
+We essentially want to repeat the general testing procedure twice for this test. The procedure is as follows:
+1. Send both SMAC_EQUALS_DMAC and SIP_LINK_LOCAL packets to the device, verify that the counts are correct.
+2. Remove SIP_LINK_LOCAL as a drop reason from the counter, verify that it is cleared.
+3. Send both SMAC_EQUALS_DMAC and SIP_LINK_LOCAL packets to the device, verify that both are counted by RX_DRP but only SMAC_EQUALS_DMAC is counted by the drop counter.
```
yvolynets-mlnx: Can you please also clarify "counted by the drop counter". Because there is mentioned "RX_DRP" and "counted by the drop counter".

doc/drop_counters/drop_counters_test_plan.md
```diff
@@ -0,0 +1,584 @@
...
+###[ Ethernet ]###
+  dst = [auto]
+  src = [auto]
+  type = 0x800
+###[ IP ]###
+    version = 4
+    ttl = [auto]
+    proto = tcp
+    src = 169.254.0.0
+    dst = 2.2.2.2
```
yvolynets-mlnx: In case "B" you mentioned drop reasons: SMAC_EQUALS_DMAC, SIP_LINK_LOCAL, DIP_LINK_LOCAL.
But in packet example dst IP is not "DIP_LINK_LOCAL". Is it correct?

daall: That's correct, I dropped DIP_LINK_LOCAL to avoid confusion.

doc/drop_counters/drop_counters_test_plan.md
daall: Updated.

doc/drop_counters/drop_counters_test_plan.md
daall: Should be clarified now!

yxieca:
One small suggestion is to re-number the test cases to match the scope definition.

e.g. 4.<scope number>.x.y

Did all scope get covered?

PR #536, opened 2019-12-10
PDDF v2.0:  1) Support for 2.0 platform APIs,   2) Support for BMC enabled platforms

fk410167:
PDDF v2.0 HLD


fk410167:
PDDF v2.0 HLD

This documents describes the enhancements made to the Platform Driver Development Framework (PDDF). New changes include,
1) Support for 2.0 Platform APIs:
Previous PDDF version was based out of 1.0 Platform APIs which use user-space plugins to access the HW SysFS attributes. New version of PDDF provides generic device object-calsses derived form the 2.0 platform base class. Actual per-platform object-classes would further be derived form the above mentioned PDDF generic classes.

2) Support for BMC platforms:
New PDDF version supports the attribute get/set for PSU, FAN and thermal components if they are controlled by the BMC.

PR #538, opened 2019-12-11
GEARBOX HLD

dan-arsenault:
Initial High Level Design document for GEARBOX after internal review.

This is the new Gearbox Manager functionality and infrastructure in utilizing the new OCP External PHY Abstraction Interface for SONiC.


inspurSDN:
@dan-arsenault 
I study the gearbox proposal that you submitted and see you have a plan to implement line card hot swapping in the future.

So, I think I can integrate our proposal #560 based on your gearbox design and SAI.

I will adjust the line card notification flow via sysfs to pmon and describe PHY mapping with line card in profile.

What do you think of this proposal? Maybe I can ask Xin to create a sub-group to discuss this.

dan-arsenault:
Your proposal should be compatible with the current Gearbox solution. As you've indicated, the hot-swap needs to produce the events for which PORTSORCH would consume. PORTSORCH would then resolve or apply the appropriate configuration.

inspurSDN:
Thanks for your reply, I still have some question about the gearbox solution. How do you know the download firmware procedure is ready to PHY? if NOS_platform_config describes there are multiple PHY, but at initialization time, there is only one PHY what will happen?  and I didn't find PHY re-initialize procedure, does it work for gearbox?

doc/gearbox/gearbox_mgr_design.md
```diff
@@ -0,0 +1,624 @@
...
+      "system_fec": "none",
+      "system_loopback": "none",
+      "system_speed": "25000",
+      "system_training": "false"
+    }
+  }
+}
+~~~
+
+### 3.4.3 STATE DB
```
stephenxs, line 342: Do we need to add a table in STATE DB  in case of something out of orchagent needs to know gearbox info? Or do we have another way to achieve that?

dan-arsenault: Hello inspurSDN and stephenxs;
Your questions are well received and should be considered for future design and development. Firmware download is not currently supported and there is no 'state' requirements at this time. The current HLD is focused on the initial configuration of a Gearbox with an emphasis to provide an initial infrastructure for vSONiC. Thank you for your participation and concerns.

PR #541, opened 2019-12-18
Container to Host Communications using D-Bus 

mikelazar:


doc/mgmt/Docker to Host communication.md
```diff
@@ -0,0 +1,204 @@
...
+| D-Bus                    | Desktop Bus: https://en.wikipedia.org/wiki/D-Bus  |
+
+# 1 Feature Overview
+
+The management framework runs within a Docker container, and performs actions
+translating the user CLIs or REST requests to actions. Most of these actions
+perform some operation on the Redis database, but some of them require
+operations to be done on the host, i.e., outside the container. This document
+describes the host server, and translib API that are used to communicate between
+the Docker container and the host.
```
dev-aviz: It's greatly useful. However, are we restricting the sonic users to use MGMT container to enjoy the DBUS framework? Let's say from xyz container user wanted to communicate to Host, user should be able to use it as a plug a play?

mikelazar: As mentioned in the document, any container that mounts /var/run/dbus (-v run/create docker option) can use Linux host D-Bus services in general - any pre-existing D-Bus services,  and not just "Host" services developed for the management container. Using /var/run/dbus is a choice of the container developer, and it is completely outside the scope of this feature. D-Bus is an existing Linux feature, and bind mount (-v option) a docker feature.

dev-aviz: Thanks Mike. Today we should be able to program most of the switch configuration through redis right? Can you shed some light on what are the use cases/configurations are trying to solve using   D-Bus?

sramkrishna: Given the talk about security and DBus and Containers - it's worth having a conversation with the flatpak community who have already implemented this called 'portals' of which we do consider security implementations. Happy to connect.

doc/mgmt/Docker to Host communication.md
```diff
@@ -0,0 +1,235 @@
...
+## 2.1 Target Deployment Use Cases
+
+All deployments
+
+## 2.2 Functional Description
+
+This feature enables management applications to issue
+requests to the host to perform actions such as:
+* image install / upgrade
+* ZTP enable/disable
```
dev-aviz, line 115: why D-Bus used specific listed services/activities alone, why can't it use for all the services? Can you add some guidelines to the document?

doc/mgmt/Docker to Host communication.md
```diff
@@ -0,0 +1,235 @@
...
+* create show-tech tar file using existing show-tech script
+* config save/reload using existing scripts
+
+# 3 Design
+## 3.1 Overview
+
+The feature extends the SONiC management framework to add a D-Bus service on the host. This service will register against a known endpoint, and will service requests to the endpoint. Application modules will add "applets" to the host
+service, which will automatically register their endpoints (D-Bus objects and methods). 
+
+The client application module in a container (e.g. management or telemetry) can use the APIs provided in Translib to send the request to the host - essentially, make a (remote) D-Bus method call, and either wait for the response (if the request was synchronous), or receive a channel and wait for the request to return the response on the
```
dev-aviz, line 126: How to secure the users shouldn't mis use the critical host access to commands like reboot? Can you provide some guidelines around it?

doc/mgmt/Docker to Host communication.md
```diff
@@ -0,0 +1,235 @@
...
+
+All deployments
+
+## 2.2 Functional Description
+
+This feature enables management applications to issue
+requests to the host to perform actions such as:
+* image install / upgrade
+* ZTP enable/disable
+* initiate reboot and warm reboot using existing scripts
```
dev-aviz, line 116: How does host communicate to ex: reboot/warm reboot execution status to clients (containers issued commands to host)? How does D-Bus helps here?

doc/mgmt/Docker to Host communication.md
```diff
@@ -0,0 +1,235 @@
...
+    <deny receive_sender="org.SONiC.HostService"/>
+    <allow send_destination="org.SONiC.HostService.showtech"/>
+    <allow receive_sender="org.SONiC.HostService.showtech"/>
+  </policy>
+```
+
+
+### 3.1.2 Command Logging
+
+It is possible to track and log the user name and the command that the user has requested.
```
dev-aviz, line 168: Will the design provide some D-Bus service infrastructure level audit mechanism who does what w.r.t host commands like reboot/warm reboot etc.

toravir:
@mikelazar does these guidelines apply when a process running on the host wants to initiate communication with a docker. I am working on a process and considering running it on the host and having it directly communicate with the redis container to access db.

toravir:
@mikelazar what is the another means to access Redis DB ? I am considering running PTP on Sonic and the recommendation is to run it on the host - https://github.com/sonic-net/SONiC/blob/master/doc/ptp/ptp.md

mikelazar:
> @mikelazar what is the another means to access Redis DB ? I am considering running PTP on Sonic and the recommendation is to run it on the host - https://github.com/sonic-net/SONiC/blob/master/doc/ptp/ptp.md

@toravir In general, please ask this type of questions using the SONiC mailing lists. If I recall correctly, Redis DB/SONiC is configured to use a Unix domain socket for client (application) to server communications. That socket is available for all containers, and for the host as well. So you can just use the Redis DB client library (Python or Hiredis). There might be a SONiC Redis DB wrapper - but I do not recall exactly. However, please use that for SONiC you need to define a DB schema using YANG. 

Also, not sure who made the recommendation to run PTP on the host. This might be needed ONLY if you need to communicate with a non-standard Linux kernel module/driver directly (and even that's debatable in the case of SONiC). However, I do not see that dependency in the design document (the PHC driver seems to be standard in kernels >= 5.8).  The only thing required for the host OS is to load the PHC driver, when PTP is enabled.

Anyway, this is something to discuss in the SONiC community meetings (and possibly using the SONiC mailing lists).

PR #544, opened 2019-12-19
North bound interface (CLI/REST clients) support for configuring FRR BGP

venkatmahalingam:
1) Extend and provide unified configuration and management capability for FRR-BGP features used in SONiC. 
2) Allow the user to configure & manage FRR-BGP using SONiC Management Framework with Open Config data models via REST, gNMI and also provides access via SONiC Management Framework CLI as well.


doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+  2. Syslog messages are added at appropriate places to help trace a failure.
+  3. Leverages existing debug mechanism & framework if any.
+
+# 7 Warm Boot Support
+
+This enhancement to FRR-BGP Unified management framework does not disrupt data plane traffic during warmboot. No special handling required for warmboot.
+
+# 8 Scalability
+
+# 9 Unit Test
```
lguohan: please enable the testing part. the description of test is not enough.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+### 1.1.1 Functional Requirements
+
+  1. Extend Unified mode for full FRR-BGP config and management in SONiC
+  2. Extend sonic-cfggen, bgpcfgd and integrate with FRR-BGP for features supported in SONiC
+  3. Support for retrieval of FRR-BGP state and statistics information
+  4. For backward compatibility retain access to FRR UI (vtysh) for managing features that are NOT in conflict with SONiC features
+
+### 1.1.2 Configuration and Management Requirements
+
+1. Support Open Config data models for BGP config and Management
```
pavel-shirshov: Can you please put urls to the models you want to support?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+2. Provide IS-CLI/gNMI/REST support for config and management of FRR-BGP features used in SONIC
+3. Enhance with Custom YANG models for features used in BGP that are not supported via Open Config data model
```
pavel-shirshov: why do we need to introduce the `Custom YANG models` in addition to the OpenConfig data model?
Can we augment OpenConfig data model? Can we create one custom YANG model?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+4. Define ABNF schema for BGP features used in SONiC
+
+### 1.1.3 Scalability Requirements
+N/A
+
+### 1.1.4 Warm Boot Requirements
+As state and statistics information is retrieved from FRR-BGP on demand there is no Warm Boot specific requirements for this feature.
```
pavel-shirshov: what about config and management?
what about some db schema changed?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+## 1.2 Design Overview
```
pavel-shirshov: Validation here. What is that?
is it yang validation?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+* This enhancement takes comprehensive approach to support BGP features used in SONiC:
+	* Standard based YANG models and custom YANG models
+	* Open API spec
+	* Industry standard CLI
+	* Config Validation
+* REST server, gNMI server, Transformer methods - all in Go
+* Marshalling and unmarshalling using YGOT
+* Redis updated using CAS(Check-and-Set) trans. (No locking, No rollback)
+* Config Validation by using YANG model from ABNF schema
```
pavel-shirshov: That config validation will validate not so many parameters. As I understand it can validate data types and references, but not semantic

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+## 2.2 Functional Description
+Provide GNMI and REST support for config get/set, state get and statistics get,  CLI config and show commands for FRR-BGP features used in SONiC.
+
+
+# 3 Design
+## 3.1 Overview
+The extended unified config and management framework for FRR-BGP in SONiC is represented in below diagram.
+
+![FRR-BGP Unified Mgmt Framework](images/FRR-BGP-Unified-mgmt-frmwrk.png)
```
pavel-shirshov: wrong url

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+1. Transformer common app owns the Open config data models related to BGP (which means no separate app module required for handling BGP open-config and augmented YANG objects).
```
pavel-shirshov: can you please elaborate?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+2. Provide annotations for required objects so that transformer core and common app will take care of handling them.
```
pavel-shirshov: can you please elaborate? Why do we need these annotations?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+3. Provide transformer methods as per the annotations defined to take care of model specific logics and validations.
+
+4. Define SONiC YANG and Redis ABNF schema for the supported Open Config BGP models & objects.
+
+5. KLISH CLI and REST clients provide extensive BGP configurations and hence there should not be any need for BGP configurations via vtysh.
```
pavel-shirshov: Earlier you mentioned to use vtysh to configure 'that are NOT in conflict with SONiC features'

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+6. In bgpcfgd register for Redis DB events for the BGP and other related objects, so as to translate the Redis DB events to FRR-BGP CLI commands to configure FRR-BGP, similarly, separate config daemons can be present to configure individual features like OSPF, BFD..etc
+
+7. Update /usr/share/sonic/templates/bgpd.conf.j2 template for new FRR-BGP configurations supported in SONiC which will be used by sonic-cfggen to generate /etc/frr/bgpd.conf file.
```
pavel-shirshov: how it should be updated? Why current doesn't work?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+2. Provide annotations for required objects so that transformer core and common app will take care of handling them.
+
+3. Provide transformer methods as per the annotations defined to take care of model specific logics and validations.
+
+4. Define SONiC YANG and Redis ABNF schema for the supported Open Config BGP models & objects.
+
+5. KLISH CLI and REST clients provide extensive BGP configurations and hence there should not be any need for BGP configurations via vtysh.
+
+6. In bgpcfgd register for Redis DB events for the BGP and other related objects, so as to translate the Redis DB events to FRR-BGP CLI commands to configure FRR-BGP, similarly, separate config daemons can be present to configure individual features like OSPF, BFD..etc
```
pavel-shirshov: What if some objects relies on other object. Or changing some objects requires to restart something?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+;Defines BGP globals table
+;
+;Status: stable
+
+key                                  = BGP_GLOBALS|vrf ;
+vrf                                  = 1\*15VCHAR ; VRF name
+local_asn                            = 1*10DIGIT ; Local ASN for the BGP instance
+router_id                            = \*IPv4prefix ; Router ID IPv4 address
+load_balance_mp_relax                = "true" / "false" ;
+grace_restart                        = "true" / "false" ;
```
pavel-shirshov: graceful_restart

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+  * BGP router & address family configurations
+  * BGP neighbor & address family configurations
+  * BGP peer group & address family configurations
+  * BGP listen prefix configuration for peer group
+  * Route map configurations
+  * Route redistribute configurations
+  * Route policy sets (prefix list/community/ext-community/as-path/neighbor-set/tag-set configurations
+
+Enhanced following table to configure additional attributes:
```
pavel-shirshov: Is it possible to autogenerate the schema from the yang model?
It will give us names and types automatically.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+allow_asin                     = "true" / "false" ;  Accept as-path with my AS present in it
+allow_as_origin                = "true" / "false" ; Only accept my AS in the as-path if the route was originated in my AS
+capability_orf_send            = "true" / "false" ; Capability to receive the outbound route filtering from this neighbor
+capability_orf_receive         = "true" / "false" ; Capability to send the outbound route filtering to this neighbor
+capability_orf_both            = "true" / "false" ; Capability to send and receive the outbound route filtering to/from this neighbor
+route-server-client            = "true" / "false" ; Configure a neighbor as Route Server client
+```
+
+#### 3.2.1.6 BGP_PEER_GROUP
+The existing BGP_PEER_RANGE (peer group) table does not have vrf-name as the key (but added as the field in the table as per VRF HLD) but FRR and open config models have the peer-group under VRF context, so, the new table BGP_PEER_GROUP has been introduced for configurations from management framework (This is not a backward compatible change, expecting the user to migrate to this table in the near future).
```
pavel-shirshov: BGP_PEER_RANGE is about dynamic bgp peers, it is not about bgp peer groups. SONiC doesn't have peer groups defined in DBs

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+as_path_member = string list; AS path list
+                            ;Acceptable List of as paths "string, string"
+````
+
+### 3.2.2 APP DB
+N/A
+
+### 3.2.3 STATE DB
+No changes to State DB, State and statistics information will be retrieved directly from FRR-BGP.
```
pavel-shirshov: I'd suggest to use db as a source.
otherwise someone could run the requests many times and dos frr

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+### 3.6.2 CLI
+  1. For all configuration commands, the CLI request is converted to a corresponding REST client SDK request based on the Open Config data model that was generated by the Swagger generator, and is given to the REST server.
+
+  2. From there on it will follow the same path as a REST config request for create, update and delete operations.
+
+  3. The Swagger generated REST server handles all the REST requests from the client SDK and invokes a common handler for all the create, update, replace, delete and get operations along with path and payload. This common handler converts all the requests into Transformer arguments and invokes the corresponding Transformer APIs.
+
+  4. For show commands, the CLI request is converted to a corresponding REST client SDK get request based on Open Config data model's config or state object on a case by case basis.
+
+  5. For show commands that requires retrieval of the data that doesn't contain any state information (information only based on the configuration), the backend callback will fetch the data from CONFIG_DB.
```
pavel-shirshov: what if configuration was changed by frr or vtysh?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+
+### 4.3.1 REST Get Sequence - Config information
+![FRR-BGP-REST-GET-SEQUENCE1](images/frr-bgp-rest-get-sequence1.jpg)
+
+### 4.3.2 REST Get Sequence - State/Statistics
+![FRR-BGP-REST-GET-SEQUENCE2](images/frr-bgp-rest-get-sequence2.jpg)
+
+# 5 Error Handling
+
+The complete validation of BGP parameters is performed in the SONiC management framework based on open config BGP/Routing policy YANGs and against corresponding SONiC YANGs. Appropriate error code is returned for invalid configurations or on failures due to a dependency from management framework and bgpcfgd modules and the same is logged into the syslog, it is not expected for bgpcfgd (FRR CLI command execution) to return validation errors. However, such errors are logged into syslog for debugging purposes.
```
pavel-shirshov: 1. It's impossible to validate frr configuration using only YANG models. There is no semantics in the models
2. what if bgpcfgd reports any errors - rest api will show no errors. I think we must show errors from bgpcfgd in the api

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+| RIB                      | Routing Information Base            |
+| PBR                      | Policy based routing                |
+| NBI                      | North Bound Interface               |
+
+
+
+# 1 Feature Overview
+
+This feature extends and provides unified configuration and management capability for FRR-BGP features used in SONiC. This allows the user to configure & manage FRR-BGP using SONiC Management Framework with Open Config data models via REST, gNMI and also provides access via SONiC Management Framework CLI as well.
+
```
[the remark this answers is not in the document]
venkatmahalingam: What extra information is needed exactly? I believe, the above sentence clearly talks about the support that we provide with mgmt-framework for BGP configurations.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+This document describes the high level design of FRR-BGP Extended Unified Configuration and Management feature.
+
+## Definition/Abbreviation
+
+### Table 1: Abbreviations
+| **Term**                 | **Meaning**                         |
+|--------------------------|-------------------------------------|
+| FRR                      | Free Range Routing Stack            |
+| CVL                      | Config Validation Library           |
+| VRF                      | Virtual routing forwarding          |
```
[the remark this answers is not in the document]
venkatmahalingam: will fix it.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: There are bunch of URLs (100+) that we support for BGP configs, adding every URL will affect the readability of this document, rather, referring the YANG itself will help to know what containers/leafs...we support from OC for BGP configs.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: Will change the sentence, we augment OC BGP model (openconfig-bgp-ext.yang) for any configs missing in OC.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: If this table is not for static peer-groups, not sure why we're using this table to generate the peer-group configs in the below file.

https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-fpm-frr/bgpd.conf.default.j2

{% if BGP_PEER_RANGE %}
{% for bgp_peer in BGP_PEER_RANGE.values() %}
  neighbor {{ bgp_peer['name'] }} peer-group
  neighbor {{ bgp_peer['name'] }} passive
{% if bgp_peer['peer_asn'] is defined %}
  neighbor {{ bgp_peer['name'] }} remote-as {{ bgp_peer['peer_asn'] }}
{% else %}
  neighbor {{ bgp_peer['name'] }} remote-as {{ constants.deployment_id_asn_map[DEVICE_METADATA['localhost']['deployment_id']] }}
{% endif %}
  neighbor {{ bgp_peer['name'] }} ebgp-multihop 255
  neighbor {{ bgp_peer['name'] }} soft-reconfiguration inbound

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: Yes, our long term goal is to push FRR contents into state DB and read them from mgmt-framework container, but for now, we fetch information directly from FRR BGP.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: For now, we use syslog to report an error for invalid configs, looks like currently we dont report any BGP config error to the user from bgpcfgd, do you have any suggestion?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: The recommendation is to use SONIC NBIs (CLI/REST client) for BGP configs, if user does some BGP configs in VTYSH, we may expect to see additional information in the show command.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: Do we have any tool available for schema auto generation from YANG model already?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: will fix it

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: We have the cache in bgpcfgd, this will help to retain the information until the dependent data is available, also, it helps to delete the configs upon field delete in BGP tables.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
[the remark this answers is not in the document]
pavel-shirshov: I think we should make it clear why do we need to introduce this feature.
Currently I see a lot of technology names, but I don't see any advantages.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
pavel-shirshov: No. We don't have such tool.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
pavel-shirshov: Not only static peers could have peer-group defined for them.
Check following
https://github.com/Azure/SONiC/wiki/Configuration

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
pavel-shirshov: I don't think it is good idea.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
pavel-shirshov: But we should be ready for this. The user 100% will change the configuration of the frr. We must be ready for that and have strategy how to work with it.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
pavel-shirshov: My opinion is to create working configuration for bgpd. syslog report will not work for me. How big is a chance that someone with REST API will check syslog output on the box during configuration? We must give correct output to users.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: Thanks, will update manually for now.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: Yes, we're aware of this, since this needs an FRR enhancement to push state information to state DB, will take this up later.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: We're aware of this, we recommend the customers not to use VTYSH for BGP configs, if customers use the both VTYSH and SONiC CLI, it's upto customer to resolve the config. conflicts if any.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: ok, thanks, will remove this sentence.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: Sure

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: With the help of config cache in bgpcfgd, we resolve the config dependency issues, but other than that all other validations are expected to be done in mgmt-framework backend code, I dont see any reason for throwing an error to syslog from bgpcfgd for valid configs, if there are any negative configs and we should throw error for those cases, we can handle later as and when we encounter them..what do you say?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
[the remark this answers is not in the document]
venkatmahalingam: In short, we are implementing REST interface for BGP configs via mgmt-framework, hope this is clear.

venkatmahalingam:
> 1. Please convert all pictures to .png. jpeg doesn't work good for diagrams
Sure.
> 2. Suggest to split the proposal to two parts: north-bound to DB, DB to SONiC FRR
> 3. Suggest to use special utility to generate the database schema from YANG model automatically. Don't write the schema manually.
>    .4. Suggest to rewrite bgpcfgd in go
> 4. Suggest to introduce validation loop: first northbound interface check syntax rules, then bgpcfgd uses frr to validate generated bgp configuration.
> 5. Suggest to return validation errors from bgpcfgd back to the requested. syslog message is not enough

Thanks for all your suggestions, will discuss internally and take up if required now.
> 6. What if some parameter is already configured and there is conflict of something being already in the database. What should we do? overwrite? remove old parameter and create new parameter?
If there are multiple config sources, it's hard to keep the configs in sync, for now, the short answer is, it is beyond our scope of the work.

> 7. What if some parameters are depends on the parameters which is changed by some rest request. How we can be sure we work with the parameters successfully?
We handle the dependencies in CVL layer (mgmt-framework), if there are any specific example, let us know, will try to see if that is already resolved.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
pavel-shirshov: In your design bgpcfgd is not connected to your frontends.
bgpcfgd reads the database only. Your design doesn't have any feedback mechanism from bgpcfgd to the frontends. I don't agree with your solution to output issues into the syslog from bgpcfgd. That means that the frontends will never see the issues.

venkatmahalingam: I understand that the current design does not have any feedback mechanism, will explore options for notifying the errors to user (front-end) in the future.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
```diff
@@ -0,0 +1,1069 @@
...
+## 1.1 Requirements
+
+
+### 1.1.1 Functional Requirements
+
+  1. Extend Unified mode for full FRR-BGP config and management in SONiC
+  2. Extend sonic-cfggen, bgpcfgd and integrate with FRR-BGP for features supported in SONiC
+  3. Support for retrieval of FRR-BGP state and statistics information
+  4. For backward compatibility retain access to FRR UI (vtysh) for managing features that are NOT in conflict with SONiC features
+
```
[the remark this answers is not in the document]
venkatmahalingam: This HLD primarily talks about the REST intf. configurations for BGP features and the enhancements on sonic-cfggen and bgpcfgd mainly for the BGP configs, will change the contents accordingly.

Backward compatibility mainly for the users who want to use VTYSH not mgmt-framework for their BGP configurations, so, the access to VTYSH will continue to be there.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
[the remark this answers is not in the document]
mikelazar: The idea is that REST and gNMI are two of the programmatic API's of choice for automation. This is why we implement them, not just for BGP, but for other features as well. This is actually the main purpose of the SONiC management framework in general.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
mikelazar: As a note, SONiC in general does not provide a means to notify front end users of back end errors (this not just an issue of this feature). It seems that there is work underway (another activity) to provide a means for back-end applications to notify front-end applications when an error occurs at the backend. We'll look into using that new framework when it becomes available.

My understanding is that the overall SONiC design, both for integrating 3rd party applications (e.g. DHCP and teamd etc.), and communication between its native components, uses the following approach:
-	A front end application (e.g. SONiC CLI/CLICK) or some other application performs configuration validation (semantic and syntactic)
-	If the configuration is valid, it is written to the Redis DB, which publishes DB change events
-	A back-end applications listens to those DB change events and pushes the changes to the 3rd party app (DHCP, teamd etc.) or a native SONiC component. 
o	Validation needs to happen at the front end – by the time the configuration data reached the back end, it has already been written to the DB

In the context of this feature, the SONiC Mgmt Framework (BGP REST and CLI implementation) is the front-end application, and bgpcfgd is the backend application. 
o	If we wrote data into FRR first (the only means to fully validate BGP config) – before writing data to the Redis DB – we’d go against the overall concepts of the SONiC architecture.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: We do read data from config-DB and generate FRR configs upon BGP container startup, do we expect something else?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: yes, will fix it.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: CVL (config validation library) as part of mgmt-framework validates data types, references..etc...and transformer functions capable of validating semantics.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: will fix it

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: This is default config method now in mgmt-framework, we dont need to mention it, will remove it.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: This is annotations for YANG model to map OC YANG objects to mgmt-framework transformer functions to handle redis DB write/get/delete operations.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: Yes, if any user wants the programmatic interface (REST) or KLISH for configuring FRR BGP, we support extensive BGP configurations via OC model using mgmt-framework and there is no need to use VTYSH.

In general, if programmatic intf is not a requirement or support is not available thru mgmt-framework for some SONiC features, user can use VTYSH directly for configuring FRR.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: Since we introduced lot of fields in various new/old BGP config tables, we had to enhance bgpd.conf.j2 file to generate FRR configs, hope this helps.

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
pavel-shirshov: FRR configs includes bfd, zebra and staticd configurations too.
Also during warm reboot we could find some parameters changed. Should we do anything with that?

doc/mgmt/SONiC_Design_Doc_Unified_FRR_Mgmt_Interface.md
venkatmahalingam: Sure, will add the UT scripts in the code.

venkatmahalingam:
@lguohan Please review the changes again, if everything looks good, merge the PR. Thanks.

lguohan:
merge this as it is. the yang model discussion for this is still ongoing in the community.

PR #550, opened 2020-01-08
Management Framework HLD - Phase II

anand-kumar-subramanian:
Management Framework HLD - Phase II contains the following

- Translib API changes and new APIs for REST and Telemetry to consume.
- RESTCONF yang library and other enhancements
- CVL API changes, platform and custom validation support


doc/mgmt/Management Framework.md
```diff
@@ -651,63 +845,111 @@ REST server will accept concurrent requests. Translib provides appropriate locki
...
+RESTCONF Notification are not supported by framework. Clients can use gNMI for monitoring and notifications.
 
-###### 3.2.2.4.12.3 User Authentication
+###### 3.2.2.4.12 Authentication
 
-In this mode REST server expects the client to provide user credentials in every request. server will support HTTP Basic Authentication method to accept user credentials.
+REST server supports following authentication modes.
 
-REST server will integrate with Linux PAM to authenticate and authorize the user. PAM may internally use native user database or TACACS+ server based on system configuration. REST write requests will be allowed only if the user belong to admin group. Only read operations will be allowed for other users.
+* HTTP Basic authentication (username/password authentication)
```
dev-aviz, line 943: How does it integrate with HamD & tacacs/radius ?

doc/mgmt/Management Framework.md
```diff
@@ -1526,11 +1934,24 @@ Example of 'deviation' :
 ###### 3.2.2.8.2.3.2.2 Platform data is available through APIs
 
 * If constraints cannot be expressed using YANG syntax or platform data is available through feature/component API (APIs exposed by a feature to query platform specific constants, resource limitation etc.), custom validation needs to be hooked up in SONiC YANG model through custom YANG extension.
```
dev-aviz, line 1936: Can you provide some example how to use CVL for custom validations?

doc/mgmt/Management Framework.md
```diff
@@ -1029,7 +1335,58 @@ These utilities methods provides below mentioned common operations on the YGOT s
         string - object field name of the given target object
         error - error object to describe the error if this methods fails to perform the desired operation, otherwise nil
 
-###### 3.2.2.6.4 DB access layer
+###### 3.2.2.6.4 YANG Model Versioning
+
+###### 3.2.2.6.4.1 Version Management
```
dev-aviz, line 1340: which rest module does the version management ?

doc/mgmt/Management Framework.md
```diff
@@ -841,37 +1093,80 @@ Translib is a library that adapts management server requests to SONiC data provi
         func Replace(req SetRequest) (SetResponse, error)
         func Delete(req SetRequest) (SetResponse, error)
         func Get(req GetRequest) (GetResponse, error)
-        func Subscribe(paths []string, q *queue.PriorityQueue, stop chan struct{}) ([]*IsSubscribeResponse, error)
-        func IsSubscribeSupported(paths []string) ([]*IsSubscribeResponse, error)
+        func Action(req ActionRequest) (ActionResponse, error)
+        func Bulk(req BulkRequest) (BulkResponse, error)
```
dev-aviz, line 1097: Do you have REST performance numbers to share to community? ex: add-delete-add vlan workflows etc..

doc/mgmt/Management Framework.md
```diff
@@ -1295,15 +1668,19 @@ The function prototypes for external transformer callbacks are defined in the fo
...
+        curDb db.DBNum
+        ygRoot *ygot.GoStruct
+        uri string
+        requestUri string //original uri using which a curl/NBI request is made
+        oper int
+        key string
+        dbDataMap *map[db.DBNum]map[string]map[string]db.Value
+        subOpDataMap map[int]*RedisDbMap // used to add an in-flight data with a sub-op
+        param interface{}
+        txCache *sync.Map
```
dev-aviz, line 1682: Can you shed light on what is txCache and how does it use and what is the use case we are trying to solve here?

doc/mgmt/Management Framework.md
```diff
@@ -1459,6 +1847,8 @@ CVL uses SONiC YANG models written based on ABNF schema along with various const
 
 SONiC YANG can be used as Northbound YANG for management interface by adding other data definitions such as state data (read only data), RPC or Notification as needed.Such YANG models are called SONiC NBI YANG models. Since CVL validates configuration data only, these data definition statements are ignored by CVL. During build time CVL YANG is actually generated from SONiC NBI YANG models with the help of CVL specific pyang plugin. 
 
+CVL supports multiple user-defined Redis database instances based on the [Multi DB instance HLD](https://github.com/Azure/SONiC/blob/master/doc/database/multi_database_instances.md) specification. During CVL initialization time, the DB configuration file is read from the predefined location and details are stored in internal data structure. CVL implements internal API for connecting to a Redis DB instance using the details stored in its internal data structure.
```
dev-aviz, line 1850: what are the use case for multi redis db ? how does it matter w.r.t management framework?

doc/mgmt/Management Framework.md
```diff
@@ -651,63 +845,111 @@ REST server will accept concurrent requests. Translib provides appropriate locki
...
+
+API Versioning is not supported for manual OpenAPI definitions in this release.
+Accept-Version header value will be ignored even if specified.
 
 ###### 3.2.2.4.7 RESTCONF Entity-tag
 
-REST server will support RESTCONF entity-tag and last-modified timestamps in next release. server will not process or send corresponding request, response headers in first release.
-Note that entity-tag and last-modified timestamps will be supported only for top level datastore node (/restconf/data). Per resource entity tags and timestamps will not be supported. Global entity tag and timestamp are used for configuration resources.
+REST server does not maintain entity-tag and last-modified timestamp for configuration data nodes.
+GET and HEAD responses does not include "ETag" and "Last-Modified" headers.
```
dev-aviz, line 869: what is the significance of ETag  & Last-Modified exclusion from the rest URLs?

doc/mgmt/Management Framework.md
```diff
@@ -651,63 +845,111 @@ REST server will accept concurrent requests. Translib provides appropriate locki
...
+REST server supports "GET /restconf/data/ietf-restconf-monitoring:restconf-state/capabilities" API to
+advertise its capabilities as described in [RFC8040, section 9.1](https://tools.ietf.org/html/rfc8040#section-9.1).
+REsponse includes below mentioned capability information.
+
+    urn:ietf:params:restconf:capability:defaults:1.0?basic-mode=report-all
+
+###### 3.2.2.4.9 RESTCONF Query Parameters
 
-Only one mode can be active at a time. Administrator can choose the authentication mode through ConfigDB REST_SERVER table entry. See [DB Schema](#3_2_2_4_14-db-schema) section.
+RESTCONF Query Parameters will be supported in future release. All query parameters will be ignored by REST server in this release.
```
dev-aviz, line 923: Does the design support pagination ? The possible usecase could be - for the long API responses, basically to consumek by any UX applications

doc/mgmt/Management Framework.md
```diff
@@ -167,7 +175,7 @@ Management framework is a SONiC application which is responsible for providing v
     2. Custom YANG models ([SONiC YANG](https://github.com/Azure/SONiC/blob/master/doc/mgmt/SONiC_YANG_Model_Guidelines.md))
     3. Industry-standard CLI / Cisco like CLI
```
dev-aviz, line 176: Does management framework allows sonic cli & ISCLI intertwined? -config made through sonic cli shows up in ISCLI viceversa?

doc/mgmt/Management Framework.md
sachinholla: REST server uses ssh to host with the username and password obtained from HTTP header. This implementation is already present in community code.

doc/mgmt/Management Framework.md
sachinholla: Version management logic will be in translib -- it is common to both REST and gNMI services.

doc/mgmt/Management Framework.md
sachinholla: Mgmt framework code has its own DB access APIs (there is no swss DB APIs in Go). It should honor the new database config file as described in multi db HLD.

doc/mgmt/Management Framework.md
sachinholla: REST client cannot use RFC7232 style conditional requests now. It would have been useful for clients that cache the server data. Conditional requests allow clients to get data only if it is modified or set only if not modified etc..

doc/mgmt/Management Framework.md
sachinholla: RESTCONF does not have provision for pagination. So the Mgmt REST server does not support any "page cookie" like query parameters. We plan to add HTTP/1.1 chunked transfer or HTTP/2 stream support in future releases. This should ease the server load while servicing large response payload.

App developer can provide old style cookie based pagination through yang custom RPCs. Clients can specify page cookie in request body and not query parameter. However this will be an app specific API and not generic server-wide API.

doc/mgmt/Management Framework.md
sachinholla: Yes, the framework allows it; but it depends on the command implementation. You can mix the CLI channels as long as they operate on same set of tables. So, it is better to stick to one.

doc/mgmt/Management Framework.md
kwangsuk: The txCache (transaction cache) is a placeholder to cache data between transformer callbacks. 
It can be used for the case that a callback need some optimization by caching data and passing to other callbacks for subsequent use of the same data retrieved from the txCache of the inParams. The txCache is bound to the given transaction scope.

doc/mgmt/Management Framework.md
anand-kumar-subramanian: It is as simple as specifying the custom validator function in sonic yang and implementing the same function(in go) in the custom validation file.

sonic yang:
sonic-ext:custom-validation ValidatePtp;

go function:
func (t *CustomValidation) ValidatePtp(
    vc *CustValidationCtxt) CVLErrorInfo {
}

chenkelly:
Hi @anand-kumar-subramanian 
Are the enhancements described in the HLD fully developed?
For example, in 3.1.2.2 REST said "REST client will use the OpenAPI generated client SDK to send the request to the REST server."
After we checked current mk file (https://github.com/Azure/sonic-mgmt-framework/blob/master/models/openapi_codegen.mk#L26), still using swagger-codegen.
In 3.2.1.1.5 Special handling, don't see related processing about 'maxItems' and 'minItems'.
We will be grateful for any help you can provide.

faraazbrcm:
@chenkelly 
Are the enhancements described in the HLD fully developed?

The HLD content and the code present in Github seem to be not in sync. Recently Server moved to the OpenAPI generator (https://github.com/OpenAPITools/openapi-generator).

As per 3.1.1 in HLD - "In case of OpenAPI spec, it is directly given to the OpenAPI generator to generate the REST client SDK and REST server stubs in golang. In this case the REST server takes care of validating the incoming request to be OpenAPI compliant before giving the same to Translib."

The maxItems and minItems should be supported. The expectation is that generated stub should contain the validation code.

PR #557, opened 2020-01-28
ASIC Thermal Monitoring HLD

padmanarayana:
This document describes the high level design of a poller for ASIC thermal sensors.

### DIFF
--- a/doc/asic_thermal_monitoring_hld.md
+++ b/doc/asic_thermal_monitoring_hld.md
@@ -1,0 +1,255 @@
+# ASIC thermal monitoring High Level Design
+### Rev 0.1
+## Table of Contents
+
+## 1. Revision 
+Rev | Rev	Date	| Author	| Change Description
+---------|--------------|-----------|-------------------
+|v0.1 |01/10/2019  |Padmanabhan Narayanan | Initial version
+
+## 2. Scope
+ASICs typically have multiple internal thermal sensors. This document describes the high level design of a poller for ASIC thermal sensors. It details how the poller may be configured and the export of thermal values from SAI to the state DB.
+
+## 3. Definitions/Abbreviations
+
+Definitions/Abbreviation|Description
+------------------------|-----------
+ASIC|Application Specific Integrated Circuit. In SONiC context ASIC refers to the NPU/MAC.
+PCIe|Peripheral Component Interconnect express
+SAI| Switch Abstraction Interface
+SDK| Software Development Kit
+NOS| Network Operating System
+
+## 4. Overview
+
+Networking switch platforms are populated with a number of thermal sensors which include exteral (i.e. onboard) as well as internal (those located within the component, e.g. CPU, ASIC, DIMM, Transceiver etc..) sensors. Readings from both the external as well as the internal sensors are essential inputs to the thermal/fan control algorithm so as to maintain optimal cooling. While drivers exist to retrive sensor values from onboard and other internal sensors, the ASIC based sensor values are currently retrieved thru the ASIC's SDK. SAI provides the following attributes to retrive the ASIC internal sensors:
+
+|Attribute|Description|
+|---|------|
+|SAI_SWITCH_ATTR_MAX_NUMBER_OF_TEMP_SENSORS| Maximum number of temperature sensors available in the ASIC |
+|SAI_SWITCH_ATTR_TEMP_LIST|List of temperature readings from all sensors|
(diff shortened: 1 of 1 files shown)


jleveque:
@keboliu, @Junchao-Mellanox: Please review.

doc/asic_thermal_monitoring_hld.md
```diff
@@ -0,0 +1,255 @@
...
+            retrieve
+
+        Returns:
+            An object dervied from AsicBase representing the specified ASIC
+            module
+        """
+
+```
+
+The ThermalBase() implementation for ASIC sensors will return the latest values exported by the SwitchOrch to the stateDB's TEMPERATURE_INFO table.
```
keboliu: looks strange to me, the current idea is that platform API responsible to get the thermal sensor info from the lower layer and populate it to DB, and state DB info will feed CLI/SNMP or telemetry etc. but here you are suggesting platform API to read sensor info from DB?

doc/asic_thermal_monitoring_hld.md
```diff
@@ -0,0 +1,255 @@
...
+
+|Attribute|Description|
+|---|------|
+|SAI_SWITCH_ATTR_MAX_NUMBER_OF_TEMP_SENSORS| Maximum number of temperature sensors available in the ASIC |
+|SAI_SWITCH_ATTR_TEMP_LIST|List of temperature readings from all sensors|
+|SAI_SWITCH_ATTR_AVERAGE_TEMP|The average of temperature readings over all sensors in the switch|
+|SAI_SWITCH_ATTR_MAX_TEMP|The current value of the maximum temperature retrieved from the switch sensors|
+
+A configurable ASIC sensors poller is introduced that periodically retrieves the ASIC internal sensor values via SAI APIs and populates the state DB. These values may be used by the thermal control functions (via the ThermalBase() platform APIs), SNMP/CLI or other Telemetry purposes.
```
keboliu: How to avoid duplicated entries if on some platform ASIC thermals can be provided both from platform API and SAI?

padmanarayana: I don't think platform API and SAI are necessarily exclusive (to generate duplicate entries): if there is no SAI support, the poller would be disabled and the platform API is going to query the ASIC's SDK to populate the values (as part of "SONiC Thermal Control Design"). If there is SAI support, the poller would be the preferred way.

doc/asic_thermal_monitoring_hld.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+The TEMPERATORE_INFO stateDB table will be expanded to hold the ASIC internal temperatures. The object_name would be:
+asic_<0..(n-1)>_ internal_<0..(N-1)> where "n" is the number of ASIC and "N" is the number of internal sensors on the ASIC.
+...
+
+A new ASIC_SENSORS stateDB table is added which captures the following additional details:
+
+```
+; Defines thermal information for an ASIC
```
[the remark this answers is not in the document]
padmanarayana: The schema is aligned "The TEMPERATORE_INFO stateDB table will be expanded to hold the ASIC internal temperatures.".

It should be possible to deduce the number of internal sensors based on the TEMPERATORE_INFO entries. Also, if SAI for the ASIC supports the individual sensor list, average_temperature and maxinum_temperature may not be needed. My thought was that there may be ASICs that support only max or avg.

doc/asic_thermal_monitoring_hld.md
padmanarayana: The objective is to present ASIC internal sensors to any advanced thermal control algorithm. The reason it appears dissimilar to the standard platform sensors is because the intent was not to let pmon (or any platform daemon) make SAI calls (and create a new dependency) since this is a unique case. Hence letting the poller be part of SwitchOrch.

jleveque:
@keboliu: Please review responses.

liat-grozovik:
@Junchao-Mellanox and @keboliu appreciate your feedback
- Is there any reason to have an additional thermal monitoring while there is an already valid solution for thermal control and new CLIs for doing so?
- If there is something missing I believe it should be added on top of the existing functionality and not create new and parallel one.

jleveque (approved):
Looks good to me. Please wait for other reviewers.

keboliu (approved):

sujinmkang:
https://github.com/Azure/sonic-buildimage/pull/6556
https://github.com/Azure/sonic-swss/pull/1517
https://github.com/Azure/sonic-swss-common/pull/419

PR #563, opened 2020-02-19
PFC Asymmetric test plan pytest base

yvolynets-mlnx:
This document describes the high level design of PFC Asymmetric test cases based on pytest framework.

### DIFF
--- a/doc/pfc_asym/PFC_Asymmetric_Test_HLD.md
+++ b/doc/pfc_asym/PFC_Asymmetric_Test_HLD.md
@@ -1,0 +1,423 @@
+# Asymmetric PFC Test Plan
+
+* [Overview](#Overview)
+   * [Scope](#Scope)
+   * [Testbed](#Testbed)
+* [Setup configuration](#setup-configuration)
+* [Existed modules refactoring](#existed-modules-refactoring)
+* [Python  modules to setup and run test](#python-modules-to-setup-and-run-test)
+   * [Python modules](#python-modules)
+   * [PFC storm global variables](#pfc-storm-global-variables)
+   * [PTF test case execution](#ptf-test-case-execution)
+   * [Pytest fixtures](#pytest-fixtures)
+		* [deploy_pfc_gen](#deploy_pfc_gen)
+		* [setup](#setup)
+		* [pfc_storm_template](#pfc_storm_template)
+		* [pfc_storm_runner](#pfc_storm_runner)
+		* [enable_pfc_asym](#enable_pfc_asym)
+* [Test](#Test)
+* [Test cases](#Test-cases)
+
+## Overview
+The purpose is to test functionality of Asymmetric PFC on the SONIC based DUT, closely resembling production environment.
+
+### Scope
+The test is targeting a running SONIC system with fully functioning configuration. The purpose of the test is to perform functional testing of Asymmetric PFC on SONIC system. There will be reused existed PTF test suite for PFC Asymmetric which is located at https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/test/files/saitests/pfc_asym.py.
+
+### Testbed
+The test will run on the following testbeds:
+* T0
+
+### DUT image
+RPC image is required
+
+## Setup configuration
+No setup pre-configuration is required, test will configure and clean-up all the configuration.
+
+## Existed modules refactoring
+There is already existed PFC asymmetric test cases which use PTF to send traffic.
+
+Test suite location
+
+https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/test/files/saitests/pfc_asym.py
+
+It requires several updates:
+
+1. Packets sending speed can be increased by using multiprocessing instead of multithreading library.
+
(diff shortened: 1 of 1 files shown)


doc/pfc_asym/PFC_Asymmetric_Test_HLD.md
```diff
@@ -0,0 +1,423 @@
...
+		* [pfc_storm_runner](#pfc_storm_runner)
+		* [enable_pfc_asym](#enable_pfc_asym)
+* [Test](#Test)
+* [Test cases](#Test-cases)
+
+## Overview
+The purpose is to test functionality of Asymmetric PFC on the SONIC based DUT, closely resembling production environment.
+
+### Scope
+The test is targeting a running SONIC system with fully functioning configuration. The purpose of the test is to perform functional testing of Asymmetric PFC on SONIC system. There will be reused existed PTF test suite for PFC Asymmetric which is located at https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/test/files/saitests/pfc_asym.py.
```
liat-grozovik: just to clarify.
currently we have an ansible test for this functionality. the ansible wrapper will call the new pytest code for backwards compatibility but we can run pytest without this ansible wrapper, right?
in that case is it better to move the existing pfc_asym.py to better place in the source tree or this is the proper place for it?

doc/pfc_asym/PFC_Asymmetric_Test_HLD.md
```diff
@@ -0,0 +1,423 @@
...
+
+### Testbed
+The test will run on the following testbeds:
+* T0
```
liat-grozovik: why only T0? if this is indeed only for T0 only why not all the T0-x options?

doc/pfc_asym/PFC_Asymmetric_Test_HLD.md
```diff
@@ -0,0 +1,423 @@
...
+
+Test suite location
+
+https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/test/files/saitests/pfc_asym.py
+
+It requires several updates:
+
+1. Packets sending speed can be increased by using multiprocessing instead of multithreading library.
+
+https://github.com/Azure/sonic-mgmt/blob/master/ansible/roles/test/files/saitests/pfc_asym.py#L83
```
liat-grozovik: can you explain why this is needed? what does it wishes to fix/handle?

doc/pfc_asym/PFC_Asymmetric_Test_HLD.md
```diff
@@ -0,0 +1,423 @@
...
+## Test cases
+
+### Test case # 1 – Asymmetric PFC Off TX pause frames
+#### Test objective
+Asymmetric PFC is disabled. Verify that DUT generates PFC frames only on lossless priorities when asymmetric PFC is disabled
+
+#### Used fixtures
+setup
+
+#### Test steps
```
liat-grozovik: the test should run on different SKUs, right? in the case where we have more than 32 ports, how will it work?
are you also using the port channels?

doc/pfc_asym/PFC_Asymmetric_Test_HLD.md
yvolynets-mlnx: Regarding
"currently we have an ansible test for this functionality. the ansible wrapper will call the new pytest code for backwards compatibility but we can run pytest without this ansible wrapper, right?" - Correct

Regarding
"in that case is it better to move the existing pfc_asym.py to better place in the source tree or this is the proper place for it?"
Right now test cases which are directly executed in PTF docker are stored in the same common folder. So for now I suppose better will be to store them as it is

doc/pfc_asym/PFC_Asymmetric_Test_HLD.md
yvolynets-mlnx: Yes test can run on different SKUs.

When on DUT will be more then 32 ports available it is possible to specify how much ports to use/test in the current setup by using external option for test cases run.

Regarding - "are you also using the port channels?"
PFC Asymmetric is not enabling on port channels, regarding requirements for Asymmetric PFC "PFC_ASYM will be set to ON on T0 switches, on interfaces facing the servers. It will be OFF on T1 facing interfaces".

In the current test cases port channels are using as connection with leaf switches, so traffic will be sent between port channel and server interfaces.

doc/pfc_asym/PFC_Asymmetric_Test_HLD.md
yvolynets-mlnx: Theoretically I/O operations will be executed faster by using separate process rather then python threading because of python GIL.

yvolynets-mlnx: As test bed is very limited by traffic generation possibilities, it can be helpful in some cases.

doc/pfc_asym/PFC_Asymmetric_Test_HLD.md
yvolynets-mlnx: Fixed.

PR #567, opened 2020-02-29
Support redis databases in multiple namespaces

lguohan:


doc/database/multi_namespace_db_instances.md
```diff
@@ -0,0 +1,495 @@
...
+
+## The Multi NPU/namespace architecture 
+
+In the Multi NPU devices, the services could be broadly classified into
+
+	* Global services like database, snmp, pmon, telemetry running in the dockers running in the linux "host" network namespace.
+	
+	* NPU specific services like database, swss, syncd, bgp, teamd, lldp etc which runs in a separate "namespace" created. 
+	  The method used currently to seggregate services per NPU is "linux network namespace" and there is a one-to-one 
+	  mapping between the number of NPUs and linux network namesapces created.
```
lguohan: format issue here.

doc/database/multi_namespace_db_instances.md
```diff
@@ -0,0 +1,495 @@
...
+
+There will be a config_db.json file per database service, it will be named "/etc/sonic/config_db.json" for the "global DB" and /etc/sonic/config_db{NS}.json for the database service in the {NS} namespace.
+
+## Enhancements to Multi-DB design to support Multiple namespaces
+
+The file database_config.json contains the startup configuration which dictates the redis server host/port/unix_socket configurations + various available databases ( eg: APP_DB, CONFIG_DB ) etc. With the introduction of multiple namespaces, there is a need for separate database_config.json per namespace. This config file is created in the "working redis directories" during the database docker service startup. 
+
+The database docker service started in the linux network namespace {NS} will use the "/var/run/redis{NS}" as the "working redis directory" to create the various files like redis.sock, sonic-db/database_config.json etc. For "globalDB" database service it would be at "/var/run/redis".
```
lguohan: the database docker service does not have to be started in the linux network namespace. suggest to change to "database service" for a NPU linux namespace should use /var/run/redis{NS{. as the working directory.

doc/database/multi_namespace_db_instances.md
```diff
@@ -0,0 +1,495 @@
...
+
+* The startup config file database_config.json is modified to have 'default' namespace DB instances + external references to the database instances present (if any) in database dockers running in other namesapces. There is a new attribute introduced, 
+
+	* "INCLUDES" to store external database_config.json file references. Each INCLUDES entry will contain the 
+	  "namespaceID" along with the "relative path" which tells the file location of database_config.json. 
+	  
+This is an example of the database_config.json with 3 external DB references in 3 namespaces refered with namespaceID "0", "1" & "2".
+
+```json
+{
```
lguohan: the json file is not clean. I suggest to have only includes in the global database json. you can put the global db configuration into a separate file and use include to include it.

doc/database/multi_namespace_db_instances.md
```diff
@@ -0,0 +1,495 @@
...
+        return got_data
+
+
+```
+Note: There are many other utilities which needs this additional parameter, will be updated in a similar fashion.
+
+## Design of C++ Interface :  DBConnector()
+The C++ DBConnector interface is mainly used in the process context of swss/syncd etc in the dockers in their respective namespaces. In the namespaces currently we don't have any external DB refereces, hence not planning to extend the C++ DBConnector to handle multiple database_config.json files.
+
+Will be taken up as the next phase activity to get this class also at par with the python SonicDBConfig/SonicV2Connector modules.
```
lguohan: this is a must because we are deprecating the pure swsssdk library and plan to python-swsscommon library for all python related applications.

doc/database/multi_namespace_db_instances.md
judyjoseph: I have explicitly added details of the database_config.json file which will get generated for the global DB docker service and the NPU namespace DB docker service.  

There is no particular file like global DB database_config.json file. The current approach I have taken is to generate the json file depending on where the database docker service is running (either in linux host, or NPU namesapce )

Please find that the "INCLUDES" section is only present for the globalDB database_config.json (created at /var/run/redis/sonic-db) as the applications like snmp,pmon running in the dockers in the Linux host needs to talk to DB in other NPU namespaces. 

The database_config.json file for the NPU namespace docker won't have INCLUDES section. It will only have the default local DB instances, as the applications running in the namespace like swss, syncd, bgp etc will only talk to its local DB instances.

judyjoseph: Is there a use case where the applications/dockers running in a “particular NPU namespace” will need to access the DB instances in the database dockers running in either “linux host namespace” OR “other NPU namespaces” ? 

Currently my assumption is that only the applications running in the host, like snmp, CLI, pmon etc “needs to access” the DB instances running in other NPU namespaces.

Or should we make it more generic, like any application running in any docker in host/NPU namespace should be able to access the DB instance running in a different NPU namespace ? Please could you share your thoughts. 

If Yes, then I understand your ask. I would change accordinly and update the doc. Thanks

doc/database/multi_namespace_db_instances.md
```diff
@@ -0,0 +1,518 @@
...
+        },
+        "CONFIG_DB" : {
+            "id" : 4,
+            "separator": "|",
+            "instance" : "redis"
+        },
+    },
+    "INCLUDES" : [
+        {
+	    "namespace_id" : "0",
```
lguohan: namespace_id -> namespace

doc/database/multi_namespace_db_instances.md
lguohan: example

```
{
   [
       {
	    "config" : "../redis/sonic-db/database_config.json"
	},
	{
	    "namespace" : "0",
	    "config" : "../redis0/sonic-db/database_config.json"
        },
	{
	    "namespace" : "1",
	    "config" : "../redis1/sonic-db/database_config.json"
        },
    ],
    "VERSION" : "1.1"
}
```

judyjoseph: Updated the document, by adding a new file "databases.json" which has the namespace names and the database_config.json file references. The INCLUDES will have the path for this new file.
Additional attribute added in the database_config.json file "NAMESPACE" which will have the namespace name stored.

doc/database/multi_namespace_db_instances.md
qiluo-msft: Could you clarify "the database docker service does not have to be started in the linux network namespace" ?
If not in namespace, how to specify one database container instance?

For example, currently we are using `sonic-db-cli STATE_DB hset`, what will be new command link like?

doc/database/multi_namespace_db_instances.md
```diff
@@ -0,0 +1,561 @@
...
+![center](./img/redis.png)
+
+Following are the major design changes
+
+* A new file **databases.json** is introduced. It will contain the details of the namespaces and the corresponsing database_config.json files. This file would be created by the "globalDB" service in the directory "/var/run/redis/sonic-db/". In the below example, we consider a SONIC device with 3 NPUS's and hence have 3 namespaces refered with namespaceID "0", "1" & "2".
+
+```json
+[
+    {
+        "namespace" : "global",
```
qiluo-msft: >global [](start = 23, length = 6)

global is a normal string and could be used as a valid namespace name.
As guohan's sample, just skill this field for default namespace. In code, you can use empty string to represent.

doc/database/multi_namespace_db_instances.md
```diff
@@ -0,0 +1,561 @@
...
+        "config" : "../redis/sonic-db/database_config.json"
+    },
+    {
+        "namespace" : "0",
```
qiluo-msft: >"namespace" : "0", [](start = 8, length = 18)

Do we need 2 fields?
```
"namespace" : "asic0",
"id" : "0"
```

I notice related scripts using both extensively, and deduce from one to another.

doc/database/multi_namespace_db_instances.md
```diff
@@ -0,0 +1,561 @@
...
+* The startup config file database_config.json is modified to have 'default' namespace DB instances + external references to the database instances present (if any) in database dockers running in other namesapces. There are a few attributes introduced, 
+
+	* "NAMESPACE" will store the namespace to which this database_config.json belongs.
+	* "INCLUDES" to point to the path of the databases.json file.
+
+* The database_config.json file for the "globalDB" service will have the "INSTANCES"/"DATABASES" (global DB service to have only APP_DB and CONFIG_DB) running localy in globalDB database docker on linux host namespace + INCLUDES attribute to refer to the external DB instances in the database docker services running in other NPU namespaces. The sample format is as below.
+
+```json
+{
+    "NAMESPACE": "global",
```
qiluo-msft: >"NAMESPACE": "global", [](start = 4, length = 22)

Why we need this? If we can remove it, keep original version.

doc/database/multi_namespace_db_instances.md
judyjoseph: Fixed the formatting.

doc/database/multi_namespace_db_instances.md
judyjoseph: The database docker service needs to be started in some network namespace. It could be a new "net_ns" which we create, or the linux host network namespace. 

A new parameter is added to  sonic-db-cli, 
sonic-db-cli APPL_DB **-namespace <ns>** HGET VLAN_TABLE:Vlan10 mtu

if we don't specify the namespace parameter, it executes the redis command in the DB instance in the database docker running that namespace where we executed the sonic-db-cli command. 

Do let me know if we need more discussion on this.

doc/database/multi_namespace_db_instances.md
judyjoseph: Updated the doc, will take up the changes to DBConnector class as well and raise the PR for those changes.

doc/database/multi_namespace_db_instances.md
judyjoseph: updated

doc/database/multi_namespace_db_instances.md
judyjoseph: The new file **databases.json** is a static file containing the namespace names and the correspoding database_config.json file.

The way I parse the database_config.json file (in class SonicDBConfig), is load the INSTANCES and DATABASES defined directly in the file as **SonicDBConfig._sonic_db_config['default'] = json.load(read_file) ** as default DB's
and then loop through all the files defined in the **"INCLUDES" : "../redis/sonic-db/databases.json",**. 

So when I loop through the files in **"../redis/sonic-db/databases.json"** I should skip the current NAMESPACE. I added this new attribute, to tell what namespace this "database_config.json" belongs to - and it is explicit detail which I felt was better.

doc/database/multi_namespace_db_instances.md
judyjoseph: The naming convention I am refering here is as follows,

"global"  -- for linux host namespace 
"0" -- for linux network namespace for NPU 0
"1" -- for the linux network namespace for NPU 1

There is no "default" namespace as such as you pointed out. Which ever namespace you are present while running a script, executing a config/show command etc --- is the default namespace. if you run the config command from linux shell in host , the default namespace is "global". If you run the sonic-cfggen tool in the swss docker running in namespace "2", then the default namespace is "2".

Do let me know if I am able to clarify.

doc/database/multi_namespace_db_instances.md
judyjoseph: I think you are trying to make the namespace name a bit more meaningful.

The idea I have is to change the namespace names from "0", "1", "2" .. --> to "namespace"0, "namespace"1, "namespace"2 etc ... 

Currently we use the string "asic0", "asic1", "asic2" etc . But I feel the prefix "asic" is not correct for a namespace name -- as tomorrow we could support more than 1 ASIC in the same namespace.

Can we keep it as "ns0", "ns1" etc ? Please share thoughts on the naming strategy... These names will be used as input string for "-namespace" parameter being introduced to sonic-utilities, config/show commands etc.

doc/database/multi_namespace_db_instances.md
judyjoseph: Removed the references to the keyword "namespace" in the json files.

doc/database/multi_namespace_db_instances.md
judyjoseph: Used the namespace name "asic0", "asic1" etc instead of "0", "1" etc

doc/database/multi_namespace_db_instances.md
judyjoseph: Removed the references to the keyword "namespace" in the json files.

doc/database/multi_namespace_db_instances.md
judyjoseph: The document is updated, Please review.

PR #579, opened 2020-03-19
Sonic Proxy Arp support

prsunny:
High-level requirement and design proposal

### DIFF
--- a/doc/Proxy Arp.md
+++ b/doc/Proxy Arp.md
@@ -1,0 +1,75 @@
+
+# Introduction
+
+The scope of this document is to provide the requirements and a high-level design proposal for Proxy ARP support. 
+
+# Requirements
+
+The following are the high level requirements when an interface is enabled with "proxy_arp"
+
+1. Proxy arp shall be set for the interface in kernel
+2. Hardware ARP packet action for that interface must be trap to CPU and not flooded/forwarded in hardware.
+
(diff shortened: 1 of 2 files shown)


doc/Proxy Arp.md
```diff
@@ -0,0 +1,75 @@
...
+
+![](https://github.com/Azure/SONiC/blob/master/images/vxlan_hld/proxy_arp_flood.png)
+
+# Additional Notes
+1. The flooding is disabled only for those interfaces belonging to a Vnet or user-configured proxy_arp setting. The implementation shall not modify the existing behavior and shall be backward compatible. 
+2. VS test can be added to existing ```test_vnet.py``` to verify the kernel/SAI configuration.
+3. Proxy ND is not planned as part of this feature but can be extended in future based on the same approach
+4. ```/proc/sys/net/ipv4/conf/Vlan2000/proxy_arp``` is not required to be set.
+5. Reference on Vnet/Vxlan design is [here](https://github.com/Azure/SONiC/blob/master/doc/vxlan/Vxlan_hld.md)
+
```
lguohan: missing tests section.

lguohan:
can you move into arp folder?

doc/Proxy Arp.md
prsunny: done- added a note to document

PR #587, opened 2020-04-03
[docs] Media enhancements

hpersh-dell:
- Media information enhancements
- Media settings enhancements

Signed-off-by: Howard Persh <Howard_Persh@dell.com>

### DIFF
--- a/doc/media-info-enhancements/media-info.md
+++ b/doc/media-info-enhancements/media-info.md
@@ -1,0 +1,170 @@
+# Enhancements to Optical Transceiver Information Model
+## Introduction
+Currently, xcvrd parses the EEPROMs of optical media transceivers, and builds a Python data structure, a dictionary, containing transceiver information.
+
+This design is an effort to enhance and standardize the representation of transceiver information.
+## Transceiver information
+### Parsing
+Xcvrd shall parse optical media transceiver EEPROMs, by means of core, platform-independent code, and also using an optional plugin provided by a platform implementation, to handle possibly vendor-proprietary information encoded in the EEPROM.
+
+The architecture of the xcvrd code is beyond the scope this document; that is the HOW of obtaining transceiver information, this document concentrates on the WHAT (syntax and semantics) of transceiver information.
+
+This design is intended to standardize the minimum set of transceiver information; platform-dependent plug-ins may provide additional information, but all fields in this document must be present for all platforms.
+### Goals
+The goal is to have xcvrd build a Python dictionary, containing all static transceiver information (i.e. transceiver information that does not change over time) for an installed transceiver, information that is needed to:
+- properly configure the NPU for the installed transceiver, e.g. set the NPU interface type to "fiber" or "copper";
+- provide information for making a platform-dependent, port-dependent and transceiver-dependent lookup of port configuration information to be passed to the NPU, e.g. look up port pre-emphasis value(s);
+- provide information for determining possible port configurations, and validating requested port configurations, e.g. determining what, if any, breakout modes are possible on a port; and
+- display transceiver information to the user.
+
+The format of this dictionary, i.e. the member fields, and the contents of all those fields, shall be standardized, and platform-independent, so that the code to configure the NPU and display information to the user can also be platform-independent.
+
+This dictionary, once created by xcvrd, is published to the Redis DB, as a record in the TRANSCEIVER_INFO table.  NPU-configuration and CLI code will obtain transceiver information from there.
+### Current format
+An example of the transceiver information dictionary constructed by xcvrd is as follows:
+~~~
+{
+	"Connector": "No separable connector", 
+	"cable_length": "2", 
+	"cable_type": "Length Cable Assembly(m)", 
+	"encoding": "64B66B", 
+	"ext_identifier": "Power Class 1(1.5W max)", 
+	"ext_rateselect_compliance": "QSFP+ Rate Select Version 1", 
+	"hardwarerev": "A0", 
+	"manufacturename": "DELL", 
+	"modelname": "76V43", 
+	"nominal_bit_rate": "255", 
+	"serialnum": "CN0LXD0096H3A36", 
+	"specification_compliance": "{'10/40G Ethernet Compliance Code': '40GBASE-CR4'}", 
+	"type": "QSFP28 or later", 
+	"type_abbrv_name": "QSFP28", 
+	"vendor_date": "2019-06-17 ", 
+	"vendor_oui": "3c-18-a0"
+}
+~~~
+### Proposed format
+The intention is to maintain existing field names and their contents, for backward-compatibility.  New fields shall be added.
+
+The following shall be added to the dictionary:
+
+<h4>cable_breakout</h4>
+<table style="border-style:none;">
+	<tr>
+		<td valign="top">Description</td>
+		<td>Breakout of attached cable, if any, i.e. the number of cables directly attached to transceiver</td>
+	</tr>
+	<tr>
+		<td valign="top">Value(s)</td>
+		<td>integer; 0 when no directly-attached cable</td>
+	</tr>
+	<tr>
+		<td valign="top">Needed by</td>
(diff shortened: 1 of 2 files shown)


doc/media-info-enhancements/media-info.md
```diff
@@ -0,0 +1,170 @@
...
+	"vendor_oui": "3c-18-a0"
+}
+~~~
+### Proposed format
+The intention is to maintain existing field names and their contents, for backward-compatibility.  New fields shall be added.
+
+The following shall be added to the dictionary:
+
+<h4>cable_breakout</h4>
+<table style="border-style:none;">
```
lguohan, line 51: can you change to markdown format for this section?

doc/media-info-enhancements/media-info.md
```diff
@@ -0,0 +1,170 @@
...
+	<tr>
+		<td valign="top">Needed by</td>
+		<td>Validation of port breakout configurations, i.e. whether or not transceiver supports breakout</td>
+	</tr>
+</table>
+<h4>cable_length_detailed</h4>
+<table style="border-style:none;">
+	<tr>
+		<td valign="top">Description</td>
+		<td>Length of directly-attached cable, in metres</td>
```
lguohan, line 69: i can see the cable_length already in the schema, what is the difference here?

doc/media-info-enhancements/media-info.md
```diff
@@ -0,0 +1,170 @@
...
+</table>
+<h4>form_factor</h4>
+<table style="border-style:none;">
+	<tr>
+		<td valign="top">Description</td>
+		<td>Detailed transceiver type, i.e. form factor</td>
+	</tr>
+	<tr>
+		<td valign="top">Value(s)</td>
+		<td>string, one of "SFP", "SFP+", "SFP28", "QSFP+", "QSFP28", "QSFP28-DD", "QSFP56-DD", "RJ45"</td>
```
lguohan, line 160: missing osfp, sfp56

lguohan: there is a type_abbrv_name, what is the augment here?

doc/media-info-enhancements/media-info.md
```diff
@@ -0,0 +1,170 @@
...
+	</tr>
+	<tr>
+		<td valign="top">Needed by</td>
+		<td>Layer-1 configuration lookup, e.g preemphasis</td>
+	</tr>
+	<tr>
+		<td valign="top">Notes</td>
+		<td>This field is intended as a replacement for the existing "type_abbrv_name" field.  That field does not fully discriminate a transceiver "type".  For example, a QSFP-DD transceiver may be a QSFP28-DD or QSFP56-DD.  Also, other types, such as "SFP28", are missing.<br/>Rather than introduce new values, such as "SFP28", "QSFP-DD28" and "QSFP-DD56" for the existing field, and deprecating old values, such as "QSFP-DD", a new field will be introduced, that fully discriminates a "type", and the old field shall be deprecated.</td>
```
lguohan, line 168: i do not understand rationale to introduce a new field here. You can deprecate the QSFP-DD value, and introduce new values such as QSFP-DD28 and QSFP-DD56. why do you need to deprecate the whole field?

doc/media-settings-enhancements/media-settings-plugin.md
```diff
@@ -0,0 +1,20 @@
...
+## Current implementation
+In xcvrd, the function get_media_settings_key() is used to form a list of two strings.  These strings are used as lookup keys for the dictionary created from the file media_settings.json.  
+The current implementation of get_media_settings_key() constructs two keys, as follows:
+- the first key is of the form _manufacturer_-_partnumber_, where _manufacturer_ and _partnumber_ are the name of the transceiver manufacturer and the manufacturer's part number, respectively, which are read from the transceiver EEPROM, and
+- the second key if of the form _type_-_compliancecode_-_length_, where _type_ is the transceiver type (e.g. SFP, SFP28, QSFP, etc.), _compliancecode_ is the physical layer standard (e.g. 10GBASE-SR, 40GBASE-CR4), and _length_ is the length of the attached cable, if applicable.
+# Enhancement
+Xcvrd shall be modified such that it will check for the presence of a platform-dependent plugin, which shall contain an implementation of the function get_media_settings_key().  Xcvrd shall check for the presence of this plugin, and, if found, use it in preference to the built-in function get_media_settings_key().
```
lguohan, line 11: what is the motivation to introduce the platform-dependent plugin. it is unclear to me.

doc/media-info-enhancements/media-info.md
hpersh-dell: Please refer to the section "Notes" below.
The existing "cable_length" entry is an integer, and so cannot represent lengths that are not whole numbers.

doc/media-info-enhancements/media-info.md
hpersh-dell: This list is for the first version -- it is not intended to be complete.  As new media form factors are created, or supported, new data values will be added.

doc/media-info-enhancements/media-info.md
hpersh-dell: It is purely a judgement call.  The main goal is to maintain backward compatibility.  Whether backward compatibility is done by deprecating an existing field and introducing a new field, or by deprecating values and introducing  new values in an existing field -- both accomplish the same.  It was felt that deprecating a field makes it more "obvious"...

doc/media-settings-enhancements/media-settings-plugin.md
hpersh-dell: The current code in xcvrd to form the media settings lookup keys does not work -- for example, for a 1 m SFP28 DAC cable, it returns a key of "SFP28", no indication that it is a copper cable, no information about the the cable length, and both copper/fiber information and cable length information are needed to make a preemphasis lookup.
In order to maintain backward compatibility, i.e. still be able to process media_settings.json in the current way for existing platforms, an optional plugin for xcvrd can be added, such that platforms that cannot work with the existing key-formation logic may properly look up their media settings.
The alternative is to have in-line code in xcvrd that both has the the existing "broken" logic, and new logic, to properly form lookup keys.

lguohan: this looks like a generic improvement, it should not be platform dependent. can you describe the motivation in the doc. also I suggest to move to the same folder as the previous design doc.

lguohan: prefer inline code.

hpersh-dell: The real issue backward compatibility.

While in principle, there is no reason that mainline xcvrd code should not be able to construct a lookup key that will account for all possible media types in used in all platforms, the fact of the matter is that the current implementation of get_media_settings_key() is flawed, and media_settings.json files for existing platforms rely on the current method for building the key.  Therefore, in order to maintain compatibility with existing platforms, an lookup implemented in mainline code must (1) build a key in the old style, (2) attempt a lookup, (3) if a hit is not found, build a key in a newer style, and (4) attempt a lookup.  That newer style could be something similar to the existing style, but more complete, i.e. QSFP28-100GBASE-CR-1M for a 1 m 100G DAC cable.  Also, any entry "Default" entry in the media_settings.json file would not be used unless both the lookup for an old-style and new-style key failed.
This need to keep backward compatible was making xcvrd rather convoluted -- moving to a platform-dependent plugin seemed cleaner.  Also, media_settings.json is a platform-dependent file... so the fragment of code to generate the lookup key for it being platform-dependent is not wholly unreasonable...

doc/media-info-enhancements/media-info.md
keboliu: From the specification(SFF8472, 8436,...) perspective, shall the cable length be an integer or float? If it shall be a float then this should be fixed in the parsing code?

hpersh-dell:
Closing pull request, media-settings key plug-in idea is abandoned.

PR #590, opened 2020-04-08
SONiC configuration setup enhancements to provide installer migration…

rajendra-dendukuri:
… hooks capability

The following new features are being added to the config-setup service:
1. Provide a second set of configuration pre hooks as part of the installer
image. This allows the user to fix issues caused by the migration-pre
scripts already installed on the switch.

2. Both the config-migration-pre and config-migration-post are migrated to
the newly installed image. This allows end user installed scripts to be
carried over to the newly installed image automatically to make it easier
for subsequent configuration migrations.

Signed-off-by: Rajendra Dendukuri <rajendra.dendukuri@broadcom.com>


doc/ztp/SONiC-config-setup.md
```diff
@@ -182,27 +185,29 @@ fi
...
+
+Later when the switch boots into the newly installed image, a file */tmp/pending_config_migration* is created by rc.local service and config-setup service detects its presence and starts the configuration migration action. The backed up files in /host/old_config are then restored to /etc/sonic directory and the restored startup-config file is loaded to Config DB. The *config-migration-pre-hooks* and the *config-migration-post-hooks.d* are also restored to the newly installed image from their backed up copies. If the new image contains files in the migration hooks directories with the same filename as the backed up files, they are not overwritten and instead the files in the newly installed image will be used for configuration migration. This allows the newer version images to provide updated migration logic and deprecate the older (possibly incorrect) implementations.
```
keboliu, line 196: If my understanding is correct, in current implementation the startup_config migration is performed by the '/usr/bin/db_migrator.py' script,  what's the plan for this part of function? will the db_migrator.py be kept or have something new to replace it?

rajendra-dendukuri: The /usr/bin/db_migrator.py is responsible to migrate the contents of the startup_config. The config-setup.service, which calls /usr/bin/config-setup is responsible to migrate the startup_config file. The config-setup service calls db_migrator.py. They both work together.

PR #603, opened 2020-04-16
201911 SONiC Added management vrf design document

kannankvs:


doc/mgmt/sonic_stretch_management_vrf_design.md
```diff
@@ -0,0 +1,400 @@
...
+
+##### UDP Application Daemons
+Linux 4.9 does not support `udp_l3mdev_accept`; corresponding patch is back ported to SONiC Linux 4.9. With this patch, UDP applications will be able to receive and process packets from both data VRF and management VRF. UDP being a connectionless protocol, these is no concept of accepting the connection through the same incoming port (VRF) and hence UDP applications will not be able to specify the interface through which the reply has to be sent. i.e. When UDP applications send a reply reply packet, there is no reference to the port in which the request packet had arrived. Hence the Linux stack will try to route this packet via default routing table, even if request packet had arrived via management port. Such UDP application daemon code needs to be modified to use the incoming port reference and use it for sending replies via the same VRF through which the request had arrived.
+
+##### IP Application Daemons
+Most IP applications (like PING) always have the incoming port/VRF association with the reply packet and hence they will continue to work without any modification. 
+
+#### Applications Originating From the Device
```
a-barboza, line 254: Will there be any description regarding "docker pull" when Management VRF is enabled? Currently the cgexec does not seem to work for it, as it is doing all its lookups, and fetches (DNS, HTTPS) using the default VRF. The current workaround seems to be by  adding a rule to the routing policy database to lookup the Management VRF, if route cannot be found. Thinking out loud here, perhaps an option inside the MGMT_VRF_CONFIG table to allow for this?

kannankvs: No. There wont be any description for "docker pull" since the original scope during development was for "snmp" and "ntp" only. 
Once if this is merged, any contributor is welcome to enhance the design and the code.

PR #605, opened 2020-04-21
[QoS] Dynamic buffer calculation

stephenxs:
We're going to improve the way we calculating headroom size. Originally we generate headroom size by looking up the predefined lookup table.
We will have the following groups of parameters
- List of SONiC configuration (such as speed and cable length)
- List of pg configuration (such as cell size, IPG, MTU, Peer response time, small packet size percentage)

Based on the parameters and a well-known formula the code in buffer manager will do the calculation and not take it from pre-defined values as we have today.
On top of that, need an ability to override headroom and do not calculate it in the code.


liat-grozovik:
@neethajohn 
As review was done long time ago, feedback provided and updated as well, following CLI improvements during implementation and no additional comments i suggest to merge this one.
Any concerns?

liat-grozovik (approved):

neethajohn (approved):

PR #606, opened 2020-04-22
Sonic CoPP config and management

prsunny:
- Introduce CoPP Manager
- Refer Feature Table before installing Traps

### DIFF
--- a/doc/copp/CoPP Config and Management.md
+++ b/doc/copp/CoPP Config and Management.md
@@ -29,6 +29,8 @@ A new schema is defined for COPP tables that seperates Queue/Policer groups and 
 ```
 key = "COPP_GROUP|name"
 queue         = number; strict queue priority. Higher number means higher priority.
+action        = packet_action; trap action which will be applied to all trap_ids for this group.
+priority      = trap_priority
 
 ;Settings for embedded policer. 
 meter_type  = "packets" | "bytes"
@@ -44,8 +46,8 @@ red_action           = packet_action
 ```
 ```
 key = "COPP_TRAP|name"
+trap_ids      = name ; list of trap ids
 trap_group    = name ; copp group name
-trap_action   = packet_action; trap action which will be applied to all trap_ids.
 genetlink_name       = genetlink_name ;[Optional] "psample" for sFlow 
 genetlink_mcgrp_name = multicast group name; ;[Optional] "packets" for sFlow 
 ```
@@ -64,7 +66,7 @@ state        = "ok"
 Introduce a *new* CoPP manager, that subscribes for the Config DB CoPP Tables and Feature Tables. Based on the feature enablement, ```coppmgr``` handles the logic to resolve whether a CoPP table shall be written to APP DB for orchagent consumption. Inorder to reduce changes to copporch and for backward compatibility during warmboot, ```coppmgr``` shall use the existing APP_DB schema and implement internal logic to convert the proposed ConfigDB entries to APP DB entries. Similar to existing swss managers, an entry with state "ok" shall be added to STATE_DB.
 
 ### copporch
-```copporch``` shall only be a consumer of APP DB CoPP Table. It is not expected to handle feature logic and the current handling of features like sFlow, NAT shall be revisited and removed to be added as part of ```coppmgr```. However `copporch` must be able to handle any new trap_id getting added or removed from an existing CoPP table. 
+```copporch``` shall only be a consumer of APP DB CoPP Table. It is not expected to handle feature logic and the current handling of features like sFlow, NAT shall be revisited and removed to be added as part of ```coppmgr```. However `copporch` must be able to handle any new trap_id getting added or removed from an existing CoPP table, and handle attribute value set for a trap group
 
 ### swssconfig
 Handling of CoPP config json file shall be removed from ```dockers/docker-orchagent/swssconfig.sh```
@@ -95,6 +97,11 @@ As part of the post start action, the Config DB shall be loaded with default CoP
 It is desirable to have warmboot functionality from previous release versions of Sonic. Since the existing schema has COPP Group name/key with protocol names (e.g `"COPP_TABLE:trap.group.bgp.lacp"`, there is a limitation in adding any new protocol or trap to an existing CoPP group and seamlessly migrate. However, with this proposal, the implementation is to do a migration of APP DB entries to new schema.
 
 In addition, the implementation must ensure that backward compatibility is maintained. If the system is boot-up with an *old* config file, the default CoPP tables are to be loaded from ```init_cfg.json``` and expected to work seamlessly.
+
+## Limitations
+1. In case of downgrade, the config_db entries shall be present as stale entries as there is no subscribers for the table. Functionality would be same as supported by the downgraded version
+2. This proposal expects the table names to be consistent across multiple releases. 
+3. User is expected to resolve any conflicts, say for a trap id or group, that arises due to default values from the Sonic binary vs same trap or group currently or previously configured by the user.
 
 ## CLI
 `show` commands to display CoPP group and CoPP entries shall be provided as part of this feature implementation.
@@ -136,10 +143,14 @@ The following flow captures CoPP manager functionality.
 
         "COPP_GROUP|queue4_group1": {
             "queue": "4",
+            "action":"trap",
+            "priority":"4",
         },
         
         "COPP_GROUP|queue4_group2": {
             "queue": "4",
+            "action":"copy",
+            "priority":"4",
             "meter_type":"packets",
             "mode":"sr_tcm",
             "cir":"600",
@@ -149,27 +160,23 @@ The following flow captures CoPP manager functionality.
 
         "COPP_TRAP|bgp": {
             "trap_ids": "bgp,bgpv6",
-            "trap_action":"trap",
-            "trap_priority":"4",
             "trap_group": "queue4_group1"
         },
 
         "COPP_TRAP|lldp": {
             "trap_ids": "lldp",
-            "trap_action":"trap",
-            "trap_priority":"4",
             "trap_group": "queue4_group1"
         },
        
         "COPP_TRAP|arp": {
             "trap_ids": "arp_req,arp_resp,neigh_discovery",
-            "trap_action":"copy",
-            "trap_priority":"4",
             "trap_group": "queue4_group2"
         },
 
         "COPP_GROUP|queue1_group1": {
             "queue": "1",
+            "action":"trap",
+            "priority":"1",
             "meter_type":"packets",
             "mode":"sr_tcm",
             "cir":"6000",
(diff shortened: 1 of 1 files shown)


prsunny:
@ritalhui, @dgsudharsan, @padmanarayana, please help review

doc/copp/CoPP Config and Management.md
```diff
@@ -0,0 +1,111 @@
...
+
+With the new proposal, the CoPP tables shall be loaded to Config DB instead of APP DB. The default CoPP json file shall be prebuilt to the image and loaded during initialization. Any Config DB entries present shall be configured overwriting the default CoPP tables. This also ensures backward compatibility. At a high-level, the following are the expected changes:
+
+### Schema Changes
+
+### Config DB
+
+```
+key = "COPP|name"
+name_list     = name | name,name_list
```
dgsudharsan, line 30: Can we have a name generically defined instead of being a list of trap ids like in copp config file? For example like COPP|trap_group1. This would be helpful in maintaining the name consistent across upgrades and in scenarios where we add new trap ids to the group.

doc/copp/CoPP Config and Management.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Design Proposal
+
+## Current behaviour
+
+During SWSS start, the prebuilt [copp.json](https://github.com/Azure/sonic-swss/blob/201911/swssconfig/sample/00-copp.config.json) file is loaded as part of start script [swssconfig.sh](https://github.com/Azure/sonic-buildimage/blob/201911/dockers/docker-orchagent/swssconfig.sh) and written to APP DB. [CoppOrch](https://github.com/Azure/sonic-swss/blob/201911/orchagent/copporch.cpp) then translates it to Host trap/policer tables and programmed to SAI.
+
+## Proposed behaviour
+
+With the new proposal, the CoPP tables shall be loaded to Config DB instead of APP DB. The default CoPP json file shall be prebuilt to the image and loaded during initialization. Any Config DB entries present shall be configured overwriting the default CoPP tables. This also ensures backward compatibility. At a high-level, the following are the expected changes:
```
dgsudharsan, line 22: Can you please elaborate on the case where config has been removed from config_db.json but still exists in init_cfg.json? What is the suggested mechanism to remove a trap group as a whole

doc/copp/CoPP Config and Management.md
```diff
@@ -0,0 +1,111 @@
...
+
+### StateDB
+
+```
+key = "COPP_TABLE|name"
+state        = "ok"
+```
+
+### coppmgr
+Introduce a *new* CoPP manager, that subscribes for the Config DB CoPP Tables and Feature Tables. Based on the feature enablement, ```coppmgr``` handles the logic to resolve whether a CoPP table shall be written to APP DB for orchagent consumption. In case if the feature requires adding only few attributes to an existing CoPP Table, ```coppmgr``` shall add the respective attributes to APP DB entry when the feature is enabled. Similar to existing swss managers, an entry with state "ok" shall be added to STATE_DB.
```
wendani, line 59: Any application in particular that is interested in CoPP STATE_DB state?

doc/copp/CoPP Config and Management.md
```diff
@@ -0,0 +1,211 @@
...
+pir         = number ;packets or bytes depending on the meter_type value
+green_action         = packet_action
+yellow_action        = packet_action
+red_action           = packet_action
+```
+```
+key = "COPP_TRAP|name"
+trap_group    = name ; copp group name
+trap_action   = packet_action; trap action which will be applied to all trap_ids.
+genetlink_name       = genetlink_name ;[Optional] "psample" for sFlow
```
michaelli10, line 49: Add trap_priority field.

doc/copp/CoPP Config and Management.md
```diff
@@ -0,0 +1,211 @@
...
+With the new proposal, the CoPP tables shall be loaded to Config DB instead of APP DB. The default CoPP json file shall be prebuilt to the image and loaded during initialization. Any Config DB entries present shall be configured overwriting the default CoPP tables. This also ensures backward compatibility. At a high-level, the following are the expected changes:
+
+### Schema Changes
+
+A new schema is defined for COPP tables that seperates Queue/Policer groups and Traps. More details are in the "Examples" section
+
+### Config DB
+```
+key = "COPP_GROUP|name"
+queue         = number; strict queue priority. Higher number means higher priority.
```
michaelli10, line 31: Does it make sense to move the queue to COPP_TRAP table?  Then COPP_GROUP reference could become a POLICER_TABLE reference in COPP_TRAP.

Decoupling queue from policer helps avoid extra COPP_GROUPs from being defined when different policer rates are required but same queue is needed or visa versa.

doc/copp/CoPP Config and Management.md
prsunny: Covered in the updated version

doc/copp/CoPP Config and Management.md
prsunny: This could require further changes and also may affect warmboot. If not with major advantage, prefer to keep it this way

doc/copp/CoPP Config and Management.md
prsunny: Added and moved to copp_group

doc/copp/CoPP Config and Management.md
```diff
@@ -0,0 +1,211 @@
...
+
+The following flow captures scenarios for ```boot-up``` sequence and ```config reload```. Default CoPP Tables shall be present in ```init_cfg.json``` and if the same entry is present in ```config_db.json```, it is expected to be overwritten by ```config_db``` entry. This model ensures user-configuration gets priority over default configuration.
+
+![](https://github.com/Azure/SONiC/blob/master/images/copp/copp_init.png)
+
+## Copp Manager flow
+
+The following flow captures CoPP manager functionality. 
+
+![](https://github.com/Azure/SONiC/blob/master/images/copp/CoppManager_1.png)
```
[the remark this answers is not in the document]
prsunny: Based on our review, moved action and priority to copp_group

doc/copp/CoPP Config and Management.md
prsunny: Not really, but just following the existing convention

doc/copp/CoPP Config and Management.md
```diff
@@ -0,0 +1,214 @@
...
+
+# Examples
```
michaelli10, line 132: It would help clarify the CoPPMgr behavior if you had an equivalent AppDB configuration next to this new ConfigDB version.

doc/copp/CoPP Config and Management.md
```diff
@@ -0,0 +1,214 @@
...
+### Schema Changes
+
+A new schema is defined for COPP tables that seperates Queue/Policer groups and Traps. More details are in the "Examples" section
+
+### Config DB
+```
+key = "COPP_GROUP|name"
+queue         = number; strict queue priority. Higher number means higher priority.
+action        = packet_action; trap action which will be applied to all trap_ids for this group.
+priority      = trap_priority
```
michaelli10, line 33: For consistency with AppDB, can we change "action" to "trap_action" and "priority" to "trap_priority"?

doc/copp/CoPP Config and Management.md
prsunny: Sure, added to examples section

dgsudharsan:
The Recent changes look good to me

doc/copp/CoPP Config and Management.md
```diff
@@ -0,0 +1,336 @@
...
+
+For this design, considering the limitations, proposal is to go ahead with second option. 
+
+## Warmboot and Backward Compatibility
+It is desirable to have warmboot functionality from previous release versions of Sonic. Since the existing schema has COPP Group name/key with protocol names (e.g `"COPP_TABLE:trap.group.bgp.lacp"`, there is a limitation in adding any new protocol or trap to an existing CoPP group and seamlessly migrate. However, with this proposal, the following options are considered:
+
+1. The implementation is to do a migration of APP DB entries to new schema.
+2. Let ```db_migrator``` remove all previously saved APP_DB entries and let syncd reconcile and remove the TRAP entries during warmboot. Later, ```coppmgr``` can reapply the traps afresh. Also remove COPP_TABLES from being saved in [backup_database](https://github.com/Azure/sonic-utilities/blob/master/scripts/fast-reboot#L234). This may require additional tests to confirm system behaviour during warmboot. 
+3. Remove '00-copp.config.json' from being included for checksum calculation in ```files/build_scripts/generate_asic_config_checksum.py```
```
abdosi, line 111: @prsunny even though we remove copp.json from checksum calculation it will still throw error on warm-reboot from old image.

Eg: current_image let's say 201911 next_image 202011 (that have new copp design)

when we issue warm-boot on 201911 image to go to 202011 asic checksum validation done by 201911 will fail as it will try to get checksum of 2020 image but it will not be same as 201911.

prsunny: Agree, but I think, would recommend to use "warm-reboot-f".

prsunny:
@dgsudharsan , @michaelli10 , @wendani , can you please sign off?

doc/copp/CoPP Config and Management.md
```diff
@@ -0,0 +1,336 @@
...
+            fi
+        fi
+```
+However, this approach has the following limitations:
+
+In warmboot scenarios, as init_cfg.json is not read during warmboot and may result in ***new*** CoPP Tables in the "to" image not getting applied.
+In coldboot, if user saves the config and the new release happened to have new CoPP values, it will not be applied since previous default values are present in config_db
+
+2. ```coppmgr``` reading directly from ```copp_cfg.json``` and apply the default CoPP Tables. 
+
```
michaelli10, line 103: I have implemented a 3rd option that addresses a limitation of the first option so that we can load the default COPP configs in the ConfigDB to keep inline with SONiC architecture and some requirements for config CLIs.  

To address the issue of an old COPP default config stored in config_db.json file (on config save) overwriting the new image COPP default config with new values, we are comparing the default COPP config file with the ConfigDB entries in sonic-cfggen and only writing back config modifications to the config_db.json file.   No COPP config will be stored in config_db.json if no modifications have been made.   Only "user modifications" will be stored back in the config_db.json.   This change in sonic-cfggen is generic enough for other use cases as well.  Let me know if this 3rd option is a possibility for this feature.

prsunny: The thought is, lets not have default in config_db at all. In case of warmboot, we will again have the issue if entries are in config_db and new image has a different default value. In which case you've to add the similar logic in the warmboot script before saving the redis_db to skip saving default values. Also I think, sonic-cfggen should not be used to decide which TABLES must be written to config_db.json file vs not written to file.

michaelli10: Ok, we can have a separate discussion on this specific issue.  The rest of the doc looks good.

PR #610, opened 2020-04-27
Fine Grained ECMP HLD

anish-n:
Fine Grained ECMP HLD which introduces a means of fine grained control over a set of next-hops


doc/ecmp/fine_grained_next_hop_hld.md
```diff
@@ -0,0 +1,306 @@
...
+
+# 2 Modules Design
+
+## 2.1 Config DB
+Following new tables will be added to Config DB. Unless otherwise stated, the attributes are mandatory.
+```
+FG_NHG|{{fg-nhg-group-name}}:
+    "bucket_size": {{hash_bucket_size}}
+
+FG_NHG_PREFIX|{{IPv4 OR IPv6 address}}:
```
rck-innovium, line 116: FG_NHG_PREFIX|{{IPv4 OR IPv6 address prefix}}:

doc/ecmp/fine_grained_next_hop_hld.md
```diff
@@ -0,0 +1,306 @@
...
+```
+
+Please refer to the [schema](https://github.com/Azure/sonic-swss/blob/master/doc/swss-schema.md) document for details on value annotations. 
+
+
+## 2.2 State DB
+Following new table will be added to State DB. Unless otherwise stated, the attributes are mandatory.
+FG_ROUTE_TABLE is used for some of the show commands associated with this feature as well as for warm boot support.
+```
+FG_ROUTE_TABLE|{{IPv4 OR IPv6 address}}:
```
rck-innovium, line 156: FG_ROUTE_TABLE|{{IPv4 OR IPv6 address prefix}}:

kannankvs:
Just curious whether there is an IETF RFC/draft for handling this requirement.

prsunny:
@anish-n , please add some notes on kernel routes vs hardware route.  AFAIK, FG ECMP is a hardware capability and in this scenario, kernel routes shall have the regular ecmp nexthop group whereas hardware will be programmed with the FG ecmp nexthops. On another note, this is not related to regular static route feature.

doc/ecmp/fine_grained_next_hop_hld.md
```diff
@@ -0,0 +1,306 @@
...
+## 1.1 Functional requirements
+This section describes the SONiC requirements for Fine Grained ECMP next-hop groups 
+
+At a high level the following should be supported:
+
+Phase #1
+- Should be able to configure a Fine Grained ECMP group which defines a static means of redistribution of ECMP upon next-hop modifications based upon the fine grained ECMP SAI proposal
+- Standard route modifications should enable special ECMP behavior for those prefixes which desire Fine Grained ECMP. For all other prefixes the standard ECMP behavior should apply. 
+- Ability to enable consistent hashing via Fine grained ECMP for a statically defined ECMP group
+- Ability to specify a group(bank) in which ECMP redistribution should be performed out of a set of available next-hops
```
rck-innovium, line 75: Can you explain the notion of 'bank'. It is not explained in the SAI spec.

doc/ecmp/fine_grained_next_hop_hld.md
```diff
@@ -0,0 +1,306 @@
...
+
+# 1 Requirements Overview
+## 1.1 Functional requirements
+This section describes the SONiC requirements for Fine Grained ECMP next-hop groups 
+
+At a high level the following should be supported:
+
+Phase #1
+- Should be able to configure a Fine Grained ECMP group which defines a static means of redistribution of ECMP upon next-hop modifications based upon the fine grained ECMP SAI proposal
+- Standard route modifications should enable special ECMP behavior for those prefixes which desire Fine Grained ECMP. For all other prefixes the standard ECMP behavior should apply.
```
zhenggen-xu, line 73: Not sure if specifying the prefixes to use this feature is a requirement or just the ways to simply the design.  This feature seems to be more tied to the nexthops rather than routes. To manage the static prefixes in configDB is challenging in most networks as routes are normally dynamic learnt, not to mention the potential inconsistency of  routing stack and SONiC config. However, the nexthops are mostly statically configured.  I would think we can make this feature to follow that nature.  I,E, configuration on SONiC are most to enable FG-ECMP for nexthops,  and the FG-ECMP-Group is automatically created if all members are nexthops with FG-ECMP enabled (otherwise, standard ECMP group).  During the route changes, it just uses or creates the ECMP group based on the nexthops the route uses as above.    Does this make any sense?

doc/ecmp/fine_grained_next_hop_hld.md
anish-n: Updated, thanks!

doc/ecmp/fine_grained_next_hop_hld.md
anish-n: Updated, thanks!

anish-n:
> Just curious whether there is an IETF RFC/draft for handling this requirement.

I hadn't referred to any particular IETF RFC, but did find the following upon lookup: https://tools.ietf.org/html/rfc2992 
At a high level it appears to be kind of similar to the approach here, but I haven't gone over it in a very detailed manner.

anish-n:
> fgnhgorch talks about handling "BGP route modifications". Can we assume that the same design can be used for static routes also? i.e. Both the route is static and the FgEcmp is static.

This feature comes into play at the final layer of programming routes to the ASIC_DB and as such all routes(BGP, static etc) would work alongside this feature.

doc/ecmp/fine_grained_next_hop_hld.md
anish-n: Added some more details to the use case and details section. The example section covers the bank in terms of redistribution too. Let me know if you need further information on it.

anish-n:
> @anish-n , please add some notes on kernel routes vs hardware route.  AFAIK, FG ECMP is a hardware capability and in this scenario, kernel routes shall have the regular ecmp nexthop group whereas hardware will be programmed with the FG ecmp nexthops. On another note, this is not related to regular static route feature.

Thanks, added this information into the "details" section.

doc/ecmp/fine_grained_next_hop_hld.md
```diff
@@ -0,0 +1,306 @@
...
+| FG_NHG component         | SAI attribute                                         |
+|--------------------------|-------------------------------------------------------|
+| Fine Grained ECMP type   | SAI_NEXT_HOP_GROUP_TYPE_FINE_GRAIN_ECMP               |
+| Length of ECMP group     | SAI_NEXT_HOP_GROUP_ATTR_CONFIGURED_SIZE               |
+| Next-hop ids             | SAI_NEXT_HOP_GROUP_MEMBER_ATTR_NEXT_HOP_ID            |
+| Specify nh index         | SAI_NEXT_HOP_GROUP_MEMBER_ATTR_INDEX                  |
+
+
+
+# 3 Flows
```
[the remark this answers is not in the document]
anish-n: Added, thanks

doc/ecmp/fine_grained_next_hop_hld.md
anish-n: This makes sense and we had considered both options(with and without prefix), from an implementation stand point both would work. We thought that adding the prefix gives more granularity to the configuration: A user has special ecmp behavior for a prefix. At a high level I think that ecmp defined for a prefix so I thought it may make sense to make this config prefix specific. Further it also makes things clearer in terms of expected behavior if we get prefixes with next-hops which are absent from FG_NHG member definitions: ignore those since FG_NHG is a hw feature which makes the final determination of ASIC next-hops based on config pushed by user. Does that make sense? I will also spend some time and rethink if we should remove the prefix.

anish-n: @zhenggen-xu, let me know what you think. We discussed with @lguohan and others as well, along with the reasons highlighted above, the prefix approach may have an advantage for future oriented use cases, for example, special ecmp behavior is set for all prefix announcements bearing a special community tag, given that the community tag is made a part of the db entry. The next-hop approach would then not be compatible with this use case and the prefix based approach would be. Does that make sense?

anish-n: @zhenggen-xu discussed some more on this. We propose to introduce a FG_NHG 'mode' as a future schema change to support this behavior based on any future usecase requirements. The mode specifies whether fgnhg will use a prefix based or next-hop based filtering to determine fine grained behavior:
FG_NHG|{{fg-nhg-group-name}}:
    "bucket_size": {{hash_bucket_size}}
    **"mode":{{prefix/next-hop}}**
If/when the next-hop based implementation is desired, the above schema change may be made and configurable next-hop based filtering enabled. The default mode could remain prefix based in order to allow for backwards compatibility.

abdosi: @anish-n mode will be in TO-do list for future ? Are you planning to update document/schema in the PR ?

doc/ecmp/fine_grained_next_hop_hld.md
```diff
@@ -0,0 +1,316 @@
...
+## 2.5 SAI
+The below table represents main SAI attributes which shall be used for Fine Grained ECMP
+
+
+###### Table 3: Fine Grained ECMP SAI attributes
+| FG_NHG component         | SAI attribute                                         |
+|--------------------------|-------------------------------------------------------|
+| Fine Grained ECMP type   | SAI_NEXT_HOP_GROUP_TYPE_FINE_GRAIN_ECMP               |
+| Length of ECMP group     | SAI_NEXT_HOP_GROUP_ATTR_CONFIGURED_SIZE               |
+| Next-hop ids             | SAI_NEXT_HOP_GROUP_MEMBER_ATTR_NEXT_HOP_ID            |
```
abdosi, line 217: @anish-n Please update with SAI_NEXT_HOP_GROUP_ATTR_REAL_SIZE

doc/ecmp/fine_grained_next_hop_hld.md
anish-n: Added, thanks

doc/ecmp/fine_grained_next_hop_hld.md
anish-n: Updated, thanks

PR #617, opened 2020-05-21
Revisited Vnet default handling for proxy_arp

prsunny:
Revisited Vnet default handling for proxy_arp


prsunny:
@volodymyrsamotiy - please review

PR #618, opened 2020-05-21
[MultiDB] update warmboot design

dzhangalibaba:
* add warmboot design for database backup

Signed-off-by: Dong Zhang d.zhang@alibaba-inc.com


doc/database/multi_database_instances.md
```diff
@@ -820,36 +820,159 @@ EXEC_WITH_ERROR_THROW(cmd, res);
...
+  CPU min MHz:         1200.0000
+  BogoMIPS:            4787.94
+  Virtualization:      VT-x
+  L1d cache:           24K
+  L1i cache:           32K
+  L2 cache:            1024K
+  NUMA node0 CPU(s):   0-3
+  ```
+  
+  - [x] Third part script like redis-dump/load is very slow, usually takes **~23s** when data size is **~40K**
```
qiluo-msft, line 890: Could you make sure you also benchmark the unix socket for each method, and list the best between TCP and unix socket.

doc/database/multi_database_instances.md
```diff
@@ -820,36 +820,159 @@ EXEC_WITH_ERROR_THROW(cmd, res);
...
+  import redis
+  
+  dblists = swsssdk.SonicDBConfig.get_dblist()
+  for dbname in dblists:
+      dbsocket = swsssdk.SonicDBConfig.get_socket(dbname)
+      #dbport = swsssdk.SonicDBConfig.get_port(dbname)
+      dbid = swsssdk.SonicDBConfig.get_dbid(dbname)
+      dbhost = swsssdk.SonicDBConfig.get_hostname(dbname)
+
+      r = redis.Redis(host=dbhost, unix_socket_path=dbsocket, db=dbid)
```
qiluo-msft, line 913: Why mix TCP socket and unix socket?

doc/database/multi_database_instances.md
```diff
@@ -820,36 +820,159 @@ EXEC_WITH_ERROR_THROW(cmd, res);
...
+
+  ```python
+  #!/usr/bin/python
+  from __future__ import print_function
+  import sys
+  import swsssdk
+  import redis
+  
+  dblists = swsssdk.SonicDBConfig.get_dblist()
+  for dbname in dblists:
```
qiluo-msft, line 907: Why not iterate all DB instances, and move all code for one instance into LUA script.
Even more, pre-load the script into redis to minimize the critical path.

doc/database/multi_database_instances.md
dzhangalibaba: tcp and unix socket almost have the same performance for this case

doc/database/multi_database_instances.md
dzhangalibaba: it is not mix, if unix_socket_path is there, it always use unix socket, the host and port parameters are ignored. I leave host there since sometimes I changed  unix_socket_path to port for switching mode.

doc/database/multi_database_instances.md
dzhangalibaba: iterating  all DB instances is also fine. But the performance for this case is almost the same as   iterating all database. Also, the performance for pre-load is the same as experiment shown, since the lua script is very short and simple and only used once if we  iterate  all DB instances.

doc/database/multi_database_instances.md
```diff
@@ -820,36 +820,193 @@ EXEC_WITH_ERROR_THROW(cmd, res);
...
 
-Discussed with Guohan's team offline, we won't support all the warmreboot cases when the database\_config.json file changed. We only want to accept the database instances spliting situation. For example , before warmrebbot, there is two instances and after there are four isntances, we can use rdb file to restoring all data in four instances and then write a logic to flush unnecessary database on each instance. This logic will be done in a new script which will be executed after warmreboot database restoration.  
+Today we already assigned each database name with a unique number (APPL_DB 0, ASIC_DB 1, ...) and assign them into different redis instances in design. This makes it possible to migrate all data in all redis instances into one redis instance without any conflicts. Then we can handle this single redis instance the same as what we did today, since today we are only use single redis instance. So the poposed new idea is as below steps:
```
yxieca, line 839: I think this grant scheme is ok when multiple DB instances are used for load balancing, meaning that they have different DB IDs.

Not sure if this assumption still true for multi-ASCI scenario. Potentially, there we could have multiple DB instances contains same DB IDs/Table names.

judyjoseph: Adding a bit more thought to Ying's comment above - with the multi-AISC architecture now we would have to run "this warmboot logic" per namespace .. where in each namespace database docker could have one ore more redis instances.

doc/database/multi_database_instances.md
```diff
@@ -820,36 +820,193 @@ EXEC_WITH_ERROR_THROW(cmd, res);
...
+  admin@ASW-7005:~$ redis-cli -p 6381 info | grep Keyspace -A 10
+  # Keyspace
+  db0:keys=40012,expires=0,avg_ttl=0
+  db1:keys=42834,expires=0,avg_ttl=0
+  db2:keys=7191,expires=0,avg_ttl=0
+  db3:keys=103,expires=0,avg_ttl=0
+  db4:keys=99,expires=0,avg_ttl=0
+  db5:keys=3145,expires=0,avg_ttl=0
+  db6:keys=365,expires=0,avg_ttl=0
+  ```
```
judyjoseph, line 1001: In this output, find that the "DB names" are used to identify the databases after migration into a single instance, eg: db0, db1 etc ? 
If that is the case , even with current single database docker scheme -- we need to ensure that the database names are unique across different redis instances. eg: COUNTERS_DB cannot be there in redis instance0 and redis instance1.

dzhangalibaba: in single DB case, today  database names are unique across different redis instance, this is why this approach is proposed.

doc/database/multi_database_instances.md
dzhangalibaba: As Judy said, multi ASIC has multi namespaces/dockers, each namespace should have the same logic.

dzhangalibaba:
Is it OK to go ahead with this proposal ? Let me know if I can change the codes and make progress @qiluo-msft @lguohan @yxieca @judyjoseph

qiluo-msft:
@judyjoseph Could you help review again?

doc/database/multi_database_instances.md
judyjoseph: Sure, thanks - it should be ok then ! I am wondering if we need to add some validation to enforce this in SonicDBConfig or so when we parse the database_config file ?

judyjoseph:
LGTM as well, the only feedback I have relating to multi-ASIC scenario is that we would need more CPU cycles for doing this DB save and restore/flushdb activity in different namespaces .. Need to do this migration in parallel ( like spawning multiple threads etc ) across host + namespace redis servers.

doc/database/multi_database_instances.md
```diff
@@ -820,36 +820,193 @@ EXEC_WITH_ERROR_THROW(cmd, res);
...
+Today we already assigned each database name with a unique number (APPL_DB 0, ASIC_DB 1, ...) and assign them into different redis instances in design. This makes it possible to migrate all data in all redis instances into one redis instance without any conflicts. Then we can handle this single redis instance the same as what we did today, since today we are only use single redis instance. So the poposed new idea is as below steps:
 
-![center](./img/db_restore_new.png)
+1. When we normally start multiple redis instances, besides all the necessary and used redis isntances, we add one more unused/spare redis instance configration in database_config.json. This unused/spare redis instance is not accessed by any application and it is empty.
+2. When we try to do warmboot, we migrate the data on all used redis instances into the unused/spare/empty redis instance. In this way all the data are in one redis instance and we can issue "redis-cli save" now to generate a single rdb file containing all the data. After this point, everything is the same as what we did today for single redis instance,  copying rdb file to WARM_DIR and so on.
```
lguohan, line 842: wording: In the shutdown stage of warmboot, we migrate ...

dzhangalibaba: > wording: In the shutdown stage of warmboot, we migrate ...

updated, let me know if other concern

doc/database/multi_database_instances.md
qiluo-msft, line 839: >Today we already assigned each database name with a unique number (APPL_DB 0, ASIC_DB 1, ...) and assign them into different redis instances in design [](start = 0, length = 150)

@dzhangalibaba This is tricky/hidden information for a long run. Please help add a unit test or vs test or PR checker to gatekeeper every version handle this assumption.

xjasonlyu:
Hi @dzhangalibaba 

I noticed that you said “We tried to create two database instances and separate the huge write into two database instances. The test result shows the performance (time) improved 20-30%” in the motivation section, and I wonder if you can share more specific test cases or any other ways to simulate the tests would be very helpful!
 
Also, the “Third part script like redis-dump/load is very slow, usually takes ~23s when data size is ~40K, TCP and UNIX SOCKET connections almost take the same time.” you mentioned in the wiki, what does it mean that the data size is 40K (40KB in json file size or 40,000 keys)? I tried to reproduce the test but got far different results, so I guess it might be a problem with the test data I used.

dzhangalibaba:
Hi Jason,

I noticed that you said “We tried to create two database instances and
separate the huge write into two database instances. The test result shows
the performance (time) improved 20-30%” in the motivation section, and I
wonder if you can share more specific test cases or any other ways to
simulate the tests would be very helpful!

- We focused on the route download performance. For testing, what we did
before is to generate a large json file containing 40K route entries. Then
we use swssconfig tool to load this json file into APP_DB, then orchagent
will do the job to program ASIC_DB / ASIC, after that, we measured how long
it takes to make all those 40K entries programmed in ASIC. We found the
performance improved ~30% using multiDB. I made some demos earlier to Qi
Luo/Shi (MSFT team), Zhenggen Xu(Linkined team), you can check with them
for what I did before as well if that's convenient for you.


Also, the “Third part script like redis-dump/load is very slow, usually
takes ~23s when data size is ~40K, TCP and UNIX SOCKET connections almost
take the same time.” you mentioned in the wiki, what does it mean that the
data size is 40K (40KB in json file size or 40,000 keys)? I tried to
reproduce the test but got far different results, so I guess it might be a
problem with the test data I used.

- it means the entries/keys in json file.



Best,

DONG

PR #623, opened 2020-05-26
Fine grained ECMP test plan

anish-n:


abdosi (approved):
I would suggest to add test case that verify Minigraph changes also as part of pytest.

PR #624, opened 2020-05-27
SONiC System Health Monitor HLD

keboliu:
Add an HLD doc for SONiC System Health Monitor Service.


liat-grozovik:
@lguohan as the feature is ready and PRs are also available we should be able to merge this one. the implementation is based on the recent changes following the community feedback.

PR #639, opened 2020-06-30
Distributed VOQ HLD

skeesara-nokia:
This is a design document proposal for Distributed VOQ from the Nokia-SONiC team


doc/voq/voq_hld.md
```diff
@@ -0,0 +1,879 @@
...
+
+The figures below show the components of this architecture on a modular chassis with separate line cards, each with multiple Forwarding asics, control card and multiple Switch Fabric asics. The terms Line Card and Control Card are architectural and not necessarily separate phyiscal elements of a Distributed VOQ system (though a modular chassis typically implements them as separate physical elements). For example it is possible that the entire VOQ system consists of a single card. Where applicable it is assumed that Control Card and Line Card components can exist on the same card (or fixed configuration device).
+
+ ![](../../images/voq_hld/voq-bigpicture.png)
```
kartik-arista, line 93: This is a much more comprehensive diagram than I have in the Arch document pull/622 and likely belongs there. Since I don't have access to the source, I would welcome it if the authors could amend the Arch document with this diagram as a further revision.

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,879 @@
...
+
+Primary motivation for this work is to support SONiC on modular chassis systems using VOQ-SAI capable hardware. But this spec does not preclude support for other form factors - for example a fixed configuration system with two VOQ-SAI asics connected back-to-back using their fabric ports. This document also does not specifically cover how shared peripheral components (Fans, Power Supply, System LED etc ..) are managed on a modular chassis supporting SONiC. It is expected that the topic will be covered in a separate design document.
+
+# Definitions/Abbreviation
+###### Table 1: Abbreviations
+| Term                     | Description                    |
+|--------------------------|--------------------------------|
+| VOQ                      | Virtual Output Queue           |
+
+# Reference Architecture
```
kartik-arista, line 88: This would be a good place to point to the Architecture doc

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,879 @@
...
+| System Ports             | 4K                          |
+| VOQ Neighbors            | 16k                         |
+
+The initial scaling should at least support the requirements of a "large" modular chassis deploying ports based routing interfaces. Virtual interfaces (VLAN based routing interfaces) and directly attached hosts are still supported, but achieving higher scaling for those deployments is not a goal of the initial implementation.
+
+## 1.8 Warm Restart requirements
+At this time warm restart capability is not factored into the design. This shall be revisited as a Phase #2 enhancement
+
+# 2 Modules Design
+## 2.1 Configuration Flow Options for System Ports
```
kartik-arista: The approach of using FSI CONFIG_DB to capture the global configuration including system ports (and LAGs in the future) will be documented in the Arch doc pull/622. The details including the schema can be presented here.

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,879 @@
...
+
+ ![](../../images/voq_hld/sonic-instance-system-port-config-flow.png)
+
+## 2.2 Config DB
+
+### 2.2.1 DEVICE_METADATA
+ The **existing** DEVICE_METADATA table is enhanced to add new entry to have VOQ related parameters
+ 
+```
+DEVICE_METADATA|{"voq_db"}
```
kartik-arista, line 208: This will be captured in the Arch doc instead.

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,879 @@
...
+         "vrf": "1"
+      },
+      "Ethernet129:20.1.0.2": {
+         "neigh": "0b2:16:0b:00:00:01",
+         "vrf": "1"
+      }
+    }
+}
+```
+
```
kartik-arista, line 873: The testing section from the Arch document can be migrated here because it is more relevant. We can also include other tests that we have specified for route/neighbor creation here.

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,879 @@
...
+# 5 References
```
kartik-arista, line 874: The Future work section in the Arch doc can also be migrated here because it links with the SYSTEM_PORT schema etc.

kartik-arista: Here is a blurb that can be incorporated:
# 4 Future Work

## 4.1 Dynamic System Ports

Dynamic system port support is required to support the following forwarding scenarios

* Addition of a new forwarding device into an existing running system
* Replacing a forwarding device with another device of a different hardware SKU (such as replacing a linecard with a new linecard of a different SKU in a chassis slot).

Both these scenarios can be supported smoothly as long as the global system port numbering scheme is maintained and the modifications to system ports can be performed without impacting the System Port IDs of the running sys\
tem.

Support for dynamic system ports requires SAI support for the `create_port` and `remove_port` calls.

### Overview of PortsOrch changes

 * Subscribe to the SYSTEM_PORT table in PortsOrch
 * React to changes in the SYSTEM_PORT table
 * Make the appropriate `create_port` and `remove_port` calls.

In addition, forwarding features that are dependent on System Ports need to react to these changes and reprogram the related forwarding plane state such as routing nexthops, LAG membership etc.

doc/voq/voq_hld.md
kartik-arista: You can refer to the Arch document section and add something along the lines of

## 3.2 Distributed VoQ functionality

### 3.2.3 test_virtual_chassis.py

Is the top level test driver that executes testcases against the virtual chassis.

Sysport handling is tested by test_chassis_sysport which validates that
* System ports can be populated in CONFIG_DB
* Orchagent programs the correct SAI Redis ASIC_DB state to represent the configured sysport in a multi-instance environment

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,874 @@
...
+4. Switch_ID for each asic
+
+## 1.4 Orchagent requirements
+### Switch Creation
+ - Should create the switch with the correct values for max_cores and switch_id
+ - Should create the switch with complete list of system port for all the system ports during initialization
+
+### PortsOrch:
+ - Should be SystemPorts aware
+ - Should initialize system ports list and create host interfaces for system ports
```
rck-innovium, line 130: Please update that host i/fs are created ONLY for local ports and not system ports.

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,874 @@
...
+
+The figures below show the components of this architecture on a modular chassis with separate line cards, each with multiple Forwarding asics, control card and multiple Switch Fabric asics. The terms Line Card and Control Card are architectural and not necessarily separate phyiscal elements of a Distributed VOQ system (though a modular chassis typically implements them as separate physical elements). For example it is possible that the entire VOQ system consists of a single card. Where applicable it is assumed that Control Card and Line Card components can exist on the same card (or fixed configuration device).
```
rck-innovium, line 91: Would be helpful if both pull/622  and pull/639 use the same terminology. For eg. either use FSI or  linecard;  SSI or control-card

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,874 @@
...
+```
+
+No changes in the schema of other CONFIG_DB tables. The name of the system ports used as key in the SYSTEM_PORT table is unique across chassis. The system_port_name can be same as the PORT table name as long as the PORT table name is unique across the chassis or it can be any character string which does not include ":". If a system port is used as inband interface, the name of that system port must be a name that will be accepted by kernel for netdevice name.
+
+For router interface and address configurations for the system ports, the existing INTERFACE table in CONFIG_DB is used.
+
+Please refer to the [schema](https://github.com/Azure/sonic-swss/blob/master/doc/swss-schema.md) document for details on value annotations. 
+
+## 2.3 VOQ DB
+This is a **new** database which resides in global redis server accessible by all sonic instances. This database is modeled as Application DB. The OrchAgent in all the sonic instances will write their local NEIGH to global VOQ_DB with hardware identifier's. The OrchAgent also reads the NEIGH, SYSTEM_PORTCHANNEL, PORTCHANNEL_MEMBER tables of all sonic instances with hardware information such as hardware index.
```
rck-innovium, line 308: Same terminology differences between Arch doc and this doc:  “Global DB”  vs "VOQ DB".

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,874 @@
...
+switch_id                             = 1*4DIGIT                            ; switch id
+
+```
+
+## 2.5 Orchestration agent
+### VOQ Switch Creation
+Prior to switch creation - OrchAgent determines whether or not it is a VOQ Switch by checking if VOQ specific information is present in the APP DB. It could do this by checking for the presence of my_switch_id (or max_cores or connection information for VOQ System DB) in the VOQ System Information. If it is not a VOQ Switch - switch creation goes ahead as it does currently. VOQ Switch creation requires additional information - max_cores, my_switch_id and system port list (see previous srctions for table names). It waits until this information is available from the APP DB before going ahead with switch creation. 
+### Portsorch
+This is made aware of voq system port. During PortOrch initialization, portsorch makes a list of all the system ports created during switch creation (above). After "PortInitDone" for all the local ports, portsorch adds system ports to ports list and creats host interfaces for all the system ports. PortOrch PortChannel processing is made aware of system port channels.
```
rck-innovium, line 369: creats host interfaces for all the local and not **system** ports

doc/voq/voq_hld.md
skeesara-nokia: Will fix the terminology once the changes to PR-622 are completed

doc/voq/voq_hld.md
skeesara-nokia: Agreed. Will make the changes in the update to PR-639

doc/voq/voq_hld.md
skeesara-nokia: This section has been updated in the latest revision to add a reference to the architecture document

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,809 @@
+# VOQ SONiC
```
eswaranb, line 1: Should we move all the files to the 'chassis' directory? 'voq' is a pretty non-obvious term and I think it's better for this set of docs to be under /doc/chassis than /doc/voq. What do you think?

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,809 @@
...
+ ![](../../images/voq_hld/control-card-system-port-config-flow.png)
+
+## 2.2 Config DB
+
+### 2.2.1 DEVICE_METADATA
+ The **existing** DEVICE_METADATA table is enhanced to add new entry to have VOQ related parameters
+ 
+```
+DEVICE_METADATA: {
+   "voqsystem": {
```
eswaranb, line 182: Do we need the "voqsystem" hierarchy?

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,809 @@
...
+5.  Host IP reachability between the SONiC instances in the system over the datapath.
+6.  Routing protocol peering between SONiC instances over the datapath.
+7.  Static provisioning of System Ports in the VOQ System Database.
+8.  Dynamic discovery of Routing Interfaces and Neighbors on other asics via the VOQ System Database.
+9.  Automatic determination of Switch_Id for each asic (optional).
+
+## 1.2 Platform requirements
+Every ASIC in the system needs be assigned a unique Switch_ID. Each ASIC consumes as many consecutive switch_id values as it has cores. So the next valid switch_id assignment should skip ahead by a number equal to the number of cores in the asic. This can be done via an additional configuration attribute. Another option is atuomatic determination of sWwitch_ID. Vendor supplied platform specific components are aware of details (line card number, asic number with in the line card, number of cores for the asic etc...) which allow them to automatically calculate the Switch_ID for each asic. For this reason it is proposed that the PMON could populate the Switch_ID for an asic into the APPL_DB.
```
eswaranb, line 87: In a later section, the DEVICE_METADATA section carries the switch_id. Is the proposal still to have PMON populate the switch_id?

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,809 @@
...
+### IntfsOrch:
+- Should create/delete/update router interfaces against local-asic ports
+- Should create/delete/update remote-asic router interfaces on the local switch
+
+### NeighOrch:
+ - Should be aware of Neighbors on other asics
+ - Should be able to create/delete/update "remote-asic" Neighbors on the local switch.
+ - Should be able to create/delete/update next hops for "remote-asic" neighbors on the local switch.
+
+### VOQ DB sync
```
eswaranb, line 114: Do we still need this?

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,809 @@
...
+            "instance" : "voq-redis"
+        },
+```
+        
+
+### 2.2.2 System Port Table
+A **new** table for system port configuration
+
+```
+SYSTEM_PORT:{{system_port_name = PORT.port_name}}
```
eswaranb, line 233: I think we should name the SYSTEM_PORT like this - 

ASIC_NAME|LOCAL_PORT_NAME

This way, we don't need to keep a separate local name for the interface in this table and all combinations of information is available everywhere. The key for something like SYSTEM_INTERFACE can also follow the same pattern.

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,813 @@
...
+DEVICE_METADATA: {
+   "voqsystem": {
+    "switch_type": {{switch_type}}
+    "switch_id": {{switch_id}}
+    "max_cores" : {{max_cores}}
+   }
+}
+```
+
+VOQ_DB will be specified using database_config.json.j2 and it will generated in control card and linecard differently. Control card will be the server and line card will be running clients only. DBConnectors will use this config to connect correct VOQ_DB.
```
kartik-arista, line 190: As per the arch document in pull/622 and based on verbal discussion, the conclusion was to name the shared database in the control card/supervisor as Global DB and the instance that stores the shared state as GLOBAL_APP_DB

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,813 @@
...
+{%- endif %}
+    },
+    "DATABASES" : {
+        "APPL_DB" : {
+            "id" : 0,
+            "separator": ":",
+            "instance" : "redis"
+        },
+        ...
+        "VOQ_DB" : {
```
kartik-arista, line 221: As mentioned above, this should be GLOBAL_APP_DB

doc/voq/voq_hld.md
kartik-arista: Changes to PR-622 are done (no further changes are planned).

doc/voq/voq_hld.md
vganesan-nokia: This is one suggested model out of 2 alternate models. We can have any string describing the slot, asic and port name in any order for the system port name. Only requirement is we should not have ":" in the string. This is mentioned in the note below the alternate model (after OR).

doc/voq/voq_hld.md
vganesan-nokia: No. Syncing is done in intfsorch and neighorch. No separate process. We'll update the hld.

doc/voq/voq_hld.md
vganesan-nokia: Agreed. We'll update the hld

doc/voq/voq_hld.md
vganesan-nokia: Agreed. We'll update the hld

doc/voq/voq_hld.md
vganesan-nokia: We  believe yes. I would be clearer to group all the chassis/voq related meta data in a separate entry.

doc/voq/voq_hld.md
eswaranb: Let's stick to voq directory.

doc/voq/voq_hld.md
eswaranb: We will name switch_type = voq or fabric. Other names can remain the same.

doc/voq/voq_hld.md
eswaranb: Recommendation - 
SYSTEM_PORT|Linecard3.1|Ethernet0
SYSTEM_PORT|Linecard4.0|Ethernet0

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,801 @@
...
+## 1.2 Platform requirements
+Every ASIC in the system needs be assigned a unique Switch_ID. Each ASIC consumes as many consecutive switch_id values as it has cores. So the next valid switch_id assignment should skip ahead by a number equal to the number of cores in the asic. This can be done via an additional configuration attribute. Another option is atuomatic determination of sWwitch_ID. Vendor supplied platform specific components are aware of details (line card number, asic number with in the line card, number of cores for the asic etc...) which allow them to automatically calculate the Switch_ID for each asic. For this reason it is proposed that the PMON could populate the Switch_ID for an asic into the APPL_DB.
+
+## 1.3 Configuration requirements
+Phase-1 of the Distributed VOQ System should support static configuration of the following
+1. Connection parameters for the Central VOQ System Database (IP address, Port)
+2. All the system ports in the entire system. The configuration of each system port should specify - system_port_id, switch_id, cored_index, core_port_index and speed.
+3. Maximum number of cores in the system.
+4. Switch_ID for each asic
+
```
kperumalbfn, line 94: For features like NAT where the NAT entries are derived from global pool, I think SNAT and DNAT trap packets should be punted to supervisor CPU. Is there any SAI proposal to add the supervisor CPU for hostif TRAP entries?

doc/voq/voq_hld.md
skeesara-nokia: It might be better if you can add this part to the document. Let me know

doc/voq/voq_hld.md
```diff
@@ -0,0 +1,797 @@
...
+For router interface and address configurations for the system ports, the existing INTERFACE table in CONFIG_DB is used.
+
+Please refer to the [schema](https://github.com/Azure/sonic-swss/blob/master/doc/swss-schema.md) document for details on value annotations. 
+
+## 2.3 CHASSIS_APP_DB
+This is a **new** database which resides in chassis redis server accessible by all sonic instances. This database is modeled as Application DB. The OrchAgent in all the sonic instances will write their local information such as INTERFACE, NEIGH, PORTCHANNEL and PORTCHANNEL_MEMBER tables to CHASSIS_APP_DB with hardware identifier's. The OrchAgent also reads sync-ed remote info such as SYSTEM_INTERFACE, SYSTEM_NEIGH, SYSTEM_PORTCHANNEL and SYSTEM_PORTCHANNEL_MEMBER tables of all sonic instances with hardware information such as hardware index.
+
```
rlhui, line 262: It's better to move this section "Chassis_app_db" up a bit, as the previous section also mentions "chassis_app_db".  Also can we make it clear whether chassis_app_db is only in control card, or will be in line cards as well?

rlhui:
Are all action items discussed in chassis subgroup for this HLD closed/updated in doc if needed?  E.g. "AI: Update Switch_ID can be computed using the SKU information. If the config_db is automatically created using the chassis map, Switch_ID can be auto-generated as well"?

doc/voq/voq_hld.md
eswaranb: Srikanth, can you please do the edits necessary? To clarify, chassis_app_db will only run in the control card.

eswaranb:
> Are all action items discussed in chassis subgroup for this HLD closed/updated in doc if needed? E.g. "AI: Update Switch_ID can be computed using the SKU information. If the config_db is automatically created using the chassis map, Switch_ID can be auto-generated as well"?

The switch_id is provided as part of minigraph of the linecards under the DEVICE_METADATA section. I don't think our current proposal is to auto-generate switch_ID as part of Sonic.

rlhui (approved):

PR #643, opened 2020-07-01
HLD for audit-log feature

spenugondaa:
Contains HLD for audit-log feature. 

Audit log helps incident analysis and contains user login/logout activity and command history. The log is stored at /var/log.
Two commands help managing the log: 'show audit-log' and 'clear audit-log'.
The 'show audit-log' displays contents of the file with 'all' option to display all the contents.
The 'clear audit-log' as the name indicates,  clears the audit log.
The file is logrotate'd.


spenugondaa:
Not sure why so many files ended up in the PR. Will raise new one.

PR #648, opened 2020-07-14
[fwutil.md] Fwutil Auto-update support

sujinmkang:
Update fwutil to support for the automatic fw update with boot/action option based on platform_components.json


doc/fwutil/fwutil.md
```diff
@@ -392,18 +392,50 @@ _sonic-buildimage/device/<platform_name>/<onie_platform>/platform_components.jso
...
+{
+    "chassis": {
+        "Chassis1": {
+            "component": {
+                "BIOS": {
+                    "firmware": "/usr/share/sonic/<platform_name>/<onie_platform>/fw_update/bios.bin",
+                    "script": "/usr/share/sonic/<platform_name>/<onie_platform>/fw_update/bios_fw_update",
+                    "version": "0ACLH003_02.02.010"
+                    "required_reboot": "None"
+                    "immediate_action": "No"
```
nazariig, line 418: @sujinmkang IMHO, this should be fetched automatically by Platform API if needed.
The config file should be simple and clear.

**Use case:** the same firmware package may have a different behaviour based on device vendor/model.
**Example:** SSD with/without on-the-fly update mode

doc/fwutil/fwutil.md
```diff
@@ -340,7 +340,7 @@ The purpose of the update commands group is to provide an interface
 for automatic FW installation of various platform components.
 
 Automatic FW installation requires platform_components.json to be created and placed at:  
-_sonic-buildimage/device/<platform_name>/<onie_platform>/platform_components.json_
+_sonic-buildimage/device/<platform_name>/<onie_platform>/platform_fw_updatecomponents.json_
```
nazariig, line 343: @sujinmkang Typo? Maybe it's better to have something like `platform_fw_updates`?

doc/fwutil/fwutil.md
```diff
@@ -374,15 +374,15 @@ _sonic-buildimage/device/<platform_name>/<onie_platform>/platform_components.jso
...
                     "version": "0ACLH003_02.02.010"
                 },
                 "CPLD": {
-                    "firmware": "/etc/<platform_name>/fw/<onie_platform>/chassis1/cpld.bin",
+                    "firmware": "/usr/share/sonic/<platform_name>/<onie_platform>/fw_update/cpld.bin",
                     "version": "10"
                 },
                 "FPGA": {
-                    "firmware": "/etc/<platform_name>/fw/<onie_platform>/chassis1/fpga.bin",
+                    "firmware": "/usr/share/sonic/<platform_name>/<onie_platform>/fw_update/fpga.bin",
```
nazariig, line 385: @sujinmkang Not sure whether this is the right place for fw binaries. Also this may overlap with the sonic_device_data package contents.

doc/fwutil/fwutil.md
```diff
@@ -454,6 +486,24 @@ Warning: <firmware_update_notification>
...
+Warning: <firmware_update_notification>
+...
+FW update in progress ...
+...
+root@sonic:~# fwutil --yes --image=next --boot=any --immediate=no
+Warning: <firmware_update_notification>
+...
+FW update in progress ...
+...
+root@sonic:~# fwutil --yes --image=next --boot=any --immediate=yes --component=ssd
```
nazariig, line 501: @sujinmkang What about modular platforms? AFAIK, there is no clear restrictions on component names, which means that different modules can share the same component name. And what should be the expected behaviour in that case?

doc/fwutil/fwutil.md
```diff
@@ -454,6 +486,24 @@ Warning: <firmware_update_notification>
 New FW will be installed, continue? [y/N]: N
 Aborted!
 ```
+3. Non modular chassis platform
+```bash
+root@sonic:~# fwutil --yes --image=next --boot=cold --immediate=yes
```
nazariig, line 491: @sujinmkang I suggest to use something like: `fwutil update components fw -y --image=next -b fast`

doc/fwutil/fwutil.md
```diff
@@ -490,6 +540,24 @@ Aborted!
...
+### 3.4.2 Modular chassis platform
+
+TBD
+
+### 3.4.3 Auto FW update based on the configuration file
+
+1. Find the available firmware based on boot type and immediate action type from platform specific fw update configuration file. 
+   Exit if no configuration file exists.
+2. Update the firmware using the script if it's specified in the configuration. Otherwise, fwutil will use the platform api to update the firmware.
+   Exit if the update fails in any step.
```
nazariig, line 558: @sujinmkang I would like to propose a new CLI tree:
```
fwutil
|--- show
|    |--- version
|    |--- status
|    |--- updates -i|--image=<current|next> -c|--config=<path_to_platform_fw_updates>
|
|--- install
|    |--- chassis
|    |    |--- component <component_name>
|    |         |--- fw -y|--yes <fw_path>
|    |
|    |--- module <module_name>
|         |--- component <component_name>
|              |--- fw -y|--yes <fw_path>
|
|--- update
     |--- chassis
     |    |--- component <component_name>
     |         |--- fw -y|--yes -f|--force -i|--image=<current|next> -c|--config=<path_to_platform_fw_updates>
     |
     |--- module <module_name>
     |    |--- component <component_name>
     |         |--- fw -y|--yes -f|--force -i|--image=<current|next> -c|--config=<path_to_platform_fw_updates>
     |
     |--- components
          |--- fw -y|--yes -f|--force -i|--image=<current|next> -b|--boot=<cold|fast|warm> -c|--config=<path_to_platform_fw_updates>
```

nazariig: @sujinmkang Some thoughts:
1. In case platform API is not implemented (plugin mode) the fwutil CLI won't be fully available - only automatic installation mode. That's something which need to be highlighted in this design.
2. Platform API should be extended with the new boot argument.
3. How to handle `version` and `script` fields in case Platform API is not implemented

doc/fwutil/fwutil.md
```diff
@@ -462,6 +498,278 @@ Aborted!
...
+    nextimage = bootloader.get_next_image()
+    If [[ nextimage == curimage ]]; then
+        NEXT_BOOT_IMAGE=current
+    else
+        NEXT_BOOT_IMAGE=next
+    fi
+    ...
+    if [[ -x ${PLATFORM_FW_UPDATE} ]]; then
+        debug "updating platform fw for ${REBOOT_TYPE}"
+        ${PLATFORM_FW_UPDATE} auto-update --yes --image={NEXT_BOOT_IMAGE} --boot=${REBOOT_TYPE}
```
rkdevi27, line 719: Hi Sujin,

In what scenarios the fwutil command will include a reboot/powercycle? only in cases of immediate_(fast/warm/cold/powercycle)?

Also in few cases the upgrade may require unmount of filesystems and better to include them in the flow of reboot after unmounting filesystems. In that case fwutil can do just the placing of files in correct location and the actual upgrade will be done only as part of the reboot script. In those scenarios if this command just places the file, in what ways the actual upgrade can be included in these scripts?

doc/fwutil/fwutil.md
```diff
@@ -462,6 +540,200 @@ Aborted!
...
+The task file will be platform-specific.
+
+**Example:**
+1. `platform_component.json` with platform specific utility.
+```json
+{
+    "component": {
+        "BIOS": {
+            "firmware": "bios.bin",
+            "version": "0ACLH003_02.02.010",
```
padmanarayana, line 642: boot option needs to be added

doc/fwutil/fwutil.md
```diff
@@ -462,6 +540,200 @@ Aborted!
...
+CPLD firmware update is going to be done with this command and the CPLD completion activity will be performed during cold reboot
+once all reboot processes are finished. The CPLD completion activity for this case is a power cycle triggered by the hw register setting.
+```bash
+root@sonic:~# fwutil auto-update fw --yes --image=next --boot=cold
+Available FW updates with fast-reboot from next boot image : SSD and CPLD
+SSD firmware update will be done during cold-reboot with version 5 (current version 4)
+...
+CPLD firmware update is in progress with version 5 (current version 4)
+...
+CPLD firmware update is completed with version 5 (current version 4)
```
padmanarayana, line 614: So, in the CPLD example above, the fwutil auto-update will be functionally equivalent to the existing "fwutil install" command, please mention this in the description.

doc/fwutil/fwutil.md
```diff
@@ -462,6 +540,200 @@ Aborted!
...
+        }
+    }
+}
+```
+
+The Utility can have two passing arguments and one option for actions from fwutil.
+1.	Firmware
+2.	Boot action following after the firmware auto-update process
+3.	Fwutil action : -a(--autoupdate)
+**Utility can be supported for other platform api substitues like `compoenent_update` and `compoenent_install` with |-u(--update)|-i(--install)**
```
padmanarayana, line 662: typo

doc/fwutil/fwutil.md
```diff
@@ -462,6 +540,200 @@ Aborted!
...
+
+The expected utility response can be as follows.
+```json
+"True" : "completed"
+"True" : "scheduled"
+"False" : Error
+```
```
rkdevi27, line 669: The "fwutil show auto-update" command has status as "updated but need a power-cycle". Please explain how these status will be derived from these utility return strings described above.

doc/fwutil/fwutil.md
```diff
@@ -405,6 +459,33 @@ _sonic-buildimage/device/<platform_name>/<onie_platform>/platform_components.jso
...
+                },
+                "CPLD": {
+                    "firmware": "/usr/share/sonic/<platform_name>/<onie_platform>/fw_update/cpld.bin",
+                    "version": "10",
+                    "boot": "powercycle"
+                },
+                "FPGA": {
+                    "firmware": "/usr/share/sonic/<platform_name>/<onie_platform>/fw_update/fpga.bin",
+                    "version": "5",
+                    "boot": "cold"
```
rkdevi27, line 481: Can this boot option be removed from this json file since the platform utility will handle the boot-type passed by the user as explained below.

doc/fwutil/fwutil.md
```diff
@@ -85,18 +95,18 @@ This document provides general information about FW utility implementation in SO
 
 ## 1.1 Feature overview
 
-A modern network switch is a sophisticated equipment which consists of many auxiliary components  
-which are responsible for managing different subsystems (e.g., PSU/FAN/QSFP/EEPROM/THERMAL)
```
nazariig, line 89: @sujinmkang why do we need to remove these whitespaces? Basically in `.md` two space notation is used to start a new line which allows text separation.

doc/fwutil/fwutil.md
```diff
@@ -158,20 +173,21 @@ In order to improve scalability and performance a modern network switches provid
 1. Non modular chassis platforms
 2. Modular chassis platforms
 
-Non modular chassis platforms may contain only one chassis.
```
nazariig, line 161: @sujinmkang why do we need to remove these whitespaces? Basically in .md two space notation is used to start a new line which allows text separation.

doc/fwutil/fwutil.md
```diff
@@ -158,20 +173,21 @@ In order to improve scalability and performance a modern network switches provid
...
 Basically each chassis/module may contain one or more components (e.g., BIOS/CPLD/FPGA).
 
-SONiC platform API provides an interface for FW maintenance operations for both modular and  
-non modular chassis platforms. Both modular and non modular chassis platforms share the same platform API,  
+SONiC platform API provides an interface for FW maintenance operations for both modular and
+non modular chassis platforms. Both modular and non modular chassis platforms share the same platform API,
 but may have different implementation.
 
 SONiC FW utility uses platform API to interact with the various platform components.
+SONiC FW utility extends to support for the automatic firmware update based on <platform>_components.json and next reboot option.
```
nazariig, line 190: @sujinmkang IMHO, it's better to mention here that besides Platform API FW tools can also use plugin concept when all the auxiliary stuff is provided within a package (flash tools, scripts, etc.)

doc/fwutil/fwutil.md
```diff
@@ -252,35 +271,69 @@ Chassis1           BIOS        0ACLH004_02.02.007  Chassis BIOS
...
+          Module1  CPLD        <image_path>/cpld.bin  5 / 10                                   update is required  N/A
+                   FPGA        <image_path>/fpga.bin  5 / 5                                    up-to-date          N/A
+```
+
+4. Custom FW Package when the "boot" key and value are available in `platform_components.json`
+```bash
+root@sonic:~# fwutil show updates --fw-image=fw_update.tar.gz
+Component   Firmware               Version (current/available)              Status              Required Boot Action
+----------  ---------------------  ---------------------------------------  ------------------  -----------------------
+BIOS        <image_path>/bios.bin  0ACLH004_02.02.007 / NA                  update is required  None
```
nazariig, line 311: @sujinmkang do we want to print the entire line in case firmware update is not available? Does it make sense?

doc/fwutil/fwutil.md
```diff
@@ -252,35 +271,69 @@ Chassis1           BIOS        0ACLH004_02.02.007  Chassis BIOS
...
+CPLD        <image_path>/cpld.bin  5 / 10                                   update is required  Power cycle
```
nazariig, line 312: @sujinmkang do we need `<image_path>` prefix here? IMHO, the firmware binary is a part of tarball, so no need to specify full path - name+extension should be enough.

doc/fwutil/fwutil.md
```diff
@@ -252,35 +271,69 @@ Chassis1           BIOS        0ACLH004_02.02.007  Chassis BIOS
...
+SSD         <image_path>/ssd.bin   4 / 5                                    update is required  Cold reboot, Fast reboot
+```
+
+5. Auto-update status
+```bash
+root@sonic:~# fwutil show auto-updates status
+Component   Firmware               Version (current/available)              Status
+----------  ---------------------  ---------------------------------------  -------------------------
+CPLD        <image_path>/cpld.bin  5 / 10                                   updated but need a power-cycle
+SSD         <image_path>/ssd.bin   4 / 5                                    scheduled in cold reboot
```
nazariig, line 323: @sujinmkang where do we plan to store states like these: updated but need a power-cycle, scheduled in cold reboot

doc/fwutil/fwutil.md
```diff
@@ -252,35 +271,69 @@ Chassis1           BIOS        0ACLH004_02.02.007  Chassis BIOS
...
+FPGA        <image_path>/fpga.bin  5 / NA                                   up-to-date          Cold reboot
+SSD         <image_path>/ssd.bin   4 / 5                                    update is required  Cold reboot, Fast reboot
+```
+
+5. Auto-update status
+```bash
+root@sonic:~# fwutil show auto-updates status
+Component   Firmware               Version (current/available)              Status
+----------  ---------------------  ---------------------------------------  -------------------------
+CPLD        <image_path>/cpld.bin  5 / 10                                   updated but need a power-cycle
```
nazariig, line 322: @sujinmkang who will be responsible for that power cycle? What is the flow?

doc/fwutil/fwutil.md
```diff
@@ -252,35 +271,69 @@ Chassis1           BIOS        0ACLH004_02.02.007  Chassis BIOS
...
 1. -i|--image - show updates using current/next SONiC image
+2. -z|--fw-image - show updates using custom FW package
 
 **Note:** the default option is _--image=current_
 
 #### 2.2.2.2 Install commands
 
 ##### 2.2.2.2.1 Overview
 
-The purpose of the install commands group is to provide an interface
```
nazariig, line 283: @sujinmkang why do we need to remove these whitespaces? Basically in .md two space notation is used to start a new line which allows text separation.

doc/fwutil/fwutil.md
```diff
@@ -252,35 +271,69 @@ Chassis1           BIOS        0ACLH004_02.02.007  Chassis BIOS
...
+----------  ---------------------  ---------------------------------------  ------------------  -----------------------
+BIOS        <image_path>/bios.bin  0ACLH004_02.02.007 / NA                  update is required  None
+CPLD        <image_path>/cpld.bin  5 / 10                                   update is required  Power cycle
+FPGA        <image_path>/fpga.bin  5 / NA                                   up-to-date          Cold reboot
+SSD         <image_path>/ssd.bin   4 / 5                                    update is required  Cold reboot, Fast reboot
+```
+
+5. Auto-update status
+```bash
+root@sonic:~# fwutil show auto-updates status
```
nazariig, line 319: @sujinmkang it seems that you have forgotten to add `auto-updates` to `show` command section tree

doc/fwutil/fwutil.md
```diff
@@ -462,6 +516,218 @@ Aborted!
...
+**Note:**
+1. FW update will be disabled if component definition is not provided (e.g., 'BIOS': { })
+2. FW version will be read from image if `version` field is not provided
+
+##### 2.2.2.4.2 Description
+
+**The following command updates FW Automatically with installing the available component firmware, creating task(s) for update during reboot:**
+1. auto-update command for installation and reboot task creation
+```bash
+root@sonic:~# fwutil auto-update fw --yes --image=<current|next> --boot=<boot_type>
```
nazariig, line 555: @sujinmkang what if this will be called multiple times with different reboot types? Do we expect previously scheduled task will be cancelled?

nazariig, line 555: @sujinmkang is it implicitly assumed to first update firmware of the components which don't need any reboot action (support on-the-fly updates)? Or we want to be robust and process only that components which match specified boot type?

doc/fwutil/fwutil.md
```diff
@@ -195,16 +211,19 @@ fwutil
...
+|    |
+|    |--- module <module_name>
+|         |--- component <component_name>
+|              |--- fw -y|--yes -f|--force -i|--image=<current|next>
+|
+|--- auto-update
+     |--- fw -z|--image=<current|next> --b|--boot=<any|none|fast|warm|cold|powercycle>
+     |--- fw -z|--fw-image=<fw_package.tar.gz> --b|--boot=<any|none|fast|warm|cold|powercycle>
 
-**Note:** <fw_path> can be absolute path or URL
```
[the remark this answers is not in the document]
sujinmkang: I will add this back.

doc/fwutil/fwutil.md
sujinmkang: good point.

doc/fwutil/fwutil.md
sujinmkang: good point.

doc/fwutil/fwutil.md
sujinmkang: @nazariig no we're not going to allow another auto-update if the previous auto update was successful since the previous auto-update could have dependency with boot-type. As an example, the previous auto-update was called for cold reboot type which could've installed the cpld firmware but if a new auto-update request is initiated with fast-reboot, we don't want to proceed. So fwutil check if any task file is available, then a new auto-update request will not be allowed.

sujinmkang: @nazariig It is assumed to update all components with any given boot-type. If the component doesn't need any reboot action, then it should be updated in any boot action.

doc/fwutil/fwutil.md
sujinmkang: @nazariig I will update.

doc/fwutil/fwutil.md
sujinmkang: @nazariig Power cycle could be done in platform_reboot_plugin based on the task file similar to how we did in 201811 branch. I will add that part. Do you have any other suggestion?

doc/fwutil/fwutil.md
sujinmkang: @nazariig sorry for the delayed pr update! please find the task file section for this information.

doc/fwutil/fwutil.md
sujinmkang: reverted.

doc/fwutil/fwutil.md
```diff
@@ -195,16 +214,20 @@ fwutil
...
-```
+|    |--- chassis
+|    |    |--- component <component_name>
+|    |         |--- fw -y|--yes -f|--force -i|--image=<current|next>
+|    |
+|    |--- module <module_name>
+|         |--- component <component_name>
+|              |--- fw -y|--yes -f|--force -i|--image=<current|next>
+|
+|--- auto_update
```
nazariig, line 225: @sujinmkang `auto-update`

doc/fwutil/fwutil.md
```diff
@@ -257,22 +280,55 @@ Chassis   Module   Component   Firmware               Version (current/available
...
+                    SSD         <fwpackage_path>/ssd.bin       4 / 5                                    update is required
+```
+
+5. Auto-update status
+```bash
+root@sonic:~# fwutil show auto_update_status
+Firmware auto-update performed for cold reboot
+Component      Status    Info
+-------------  --------  ---------
+MSN2700/SSD    True      installed
```
nazariig, line 325: @sujinmkang This is not too much informative:
```
Status
--------
True
```
It would be much better to have something like `status=installed/updated/scheduled`:
```
Status    Info
--------  ---------------------
updated   No further actions required
scheduled Pending cold reboot
```

doc/fwutil/fwutil.md
```diff
@@ -257,22 +280,55 @@ Chassis   Module   Component   Firmware               Version (current/available
...
 --------  -------  ----------  ---------------------  ---------------------------------------  ------------------
 Chassis1           BIOS        <image_path>/bios.bin  0ACLH004_02.02.007 / 0ACLH004_02.02.010  update is required
                    CPLD        <image_path>/cpld.bin  5 / 10                                   update is required
                    FPGA        <image_path>/fpga.bin  5 / 5                                    up-to-date
+                   SSD         <image_path>/ssd.bin   4 / 5                                    update is required
           Module1  CPLD        <image_path>/cpld.bin  5 / 10                                   update is required
                    FPGA        <image_path>/fpga.bin  5 / 5                                    up-to-date
 ```
 
+4. Custom FW Package when the "boot" key and value are available in `platform_components.json`
```
nazariig, line 310: @sujinmkang What would be the CLI output if we don't have these available:
`when the "boot" key and value are available`

doc/fwutil/fwutil.md
```diff
@@ -195,16 +214,20 @@ fwutil
...
+|    |    |--- component <component_name>
+|    |         |--- fw -y|--yes -f|--force -i|--image=<current|next>
+|    |
+|    |--- module <module_name>
+|         |--- component <component_name>
+|              |--- fw -y|--yes -f|--force -i|--image=<current|next>
+|
+|--- auto_update
+     |--- fw -i|--image=<current|next> --b|--boot=<any|none|fast|warm|cold|powercycle>
+     |--- fw -z|--fw-image=<fw_package.tar.gz> --b|--boot=<any|none|fast|warm|cold|powercycle>
```
nazariig, line 227: @sujinmkang How do you expect to handle power cycle scenario? SONiC can only distinguish between cold/fast/warm reboots.

doc/fwutil/fwutil.md
```diff
@@ -184,6 +201,8 @@ fwutil
 |    |--- version
 |    |--- status
 |    |--- updates -i|--image=<current|next>
+|    |--- updates -z|--fw-image=<fw_package.tar.gz>
```
nazariig, line 204: @sujinmkang what is the expected behaviour when --image=next and --fw-image=<fw_package.tar.gz> are provided simultaneously?

doc/fwutil/fwutil.md
```diff
@@ -462,6 +519,321 @@ Aborted!
...
+        firmware_path,
+        boot_type
+    )
+    output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
+...
+```
+From the above `platform_components.json` example, if the platform needs the SSD firmware update, then the auto-update can be triggered with following command.
+`$PWD/ssd_fw_update -a $PWD/SSD.bin fast` 
+
+##### 2.2.2.4.3.2 Platform Firmware Update Reboot Handle Plugin
```
nazariig, line 787: @sujinmkang please elaborate for a different flows (cold/warm/fast): at which point do you expect to call this plugin?

doc/fwutil/fwutil.md
```diff
@@ -462,6 +519,321 @@ Aborted!
 
 **Note:** the default option is _--image=current_
 
+#### 2.2.2.4 Auto-update commands
+
+##### 2.2.2.4.1 Overview
```
nazariig, line 524: @sujinmkang in case of boot_action=any what should be the expected behaviour if platform component supports both cold and fast reboot? Do we expect platform API will add this item to both cold_* and fast_* task lists?

doc/fwutil/fwutil.md
```diff
@@ -462,6 +519,321 @@ Aborted!
...
+    }
+}
+```
+
+##### 2.2.2.4.3.1 Platform Component Firmware Update Utility Interface Requirement
+
+The Utility should support the minimum requirements to perform the fwutil auto-update interface. 
+The minimum requirement is that the component api's get_frimware_version, get_firmware_update_notification(), and auto_update_firmware() needs to be supported by the utility.
+Here are the interface requirements to support them.
+1. {utility} -s(--status) : able to retrieve the current firmware version:
```
nazariig, line 749: @sujinmkang how do we expect to use this info in FW tools when we don't have a way to read the original (before the update) firmware version?

doc/fwutil/fwutil.md
```diff
@@ -462,6 +519,321 @@ Aborted!
...
+```
+
+##### 2.2.2.4.4 Auto-update Use Cases
+
+##### 2.2.2.4.4.1 Standalone firmware update during run time
+Since FWutil auto-update can support the firmware update using a custom firmware package, this will allow the firmware update to be available during run time.
+
+SONiC device can have a component firmware updated without interrupting the data plan if the component firmware update doesn't need any boot action.
+```bash
+root@sonic:~# fwutil auto-update fw --fw-image=aboot-fw-update.tar.gz --boot=none
```
nazariig, line 804: @sujinmkang What if the `cold/fast/warm` boot was specified as the desired action but some of the platform components do support firmware upgrade without dataplane interruption? Do we expect the update will take place implicitly or we still do want to do only with `--boot=none`?

doc/fwutil/fwutil.md
sujinmkang: Good catch. we are not going to have the boot key in platform_components.json. The platform auto-update api will perform the expected firmware update based on the boot type.  So we don't need to have the boot key in platform_components.json.

doc/fwutil/fwutil.md
sujinmkang: that is good question. I think we should limit this boot option to non/fast/warm/cold.

doc/fwutil/fwutil.md
sujinmkang: We should take only one option. my current implementation takes the --fw-image option over --image option. Should we return error when user gives both option for one fwutil call?

doc/fwutil/fwutil.md
sujinmkang: That is good question. Without boot option, the update which can be done without data plane interruption, should be performed implicitly. I will add the default setting for boot type.

doc/fwutil/fwutil.md
```diff
@@ -462,6 +519,321 @@ Aborted!
...
+else
+    TARGET_FW=`cat -v /tmp/$FILENAME |grep -m 1 "image_version" | sed -n "s/.*image_version=\"\(.*\)\".*/\1/p"`
+fi
+
+sonic_installer install -y /tmp/$FILENAME || FAILED=1
+sync;sync;sync || exit 14
+
+
+if [[ -x ${PLATFORM_FW_UPDATE} ]]; then
+    ${PLATFORM_FW_UPDATE} auto-update fw --yes --image=next --boot=${REBOOT_TYPE}
```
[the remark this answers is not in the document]
sujinmkang: Yes. if there is no platform component firmware update available for the boot action, no firmware update is expected. The fwutil will return a message saying "no update is available for the specific boot type"

doc/fwutil/fwutil.md
sujinmkang: As I replied in your previous question, we need to limit the boot action by none/fast/warm/cold. And in the previous comments, we don't support multiple firmware auto update for the same or other boot type.

doc/fwutil/fwutil.md
```diff
@@ -462,6 +519,321 @@ Aborted!
...
+The Utility should support the minimum requirements to perform the fwutil auto-update interface. 
+The minimum requirement is that the component api's get_frimware_version, get_firmware_update_notification(), and auto_update_firmware() needs to be supported by the utility.
+Here are the interface requirements to support them.
+1. {utility} -s(--status) : able to retrieve the current firmware version:
+    - equivalent to the component api's get_firmware_version()
+    - response : installed firmware version
+2. {utility} -n(--notification) : able to provide the update complete action :
+    - equivalent to the component api's get_firmware_update_notification(image_path) 
+    - response : the required action to complete the component installation
+3. {utility} -a(--autoupdate) : able to perform the auto-update action :
```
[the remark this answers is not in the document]
sujinmkang: this utility will replace the platform api so I was expecting to pass the firmware path and boot type without any separate option. I will clarify this part.

doc/fwutil/fwutil.md
sujinmkang: This api should retrieve the original version from the platform device. not the update firmware version.

rkdevi27 (approved):

sujinmkang:
@nazariig Can you please review this again?

doc/fwutil/fwutil.md
```diff
@@ -462,6 +509,333 @@ Aborted!
...
+The FWutil infra section below indicates the necessary changes to add support for auto update during reboot.
+The Platform plugins section describes the plugins that a platform must implement to support auto updates when the platform component APIs are not available.
+
+Automatic FW installation requires default platform_components.json to be created and placed at:
+_sonic-buildimage/device/<platform_name>/<onie_platform>/platform_components.json_
+Recommended image path is "/lib/firmware/<vendor>".
+
+Auto-update command can support the standalone custom firmware image package with --fw-image option.
+The package can be any format between ".tar" or ".tar.gz" and should have the "platform_components.json", the firmware image(s).
+The package can have the auto-update plugin if the platform doesn't support platform api or if plugin needs to be updatedplugin.
```
nazariig, line 615: @sujinmkang IMHO, the original idea of FW package was to provide an ability to do the upgrades on platforms which don't support platform API. Do we really want to use it for upgrades which involves platform API, when fwutil is already provides an infrastructure for that (e.g, this can be achived with FW deb packages)?

nazariig (changes requested):

doc/fwutil/fwutil.md
```diff
@@ -462,6 +509,333 @@ Aborted!
...
+Any scheduled component firmware update needs to be logged in "<boot_type>_fw_au_task".
+Each vendor needs to implement the reboot auto-update task handler plugin which is platform specific plugin and is expected to be named as "platform_fw_au_reboot_handle".
+The plugin will parse the "<boot_type>_fw_au_task" file and perform the update during the reboot if needed. 
+The contents of the task file can be defined by the platform vendors and be utilized by the platform specific reboot plugin.
+
+```
+{
+    "<boot_type>": [
+        {
+            "status": <true/false>,
```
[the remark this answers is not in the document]
sujinmkang: @nazariig This is how fwutil save the status of auto-update in the fw_au_status file. I'm proposing that fwutil needs to process the return code of platform component api and save the process data as this format.

doc/fwutil/fwutil.md
```diff
@@ -257,6 +280,7 @@ Chassis   Module   Component   Firmware               Version (current/available
 Chassis1  N/A      BIOS        <image_path>/bios.bin  0ACLH004_02.02.007 / 0ACLH004_02.02.010  update is required
                    CPLD        <image_path>/cpld.bin  5 / 10                                   update is required
                    FPGA        <image_path>/fpga.bin  5 / 5                                    up-to-date
+                   SSD         <image_path>/ssd.bin   4 / 5                                    update is required
```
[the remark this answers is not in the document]
sujinmkang: @nazariig All components need to support the automatic firmware update but only any component firmware update won't happen during this boot time if the auto-update cannot be done for the boot type. And user will notify the components firmware update cannot be done for the boot-type. I don't think it's necessary.

doc/fwutil/fwutil.md
sujinmkang: @nazariig I think it's good point. I understand we use the FW deb package to installed firmware which is missing from the current package. We can support both for the new firmware update based on the current image. But when we use a deb package to perform auto-update, we need 3 steps, push the firmware deb package, install the deb package and fwutil auto-update while this FW package approach needs 2 steps for auto update, push the FW package and fwutil aut-update with the FW package.

sujinmkang:
@padmanarayana @nazariig @rkdevi27 can you please review this again? I want to close this by tomorrow. Thanks!

doc/fwutil/fwutil.md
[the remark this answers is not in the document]
rkdevi27: @sujinmkang From the code changes, the utility -a interface alone has been used. So is it required for the utility to have the other interfaces implemented?

doc/fwutil/fwutil.md
```diff
@@ -195,16 +215,23 @@ fwutil
...
-               |--- fw -y|--yes -f|--force -i|--image=<current|next>
+|    |--- chassis
+|    |    |--- component <component_name>
+|    |         |--- fw -y|--yes -f|--force -i|--image=<current|next>
+|    |
+|    |--- module <module_name>
+|         |--- component <component_name>
+|              |--- fw -y|--yes -f|--force -i|--image=<current|next>
+|
+|--- auto-update
```
nazariig, line 226: @sujinmkang Please update CLI tree to use new `update all` format

doc/fwutil/fwutil.md
```diff
@@ -172,18 +187,23 @@ non modular chassis platforms. Both modular and non modular chassis platforms sh
...
 **User interface**:
+
 ```
 fwutil
 |--- show
 |    |--- version
 |    |--- status
 |    |--- updates -i|--image=<current|next>
+|    |--- updates -z|--fw-image=<fw_package.tar.gz>
+|    |--- auto_update_status
```
nazariig, line 206: @sujinmkang we need to refactor this part too since moving from `auto-update` to `update all`

sujinmkang: I think this can be changed to update_status but I hope this doesn't mislead people to think that they could check the  each component update done by `fwutil update {chassis|module <module_name>} component <component_name> fw ...`.
But I agree to change the command.

nazariig:
@sujinmkang Can we also have original documents (visio) as part of this PR? Just in case of further modification will be needed?

nazariig (approved):

doc/fwutil/fwutil.md
bbinxie: Do you know the answer about your first question? I also have this question.
####In what scenarios the fwutil command will include a reboot/powercycle? only in cases of immediate_(fast/warm/cold/powercycle)?####

PR #655, opened 2020-08-03
PMON in multi-asic platforms - an approach note.

judyjoseph:
The Multi-ASIC design was presented to community via PR https://github.com/Azure/SONiC/pull/644.

This document attempts to expand on the details of the approach taken for PMON for multi-asic platforms.


judyjoseph:
> Looks great!

Thanks Joe ! 

@lguohan, Could you please take a look.

PR #669, opened 2020-09-02
Add reboot-cause information to telemetry

sujinmkang:
To support the reboot-cause information sending via telemetry agent, this HLD proposes the state-DB schema and how to process the reboot-cause information during the reboot-cause-process.

### DIFF
--- a/doc/system-telemetry/reboot-cause.md
+++ b/doc/system-telemetry/reboot-cause.md
@@ -14,7 +14,7 @@
 #### Part 1
 During the boot, the `determine-reboot-cause` service ( previously `process-reboot-cause`) determines the last reboot-cause based on the hardware reboot-cause
 and the software reboot-cause information and creates previous-reboot-cause.txt with the information as it does currently.
-In addition, to save the history of the previous reboot-cause, `determine-reboot-cause` will save the previous
+In addition, to save the history of the previous reboot-cause, `determine-reboot-cause` service will save the previous
 reboot cause information to "/host/reboot-cause/previous-reboot-cause/" with adding timestamp at the end of file name.
 And the file will be formatted to be parsed easily.
 
@@ -48,7 +48,7 @@ admin@sonic:~$ sudo cat /host/reboot-cause/previous-reboot-cause/previous-reboot
 #### Part 2
 A new service named as `process-reboot-cause.service` which will retrieve the saved reboot-cause files and read each reboot-cause information from the files
 and save the reboot-cause information up to 10 entries to state-DB.
-Verify the information from state-DB data is available via the cli command `show reboot-history` which is extended from `show reboot-cause`.
+Verify the information from state-DB data is available via the cli command `show reboot-cause history` which is extended from `show reboot-cause`.
 
 ##### Reboot Cause Schema in state-DB
 
@@ -69,7 +69,7 @@ comment                 = STRING                         ; unstructured json for
 
 Currently `show reboot-cause` displays the last reboot-cause and performing `cat /host/reboot-cause/previous-reboot-cause.txt` to show the reboot-cause.
 This will be same as current design.
-With new design, `show reboot-history` will be added to display the previous reboot-cause information up to 10 entries from state-DB.
+With new design, `show reboot-cause history` will be added to display the previous reboot-cause information up to 10 entries from state-DB.
 
 The example shows the output of `show reboot-cause` which is same as current output and displays only the last reboot-cause.
 ```
@@ -92,9 +92,9 @@ REBOOT_CAUSE|2020_10_09_04_53_58
 ""
 ```
 
-The example shows the output of `show reboot-history` and the previous reboot cause stored in state-DB in addition to the last reboot-cause.
+The example shows the output of `show reboot-cause history` and the previous reboot cause stored in state-DB in addition to the last reboot-cause.
 ```
-admin@sonic:~$ show reboot-history 
+admin@sonic:~$ show reboot-cause history
 name                 cause        time                          user    comment
 -------------------  -----------  ----------------------------  ------  ---------
 2020_10_09_04_53_58  warm-reboot  Fri Oct  9 04:51:47 UTC 2020  admin
--- a/doc/system-telemetry/reboot-cause.md
+++ b/doc/system-telemetry/reboot-cause.md
@@ -12,37 +12,43 @@
 ### Enable sonic streaming telemetry agent to send Reboot-cause information
 
 #### Part 1
-During the boot, the process-reboot-cause processes the last reboot-cause based on the hardware reboot-cause
+During the boot, the `determine-reboot-cause`(previously `process-reboot-cause`) determines the last reboot-cause based on the hardware reboot-cause
 and the software reboot-cause information and creates previous-reboot-cause.txt with the information as it does currently.
-In addition to save the history of the previous reboot-cause, `process-reboot-cause` will save the previous
+In addition, to save the history of the previous reboot-cause, `determine-reboot-cause` will save the previous
 reboot cause information to "/host/reboot-cause/previous-reboot-cause/" with adding timestamp at the end of file name.
 And the file will be formatted to be parsed easily.
 
 The example shows the previous reboot-cause files stored in /host/reboot-cause/previous-reboot-cause/.
 ```
-$ls /host/reboot-cause/previous-reboot-cause/
-previous-reboot-cause-20200903T232033.txt
-previous-reboot-cause-20200902T101105.txt
-previous-reboot-cause-20200902T015048.txt
+admin@sonic:~$ ls /host/reboot-cause/previous-reboot-cause/
+previous-reboot-cause-2020_10_09_01_56_59.txt
+previous-reboot-cause-2020_10_09_02_00_53.txt
+previous-reboot-cause-2020_10_09_02_33_06.txt
+previous-reboot-cause-2020_10_09_04_53_58.txt
 ...
 ```
-The following example shows the content of the previous reboot-cause file - previous-reboot-cause-20200903T232033.txt.
+
+The following example shows the content of the previous reboot-cause file - previous-reboot-cause-2020_10_09_04_53_58.txt.
+```
+admin@sonic:~$ sudo cat /host/reboot-cause/previous-reboot-cause/previous-reboot-cause-2020_10_09_04_53_58.txt
+{"comment": "", "gen_time": "2020_10_09_04_53_58", "cause": "warm-reboot", "user": "admin", "time": "Fri Oct  9 04:51:47 UTC 2020"}
+```
 ```
 [
     {
-        gen_time: "20200903T232033",
-        cause: "reboot",
+        gen_time: "2020_10_09_04_53_58",
+        cause: "warm-reboot",
         user: "admin",
-        time: "Thu 03 Sep 2020 11:15:30 PM UTC",
+        time: "Fri Oct  9 04:51:47 UTC 2020"
         comment: ""
     }
 ]
 ```
 
 #### Part 2
-A new service named as `post-process-reboot-cause.service` which will retrieve the saved reboot-cause files and read each reboot-cause information from the files
+A new service named as `process-reboot-cause.service` which will retrieve the saved reboot-cause files and read each reboot-cause information from the files
 and save the reboot-cause information up to 10 entries to state-DB.
-Verify the information from state-DB data is available via the cli command `show reboot-cause history` which is extended from `show reboot-cause`.
+Verify the information from state-DB data is available via the cli command `show reboot-history` which is extended from `show reboot-cause`.
 
 ##### Reboot Cause Schema in state-DB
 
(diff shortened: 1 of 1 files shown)


doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,118 @@
...
+## Scope
+-Enable sonic streaming telemetry agent to send Reboot-cause information
+
+### Enable sonic streaming telemetry agent to send Reboot-cause information
+
+##### Part 1 
+For 1st part, process-reboot-cause copies the previous-reboot-cause.txt 
+to "/host/reboot-cause/previous-reboot-cause/" with adding timestamp 
+at the end of file name after processing the reboot-cause.
+Currently previous-reboot-cause.txt is plan text format but this file content will be formatized to be parsed easily.
```
jleveque, line 18: s/plan/plain/
s/formatized/formatted/

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,118 @@
...
+Read each reboot-cause information from saved previous-reboot-cause files
+And update reboot-cause information upto 10 entries to state-DB.
```
jleveque, line 20: s/upto/up to/

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,118 @@
...
+The example shows the previous reboot-cause files stored in /host/reboot-cause/previous-reboot-cause/.
+```
+$ls /host/reboot-cause/previous-reboot-cause/
+previous-reboot-cause-20200903T232033.txt
+previous-reboot-cause-20200902T101105.txt
+previous-reboot-cause-20200902T015048.txt
+```
+The following example shows the content of the previous reboot-cause file - previous-reboot-cause-20200903T232033.txt.
+```
+TIMESTATEMP: "20200903T232033"
```
jleveque, line 31: s/TIMESTATEMP/TIMESTAMP/

jleveque (changes requested):

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,118 @@
...
+
+Details of CLI and state-DB given below. 
+Cli command to retrieve the Reboot-cause information
+```
+$ show reboot-cause
+```
+state-DB schema to store the Reboot-cause information
+```
+; Defines information for reboot-cause
+key                     = REBOOT_CAUSE|timestamp         ; last reboot-cause processing time
```
jleveque, line 46: Suggest adding angle brackets to indicate timestamp is variable: `REBOOT_CAUSE|<timestamp>`

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,118 @@
...
+; field                 = value
+cause                   = STRING                         ; last reboot cause
+time                    = STRING                         ; time when the last reboot was initiated
+user                    = STRING                         ; user who the last reboot initiated
+comment                 = STRING                         ; unstructured json format data
+```
+Along with data new entry for timestamp will be added up to 10 entries in state_db:  
+
+```
+REBOOT_CAUSE|timestamp
```
jleveque, line 56: Suggest showing an example of 10 entries with realistic timestamps here instead

jleveque (changes requested):

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Enable sonic streaming telemetry agent to send Reboot-cause information
+
+##### Part 1 
+For 1st part, process-reboot-cause copies the previous-reboot-cause.txt 
+to "/host/reboot-cause/previous-reboot-cause/" with adding timestamp 
+at the end of file name after processing the reboot-cause.
+Currently previous-reboot-cause.txt is plain text format but this file content will be formatted to be parsed easily.
+Read each reboot-cause information from saved previous-reboot-cause files
+And update reboot-cause information up to 10 entries to state-DB.
```
prsunny, line 20: which module updates the reboot-cause information to STATE_DB?

doc/system-telemetry/reboot-cause.md
```diff
@@ -11,78 +11,73 @@
...
 ### Enable sonic streaming telemetry agent to send Reboot-cause information
 
-##### Part 1 
-For 1st part, process-reboot-cause copies the previous-reboot-cause.txt 
-to "/host/reboot-cause/previous-reboot-cause/" with adding timestamp 
-at the end of file name after processing the reboot-cause.
+#### Part 1
+During the boot, the process-reboot-cause processes the last reboot-cause based on the hardware reboot-cause
+and the software reboot-cause information and creates previous-reboot-cause.txt with the information.
+To log upto 10 entries of the previous reboot-cause, `process-reboot-cause` will save the previous reboot cause information
```
jleveque, line 17: s/upto/up to/

doc/system-telemetry/reboot-cause.md
```diff
@@ -11,78 +11,73 @@
...
+to "/host/reboot-cause/previous-reboot-cause/" with adding timestamp at the end of file name after processing
+the reboot-cause and create a symbolic link.
```
jleveque, line 19: Can you add more details about what this symbolic link is used for, its name, location, etc.?

jleveque (changes requested):

sujinmkang:
@jleveque Can you please review this HLD again?

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,116 @@
...
+        cause: "reboot",
+        user: "admin",
+        time: "Thu 03 Sep 2020 11:15:30 PM UTC",
+        comment: ""
+    }
+]
+```
+
+#### Part 2
+A new service named as `post-process-reboot-cause.service` which will retrieve the saved reboot-cause files and read each reboot-cause information from the files
```
jleveque, line 43: If we go with this approach, we might want to consider naming the first service "determine-reboot-cause" and the second service "process-reboot-cause", or similar. It's more meaningful than "post-process-reboot-cause".

doc/system-telemetry/reboot-cause.md
```diff
@@ -12,37 +12,43 @@
 ### Enable sonic streaming telemetry agent to send Reboot-cause information
 
 #### Part 1
-During the boot, the process-reboot-cause processes the last reboot-cause based on the hardware reboot-cause
+During the boot, the `determine-reboot-cause`(previously `process-reboot-cause`) determines the last reboot-cause based on the hardware reboot-cause
```
jleveque, line 15: Add the word, "service":

```
During the boot, the `determine-reboot-cause` service (previously `process-reboot-cause`) ...
```

doc/system-telemetry/reboot-cause.md
```diff
@@ -12,37 +12,43 @@
...
 and the software reboot-cause information and creates previous-reboot-cause.txt with the information as it does currently.
```
jleveque, line 16: Is the previous-reboot-cause.txt file still necessary? It seems redundant now that we will have multiple files in /host/reboot-cause/previous-reboot-cause/. I believe the only thing which reads from this file is the CLI, which will now simply reference the data in State DB.

doc/system-telemetry/reboot-cause.md
```diff
@@ -63,54 +69,85 @@ comment                 = STRING                         ; unstructured json for
 
 Currently `show reboot-cause` displays the last reboot-cause and performing `cat /host/reboot-cause/previous-reboot-cause.txt` to show the reboot-cause.
 This will be same as current design.
```
jleveque: As above, this seems redundant. `show reboot-cause` can now simply display the most recent reboot cause in State DB. Is there any good reason for retaining this?

jleveque (changes requested):

doc/system-telemetry/reboot-cause.md
sujinmkang: Currently process-reboot-cause is delayed to get the reboot-cause logged into the syslog server so I could see that the returned reboot-cause is wrong until the process-reboot-cause service is running.
The reason to keep and read the file for the last reboot-cause instead of db, the `show reboot-cause` will return the reboot-cause almost immediately after system boots. It's redundant but it can give the consistent output since boot. I was also considering to parse the other saved files to get the last reboot-cause. But I felt also that one additional file to maintain the last reboot-cause won't be that harm.

jleveque: It won't cause harm, but it adds complexity, and potential maintenance. I can understand wanting to make the reboot cause available via the CLI before the DB is populated, but is it necessary? Maybe the `show reboot-cause` command could simply return "Not available yet ... Please try again in a few minutes"? Otherwise, I would prefer the `show reboot-cause` command parses the most recent of the new saved files. This way, there is only one format to deal with.

Also, the `show reboot-cause history` command will also fail until the DB is populated, so it will have the same issue. How do you plan to deal with that?

doc/system-telemetry/reboot-cause.md
sujinmkang: @jleveque The original reason that we planed to have two services which determines the reboot cause and process the reboot cause is to determine the reboot cause as early as possible but the process of the root-cause (saving to db) could be delayed after db  is ready. If we just want to refer to the db for the cli, there is no reason to maintain two service, I think. What do you think?

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,150 @@
...
+## Scope
+-Enable sonic streaming telemetry agent to send Reboot-cause information
+
+### Enable sonic streaming telemetry agent to send Reboot-cause information
+
+#### Part 1
+During the boot, the `determine-reboot-cause` service ( previously `process-reboot-cause`) determines the last reboot-cause based on the hardware reboot-cause
+and the software reboot-cause information and `determine-reboot-cause` service will save the formatted last previous
+reboot cause information to "/host/reboot-cause/previous-reboot-cause/" with adding timestamp at the end of file name.
+`determine-reboot-cause` also will also create a symbolic link of the last reboot cause file to "/host/reboot-cause/previous-reboot-cause.txt"
```
jleveque, line 18: Do we need previous-reboot-cause.txt any longer, when we could just read the file with the most recent timestamp?

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,150 @@
...
+
+The example shows the previous reboot-cause files stored in /host/reboot-cause/previous-reboot-cause/.
+```
+admin@sonic:~$ ls /host/reboot-cause/previous-reboot-cause/
+previous-reboot-cause-2020_10_09_01_56_59.txt
+previous-reboot-cause-2020_10_09_02_00_53.txt
+previous-reboot-cause-2020_10_09_02_33_06.txt
+previous-reboot-cause-2020_10_09_04_53_58.txt
```
jleveque, line 26: Suggest removing the "previous-" prefix from these filenames. "previous-" was used to differentiate the previous reboot cause from the upcoming reboot, if it was software-initiated, but now we have a different directory structure.

jleveque (changes requested):

doc/system-telemetry/reboot-cause.md
sujinmkang: To make the `show reboot-cause`  simple, otherwise, we need to read file list in the directory and sort them before parsing the files for displaying the cli output. Before it was just one step, just "cat /host/reboot-cause/previous-reboot-cause.txt" and now it's two steps - parsing and composing as output format. If we just using the file reading from the most recent timestamp, it will be more than 4 steps - read filelist, sort them in timeorder, pase the file and formatting as output for each cli call.

doc/system-telemetry/reboot-cause.md
jleveque: Also, why do these files end in a `.txt` suffix, since they contain JSON, not plain text?

doc/system-telemetry/reboot-cause.md
jleveque: Understood. Then I'm OK with symlinking the latest file. However, this file, once symlinked, will no longer be a plain text file, so we should change the suffix.

Also, I suggest changing the naming of the dirs/files as such:

`/host/reboot-cause/previous-reboot-cause/` --> `/host/reboot-cause/history/`
`/host/reboot-cause/previous-reboot-cause.txt` --> `/host/reboot-cause/previous-reboot-cause.json` (or simply `/host/reboot-cause/previous.json`

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,150 @@
...
+## Scope
+-Enable sonic streaming telemetry agent to send Reboot-cause information
+
+### Enable sonic streaming telemetry agent to send Reboot-cause information
+
+#### Part 1
+During the boot, the `determine-reboot-cause` service ( previously `process-reboot-cause`) determines the last reboot-cause based on the hardware reboot-cause
+and the software reboot-cause information and `determine-reboot-cause` service will save the formatted last previous
+reboot cause information to "/host/reboot-cause/previous-reboot-cause/" with adding timestamp at the end of file name.
+`determine-reboot-cause` also will also create a symbolic link of the last reboot cause file to "/host/reboot-cause/last-reboot-cause.json"
```
jleveque, line 18: Prefer sticking with "previous-" rather than "last-", as it's a bit more specific, whereas "last" can have a few meanings.

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,150 @@
...
+
+The example shows the previous reboot-cause files stored in /host/reboot-cause/previous-reboot-cause/.
```
jleveque, line 20: Suggest renaming directory to `/host/reboot-cause/history/`

jleveque (changes requested):

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,150 @@
...
+Verify the information from state-DB data is available via the cli command `show reboot-cause history` which is extended from `show reboot-cause`.
+
+##### Reboot Cause Schema in state-DB
+
+Here is the definition of Reboot-cause schema which will be stored in state-DB.
+```
+; Defines information for reboot-cause
+key                     = REBOOT_CAUSE|<timestamp>         ; last reboot-cause processing time
+; field                 = value
+cause                   = STRING                         ; last reboot causek
```
jleveque, line 59: Typo: s/causek/cause/

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,150 @@
...
+        time: "Fri Oct  9 04:51:47 UTC 2020"
+        comment: ""
+    }
+]
+```
+
+#### Part 2
+A new service named as `process-reboot-cause.service` which will retrieve the saved reboot-cause files and read each reboot-cause information from the files
+and save the reboot-cause information up to 10 entries to state-DB.
+Verify the information from state-DB data is available via the cli command `show reboot-cause history` which is extended from `show reboot-cause`.
```
jleveque, line 50: Are we sticking with `show reboot-cause and `show reboot-cause history` or going with `show reboot cause` and `show reboot history`?

jleveque (changes requested):

doc/system-telemetry/reboot-cause.md
sujinmkang: `show reboot-cause` and `show reboot-cause history`

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,150 @@
...
+| 0.1 | 09/02/20 | Sujin Kang  | Initial version    |
+
+## Scope
+-Enable sonic streaming telemetry agent to send Reboot-cause information
+
+### Enable sonic streaming telemetry agent to send Reboot-cause information
+
+#### Part 1
+During the boot, the `determine-reboot-cause` service ( previously `process-reboot-cause`) determines the last reboot-cause based on the hardware reboot-cause
+and the software reboot-cause information and `determine-reboot-cause` service will save the formatted last previous
```
jleveque, line 16: s/formatted/JSON-formatted

jleveque (changes requested):

doc/system-telemetry/reboot-cause.md
```diff
@@ -0,0 +1,150 @@
+# Reboot-cause information via telemetry agent
+
+## Revision
+
+| Rev | Date     | Author      | Change Description |
+|:---:|:--------:|:-----------:|--------------------|
+| 0.1 | 09/02/20 | Sujin Kang  | Initial version    |
+
+## Scope
+-Enable sonic streaming telemetry agent to send Reboot-cause information
```
jleveque, line 10: The document describes how we will put reboot-cause information into State DB, but doesn't mention the telemetry agent. I think you should at least mention here that this document describes how we will store reboot-cause information in State DB so that it can be retrieved by the streaming telemetry agent.

jleveque (changes requested):

jleveque (approved):

PR #672, opened 2020-09-04
[synchronous-mode] Document for configuration of synchronous mode

shi-su:
Document for the configuration design of synchronous mode for review.


doc/synchronous-mode/synchronous-mode-cfg.md
```diff
@@ -0,0 +1,40 @@
...
+An illustrative figure of the synchronous mode is shown below. The synchronous mode achieves a closed-loop execution if SAI APIs from orchagent. In contrast to the previous asynchronous mode which cannot properly handle SAI API failures, the synchronous mode can gracefully handle SAI API failures by conducting the proper actions in orchagent. Therefore, the synchronous mode can substantially improve the reliability of SONiC.
+
+<img src="synchronous-mode-diagram.png" width="500">
+
+# 3. Configuration design 
+### 3.1 Configuration Requirements
+1.	Allow users to enable or disable synchronous mode via CLI.
+2.	Follow the image-specified mode when upgrading an image with cold/warm/fast-reboot if the synchronous mode configuration is not explicitly specified by users in configDB.
+3.	Use the user-specified mode when an explicit configuration of synchronous mode exists in configDB.
+4.	Support enabling/disabling the synchronous mode when deploying a minigraph.
```
qiluo-msft, line 14: >deploying a minigraph [](start = 56, length = 21)

parsing a minigraph and generating config_db.json

doc/synchronous-mode/synchronous-mode-cfg.md
```diff
@@ -0,0 +1,40 @@
...
+```
+DEVICE_METADATA|localhost
+    "synchronous_mode": "enable|disable"
+```
+There are three possibilities for the field with the behavior as follows:
+| Field value               | Behavior                          |
+|---------------------------|-----------------------------------|
+| enable                    | Enable synchronous mode.           |
+| disable                   | Disable synchronous mode.          |
+| Field does not exist      | Orchagent and syncd use the mode specified by `orchagent.sh` and `syncd-init-common.sh`, respectively. |
```
qiluo-msft, line 27: >Field does not exist [](start = 2, length = 20)

Let's be verbose. If the field does not exist, and user never use the CLI, will it appear in ConfigDB or not at all?

qiluo-msft, line 27: >specified by `orchagent.sh` and `syncd-init-common.sh` [](start = 63, length = 54)

The design doc have some details not fixed:
1. if Field does not exist, what is the behavior? Will new image change the behavior?
2. `orchagent.sh` and `syncd-init-common.sh` will use the same mode, so better to define the behavior in one place.

doc/synchronous-mode/synchronous-mode-cfg.md
```diff
@@ -0,0 +1,40 @@
...
+# 2. A brief of synchronous mode
+An illustrative figure of the synchronous mode is shown below. The synchronous mode achieves a closed-loop execution if SAI APIs from orchagent. In contrast to the previous asynchronous mode which cannot properly handle SAI API failures, the synchronous mode can gracefully handle SAI API failures by conducting the proper actions in orchagent. Therefore, the synchronous mode can substantially improve the reliability of SONiC.
+
+<img src="synchronous-mode-diagram.png" width="500">
+
+# 3. Configuration design 
+### 3.1 Configuration Requirements
+1.	Allow users to enable or disable synchronous mode via CLI.
+2.	Follow the image-specified mode when upgrading an image with cold/warm/fast-reboot if the synchronous mode configuration is not explicitly specified by users in configDB.
+3.	Use the user-specified mode when an explicit configuration of synchronous mode exists in configDB.
```
qiluo-msft, line 13: You need to provide design detail on how to achieve this.

doc/synchronous-mode/synchronous-mode-cfg.md
qiluo-msft, line 14: >. [](start = 77, length = 1)

Do you need design for [https://github.com/Azure/SONiC/wiki/L2-Switch-mode](https://github.com/Azure/SONiC/wiki/L2-Switch-mode)?

qiluo-msft (changes requested):
As comment

doc/synchronous-mode/synchronous-mode-cfg.md
shi-su: Fixed.

doc/synchronous-mode/synchronous-mode-cfg.md
shi-su: Added a description for this scenario.

shi-su: Added discussion in section 3.4 to clarify that new images can change the behavior if the field does not exist.

shi-su: As for 2, both `orchagent.sh` and `syncd-init-common.sh` follow the configuration in config_DB. The reason to make the field in config_DB empty as default and let the two files to choose their own mode with the field empty is to make sure the image can change the behavior accordingly with cold/warm/fast reboot if a user-specified mode does not exist. It is kind of like trading off this for better controllability in image upgrade. Since the mode defined by the two files are expected to change only once from async mode to sync mode, I think doing such a tradeoff should be ok.

doc/synchronous-mode/synchronous-mode-cfg.md
shi-su: Added design details about where the configuration in configDB is read and applied.

doc/synchronous-mode/synchronous-mode-cfg.md
shi-su: I think this L2 switch mode does not need specific designs for it. I added Section 3.6 to discuss about the L2 switch mode.

qiluo-msft (approved):

PR #680, opened 2020-09-21
Kube systemd

renukamanavalan:
Extend SONiC dockers management to Kubernetes.


GarrickHe:
Hello @renukamanavalan 
Got a few questions:

1. Will /swap be disabled when running in kube-mode as per k8s recommendation? If not, are we bypassing some k8s pre-flight checks?
2. Are we using the official `kubeadm` to register nodes (SONiC switches) into the cluster? Or we're doing this through some other mechanism? I see the command `sudo config kube server IP <address>`, will this wrap around `kubeadm` since that's the official way Kubernetes recommends adding nodes?
3. Will all SONiC containers live in one Pod? Or will SONiC containers follow some k8 pod design pattern when operating under kube-mode?
4. What CNI will the community use for master to node connectivity? Or is this up to the user who decides to use kube-mode?
3. Diving down to some specifics, what liveness/readiness probes will be used for SONiC Pod(s) in the manifest?

Thank you.

renukamanavalan:
> Hello @renukamanavalan
> Got a few questions:
> 
> 1. Are we using the official `kubeadm` to register nodes (SONiC switches) into the cluster? Or we're doing this through some other mechanism? I see the command `sudo config kube server IP <address>`, will this wrap around `kubeadm` since that's the official way Kubernetes recommends adding nodes?
> 2. Will all SONiC containers live in one Pod? Or will SONiC containers follow some k8 pod design pattern when operating under kube-mode? May all SONiC related pods can exist in their own namespace?
> 3. Diving down to some specifics, what liveness/readiness probes will be used for SONiC Pod(s) in the manifest?
> 4. Will a CNI be used for Pod connectivity? If yes, is there one Azure will officially use? Or is this up to the user to decide?
> 5. What version of k8s will SONiC use?
> 
> Thank you.


1) Yes, we use official kubeadm. The hostcfgd  would watch config-db and auto-join/reset using kubeadm
2) One container per pod. 
    <copy/paste from doc> _* In SONiC switches, a pod runs a single container only.*_
3) None planned. As we could be switching between local & kube, we rely on SONiC switch's inherent monitoring. But we may use some, if kube master would require, for its managing purposes. For example, Safety Check Service
4) No. There is no inter-pod communication required in SONiC usage scenarip
5) Currently we are testing on 1.18.6.

GarrickHe:
@renukamanavalan - Are we going to have SONiC-related pods in the default namespace, have a SONiC-related namespace (ie: sonic-systems), or leave this up to the user to specific in their deployment? There is a good chance that the k8 master will be orchestrating other non-SONiC pods, so from a organization perspective it will be good to have our own. I believe you can use namespaces to only allow SONiC-related containers to be deployed on SONiC-related nodes (SONiC switches).

renukamanavalan:
> @renukamanavalan - Are we going to have SONiC-related pods in the default namespace, have a SONiC-related namespace (ie: sonic-systems), or leave this up to the user to specific in their deployment? There is a good chance that the k8 master will be orchestrating other non-SONiC pods, so from a organization perspective it will be good to have our own. I believe you can use namespaces to only allow SONiC-related containers to be deployed on SONiC-related nodes (SONiC switches).

Good point. This is outside the scope of this doc. Said that, this more a requirement from the external deployment perspective.

GarrickHe:
> > @renukamanavalan - Are we going to have SONiC-related pods in the default namespace, have a SONiC-related namespace (ie: sonic-systems), or leave this up to the user to specific in their deployment? There is a good chance that the k8 master will be orchestrating other non-SONiC pods, so from a organization perspective it will be good to have our own. I believe you can use namespaces to only allow SONiC-related containers to be deployed on SONiC-related nodes (SONiC switches).
> 
> Good point. This is outside the scope of this doc. Said that, this more a requirement from the external deployment perspective.

Sure, no problem. Just something to keep in mind. I haven't given it much thought but it would be nice to explore what the implications are (if any) if a user puts all SONiC-related pods into it's own namespace since that is essentially creating a separate virtual-cluster.

MalteJ:
There is a typo in the filename: `kuberenetes-support.md`

renukamanavalan:
> There is a typo in the filename: `kuberenetes-support.md`

Thanks. Fixed it.

PR #681, opened 2020-09-30
SONiC TPID Configuration Support

gechiang:
This is the HLD proposal for the implementation of TPID configuration in SONiC.

Please also refer to the following SAI Proposal that is depended by this TPID configuration feature:
<https://github.com/opencomputeproject/SAI/pull/1089>


doc/tpid/SonicTPIDSettingHLD1.md
```diff
@@ -0,0 +1,386 @@
...
+admin@SONiC:~$ sudo config interface tpid PortChannel0002 0x9200
+HW is not capable to support PortChannel TPID config.
+```
+
+In case a port is already configured as part of a LAG, TPID config directly to the LAG member port is not allowed and will fail as following:
+```
+admin@SONiC:~$ sudo config interface tpid Ethernet4 0x9200
+Ethernet4 is already member of PortChannel0002. Set TPID NOT allowed.
+```
+
```
gechiang, line 145: During the review there was one question raised by Ravi from Innovium in regard to the handling of adding a port to LAG when the port already has some non-default TPID and how should that be handled.
Here are 2 options:
1. After the port passes all current LAG membership add pre-check (if this is present today) we shall allow the add to proceed by first implicitly modify the port's TPID value to 0x8100 right before the Add LAG member continue to execute.
2. Reject the request due to Port TPID is NOT set at "default TPID" value. This forces the user to have to explicitly restore the port's TPID to 8100 first.  After the TPID is restored to default, the user can retry the add LAG member again and it should now able to proceed if TPID was the only blocking constraint check.

Option 1 has a potential race condition issue where if the implicit set Port TPID to default arrived later than the Add member to LAG.  So the code handling need to be aware of this race condition and only update the port's config DB, App DB to the default TPID value only and NOT program this TPID with SAI through syncd when such race condition is detected.

option 2 is the safest thing to do as it also draws attention to the user that issued the command.  Perhaps it was a typo where user specified the wrong interface... OR the error generated by this option can serves as a reminder to the user that the port being requested to add to the port channel may had some other usage intended.  It gives the user a chance to ensure it is desirable configuration that the user wants to perform.   The draw back of option 2 is that it prevented the user who know what he/she is doing and just want to quickly add the port to port-channel ignoring what its previous TPID maybe...

Given the two options above, I think we should go with option 2 for now...  If anyone else has comments over this, please raise a comment.
Thanks!

doc/tpid/SonicTPIDSettingHLD1.md
```diff
@@ -0,0 +1,386 @@
...
+To simplify the handling of TPID setting to avoid complex TPID usage counters and validation checking, we have simplified it by restricting
+the configurable values to the commonly used TPID values: 0x8100 (default), 0x9100, 0x9200, and 0x88A8. If user attempt to configure any
```
msreddypaluru, line 147: How does an application know what are the supported TPID's by ASIC?  Is there any SAI support to pull these capabilities? Should it be part of switch capabilities in addition to TPID port & port channel capability?

gechiang: Please refer to the SAI proposal document for details.
you can find it here:
     https://github.com/opencomputeproject/SAI/pull/1089
In summary: The SAI proposal details that the SAI HW vendor when replying to the sai_query_attribute_capability() API that indicate it is capable of supporting the TPID attribute on PORT on or LAG means that it is able to support TPID values of 0x8100, 0x9100, 0x9200, and 0x88A8. 
If the SAI  HW vendor is not able to support these, then it should indicate it is NOT capable to handle this attribute to the Port/LAG object.
Hope this answered your question.

msreddypaluru: sounds good.

doc/tpid/SonicTPIDSettingHLD1.md
```diff
@@ -0,0 +1,386 @@
...
+
+Currently SONiC Testbed setup uses fanout switches to connect to the testbed server and running the PTF test cases. The fanout switch ports
+connecting to the DUT are configured as access ports. But the PTF test cases also need to test tagged packet functionality. In order to make
+this work, the fanout switch is configured with 802.1Q tunnel on the DUT facing ports side so that tagged packets are not to be used for service
+delimiting purpose. Having the capability for SONiC to configure the port's TPID value will allow SONiC to handle tagged packets in the same
+way. With this TPID configuration capability, these fanout switches can be converted to run in SONiC.
+
+![FanoutSwitch Pkt Flow](FanoutPktFlow.png)
+
+**Linux Kernel TPID Support**
```
sandeep-kulambi, line 63: Hi Gen-Hwa,

In case user configures a non default TPID value (say 0x9100) in such a case what happens to the control packets in the kernel will they be accepted or dropped (ex STP, ARP, ICMP etc). If they are dropped it will break the functionality of these features right ?

Thanks

doc/tpid/SonicTPIDSettingHLD1.md
rck-innovium: Agree that option#2 is safer. 
And some more things to consider for these checks:  SAI allows a platform to specify the capability independently for a physical port and PortChannel. So we need to handle the theoretical possibility where the TPID is supported on the port and not on the PortChannel. And vice-versa.

doc/tpid/SonicTPIDSettingHLD1.md
gechiang: @sandeep-kulambi ,
This HLD was to address the specific use case mentioned in the fanout switch scenario.  It also serves as a starting point for support of 802.q tunneling.  STP BPDU is untagged so it should not matter which TPID is configured.  But if you need to run PVST which needs to be tagged then you will need to re-evaluate why there is a need to change TPID for this port as the kernel does not support non-default TPID values.
Hope this clarify the expectation of this HLD.

doc/tpid/SonicTPIDSettingHLD1.md
```diff
@@ -0,0 +1,386 @@
...
+
+**Default TPID**
+----------------
+
+SONiC default TPID is 0x8100. There is no need to explicitly configure the TPID of each interface to the default TPID value unless a different
+value is desired. Once configured with a different value and the configuration is saved, the TPID value configured will persists across reboot.
+
+**SAI TPID setting Capability**
+-------------------------------
+
```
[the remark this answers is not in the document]
gechiang: @rck-innovium
The TPID setting is always applied on the OUTER TPID.  It does not change the inner TPID specification.
I believe some SAI vendors do allow setting of inner TPID but this is rarely used by most applications and is usually default to 0x8100. 
The TPID setting in the SAI proposal as well as in this HLD is applicable at PORT/LAG level. It is not at Switch level.  So having this configured at PORT/LAG level, it overrides what is set at switch level for that specified PORT/LAG.
Hope this help clarify your question.

doc/tpid/SonicTPIDSettingHLD1.md
gechiang: @rck-innovium
You are correct that the capability are independent and this HLD does not assume that the HW vendor is able to support both.  That is why in the State DB there are 2 separate attributes indicating whether TPID can be configured at PORT level and/or at LAG level.  
The requirement for support at LAG level actually placed more responsibility on the HW SAI vendor to be able to apply the TPID on each new joining member as well as reset to default TPID when a port is deleted from the LAG to become an individual port.   So my expectation is most vendor should have no problem to support at PORT level at first and later to support LAG but NOT vice versa as it make no sense to be able to support LAG but not able to support at port level... 
Rest assure this HLD when implemented will not make any assumptions for this.

doc/tpid/SonicTPIDSettingHLD1.md
sandeep-kulambi: Thanks Gen-Hwa !!

rlhui (approved):

PR #682, opened 2020-10-01
SONiC Application Extension HLD

stepanblyschak:


mikelazar:
A couple of comments:
- this approach assumes that extensions need to follow the SONiC convention (use supervisord to start processes, use a per container rsyslogd to log, use systemd unit files to start containers at system). There are useful situations/use cases where this is not true - for instance when SONiC switches are used as "compute nodes" for applications designed to work in a generic container environment. In that case, using supervisord and rsyslog is not necessarily the case - there can be one microservice / container and no rsyslog (such applications use docker log drivers to log, for instance). Furthermore, such applications are designed to benefit from the automatic restart options of docker (e.g. --restart=unless-stopped) rather than the heavier systemd unit files. In other words, this approach forces users of SONiC to design their applications to be very SONiC centric, rather than generic.
- it will also be beneficial to allow the life cycle management (install and upgrade) of extension applications to be independent of SONiC. With the current approach, the assumption is that all extensions need to be reinstalled after SONiC is upgrade - or that they be included in the SONiC installation/upgrade.

doc/sonic-application-extention/sonic-application-extention-hld.md
```diff
@@ -0,0 +1,1441 @@
...
+Thus, for such Docker images the limitations are:
+- can be locally managed only, as the requirement for kube managed features is not met.
+
+Another possibility is to allow users to provide a manifest file URL.
+
+An example of the flow for a collectd Docker image:
+```
+admin@sonic:~$ sudo sonic-package-manager repository add collectd puckel/docker-collectd
+admin@sonic:~$ sudo sonic-package-manager install collectd --manifest https://some-server/manifests/collectd/manifest.json
+```
```
venkatmahalingam, line 1312: Does the user need to update package.json for collected (3rd party docker) or it would be taken care implicitly by the sonic package manger?

venkatmahalingam: If the user wants to restrict the 3rd party docker images or non-SONiC built-in docker images to use only some % of the device storage, any option available?

doc/sonic-application-extention/sonic-application-extention-hld.md
```diff
@@ -0,0 +1,1441 @@
...
+It is possible to install 3rd party Docker images if the installer will use a default manifest. This default manifest will have only mandatory fields
+and an auto-generated service will require only "docker" service to be started. The name of the service is derived from the name of the Package in *packages.json*.
+
+Default manifest:
+```json
+{
+  "service": {
+    "name": "<package-name>",
+    "requires": "docker",
+    "after": "docker"
```
venkatmahalingam, line 1293: If the user wants to make use of the docker options say. cpu cores, memory..for a given docker, do we have any option in the manifest file?

doc/sonic-application-extention/sonic-application-extention-hld.md
```diff
@@ -0,0 +1,1441 @@
...
+  * Buf fixes
+
+-- Stepan Blyshchak <stepanb@nvidia.com>  Fri, 23 Oct 2020 12:26:08 +0300
+
+```
+
+### SONiC Docker Container Resource Restrictions
+
+This feature will allow user to specify resource restrictions for a container via FEATURE table in CONFIG DB.
+This feature is not related to SONiC Application Extension Design, as can be applied to any existing SONiC
```
venkatmahalingam, line 795: How is this FEATURE table is populated? based on manifest file? is this applicable for 3rd party docker as well?

doc/sonic-application-extention/sonic-application-extention-hld.md
```diff
@@ -0,0 +1,1441 @@
...
+
+```
+docker create {{ docker_run_options }}
+  --net=$NET \
+  --uts=host \
+  --log-opt max-size=2M --log-opt max-file=5 \
+  -v /var/run/redis$DEV:/var/run/redis:rw \
+  $REDIS_MNT \
+  -v /usr/share/sonic/device/$PLATFORM:/usr/share/sonic/platform:ro \
+  -v /usr/share/sonic/device/$PLATFORM/$HWSKU/$DEV:/usr/share/sonic/hwsku:ro \
```
venkatmahalingam, line 938: We may not need these mount volumes for 3rd party dockers as they may not depend on SONiC resources? can we make it an optional field (e.g SONiC dependent docker vs non-dependent docker) in the manifest file?

doc/sonic-application-extention/sonic-application-extention-hld.md
stepanblyschak: > Does the user need to update package.json for collected (3rd party docker) or it would be taken care implicitly by the sonic package manager?

User normally should not touch packages.json, this is internal to sonic-package-manager, so it is taken care implicitly

stepanblyschak: > If the user wants to restrict the 3rd party docker images or non-SONiC built-in docker images to use only some % of the device storage, any option available?

This is a functionality that is out of scope in this document. The reason is that it is not really a feature that can be applied for non SONiC dockers or Application Extension Dockers. Ideally, we might need to plan for general support for resource constraints for all dockers that can run in the system, something like:

```
config feature storage-quota swss 50%
```

doc/sonic-application-extention/sonic-application-extention-hld.md
stepanblyschak: This is similar to the question with disk space limit. Having these constraints in manifest might not be a good idea because one switch may have powerful enough CPU and a lots of RAM while another switch platform might be much slower.
You may want to tweak this parameters differently for different platforms, but you can not put these parameters in manifest for every possible platform.
The plan is to have an option for the user to configure resource limits and to have something similar to above:
```
config feature cpu-limit
config feature memory-limit
etc.
```
But this is out of scope for this document.

doc/sonic-application-extention/sonic-application-extention-hld.md
stepanblyschak: This is out of scope of this document. Please see above replies

doc/sonic-application-extention/sonic-application-extention-hld.md
stepanblyschak: This is mentioned in https://github.com/Azure/SONiC/pull/682/files#diff-be6a02b89b88eb9dd23104151d3b06081dbc49c58c610f5f2aa5dfc5d1d4e13aR1298.

Normally, if you don't need these volume mounts you will not specify "depends": ["database"], or any other sonic container that has a dependency on database. In this case sonic-package-manager will know that these mounts are not needed.

Please, note, 3rd party dockers are not part of phase 1 implementation

venkatmahalingam: ok, thanks.

doc/sonic-application-extention/sonic-versioning-strategy.md
```diff
@@ -0,0 +1,117 @@
...
+<img src="img/versioning-strategy.svg" alt="Figure 1. SONiC Docker Images Versioning Strategy">
+</p>
+
+***package maintainer can*** update *default-reference* in package.json in SONiC buildimage to point to a default version which will be used when user installs a package.
+
+## Base OS versioning
+
+## Base OS API that a package uses
+
+- SONiC utilities; Until CLI autogeneration, sonic utilities plugin system API incompatibility must be recorded by base OS major version increment
```
bandaru-viswanath, line 101: I think we should de-link the CLI autogeneration from this.  If we think the sonic-utilities is part of the base-os, we should list here. Alternately, we can let other individual dockers to register their dependencies on sonic-utilities.

CLI is just one form of management, and even for CLI there are multiple models (linux-like,  cisco-like etc). The base-os should not have any awareness of the CLI, or any UI.

So, I suggest we look at  whether we need sonic-util as part of base-os, independent of a specific CLI model

doc/sonic-application-extention/sonic-versioning-strategy.md
```diff
@@ -0,0 +1,117 @@
...
+feat: Introduce new methods in ConsumerTable
+
+BREAKING CHANGE: this feature breaks the Consumer/Producer based IPC
+```
+
+## SONiC Packages Versioning
+
+- Package release ***happens on SONiC release/branch-out*** or on demand for single Docker image: ***sub-release***
+- Manual version update ***is required*** when SONiC Package releases
+- Within a release major and minor version ***must*** not change
```
bandaru-viswanath, line 79: Just to aid clarity - Within a SONIC release, the major and minor versions of the feature package ***must*** not change, i.e., any newer version of the feature package for that SONIC release, remains fully backward compatible.

stepanblyschak: Yes, only bug fixes are allowed as usually for release branches

liat-grozovik:
@renukamanavalan I believe it is time to merge this PR as the code is almost done and merged. Any objection?

PR #697, opened 2020-11-03
[VOQ] Distributed VOQ LAG HLD

skeesara-nokia:


doc/voq/lag_hld.md
```diff
@@ -0,0 +1,427 @@
...
+
+## 2.2 LAG Requirements for VOQ SAI
+The following rules apply with regard to programming LAG information via SAI on a distributed VOQ system
+*  Every LAG needs to be created in SAI of all of the asics in the system. This is irrespective of which asic the member ports of a LAG "belong" to.
+*  The active member port list for each LAG should be the same in the SAI instance of every asic. Any update to the member port list of any LAG must be propagated to all of the SAI instances.
+*  The chnages made to SAI for VOQ support allow the member port list for a LAG to be specified as a list of system ports
+*  SAI allows the application layer to specify a "LAG_ID" (SAI_LAG_ATTR_SYSTEM_PORT_AGGREGATE_ID) as part of LAG creation. For a given LAG - the same value must be used on all the SAI instances.
```
vganesan-nokia, line 65: Mention dynamic lag member add/remove at runtime

doc/voq/lag_hld.md
```diff
@@ -0,0 +1,427 @@
...
+|          SYSTEM LAG ID            |
++--------+--------+--------+--------+
+| Switch |          LAG ID          |
+|   ID   |                          |
++--------+--------+--------+--------+
+
+```
+
+Changes in orchagent/portsorch include: 
+ * Assign unique system_lag_id across system by using switch_id. The encoding scheme is described as above.
```
vganesan-nokia, line 110: System lag id has a range restriction (1 to max lags supported). Having switch id encoded in the lag id will make the id bigger. Consider make this id be managed by configuration to make it unique. Revisit after checking with hardware restrictions

doc/voq/lag_hld.md
```diff
@@ -0,0 +1,427 @@
...
+
+#### teamd: teamsyncd
+
+teamsyncd will act as if its single asic sonic system and on local ports LAG management. So, there is no changes required here. 
+
+#### orchagent: portsorch
+**Local LAG**: The system_lag_id is a locally generated system wide unqiue id for each LAG. After creating local LAG, the PortChannel information is synced to SYSTEM_LAG_TABLE in centralized database CHASSIS_APP_DB. The lag members of the local LAG use local port ids. Any member addition/removal members to a local LAG is synced to SYSTEM_LAG_MEMBER_TABLE in centralized database CHASSIS_DB. While syncing lag members, unique system lag name and system port alias of the local port members are used as the key. Unique system lag name is generated using local Portchannel name and switch id in the format shown below
+```
+Switch<local switch_id>-<local PortChannel name>
+```
```
vganesan-nokia, line 92: Make the PortChannel name globally unique across the chassis in the local asic configuration itself. We do not need this encoding while syncing to CHASSIS_APP_DB

doc/voq/lag_hld.md
vganesan-nokia, line 110: When the lag id is configured, make a validation system at the configuration to make sure that it is unique across system. Proposal: take the chassis app db system lag info and populate the local state db and check with this local state db lag id info during configuration.

doc/voq/lag_hld.md
vganesan-nokia: Fixed

doc/voq/lag_hld.md
vganesan-nokia: Instead of configuring system_lag_id, acquire it from global redis (with atomic acition insured) so that all asics will get unique available id. By this we can avoid availability validation and race conditions.

doc/voq/lag_hld.md
```diff
@@ -0,0 +1,561 @@
...
+   Slot1|Asic0|PortChannel001
+```
+
+The lag members of the local LAG use local port ids. Any member addition/removal members to a local LAG is written to the SYSTEM_LAG_MEMBER_TABLE in centralized database CHASSIS_DB. The system port alias of the lag member is used when to represent a lag member in the SYSTEM_LAG_MEMBER_TABLE. 
+
+**Remote LAG**: To create LAG corresponding to a LAG on another asic, the system_lag_id will be from corresponding entry in the SYSTEM_LAG_TABLE from CHASSIS_APP_DB. The Lag members for a remote LAG will be specified using the OID of the corresponding system port objects.
+
+Changes in orchagent/portsorch include: 
+ * Port structure enhancement to store the system lag info such as system lag name (alias), system lag id and switch_id.
+ * Getting a chassis wide unique system_lag_id from centralized database using lua script that can do atomic allocation/deallocation of system_lag_id globally
```
eswaranb, line 115: Minor: This reads as if the lua script is running inside orchagent/portsorch.

doc/voq/lag_hld.md
```diff
@@ -0,0 +1,561 @@
...
+```
+
+**Schema:**
+
+```
+; Defines schema for SYSTEM_LAG_TABLE table attributes
+key                 = SYSTEM_LAG_TABLE|system lag name   ; System LAG name
+; field             = value
+system_lag_id       = 1*10DIGIT                          ; LAG id.
+switch_id           = 1*4DIGIT                           ; Switch id
```
eswaranb, line 250: Curious where switch_id will be used?

eswaranb (approved):

doc/voq/lag_hld.md
```diff
@@ -0,0 +1,565 @@
...
+#### teamd: teamsyncd
+
+* teamsyncd only deals LAG on the port of its local asic instance. So no changes are required in teamsyncd. 
+
+#### orchagent: portsorch
+**Local LAG**: portsorch obtains a system wide unique system_lag_id from centralized database CHASSIS_APP_DB. portsorch sends this system_lag_id in SAI_LAG_ATTR_SYSTEM_PORT_AGGREGATE_ID attribute to SAI while creating LAG in SAI. After creating local LAG, the PortChannel with switch_id and system_lag_id is written to the SYSTEM_LAG_TABLE in centralized database CHASSIS_APP_DB. A unique name is assigned automatically to each LAG in the system lag table in the CHASSIS_APP_DB using the following naming scheme (similar to the nameing scheme used for naming the system ports).
+```
+<Host name>|<Asic name>|<port channel name in its local asic instance>
+```
+The _Host name_ and _Asic name_ are available in CONFIG_DB device meta data attributes **hostnmae** and **asic_name** respectively.
```
mlorrillere, line 100: `asic_name` is only available on multi-asic devices. On single-asic devices the asic can be identified by the `hostname` only.

Do you plan to make `asic_name` available on single-asic devices as well? Or can a lag name simply omit <Asic name> for single-asic devices, i.e. `Slot1|PortChannel001`?

vganesan-nokia: As per current system lag implementation, if the switch type is "voq", "asic_name" is mandatory in DEVICE_METADATA['localhost'] and it should have  a valid non-zero sized value. So even for the single asic linecards, "asic_name" must be provided with a valid value.

mlorrillere: I see that it became mandatory when #1605 merged (https://github.com/Azure/sonic-swss/pull/1605/files#diff-3cba33a216c0ea2b942dc03b2638fa27a40a8b6eb81abf9f83f13e1ee7831ac4R231-R237), i.e.:
```cpp
    if (!cfgDeviceMetaDataTable.hget("localhost", "asic_name", value))
    {
        // asic_name is not configured.
        SWSS_LOG_ERROR("Asic name is not configured");
        return false;
    }
```

However I'm a bit surprised as this basically prevents system ports to be configured without `asic_name`. This doesn't seem necessary.

What is the reason to make `asic_name` mandatory even on single-asic devices?

vganesan-nokia: I agree. "asic_name" for single asic seems unnecessary. 
However, there is a reason why we need asic_name as mandatory in the system lag name. We use "|" as separator between the hostname, asic_name and port name. In the Distributed VOQ LAG HLD design discussion meeting it was told that in  the future all tables will have "|" as the table separator. So for the system lag member table the key will look something like 
**"Linecard1|Asic0|PortChannel001|Linecard1|Asic0|Ethernet1"**. This is **\<derived system lag name\>|\<system port name\>**. In the lag member processing, to extract the port channel alias and member system port name alias from the key, it was suggested to use the fixed format of system lag and system port names. So we do not want two different formats.
Please note that for now, the separator is ":" and porchannel alias and system port member alias extraction is based on this. But in the future this may not be the case.

mlorrillere: Ok. Then I'll make sure that `asic_name` is set for VoQ regardless of of the number of asics in sonic-buildimage [PR5997](https://github.com/Azure/sonic-buildimage/pull/5991).

eswaranb: I assume this is for single-chip linecards that are in a chassis with other chips. I don't think it's unreasonable to expect table entries are qualified with 'slotname|chipname' always. It's a little clunky for single-chip linecards, but not terrible and if it helps simplify code, we can do that. 

For single-chip fixed systems that are not a in chassis, we shouldn't set the switch_type to 'voq' anyway.

rlhui (approved):

PR #699, opened 2020-11-04
Generalizing config.bcm support for BRCM silicons

geans-pin:
HLD for per-switching silicon Common config for Broadcom Supported Platforms design


lguohan:
@gechiang , can you take a look?

doc/platform/common_config.md
```diff
@@ -0,0 +1,107 @@
...
+
+## 4 Serviceability and DEBUG
+### 4.1 Syslogs
+During system booting syncd initialization, the Per-switching silicon common config design will log the common config merged information to the syslogs for identifying which common config be merged to the ODM config.bcm for syncd initialization.
+
+### Examples:
+```
+Nov  3 09:19:25.373964 2020 sonic INFO syncd#supervisord: syncd Merging 
+/usr/share/sonic/hwsku/td3-ix8a-bwde-48x25G+8x100G.config.bcm with
+/usr/share/sonic/device/x86_64-broadcom_common/x86_64-broadcom_b77
```
gechiang, line 75: What is the difference between the b77 and b87 version of the common td3 config.bcm?  from chip_id you can only tell if it is TD3/TH/TD2/... etc but how can you use that to select which one of the TD3 to use?  Can you elaborate on this?

doc/platform/common_config.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+## 5 Scalability
+NA
+## 6 Unit Test
+- Check the system status to make sure syncd initial success
+- Check the syslog to make sure the common config merged to
+   config.bcm successfully
+- Check the final merged config.bcm dump from show tech dump 
+
```
gechiang, line 106: Do you have plans to extract those "common" soc properties from each of the platform specific config.bcm file into the common config.bcm file?

doc/platform/common_config.md
geans-pin: 1. The b77 and b87 is different TD3 model which is the upper 3 nibbles of the switching silicon's device ID.

2. We use the CHIP ID to select the common config folder. FYI, see the mention in HLD :
In the config_syncd_bcm() function, it will load the common silicon config.bcm file by looking for a *.bcm file in the  /usr/share/sonic/device/x86_64-broadcom_common/broadcom-sonic-{$chip_id} folder where the $chip_id is the upper 3 nibbles of the switching silicon's device ID as listed in SDK/include/soc/devids.h.

doc/platform/common_config.md
geans-pin: No, we don't have the plans to merge anything common from platform specific config.bcm to common config.bcm.
It will break our rule. In the design, when we find the same common config property existing on both platform specific and common specific. We will keep the original config in platform specific without merging it.

rlhui (commented):
if a platform have multiple brcm asics, is it handled?

gechiang:
At the community review session we raised a few questions which require you to update with clarifications.  Here are the set of items I noted down:

1. Can you add a section to discuss what is the expectation in regard to WARM Reboot?  Does this new feature has any impact during WARM boot where the new image being booted to contains new changes in the common portion?
2. Please clearly document if there are specific properties that are part of the common SOC properties and the ODM should never touch those properties in the ODM config.bcm file so that ODM should never attempt to overwrite them based on your new logic which values ODM being higher priority.  these set of  "common" SOC properties are something that ODM should never attempt to modify...
3. How do you introduce any "new" common SOC properties and the definition/purpose of each SOC property clearly so the  community become aware of those new SOC properties.  Clearly mark/indicate whether the newly introduced  SOC property is "mandatory" or can be overwritten by ODM if they do not wish to accept such behavior.?

doc/platform/common_config.md
```diff
@@ -0,0 +1,107 @@
...
+Per-switching silicon Common config feature is supported on all of the Broadcom platform if the common config files are created in the device/broadcom/x86_64-broadcom_common. Following is the current supported common configuration :
+
+```
+|-- x86_64-broadcom_b77 -- broadcom-sonic-td3.config.bcm
+|-- x86_64-broadcom_b85 -- broadcom-sonic-td2.config.bcm
+|-- x86_64-broadcom_b87-- broadcom-sonic-td3.config.bcm
+|-- x86_64-broadcom_b96-- broadcom-sonic-th.config.bcm
+|-- x86_64-broadcom_b97-- broadcom-sonic-th2.config.bcm
+|-- x86_64-broadcom_b98-- broadcom-sonic-th3.config.bcm
```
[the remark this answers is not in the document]
geans-pin: The common config of JR2 is not created so far. The framework will not handle this.

geans-pin:
My Reply inline:

> At the community review session we raised a few questions which require you to update with clarifications. Here are the set of items I noted down:
> 
> 1. Can you add a section to discuss what is the expectation in regard to WARM Reboot?  Does this new feature has any impact during WARM boot where the new image being booted to contains new changes in the common portion?

[GR] We will update the new section on the enhancement PR after this base line PR merged.
 
> 2. Please clearly document if there are specific "Mandatory" properties that are part of the common SOC properties and the ODM should never touch those properties in the ODM config.bcm file.  Based on your new logic which places ODM's bcm.config SOC properties being higher priority, we need clear indication to these set of  "Mandatory common" SOC properties that ODM should never attempt to modify...

[GS] We will update the Documents on the enhancement PR after this base line PR merged. 

> 3. How do you introduce any "new" common SOC properties and the definition/purpose of each SOC property clearly so the  community become aware of those new SOC properties.  Clearly mark/indicate whether the newly introduced  SOC property is "mandatory" or can be overwritten by ODM if they do not wish to accept such behavior.?

[GS] By default the common config feature is disabled, and ODM can enable this feature by touch the 
        common_config_support file in the $PLATFORM_DIR

PR #712, opened 2020-11-25
Next hop group split HLD

TACappleman:


doc/ip/next_hop_group_hld.md
```diff
@@ -0,0 +1,148 @@
...
+    ifname        = *PORT_TABLE.key,   ; zero or more separated by “,” (zero indicates no interface)
+    blackhole     = BIT ; Set to 1 if this route is a blackhole (or null0)
+ ```
+
+This design adds a new NEXT_HOP_GROUP_TABLE, to store next hop group information to be used by one or more routes.
+```
+### NEXT_HOP_GROUP_TABLE
+    ;Stores a list of groups of one or more next hops
+    ;Status: Mandatory
+    key           = NEXT_HOP_GROUP_TABLE:string ; arbitrary identifier for the next hop group
```
qiluo-msft, line 84: >arbitrary identifier [](start = 50, length = 20)

"arbitrary identifier" is not well defined.

doc/ip/next_hop_group_hld.md
```diff
@@ -0,0 +1,148 @@
...
+    ;Status: Mandatory
+    key           = ROUTE_TABLE:prefix
+    nexthop       = *prefix, ;IP addresses separated “,” (empty indicates no gateway)
+    ifname        = *PORT_TABLE.key,   ; zero or more separated by “,” (zero indicates no interface)
+    blackhole     = BIT ; Set to 1 if this route is a blackhole (or null0)
+ ```
+
+This design adds a new NEXT_HOP_GROUP_TABLE, to store next hop group information to be used by one or more routes.
+```
+### NEXT_HOP_GROUP_TABLE
```
qiluo-msft, line 81: >NEXT_HOP_GROUP_TABLE [](start = 4, length = 20)

Is there new applications going to use this table?

doc/ip/next_hop_group_hld.md
```diff
@@ -0,0 +1,148 @@
...
+    ;Stores a list of groups of one or more next hops
+    ;Status: Mandatory
+    key           = NEXT_HOP_GROUP_TABLE:string ; arbitrary identifier for the next hop group
+    nexthop       = *prefix,           ; IP addresses separated “,” (empty indicates no gateway)
+    ifname        = *PORT_TABLE.key,   ; zero or more separated by “,” (zero indicates no interface)
+```
+The ROUTE_TABLE is then extended to allow a reference to the next hop group to be specified, instead of the current nexthop and intf fields.
+```
+### ROUTE_TABLE
```
qiluo-msft, line 90: >ROUTE_TABLE [](start = 4, length = 11)

What if application add a route which refers to a nexthop group that hasn’t be created, how to deal with that case?

doc/ip/next_hop_group_hld.md
```diff
@@ -0,0 +1,148 @@
...
+## 3.5 SAI
+No changes are being made in SAI. The end result of what gets programmed over the SAI will be the same as currently gets programmed - the difference is the programming sequence to get there.
+
+## 3.6 CLI
+
+### 3.6.1 Show Commands
+
+The output of 'show ip route' and 'show ipv6 route' will remain unchanged - the CLI code will resolve the next hop group ID referenced in the ROUTE_TABLE to display the next hops for the routes.
+
+# 4 Warm Boot Support
```
qiluo-msft, line 135: >Warm Boot Support [](start = 4, length = 17)

What abut the fast convergence feature we have developed, does it apply to this next group design?

doc/ip/next_hop_group_hld.md
```diff
@@ -0,0 +1,148 @@
...
+
+Warm boot support must not be affected by this enhancement. Specifically, this means that (assuming the required routes don't change during warm boot), routes must be associated with the same next hop group after warm boot. The application programming APP_DB must therefore ensure that next hop groups get assigned the same identifier after reboot. This could be done by the application maintaining that information itself, or retrieving the identifiers from APP_DB at the start of warm boot.
```
qiluo-msft, line 137: >ensure that next hop groups get assigned the same identifier after reboot [](start = 274, length = 73)

How do you "ensure that next hop groups get assigned the same identifier after reboot" ?

qiluo-msft, line 137: >The application [](start = 224, length = 15)

Do you plan to change the db reconciliation logic in fpmsyncd?

doc/ip/next_hop_group_hld.md
```diff
@@ -0,0 +1,148 @@
...
+## 3.4 SyncD
+No change.
+
+
+## 3.5 SAI
+No changes are being made in SAI. The end result of what gets programmed over the SAI will be the same as currently gets programmed - the difference is the programming sequence to get there.
+
+## 3.6 CLI
+
+### 3.6.1 Show Commands
```
qiluo-msft, line 131: >Show [](start = 10, length = 4)

For the cli, we also need show ip nexthopgroup, show ip nexthop cli

doc/ip/next_hop_group_hld.md
```diff
@@ -0,0 +1,148 @@
...
+    blackhole     = BIT ; Set to 1 if this route is a blackhole (or null0)
+ ```
+
+This design adds a new NEXT_HOP_GROUP_TABLE, to store next hop group information to be used by one or more routes.
+```
+### NEXT_HOP_GROUP_TABLE
+    ;Stores a list of groups of one or more next hops
+    ;Status: Mandatory
+    key           = NEXT_HOP_GROUP_TABLE:string ; arbitrary identifier for the next hop group
+    nexthop       = *prefix,           ; IP addresses separated “,” (empty indicates no gateway)
```
qiluo-msft, line 85: >nexthop [](start = 4, length = 7)

How one next hop group is shared by multiple routes?

doc/ip/next_hop_group_hld.md
```diff
@@ -0,0 +1,148 @@
...
+
+The route orchagent will parse the new nexthop_group field, and call into the next hop group orchagent to find the corresponding SAI object ID to program as the next hop attribute.  If the next hop group does not exist, the route will remain on the list of routes to sync, and will be retried on future runs of the route orchagent.
+
+If a next hop group cannot be programmed because the data plane limit has been reached, one next hop will be picked to be temporarily used for that group.  When a route is then programmed using that next hop group, it will be informed that a temporary form of the next hop group is in use, and so the route will remain on the pending list and will retry until the correct next hop group is programmed.  This mirrors the current behaviour in the route orchagent when the next hop group limit is reached.
```
qiluo-msft, line 114: >it will be informed [](start = 215, length = 19)

In this situation, what does the route entry and next hop group entry look like?

doc/ip/next_hop_group_hld.md
```diff
@@ -0,0 +1,148 @@
...
+
+The output of 'show ip route' and 'show ipv6 route' will remain unchanged - the CLI code will resolve the next hop group ID referenced in the ROUTE_TABLE to display the next hops for the routes.
+
+# 4 Warm Boot Support
+
+Warm boot support must not be affected by this enhancement. Specifically, this means that (assuming the required routes don't change during warm boot), routes must be associated with the same next hop group after warm boot. The application programming APP_DB must therefore ensure that next hop groups get assigned the same identifier after reboot. This could be done by the application maintaining that information itself, or retrieving the identifiers from APP_DB at the start of warm boot.
+
+# 4.1 Warm Upgrade
+
+Existing applications will not be affected by this enhancement, so no warm upgrade support is needed.
```
qiluo-msft, line 141: >so no warm upgrade support is needed [](start = 64, length = 36)

If warm upgrade from old image to new image, there is no next hop group table entries at all before warm-reboot doTask(). How do you handle ?

doc/ip/next_hop_group_hld.md
TACappleman: Would just "name of the next hop group" be suitable? There are no restrictions on it, so wanted to convey that.

doc/ip/next_hop_group_hld.md
venkatmahalingam, line 81: Which module updates the NEXT_HOP_GROUP_TABLE? fpmsyncd? if yes, please capture the changes expected in that module upon getting the Zebra route updates.

prsunny: Thats a section missing in the document. As @venkatmahalingam mentioned, it should be covered in the document on how/which module shall update the NEXT_HOP_GROUP_TABLE

doc/ip/next_hop_group_hld.md
```diff
@@ -0,0 +1,148 @@
...
+    ;Status: Mandatory
+    key           = ROUTE_TABLE:prefix
+    nexthop       = *prefix, ;IP addresses separated “,” (empty indicates no gateway)
+    ifname        = *PORT_TABLE.key,   ; zero or more separated by “,” (zero indicates no interface)
+    blackhole     = BIT ; Set to 1 if this route is a blackhole (or null0)
+    nexthop_group = NEXT_HOP_GROUP_TABLE:key ; index within the NEXT_HOP_GROUP_TABLE, optionally used instead of nexthop and intf fields
+ ```
+
+## 3.3 Switch State Service Design
+### 3.3.1 Orchestration Agent
```
prsunny, line 101: Can you please add code flow for the nexthop group orch? Say if nexthop group has only one nexthop, does it translate to SAI ECMP group with one member or is it single nexthop..

doc/ip/next_hop_group_hld.md
nkelapur: From the discussion I understood that when the route changes from ECPM to single path or vice-versa, the application ( frr-zebra) needs to reprogram the route so that NH identifier will be updated correctly. ? Is this understanding correct ? if so, does frr-zebra support this behaviour already ?

doc/ip/next_hop_group_hld.md
TACappleman: Actually, we're not making any changes to fpmsyncd at all at this point, so no.

doc/ip/next_hop_group_hld.md
TACappleman: Currently frr-zebra will reprogram the route whenever its next hops change at all. If it were to change to use this new schema (which is not in scope at this time) it would need to ensure that it still updated the route when changing between single and multiple next hops

nkelapur: Thanks for the reply.  Will it be better to handle the reprogramming at orchagent level, since that is where we are differentiating between the ECMP and Non-ECPMP NH  type ?

TACappleman: It would be possible, but would require orchagent to keep track of which routes are using each next hop/next hop group, which would add a lot of storage. This is because changing between the two will change the SAI object ID, and so every route using the next hop/group would need to be updated. I would suggest keeping the current requirement for now, and removing it in future if there is a clear need to do so.

TACappleman:
We believe that all comments on this have been sufficiently dealt with, so will seek to close the request in a week. @qiluo-msft please let me know before then if this is not the case.

doc/ip/next_hop_group_hld.md
qiluo-msft: I guess you mean to use random string? I would like your implementation choice in this design doc.

doc/ip/next_hop_group_hld.md
qiluo-msft: Let me clarify my questions:
1. Will you reuse the same entry in NEXT_HOP_GROUP_TABLE for multiple routes?
2. If yes, you may use nexthop as the 'key', and there is no need to use another arbitrary identifier

doc/ip/next_hop_group_hld.md
qiluo-msft: Continue on #2: how is the "reference count" designed? Do you want to store it in Redis?

doc/ip/next_hop_group_hld.md
qiluo-msft: Let's say use add a static route, the nexthop group is not existing in SAI or AppDB table, and there are ASIC resource available. How to create this nexthop group? Waiting does not solve anything.

doc/ip/next_hop_group_hld.md
qiluo-msft: To clarify my original question:
In this situation, what does the route entry and next hop group entry look like in AppDB?

doc/ip/next_hop_group_hld.md
qiluo-msft: I am asking warm upgrade(reboot) from old image (without your feature) to new image (with your feature). If you don't handle the DB migration, the warm reboot will fail, and there will be data plane disrutpion.

doc/ip/next_hop_group_hld.md
qiluo-msft: [https://github.com/Azure/SONiC/blob/master/doc/fast-reboot/fastreboot.pdf](https://github.com/Azure/SONiC/blob/master/doc/fast-reboot/fastreboot.pdf)

doc/ip/next_hop_group_hld.md
qiluo-msft: @lguohan, any concern?

doc/ip/next_hop_group_hld.md
qiluo-msft: I don't see "Our proprietary application manages that itself" in this design doc. If you want to integrate this feature on current code base, which component is responsible? could you explain how does it take actions?

doc/ip/next_hop_group_hld.md
TACappleman: This will be maintained within orchagent. I've extended the relevant section of 3.3.1 to cover this in more detail

doc/ip/next_hop_group_hld.md
TACappleman: Yes to 1. For 2, the point of this enhancement is to not have the route table store the full set of next hops, and instead use a much smaller identifier. Using the nexthop field as the key to the table would reverse this. Also, it's possible under this system to change the set of next hops within a group while keeping the identifier the same, which wouldn't be possible if the next hops were the identifier.

doc/ip/next_hop_group_hld.md
TACappleman: If an application is using this model, then it's the application's responsibility to program the NEXT_HOP_GROUP_TABLE and ROUTE_TABLE - to program just the ROUTE_TABLE would be a bug. Note that the ROUTE_TABLE can still contain the next hop information, if an application chooses to do it that way, and in that case routeorch would manage the next hop group as is currently the case.

doc/ip/next_hop_group_hld.md
TACappleman: Done

doc/ip/next_hop_group_hld.md
TACappleman: Added a comment about this

doc/ip/next_hop_group_hld.md
TACappleman: Clarified this - there's a couple of possible methods broadly outlined ("the application maintaining that information itself, or retrieving the identifiers from APP_DB at the start of warm boot"), but exact details would be up to whoever chooses to implement that in future.

doc/ip/next_hop_group_hld.md
TACappleman: I've changed the description slightly, and added a further clarifying comment to the doc. It's entirely up to the programming application how the string is chosen - it could be deterministic, or random.

doc/ip/next_hop_group_hld.md
TACappleman: They remain the same

doc/ip/next_hop_group_hld.md
TACappleman: Since no existing application is being changed to use the new feature, I don't think it makes sense to add warm upgrade support now, as it wouldn't be testable. Warm upgrade support should be added at the point it becomes needed. I've updated the doc to this effect.

doc/ip/next_hop_group_hld.md
qiluo-msft: Discuss under another assumption (not following above), we are using dynamic routes in most cases, we don't change fpmsyncd, "ROUTE_TABLE can still contain the next hop information", everything is like before? I feel the 2-option design is too complex. Can we just design with one option?

TACappleman: In due course we'd like to move to 1 option, but removing the existing mechanism would require both changing fpmsyncd (which would be possible), and ensuring that everyone using the current mechanism but not fpmsyncd is able to change to the new mechanism

TACappleman:
@qiluo-msft Are there any further issues that need resolving on this or can the comments be resolved and the design merged in?

TACappleman:
Can this now be merged?

prsunny (approved):
lgtm, wait for other reviewers

PR #718, opened 2020-12-03
Added SNMP_USER_PERMISSIONS parameter and removed SNMP_USER_STRING parameter

tsvanduyn:
Added one new parameters: 
"SNMP_USER_PERMISSIONS": "RO|RW" to SNMP_USERS. 
This will be needed to support RW SNMPv3 in the future. 

Removed "SNMP_USER_STRING".   This is not necessary because we have this 
SNMP_USER|user1 which shows the user string as "user1" as a key.

### DIFF
--- a/doc/snmp/snmp-schema-addition.md
+++ b/doc/snmp/snmp-schema-addition.md
@@ -66,7 +66,8 @@ In the end this is used to produce */etc/snmp/snmpd.conf*.
          "SNMP_USER_AUTH_TYPE":"MD5|SHA|HMAC-SHA-2",
          "SNMP_USER_ENCRYPTION_TYPE":"DES|AES",
          "SNMP_USER_AUTH_PASSWORD":"<AUTH_PASSWORD_STRING>",
-         "SNMP_USER_ENCRYPTION_PASSWORD":"<ENCRYPTION_PASSWORD_STRING>"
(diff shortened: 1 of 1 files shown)


doc/snmp/snmp-schema-addition.md
```diff
@@ -66,23 +66,26 @@ In the end this is used to produce */etc/snmp/snmpd.conf*.
          "SNMP_USER_AUTH_TYPE":"MD5|SHA|HMAC-SHA-2",
          "SNMP_USER_ENCRYPTION_TYPE":"DES|AES",
          "SNMP_USER_AUTH_PASSWORD":"<AUTH_PASSWORD_STRING>",
-         "SNMP_USER_ENCRYPTION_PASSWORD":"<ENCRYPTION_PASSWORD_STRING>"
+         "SNMP_USER_ENCRYPTION_PASSWORD":"<ENCRYPTION_PASSWORD_STRING>",
+         "SNMP_USER_PERMISSIONS": "RO|RW"
```
SuvarnaMeenakshi, line 70: Might be better to use singular "SNMP_USER_PERMISSION" as it can only one value, right?

tsvanduyn: Agreed.  Fixed.

SuvarnaMeenakshi (approved):
LGTM

qiluo-msft (approved):

PR #736, opened 2021-01-19
SONiC Generic Update and Rollback - HLD

ghooo:


doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,617 @@
...
+- Pushing `minigraph.xml` to the device that contains the new ACL interfaces
+- Execute `sudo acl-loader update full /etc/sonic/acl.json --table_name example_acl`
+
+*DHCP*: Updating DHCP config requires the following steps:
+- Pushing `minigraph.xml` to the device that contains the new ACL interfaces
+- Generating `dhcp_servers` JSON configs from the `minigraph.xml`, and saving it a temporary file
+- Executing `sudo sonic-cfggen -j /tmp/dhcp.json --write-to-db`
+- Restart `dhcp_relay` service
+
+We have explored [SONiC CLI commands](https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md) to make configuration changes. These CLI commands result in updates to the ConfigDB which are corresponding to the CLI command executed. For example, the config `vlan add 10` will create a new row in the VLAN table of the ConfigDB. But relying on the CLI commands to do partial update is also not feasable as there is no standard way of showing the config after the update. Setting up a different update mechanism for each part of the config is very time consuming and ineffecient.
```
qiluo-msft, line 103: >feasable  [](start = 414, length = 9)

feasible

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,617 @@
...
+- Con's
+  - Can't remove existing configuration and can only be used to add/modify the existing configuration
+- Verdict
+  - Cannot be used as a rollback mechanism
+
+Since both `config load` and `config reload` are not suitable for a hitless/non-disruptive rollback, we have to look for other approaches.
+
+In this design document, we will be exploring how to standardize the way to do partial updates, how to take checkpoints and finally how to rollback the configurations.
+
+In summary, this is the flow of an update:
```
qiluo-msft, line 131: >flow [](start = 24, length = 4)

Rollback could succeed or fail in this diagram

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,617 @@
...
+
+![basic-target-design](files/basic-target-design.png)
+
+And the steps would be:
+```
+admin@sonic:~$ checkpoint mycheckpoint
+admin@sonic:~$ echo "config changes to apply to ConfigDb" > some-config-changes.patch
+admin@sonic:~$ apply-patch ./some-config-changes.patch
```
qiluo-msft, line 139: >patch [](start = 49, length = 5)

RFC6902 has defined "File extension(s):  .json-patch".
We discussed .patch.json before, but I think the RFC one is better.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,617 @@
...
+
+## 1.1 Requirements
+
+### 1.1.1 Functional Requirements
+- A single, simple command to partially update SONiC configuration according to a patch of updates
+- A single, simple command to take a checkpoint of the full current SONiC config
+- A single, simple command to fully rollback current SONiC configs with to a checkpoint
+- Other commands to list checkpoints, delete checkpoints
+- The patch of updates should follow a standard notation. The [JSON Patch (RFC6902)](https://tools.ietf.org/html/rfc6902) notation should be used
+- Config rollback should be with minimum disruption to the device e.g. if reverting ACL updates DHCP should not be affected i.e. hitless rollback
```
qiluo-msft, line 151: >minimum disruption [](start = 33, length = 18)

"minimum disruption" is more accurate than hitless or non-disruptive. Suggest make it consistent in this document.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,617 @@
...
+- User should be able to preview the configuration difference before going ahead and committing the configuration changes
+- In case of errors, the system should just report an error and the user should take care of it
+- Only 1 user session can update device config at a time i.e. no concurrent updates to configurations
```
qiluo-msft, line 154: >Only 1 user session [](start = 2, length = 19)

I think you mean "Only one session globally on the device"

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,617 @@
...
+    "1.1.1.1": {},
+    "2.2.2.2": {},
+    "3.3.3.3": {}
+  }
+}
+```
+
+and the ask to:
+- replace *port* under *Ethernet8* with *eth1*
+- add 4.4.4.4 tanother DHCP_SERVER with IP
```
qiluo-msft, line 203: >tanother [](start = 14, length = 8)

typo?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,617 @@
...
+
+The `apply-patch` command uses an ordering table which specifies the list of dependent tables that need to be processed before a table is processed. Below is an example depiction of two ConfigDB tables.
+
+```
+{
+  "VLAN" : ["VLAN_MEMBER", "VLAN_INTERFACE"],
+  "PORTCHANNEL" : ["VLAN_MEMBERSHIP", "PORTCHANNEL_MEMBER"]
+}
+```
+A dependency graph is created for all the tables which are in the JSON patch file. The dependency graph is then resolved to find the order in which the updates to tables in the JSON patch are performed. The JSON patch file is then sorted in the resolved order.
```
qiluo-msft, line 286: >The dependency graph is then resolved to find the order in which the updates to tables in the JSON patch are performed [](start = 83, length = 118)

This is the interesting part. Will you give details and example?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,617 @@
...
+  "NTP_SERVER": ["ntp-config.service", "ntp.service"]
+  "BGP_MONITORS": ["bgp"],
+  "BUFFER_PROFILE": ["swss"],
+  "RESTAPI": ["restapi"]
+}
+```
+
+This table is used to restart corresponding systemd service when a patch operation has been perform on its corresponding ConfigDB table. 
+
+#### Stage-3 Add delay after table updates for listening services to absorb changes
```
qiluo-msft, line 304: >Add delay after table updates for listening services to absorb changes [](start = 13, length = 70)

Offline discussed, delay is not a stable way to check changes absorbed. Let's depend on 'service restart' response, or use cli to verify it.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,617 @@
...
+- default-delay-between-table-updates - Some daemons are listening to ConfigDB changes, so artificial sleep to give a pause between two table updates as to ensure that all internal daemons have digested the previous update, before the next one arrives.
+
+### 3.6.2 CLI
+
+#### 3.6.2.1 Configuration Commands
+
+**apply-config**
+*Command Format*
+
+config apply-patch <*patch-filename*> [dry-run] [verbose]
```
qiluo-msft, line 501: >config apply-patch [](start = 0, length = 18)

It is good to integrated into existing `config` CLI. Could you make it consistent in this document?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,617 @@
...
+
+| Command Option                                                | Purpose                                                 |
+| -------------------------------------------------------- | ------------------------------------------------------------ |
+|<*patch-filename*> | The file of the JsonPatch file to apply which follows [JSON Patch (RFC6902)](https://tools.ietf.org/html/rfc6902) specifications |
+|dry-run | Displays the generates commands going to be executed without running them.  |
```
qiluo-msft, line 506: >dry-run [](start = 1, length = 7)

Is it an option? if yes, we may use `-d/--dry-run` ?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Good idea!

ghooo: Discussed offline, the ordering of updating the tables might be different in case of deletion vs addition

image

ghooo: Idea `i` pseudo code to process all lines to be deleted, then processing all lines to be added:
```python
apply-patch(patch):
  current_running_config = get_running_config()
  simulated_config_after_patch = simulate_apply_patch(current_running_config, patch)
  
  common_config = get_common_config(current_running_config, simulated_config_after_patch)

  # change running config to common config, this should be deletion only
  config_replace(common_config, TABLES_DELETION_ORDER)

  # once config is common config, change it to simulated_config_after_patch, this should be addition only
  config_replace(simulated_config_after_patch, TABLES_ADDITION_ORDER)
```

ghooo: I think the pseudo code will be better if just take config_replace to its own method

```python
apply_patch(patch):
  current_running_config = get_running_config()
  simulated_config_after_patch = simulate_apply_patch(current_running_config, patch)
  config_replace(simulated_config_after_patch, current_running_config )

config_replace(target_config, current_config):
  common_config = get_common_config(current_config, target_config)

  # change running config to common config, this should be deletion only
  config_replace(common_config, TABLES_DELETION_ORDER)

  # once config is common config, change it to simulated_config_after_patch, this should be addition only
  config_replace(current_config, TABLES_ADDITION_ORDER)
```

ghooo: The common idea will not work, as it will lead to bunch of deletions followed by bunch of additions. But I might encounter a case where a config cannot be deleted, and can only be replaced. So in the time of deletion/addition of this command the switch might be in a bad state.

I will instead rely on [YANG SONiC Models](https://github.com/Azure/sonic-buildimage/tree/master/src/sonic-yang-models) to fetch the dependency between the configs getting updated in the patch, and decide how they can be updated. For example:
Assume the patch is the following
```
[
 { "op": "remove", "path": "/PORT/Ethernet1" }
 { "op": "remove", "path": "/VLAN_MEMBER/VLAN1000/Ethernet1" },
 { "op": "remove", "path": "/ACL_TABLE/ssh-acl/Ports/Ethernet1" },
]
```

Checking the `YANG` models, we will find that VLAN_MEMBER and ACL are both referencing PORT. So cleaning up of the VLAN_MEMBER and ACL

The new order would be
```
[
 { "op": "remove", "path": "/ACL_TABLE/ssh-acl/Ports/Ethernet1" },
 { "op": "remove", "path": "/PORT/Ethernet1" }
 { "op": "remove", "path": "/VLAN_MEMBER/VLAN1000/Ethernet1" },
]
```

ghooo: updated to a different approach described in the PR

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
qiluo-msft: Still applicable.

---
In reply to: [560496306](https://github.com/Azure/SONiC/pull/736#discussion_r560496306) [](ancestors = 560496306)

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,705 @@
...
+```
+[
+  { "order": 1, "table": "PORT", "op": "remove", "path": "/PORT/Ethernet2" }
+  { "order": 2, "table": "VLAN_MEMBER", "op": "remove", "path": "/VLAN_MEMBER/Vlan100|Ethernet2" }
+  { "order": 3, "table": "ACL_TABLE", "op": "remove", "path": "/ACL_TABLE/NO-NSW-PACL-V4/ports/0" }
+]
+```
+This will not work, as the PORT table will complain that the port `Ethernet2` is still in use by a `VLAN_MEMBER`.
+
+The solution for that is if an operation fails, we move on to the next operation, and so on. Until we finish all the operation, then we loop over again on the operations that did not succeed in the previous cycle. If in a cycle no operation succeed, we have hit a dead-end and we report failure
```
qiluo-msft, line 343: >The solution for that [](start = 0, length = 21)

Offline discussed, this solution is not recommended. Do not depends on a running SONiC to test the dependencies blindly.

ghooo: Removed, replaced by the contract `order-patch`

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+
+*config load <config_db.json>*: This command loads the contents of config_db.json into the ConfigDB. The updates made to the ConfigDB are additive in nature and thus the new configuration state is a combination of the current running state and the partial configuration state specified by the user in the config_db.json file
+
+- Pro's
+  - Quick way to add new configuration changes
+  - It does not disrupt existing service whose configuration is not being modified. So it is non-disruptive in nature
+- Con's
+  - Can't remove existing configuration and can only be used to add/modify the existing configuration
+- Verdict
+  - Cannot be used as a rollback mechanism
```
praveen-li, line 145: This was true because we did not have backend code for all networking features, that handles delete OP.  I think, Otherwise, we can use it for rollback for any networking feature related config which supports delete operation at the backend.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+In this design document, we will be exploring how to standardize the way to do partial updates, how to take checkpoints and finally how to rollback the configurations.
+
+In summary, this is the flow of an update:
+
+<img src="files/basic-target-design.png" alt="basic-target-design" width="500px"/>
+
+And the steps would be:
+```
+admin@sonic:~$ config checkpoint mycheckpoint
+admin@sonic:~$ echo "config changes to apply to ConfigDb" > some-config-changes.json-patch
```
praveen-li, line 158: I think You meant <config changes to apply to ConfigDb> here? , i.e. the patch file will contain the real config, not the string

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+admin@sonic:~$ config apply-patch ./some-config-changes.json-patch
+admin@sonic:~$ config rollback mycheckpoint # in case of failures
```
praveen-li, line 160: Does it load the entire config back which is at < mycheckpoint > or just revert the last diff?, Maybe the answer lies in the same document, let me see.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+```
+
+## 1.1 Requirements
+
+### 1.1.1 Functional Requirements
+- A single, simple command to partially update SONiC configuration according to a patch of updates
+- A single, simple command to take a checkpoint of the full current SONiC config
+- A single, simple command to fully rollback current SONiC configs with to a checkpoint
+- Other commands to list checkpoints, delete checkpoints
+- The patch of updates should follow a standard notation. The [JSON Patch (RFC6902)](https://tools.ietf.org/html/rfc6902) notation should be used
```
praveen-li, line 170: By the way, what libraries are there which provide this functionality? I think one is python jsonDiff.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+- Only root user must be allowed to execute the commands
+- Non-root users can execute commands with dry-run option
+- Each command must provide the following sub options:
+  - "dry-run" Perform a dry-run of the command showing exactly what will be executed on the device, without executing it
+  - "verbose" Provide additional information on the steps executed
+
+### 1.1.3 Scalability Requirements
+N/A
+
+### 1.1.4 Warm Boot Requirements
```
praveen-li, line 187: I guess, while doing a warm reboot, we put the switch in maintenance (reason: s/w upgrade) mode, and if we can detect that, and doesn't allow any config to run from the outside tool.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+N/A
+
+## 1.2 Design Overview
+
+<img src="files/sonic-design.png" alt="sonic-design" width="1200"/>
```
praveen-li, line 192: While rolling back do we create a new checkpoint because of the rollback config change?

Also if the difference between the two checkpoints is none then we can avoid further steps. I feel you would have taken care of this.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+
+### 1.2.1 Basic Approach
+SONiC ConfigDB contents can be retrieved in a JSON file format. Modifying JSON file should follow some rules in order to make it straightforward for the users. Fortunately there is already a formal way of defining JSON config updates. It is called JsonPatch, and is formally defined in [RFC 6902 JSON Patch](https://tools.ietf.org/html/rfc6902).
+
+On top of ConfigDBConnector we are going to implement [RFC 6902 JSON Patch](https://tools.ietf.org/html/rfc6902). This API we will call `apply-patch`. On top of that API, we will implement the `rollback` functionality. It will simply starts by getting the diff (patch) between the checkpoint and the current running config, then it will call the API `apply-patch` to update that patch.
+
+The [JsonPatch](https://pypi.org/project/jsonpatch/) python is an open source library that already implements the [RFC 6902 JSON Patch](https://tools.ietf.org/html/rfc6902). We can leverage this library to verify patch config, generate a diff between checkpoint and current running config, verify apply-patch and rollback work as expected by simulating the final output of the update and comparing with the observed output.
```
praveen-li, line 199: cool,  you have the lib name here.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+```
+
+and the ask to:
+- replace *port* under *Ethernet8* with *eth1*
+- add 4.4.4.4 to DHCP_SERVERS
+- remove 2.2.2.2 from DHCP_SERVERS
+
+The steps would be:
+
+1) Take a checkpoint of the config
```
praveen-li, line 230: Take the content of config DB here and store it?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+[
+  { "order": 3, "table": "ACL_TABLE", "op": "add", "path": "/ACL_TABLE/NO-NSW-PACL-V4/ports/0", "value": "Ethernet2" }
+  { "order": 2, "table": "VLAN_MEMBER", "op": "add", "path": "/VLAN_MEMBER/Vlan100|Ethernet2", "value": { "tagging_mode": "untagged" } }
+  { "order": 1, "table": "PORT", "op": "add", "path": "/PORT/Ethernet2", "value": { "lanes": "65", "speed": "10000"} }
+]
+```
+
+Now let's order the operations by "order":
+```
+[
```
praveen-li, line 345: Once the entire config is verified, I think we can push the entire config in, if the backend can handle the dependency. I feel backend handles the dependency for ACL, VLAN and PORT, but not sure if this is true for all network objects .

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+#### Stage-5 Verify config rollback
+Compare the ConfigDB JSON after the update with the checkpoint JSON, there should be no differences.
+
+#### Fail-safe Action
+
+If an error is encountered during the `rollback` operation, an error is reported and the system DOES NOT take any automatic action. Rollback operation is itself an automated fail-safe action, if itself fails the caller should decide how to handle such failures e.g. generate an alert of high severity, or do `config reload`.
+
+#### Logging
+
+All the configuration update operations executed and the output displayed by the `rollback` command are stored in the systemd journal. They are also forwarded to the syslog. By storing the commands in the systemd-journal, the user will be able to search and display them easily at a later point in time. The `show rollback log` command reads the systemd-journal to display information about the `rollback` command that was previously executed or currently in progress.
```
praveen-li, line 479: Good, we need this.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+[
+  { "op": "add", "path": "/PORT/Ethernet2", "value": { "lanes": "65", "speed": "10000"} }
+  { "op": "add", "path": "/VLAN_MEMBER/Vlan100|Ethernet2", "value": { "tagging_mode": "untagged" } }
+  { "op": "add", "path": "/ACL_TABLE/NO-NSW-PACL-V4/ports/0", "value": "Ethernet2" }
+]
+```
+
+The only condition of JsonChange is that the final outcome after applying the whole JsonPatch is the same.
+
+##### 3.1.1.4.2 Order-Patch
```
praveen-li, line 545: This topological sort can be generated from the yang model, because the dependency should come from the yang model between various networking object.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+
+<img src="files/jsonpatch-operations-order.png" alt="jsonpatch-operations-order" width="800"/>
+
+But if we organize the operations into groups of JsonChange, we will have:
+
+<img src="files/jsonchanges-order.png" alt="jsonchanges-order" width="800"/>
+
+This will allow the the implementor of `apply-change` to have the freedom to optimize the operations in any order they see fit.
+
+**NOTE:** Check Patch Orderer implementation design design details in [Json_Patch_Ordering_using_YANG_Models_Design](Json_Patch_Ordering_using_YANG_Models_Design.md) document.
```
praveen-li, line 599: Okie here it is, :)

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+
```
praveen-li, line 600: Even though we can control the order of operation from outside, I feel, until and unless, backend handles the dependency we can still result in a service crash.

We have noticed some circumstances where the previous config changes are not processed by the backend and  new config changes come which are already in the pipeline and being processed before last config change.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+Since the order of executing the operation does not matter, the implementor of this component can work on optimizing the time to run the operation. For details check [3.1.1.4 Change Applier](#3115-change-applier).
+
+**NOTE:** Check Change Applier implementation design design details in [Json_Change_Application_Design](Json_Change_Application_Design.md) document.
+
+#### 3.1.1.6 ConfigDB
+SONiC is managing configuration in a single source of truth - a redisDB instance that we refer as ConfigDB. Applications subscribe to ConfigDB and generate their running configuration correspondingly.
+
+For further details on the ConfigDB, check [SONiC Configuration Database Manual](https://github.com/Azure/SONiC/wiki/Configuration).
+
+### 3.1.2 Checkpoint
```
praveen-li, line 630: I think we need the config lock here as well because nobody should be writing in the config DB when we are reading for checkpoint.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+
+show apply-patch log [exec | verify | status]
+
+| Command                     | Purpose                                                      |
+| --------------------------- | ------------------------------------------------------------ |
+| show apply-patch log exec   | Displays a log of all the ConfigDB operations executed including<br/>those that failed. In case of a failed operation, it displays an<br/>error message against the failed operation. |
+| show apply-patch log verify | Displays a log all the ConfigDB operations that failed, <br>along with an error message. It does not display the <br/>operations that were successful. |
+| show apply-patch log status | Displays the status of last successful patch application<br/>operation since switch reboot. |
+
+**rollback**
```
praveen-li, line 790: Overall it's a good start. Can we have a meeting about it because that will clarify more points and also we can share some knowledge because we have done similar stuff. Thx.

doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md
```diff
@@ -0,0 +1,703 @@
...
+- Nodes are configuration states which should be valid according to YANG models
+- Edges are the valid JsonChanges which should ignore any command ordering
+
+<img src="files/config-transitions-diagram.png" alt="config-transitions-diagram" width="1200"/>
+
+The problem now is just finding a path between the current config and the target config.  A solution to the above graph might be:
+
+<img src="files/valid-config-transitions-diagram.png" alt="valid-config-transitions-diagram" width="1200"/>
+
+Note:
```
praveen-li, line 138: Think we can have a meeting about this because I feel this can be solved in a much simpler way.

From yang models, we can find the schema dependencies (i think we can also find it using pyang sonic yang tree generation). Using schema dependencies, we can create a graph based on the current config changes and then run topological sort on that.
But as said, This is not guaranteed that backend will process these config changes in order as I know. That is why we decided to depend on APP_DB and ASIC_DB updates in between config changes under dynamic Port breakout project.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Yes, we are adding the support to delete OP in this design document.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: That's correct

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Entire config

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: [python-json-patch](https://github.com/stefankoegl/python-json-patch)

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+
+## 1.1 Requirements
+
+### 1.1.1 Functional Requirements
+- A single, simple command to partially update SONiC configuration according to a patch of updates
+- A single, simple command to take a checkpoint of the full current SONiC config
+- A single, simple command to fully rollback current SONiC configs with to a checkpoint
+- Other commands to list checkpoints, delete checkpoints
+- The patch of updates should follow a standard notation. The [JSON Patch (RFC6902)](https://tools.ietf.org/html/rfc6902) notation should be used
+- Config rollback should be with minimum disruption to the device e.g. if reverting ACL updates DHCP should not be affected i.e. minimum-disruption rollback
```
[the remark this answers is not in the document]
ghooo: No, it supports changes in multiple networking features.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+
+### 1.1.2 Configuration and Management Requirements
+- All commands argument to generated using Python-click to provide help menus and other out-of-the box features
+- Only root user must be allowed to execute the commands
+- Non-root users can execute commands with dry-run option
+- Each command must provide the following sub options:
+  - "dry-run" Perform a dry-run of the command showing exactly what will be executed on the device, without executing it
+  - "verbose" Provide additional information on the steps executed
+
+### 1.1.3 Scalability Requirements
```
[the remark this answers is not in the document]
ghooo: Good idea.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: My understanding is that when we make changes to ConfigDB there is 2 types of services:
- Services that need manual restart 
- Services that absorb the changes automatically

When you say "while doing a warm reboot", which one do you mean?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: No, why would we create a checkpoint while rolling back? In case of rollback failure, we should just report failure and the user should take care of that.

That's seems like a special case which is handled correctly by the current flow of operations. Maybe we can make the code short-circuit in such cases and return immediately.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Sure, will update.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+- add 4.4.4.4 to DHCP_SERVERS
+- remove 2.2.2.2 from DHCP_SERVERS
+
+The steps would be:
+
+1) Take a checkpoint of the config
+```
+admin@sonic:~$ config checkpoint mycheckpoint
+```
+2) Create a file on the device named `dhcp-changes.json-patch`, with the following content
```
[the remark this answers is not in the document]
ghooo: Yes, same lib can provide the diff functionality.

```python
patch = jsonpatch.JsonPatch.from_diff(current_configdb, checkpoint_configdb)
if patch:
  print("Diff found")
else:
  print("NoDiff found")

```

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,867 @@
...
+
+Human operators can also leverage the `checkpoint` and `rollback` functionalities while doing updates through the CLI using [SONiC CLI commands](https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md).
+
+## 2.2 Functional Description
+
+### 2.2.1 Apply-Patch
+The SONiC `apply-patch` command can broadly classified into the following steps
+
+#### Stage-1 JSON Patch Verification
+Using [YANG SONiC models](https://github.com/Azure/sonic-buildimage/tree/master/src/sonic-yang-models), but the format of the JSON patch is not what YANG SONiC models is built to verify. We will verify using the following steps:
```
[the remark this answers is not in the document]
ghooo: Is validating the entire config time-consuming on SONiC box? usually how long it takes to validate the config?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Backend cannot handle dependency of which objects, do you have an example we can check?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: We have isolated this concern behind its own contract called `Apply-Change`. The implementation of this contract should take care of applying the changes to ConfigDB, restart corresponding services, and making sure services have absorbed the change correctly. Check the design doc: Json_Change_Application_Design.md for details.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Sure, will add it.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Booked a meeting on 3/15

doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md
ghooo: Booked a meeting for 3/15, main challenges to ordering are:
- Ordering divides the update into steps, each step need to move the config from a valid YANG model state to another YANG model state. Do we respect this constraint?
- Interdependency between network objects should be taken care of. For example: PORT/VLAN_MEMBER are related. Deleting a PORT/VLAN_MEMBER should start with deleting the VLAN_MEMBER then deleting the PORT. But adding PORT/VLAN_MEMBER should go in opposite order, adding the PORT adding the VLAN_MEMBER
- YANG models `mandatory` can force some fields to be required, in such case they cannot be deleted/added, only replaced.
- YANG models `min-elements` and `max-elements` can force the count of some items to be in certain range.
- YANG models `choice`/`unique` fields cannot be added then deleted as adding can create a conflict with existing fields, have to always be delete then added or replaced.
- All the above constraints can be combined

We separated apply-change mechanism to its own contract, so we can focus on ordering sedately from config application. (i.e. separation of concerns). Config application includes updating ConfigDB, restarting services, verifying services have absorbed change correctly, check Json_Change_Application_Design.md for details

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Or are you saying if somehow a warm-reboot is getting executed, we should lock config changes?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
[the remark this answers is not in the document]
ghooo: Offline discussion, since we are providing the capability of deleting/listing checkpoints to the end-user. The end user can take care of storage management.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Offline discussion: The locking mechanism should be applied to config changes when doing warm-reboot, but I think this is a problem with config changes in general and not only current suggested config operation. I think it should be fixed for all operations with config-mgmt.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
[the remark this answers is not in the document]
ghooo: Offline discussion: need to analyze the performance over entire validation of a very large config (worst case scenario), if performance is bad need to optimize. Ideas for optimization include applying the new changes to YANG config tree directly as opposed ConfigDB JSON and re-building YANG config tree and evaluating it again.

doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md
ghooo: Offline discussion: The proposed approach takes into consideration using schema dependencies provided by YANG, but due to other requirements listed in the above reply, that won't be enough to allow the switch a solely topological sorting algorithm. It will be great if we can find a greedy solution to this problem, but so far there is not a clear one.

qiluo-msft (approved):

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -34,14 +34,23 @@
...
+    + [2.2.3 Rollback](#223-rollback)
       - [Stage-1 Get current ConfigDB JSON config](#stage-1-get-current-configdb-json-config-1)
       - [Stage-2 Get checkpoint JSON config](#stage-2-get-checkpoint-json-config)
       - [Stage-3 Generate the diff as JsonPatch between current config and checkpoint](#stage-3-generate-the-diff-as-jsonpatch-between-current-config-and-checkpoint)
       - [Stage-4 Apply-Patch](#stage-4-apply-patch)
       - [Stage-5 Verify config rollback](#stage-5-verify-config-rollback)
       - [Fail-safe Action](#fail-safe-action-1)
       - [Logging](#logging-1)
+    + [2.2.4 Replace](#224-replace)
+      - [Stage-1 Get current ConfigDB JSON config](#stage-1-get-current-configdb-json-config-2)
```
praveen-li, line 46: Here the stages refer to the stage of development?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -227,7 +244,7 @@ and the ask to:
 
 The steps would be:
 
-1) Take a checkpoint of the config
+1) Take the content of config DB here and store it as a checkpoint
```
praveen-li, line 247: As per my understanding, we change the config then the previous config becomes a checkpoint. But the current config,  after the recent change, that doesn't convert into a checkpoint right?.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -478,6 +495,35 @@ If an error is encountered during the `rollback` operation, an error is reported
...
+### 2.2.4 Replace
+The SONiC `replace` command can broadly classified into the following steps
+
+#### Stage-1 Get current ConfigDB JSON config
+The *ConfigDBConnector* class is used to obtain the running configuration in JSON format
+
+#### Stage-2 Get target config from the external user/system
+The external user/system to provide the full target ConfigDB config in JSON format.
+
+#### Stage-3 Validating the target config using YANG models
```
praveen-li, line 507: This can be step 2 in replace case, because if the incoming config is not valid ,we do not even have to do current step 1 which is to get the current config DB.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -219,18 +221,22 @@ admin@sonic:~$ checkpoint mycheckpoint
...
+  {
+    "op": "remove",
+    "path": "/DHCP_SERVERS/2.2.2.2"
+  }
 ]
 ```
 3) Apply patch using `apply-patch` command
 ```
-admin@sonic:~$ apply-patch ./dhcp-changes.patch.json
+admin@sonic:~$ config apply-patch ./dhcp-changes.json-patch
```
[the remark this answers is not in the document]
ghooo: From the rfc: https://tools.ietf.org/html/rfc6902#section-6

The file extension of JsonPatch is `.json-patch`
```
      File extension(s):  .json-patch
```

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: the stages are general explanation of how it should be developed

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: No, we don't create automatic checkpoints.

Creating a checkpoint is a manual process. the user has to take a checkpoint at any point in time, decided to rollback at any point it time. 

We might consider automatically taking checkpoints in the future for some operations, but that's out of the scope of this design document.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: checkpoint, and rollback are both manually initiated by the user. The user can create a checkpoint at any time, and rollback to any checkpoint at any time. the operations are decoupled and also generic and can be applied to any config.

The image just demonstrates how the replace/rollback APIs rely on the apply-patch API in a clean way.

praveen-li (commented):
I will complete entire review today/

doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md
```diff
@@ -0,0 +1,698 @@
...
+### 3.1.2 Using YANG models to Generate Other Moves
+Check [2.2.3 Generating all possible moves](#223-generating-all-possible-moves) to learn more about what are the possible moves.
+
+YANG models can help with:
+- Lines getting removed which have references, can be substituted by deleting their references first as it is required anyway (check [2.2.3.3 Generating Other Moves](#2233-generating-other-moves) for details). Getting the dependency of some data can be done using [SONiC Libynag wrapper](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-mgmt/sonic_yang.py), specifically `find_data_dependencies`.
+- Any other moves should be generated using YANG models, if YANG models data are not enough an extension to YANG models can be added to help with generating new moves.
+
+### 3.1.3 Using YANG models to Host Flags for Move Validation
+#### 3.1.3.1 Extension: create-only
+Some fields are not allowed to be modified and can only be created, such as `/PORT/Ethernet0/lanes` which is marked as `CREATE_ONLY` in [opencomputeproject/SAI/blob/master/inc/saiport.h](https://github.com/opencomputeproject/SAI/blob/master/inc/saiport.h#L624):
```
praveen-li, line 397: Any idea if we can mark in Yang model that a particular object is create-only,  I noticed the extension down later

doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md
```diff
@@ -0,0 +1,698 @@
...
+    }
+]
+```
+But the field `lanes` is `create-only` and cannot be modified i.e. replaced. The only way to update the `lanes` field is through deleting then adding its parent the `/PORT/Ethernet0`.
+
+In [port YANG model](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-port.yang), let's mark `lanes` with `create-only`:
+```diff
+ leaf lanes {
+     mandatory true;
++    ext:create-only;
```
praveen-li, line 438: We need to add processing of this in the sonic-yang-mgmt python library

doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md
```diff
@@ -0,0 +1,698 @@
...
+        "path": "/ACL_TABLE/NO-NSW-PACL-V4/ports/3",
+        "value": "Ethernet3"
+    }
+]
+```
+
+**Result Jsonchanges**
+```json
+[
+    [{"op": "remove", "path": "/ACL_TABLE/NO-NSW-PACL-V4/ports"}],
```
praveen-li, line 685: In this case, we are modifying the port field from Ethernet0 to empty, but we are not removing it right?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -25,51 +25,87 @@
...
+    + [2.2.3 Rollback](#223-rollback)
       - [Stage-1 Get current ConfigDB JSON config](#stage-1-get-current-configdb-json-config-1)
       - [Stage-2 Get checkpoint JSON config](#stage-2-get-checkpoint-json-config)
       - [Stage-3 Generate the diff as JsonPatch between current config and checkpoint](#stage-3-generate-the-diff-as-jsonpatch-between-current-config-and-checkpoint)
       - [Stage-4 Apply-Patch](#stage-4-apply-patch)
       - [Stage-5 Verify config rollback](#stage-5-verify-config-rollback)
       - [Fail-safe Action](#fail-safe-action-1)
       - [Logging](#logging-1)
+    + [2.2.4 Replace](#224-replace)
+      - [Stage-1 Get current ConfigDB JSON config](#stage-1-get-current-configdb-json-config-2)
```
praveen-li, line 46: Should we do it at stage 3, after validating the target config?

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -374,85 +419,42 @@ There are a few SONiC applications which store its configuration in the ConfigDB
...
-3) Verify expected listening services to have abosorbed the change, maybe check if they were resarted after the table is updated?
-4) ... other options?
+Currently there is no standard approach of updating ConfigDB that takes care of manual service restart and verifying services have absorbed change correctly.
 
-#### Stage-5 Apply patch, service restarts and delays
-At this stage we have built all the steps needed. The steps include JsonPatch operations, restart operations and delay operations.
+So we are going to introduce a new contract overlaying ConfigDB changes, that will take care of making correct changes to ConfigDB as well as verifying that corresponding services have absorbed the changes correctly.
 
-The JsonPatch consistes of a list operation, and the operation follows this format:
+The contract would be:
```
praveen-li, line 428: By the way, just out of curiosity, this "contract language" is taken out of some formal way of writing documents

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -493,40 +495,257 @@ If an error is encountered during the `rollback` operation, an error is reported
...
 ### 3.1.1 ApplyPatch
-![apply-patch-design](files/apply-patch-design.png)
+<img src="files/apply-patch-design.png" alt="apply-patch-design" width="1200"/>
+
+
+#### 3.1.1.1 User
+The user of the system, can simply be a human operator or a service that can talk to SONiC CLI. The user can only apply-patch if they have an admin permission. Users without admin permissions can only execute dry-run of the operations where they will be able to see the exact changes going to affect the device, without executing these changes.
+
+#### 3.1.1.2 SONiC CLI
+These are the CLI of SONiC switch to which makes it easy for the users to interact with the system. The CLI commands we are interested in are `config ...` and `show ...`, check [SONiC Command Line Interface Guide](https://github.com/Azure/sonic-utilities/blob/master/doc/Command-Reference.md) to learn more about SONiC CLI.
```
praveen-li, line 539: Here the user will see only the config changes after dry-run or they will also see the list of services that will be restarted due to the config changes.

praveen-li (commented):
Overall looks good to me, let me know if there is a section for lower-level design and the section which describes the unit test in more detail.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Good idea!

doc/config-generic-update-rollback/Json_Change_Application_Design.md
```diff
@@ -0,0 +1,400 @@
...
+```json
+    { "op": "remove", "path": "/DHCP_SERVER/192.0.0.2" },
+    { "op": "replace", "path": "/DEVICE_NEIGHBOR/Ethernet8/port", "value": "eth1" },
+    { "op": "replace", "path": "/DEVICE_NEIGHBOR/Ethernet80/port", "value": "eth1" },
+    { "op": "add", "path": "/DHCP_SERVER/192.0.0.3", "value": {} },
+```
+
+Applying JsonChange is all about the running config being converted to the target config, the steps to update can be arbitrary and are decided by the implementor of the `apply-change` interface.
+
+Our current design documents will use the "Table" as the main granular element of update. We will update "Table" by "Table" in alphabetical order. Each table update will take care of updating table entries in ConfigDB, restarting services if needed and verifying services have absorbed
```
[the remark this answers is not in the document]
ghooo: That's correctly, dependency should be handled in the patch-ordering stage.

This is the flow of the operations:
```python
apply-patch(patch):
  changes = order-patch(patch)
  for change in changes:
    apply(change)
```

doc/config-generic-update-rollback/Json_Change_Application_Design.md
```diff
@@ -0,0 +1,400 @@
...
+      "services-to-validate": ["<SERVICE1>", "<SERVICE2>" ...]
+    },
+    .
+    .
+    .
+  },
+  "services": {
+    "<SERVICE-NAME>": {
+      "restart-command": "<CLI-COMMAND>",
+      "validate-command": "<CLI-COMMAND>"
```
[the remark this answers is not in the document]
ghooo: Sure.

doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md
```diff
@@ -0,0 +1,698 @@
...
+
+  moves = get_all_moves(current_config)
+  best_moves = None
+  for move in moves:
+      if is_valid(move, current_config):
+          new_config = apply_move(move, current_config)
+          new_moves = rec(new_config, target_config)
+          if new_moves != None:
+              # Minimize on the number of moves
+              if best_moves == None or len(best_moves) > 1 + len(new_moves):
```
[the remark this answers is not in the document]
ghooo: Yes, I believe converting memorization to dynamic-programming should be the way to go. We will start by the DFS explained earlier make sure we have a working soln, then we will improve from there.

doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md
ghooo: Yes, will use an extention.

doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md
ghooo: Yes, will have to add this support.

doc/config-generic-update-rollback/Json_Patch_Ordering_using_YANG_Models_Design.md
ghooo: It will be removed.

If we are to make it empty the jsonpatch would be:
    [{"op": "replace", "path": "/ACL_TABLE/NO-NSW-PACL-V4/ports", "value":""}],

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: The plan is to see both.

We will execute the whole code as expected, and will replace the last step when actually updating configdb with a specific config or issuing a service restart/service validation commands with printing to user.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: I made it up for important parts of a function: input, output, exceptions, side-effects, assumptions. As for the list of errors these are the similar to [http patch error](https://tools.ietf.org/html/rfc5789#section-2.2)

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: Good idea

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: updated locally

doc/config-generic-update-rollback/Json_Change_Application_Design.md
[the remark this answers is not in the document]
ghooo: updated locally

doc/config-generic-update-rollback/Json_Change_Application_Design.md
```diff
@@ -0,0 +1,400 @@
...
+# 7 Warm Boot Support
+N/A
+
+# 8 Scalability
+N/A
+
+# 9 Unit Tests
+| Test Case | Description |
+| --------- | ----------- |
+| 1         | Add a new table. |
```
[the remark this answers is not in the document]
ghooo: No, I think they are pretty generic. I added an example of the last unit-test for an item with default value.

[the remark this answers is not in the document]
ghooo: updated locally

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: updated locally

ghooo: updated

doc/config-generic-update-rollback/Json_Change_Application_Design.md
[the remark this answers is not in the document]
ghooo: updated

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
ghooo: updated

praveen-li (approved):
Overall looks good to me, let me know if any pending questions, where my involvement is needed. Thx.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -0,0 +1,940 @@
...
+admin@sonic:~$ config rollback mycheckpoint # in case of failures
+```
+
+## 1.1 Requirements
+
+### 1.1.1 Functional Requirements
+- A single, simple command to partially update SONiC configuration according to a patch of updates
+- A single, simple command to take a checkpoint of the full current SONiC config
+- A single, simple command to fully rollback current SONiC configs with to a checkpoint
+- [low-priority] A single simple command to fully replace current SONiC configs with a full config provided by an external user.
```
qiluo-msft, line 186: > [low-priority] A single simple command to fully replace current SONiC configs with a full config provided by an external user. [](http://example.com/codeflow?start=1&length=127)

When investigating gnmi interface, we found another function requirement: gnmi SetRequest could contains multiple `delete`, `replace`, `update` operations, for example two `replace` operations, one under the path under /PORT, one under the path /PORTCHANNEL.

ref: [https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#343-transactions](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#343-transactions)

The "fully replace" could be a special case for above SetRequest  requirement. As a gnmi MVP implementation, I believe supporting multiple `replace` operations will be the most high priority requirement.

I also think a CLI to replace several ConfigDB tables are handy.

ghooo: GNMI input can be converted to either:
- JsonPatch format and then we call `config apply-patch`, or
- Full target ConfigDB then we call `config replace`

We should let GCU decide the correct order of execution.

PR #743, opened 2021-01-26
cpu queue stats

PrabhuSreenivasan:


liat-grozovik:
@PrabhuSreenivasan can you please provide some more information on this motivation?
it is an md file but not based on the SONIC HLD template so context is missing. Appreciate if you can align it to an existing HLD or new one

PR #754, opened 2021-03-02
Sonic yang guidelines

maheshwari-mayank:
Updated SONiC YANG Guidelines with new YANG extensions.

### DIFF
--- a/doc/mgmt/SONiC_YANG_Model_Guidelines.md
+++ b/doc/mgmt/SONiC_YANG_Model_Guidelines.md
@@ -8,7 +8,7 @@
  |:---:|:-----------:|:------------------:|-----------------------------------|
  | 1.0 | 22 Aug 2019 | Praveen Chaudhary  | Initial version                   |
  | 1.0 | 11 Sep 2019 | Partha Dutta       | Adding additional steps for SONiC YANG  |
- | 1.0 |  2 Mar 2021 | Mayank Maheshwari  | Adding details of each extension used in SONiC YANG  |
+ | 1.1 |  2 Mar 2021 | Mayank Maheshwari  | Adding details of each extension used in SONiC YANG  |
 
 ## References
 | References 		    |     Date/Version    |   		    Link     	 	|


doc/mgmt/SONiC_YANG_Model_Guidelines.md
```diff
@@ -622,7 +623,180 @@ module sonic-port {
...
+                        type string {
+                            pattern "[1-9][0-9]?|[0-9]?";
+                        }
+                    }
+                }
+            }
+        }
+```
+
+#### 5. custom-validation: Use this extension when customized code is required for validation. Function name has to be given with this extension which will be called to perform custom validation.
```
ben-gale, line 729: I think you should state that this is specific to CVL, and that the provided function must be provided in the Mgmt Framework common containers. You could also point to the Mgmt Framework HLD for more information on custom validation.

doc/mgmt/SONiC_YANG_Model_Guidelines.md
```diff
@@ -622,7 +623,180 @@ module sonic-port {
...
+                        enum NON_IPV4;
+                        enum IPV6ANY;
+                        enum NON_IPV6;
+                    }
+                }
+            }
+        }
+```
+
+#### 6. dependent-on: Use this extension to define dependency on other table. Usually dependency can be specified using 'leafref' statement. This extension should to be used for cases where it is not possible to specify dependency using 'leafref'. In below example, edge_port and link_type nodes has 'when' condition which requires to check STP mode. This has dependency on STP table but not possible to specify using 'leafref'.
```
ben-gale, line 761: Curious - why would it not be possible to specify using a leafref?

ben-gale (commented):
Looks good - just some suggested clarifications (optional)

doc/mgmt/SONiC_YANG_Model_Guidelines.md
```diff
@@ -622,7 +623,180 @@ module sonic-port {
...
+        }
+        leaf qindex {
+            type string;
+        }
+        ....
+    }
+}
+```
+
+#### 4. map-leaf: Mapping tables in Redis are defined using nested 'list'. The outer 'list' is used for multiple instances of mapping. The inner 'list' is used for mapping entries for each outer list instance. Use 'map-leaf' extension to indicate nested 'list' and specify mapping between two fields which are also key fields of inner list.
```
arlakshm, line 680: Is this extension supported  in the python based library  or only CVL ?

maheshwari-mayank: As of now, it is supported in CVL only.

PR #761, opened 2021-03-16
Event and Alarm Framework HLD

spenugondaa:
Event and Alarm framework enables applications to raise notifications in response to a condition that may impact the health of the system that operator may be interested in.
1. Define an API and allow applications to raise events
2. Store the events in the DB 
3. Through CLI, REST and gNMI, user can fetch contents of the DB tables.
4. Message corresponding to an event is sent to remote syslog host
5. Allows operator to customize severities of events and enable/disable events through event profiles.
6. Through management framework, gNMI can subscribe to receive updates as events are raised.

Corresponding Coding PRs:
[Framework](https://github.com/Azure/sonic-buildimage/pull/7813)
[Schema for Tables](https://github.com/Azure/sonic-swss-common/pull/490)
[mgmt-framework changes](https://github.com/Azure/sonic-mgmt-framework/pull/85)
[mgmt-common changes](https://github.com/Azure/sonic-mgmt-common/pull/48)

Signed-off-by: Srinadh Penugonda <Srinadh_Penugondaa@dell.com>


doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+    Create a patch for these components by adding libeventnotify library and invoke the API. This however, requires these patches need to be maintained in the code forever.
+
+2.  Listen to syslog messages
+    As many of these components raises syslog messages on an important event, a listener can be implemented to read incoming syslog messages and raise 
+    events based on the message.
+    This however is heavy on performance due to the fact that listener has to parse each syslog message. Also listener need to maintain a map of messages to 
+    event-id and need to be aware of resource and other specific details. It need to be aware of nuances of alarm raising/clearing if the component follows
+    any specific logic. 
+
+Approach 1 is preferred.
```
liat-grozovik, line 711: Unfortunately it will not allow us to upgrade properly to newer version and each upgrade will requires a lot of work and manage patches. Therefore I believe this should be agreed in advance

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+
+## 3.2 DB Changes
+### 3.2.1 CONFIG DB
+### 3.2.2 APP DB
+Table EVENTPUBSUB is used for applications to write events and for eventd to access and process them.
```
liat-grozovik, line 716: If i am not mistaken you mentioned this is stored after the system restart as of that I don't understand how you store it in APP_DB. Can you please elaborate?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+Event History Table (EVENT) and Current Alarm Table (ALARM).
+To maintain various statistics of events, these two are introduced : EVENT_STATS and ALARM_STATS.
+
+### 3.2.4 ASIC DB
+### 3.2.5 COUNTER DB
+
+## 3.3 User Interface
+### 3.3.1 Data Models
+
+The following is sonic yang for events.
```
liat-grozovik, line 729: Change sonic to SONiC

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+    +--rw event-stats
+       +--ro state
+          +--ro events?       uint64
+          +--ro alarms?       uint64
+          +--ro acked?        uint64
+          +--ro cleared?      uint64
+          +--ro non-alarms?   uint64
+```
+
+The following is sonic yang for alarms.
```
liat-grozovik, line 805: Change sonic to SONiC

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+sonic(config)# logging server <ip> [log|event] 
+```
+Note: The 'logging server' command is an existing, already supported command. 
+It is only enhanced to take either 'log' or 'event' to indicate either native syslog messages or syslog messages corresponding to events alone are sent to the remote host.
+Support with VRF/source-interface and configuring remote-port are all backward comaptible and will be applicable to either 'log' or 'event' options.
+
+#### 3.3.2.3 Show Commands
+```
+sonic# show system health
+System Health: Normal, Warning or Error
```
liat-grozovik, line 899: Do we have SONiC native CLI commands also in the management framework? 
Do you mean that the SONiC native CLI command will be extended to support the new state? if so, please provide an example of how it will be integrated into the existing command and if there is a changes on it.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+
+sonic# show event 
+Sq No   State        Name                             Timestamp                   Description
+-----   --------     ------------------------------   --------------------------- --------------
+2292    -            TAM_SWITCH_ID_CHANGE             Wed Feb 10 18:08:27 2021    processSwitchTblEvent: Received new switch-id as 3        
+2291    raised       TEMPERATURE_EXCEEDED             Wed Feb 10 18:08:24 2021    temperatureCrossedThreshold: Current temperature of sensor/2 is 76 degrees
+182     cleared      TEMPERATURE_EXCEEDED             Wed Feb 10 09:52:12 2021    temperatureCrossedThreshold: Current temperature of sensor/2 is 76 degrees
+134     raised       TEMPERATURE_EXCEEDED             Wed Feb 10 09:32:01 2021    temperatureCrossedThreshold: Current temperature of sensor/2 is 76 degrees
+        
+sonic# show event details
```
liat-grozovik, line 916: Is there a limit number of the events? If there are thousands of them will they all be displayed in one shot?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+2292    -            TAM_SWITCH_ID_CHANGE             Wed Feb 10 18:08:27 2021    processSwitchTblEvent: Received new switch-id as 3        
+2291    raised       TEMPERATURE_EXCEEDED             Wed Feb 10 18:08:24 2021    temperatureCrossedThreshold: Current temperature of sensor/2 is 76 degrees
+182     cleared      TEMPERATURE_EXCEEDED             Wed Feb 10 09:52:12 2021    temperatureCrossedThreshold: Current temperature of sensor/2 is 76 degrees
+134     raised       TEMPERATURE_EXCEEDED             Wed Feb 10 09:32:01 2021    temperatureCrossedThreshold: Current temperature of sensor/2 is 76 degrees
+        
+sonic# show event details 
+Event History Details - 2292
+-------------------------------------------
+Sequence Number:   2292
+Severity:          informational
```
liat-grozovik, line 920: I think you should align naming to be the same. In some places it is INFORMATIONAL and here it is informational. Need to have 1 type of naming convention .

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+Name:              TAM_SWITCH_ID_CHANGE
+Description:       processSwitchTblEvent: Received new switch-id as 3
+Timestamp:         Wed Feb 10 18:08:27 2021
+Source:            3
```
liat-grozovik, line 924: what is 3 means?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+
+Event Summary
+-------------------------------------------
+Raised:                 5
+Ack:                    1
+Cleared:                4
+Events:                 20
+-------------------------------------------
+
+sonic# show event severity critical
```
liat-grozovik, line 947: Suggest to have severity column for all commands so even if you filter you will see the same look and feel.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+Sq No   State        Name                             Timestamp                   Description
+-----   --------     ------------------------------   --------------------------- --------------
+2292    -            TAM_SWITCH_ID_CHANGE             Wed Feb 10 18:08:27 2021    processSwitchTblEvent: Received new switch-id as 3        
+2291    raised       TEMPERATURE_EXCEEDED             Wed Feb 10 18:08:24 2021    temperatureCrossedThreshold: Current temperature of sensor/2 is 76 degrees
+
+sonic# show event sequence-id from 2290 to 2300
+Sq No   State        Name                             Timestamp                   Description
+-----   --------     ------------------------------   --------------------------- --------------
+2292    -            TAM_SWITCH_ID_CHANGE             Wed Feb 10 18:08:27 2021    processSwitchTblEvent: Received new switch-id as 3        
+2291    raised       TEMPERATURE_EXCEEDED             Wed Feb 10 18:08:24 2021    temperatureCrossedThreshold: Current temperature of sensor/2 is 76 degrees
```
liat-grozovik, line 962: Suggest to have the State similar in all show commands. Better to have it started with capital letter.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+Source:            /psu/2
+Name:              PSU_FAULT
+Description:       /psu/2 has experienced a fault 
+Raise-time:        Wed Feb 10 08:08:24 2021
+Ack-time:          Wed Feb 10 08:10:14 2021 
+New:               true
+Acknowledged:      true
+
+sonic# show alarm summary
+Active-alarm Summary
```
liat-grozovik, line 1016: why - is needed?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+-------------------------------------------
+Total:            6
```
liat-grozovik, line 1018: Suggest Total to be the last one

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+During normal operation, eventd reads, processes whenever a new record is added to the table.
+
+When eventd is restarted, events and alarms raised by applications will be waiting in a queue while eventd is coming up.
+When eventd eventually comes back up, it reads those records in the queue.
+
+# 6 Scalability
+In this feature, scalability applies to Event History Table (EVENT). As it is persistent and it records every event generated on the system, to protect
+against it growing indefinitely, user can limit its size through a manifest file.
+By default, the size of Event History Table is set to 40k events or events for 30 days - after which, older records are discarded to make way for new records.
+
```
liat-grozovik, line 1083: Please add a reference to the show techsupport. I believe we need the configuration as well as the 'show' commands in it.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+|:---:|:-----------:|:------------------:|-----------------------------------                       |
+| 0.1 | 03/20/2021  | Srinadh Penugonda  | Initial Version                                          |
+
+# About this Manual
+This document provides general information on the implementation and functionality of Event and Alarm Framework in SONiC.
+Wherever CLI is specified, it is the CLISH cli that is referred.
+
+# Scope
+This document describes the high-level design of Event and Alarm Framework feature.
+It is not in the scope of this feature to update all the applications to raise events and alarms.
```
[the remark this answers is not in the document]
spenugondaa: As mentioned in the HLD, the scope is to define and implement the framework but not to update any applications. It is upto the application owners, new feature implementers to use the defined API and raise/clear events and alarms.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+   Out of memory, temperature crossing a threshold, and so on, are examples of conditions when the alarms are raised.
+   Such conditions are dynamic: a faulty software/hardware component encounters the above such condition and **may** come out of that situation when the condition is resolved.
+   So, these conditions have a state: RAISED, CLEARED or ACKNOWLEDGED.
+   
+   Events are sent as the condition progresses through each of these states.
+   Each of these events is characterized by "state" in addition to "severity".
+   
+   An application *raises* an alarm when it encounters a faulty condition by sending an event with a state: RAISED.
+   After the application recovers from the condition, that alarm is *cleared* by sending an event with a state: CLEARED. 
+   An operator could *acknowledge* an alarm. This indicates that the user is aware of the faulty condition.
```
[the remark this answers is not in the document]
spenugondaa: Yes.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+   All events with a RASIED state get recorded in a table, by name, "ALARM" in addition to getting recorded in Event History Table ( only events corresponding to an alarm has state ).
+   When a component that raised the alarm clears it ( by sending an event with CLEARED state ), the alarm record is removed from ALARM table.
+   An user acknowledging a particular alarm will NOT remove that alarm record from this table.
+   
+   In effect, ALARM table contains outstanding alarms that need to be cleared by those components who raised them.
+   This table is NOT persisted and its contents are cleared with a reload.
+   
+In summary, the framework provides both current and historical event status of software and physical entities of the system through ALARM and EVENT tables. 
+
+As mentioned above, each event has an important characteristic: severity. SONiC defines following severities for events.
```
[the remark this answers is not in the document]
spenugondaa: There is nothing to clear events which are one-shot. We can support clearing alarms.  I will update the HLD

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+        {
+        	name: "PORT_OPSTATUS_UPDATE",
+        	severity: "INFORMATIONAL",
+        	enable: "true"
+        }
+    ]
+}
+```
+
+Operator can download default event profile to a remote host.
```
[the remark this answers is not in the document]
spenugondaa: I added the template for an event profile just above line 162. Through event profile, operator can change default properties of events and alarms. Please refer to section 3.1.5
Yes. Eventd expects the event profiles to be placed at /etc/sonic/evprofile/. I will emphasize this in section 3.1.5

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: The pusbub table which is used by applications to publish events and eventd to receive them  is "housed" in app-db. 
HOwever, as per the review comments, I am going to migrate this to a new DB specifically for event framework. I will update the HLD.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: The command I specified here is mgmt-framework command. I wasnt aware that the native CLI has a system health command. Will look in to and update/enhance accordingly.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: In this particular case, it indicates the new value of the parameter. _Source_ ideally indicates the object/instance/entity that caused this event/alarm.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+   
+   User logging in, authentication failure, configuration changed notification are all examples of events.
+
+*  Alarms
+
+   Alarms are notifications raised for conditions that could be cleared by correcting or removal of such conditions.
+   
+   Out of memory, temperature crossing a threshold, and so on, are examples of conditions when the alarms are raised.
+   Such conditions are dynamic: a faulty software/hardware component encounters the above such condition and **may** come out of that situation when the condition is resolved.
+   So, these conditions have a state: RAISED, CLEARED or ACKNOWLEDGED.
```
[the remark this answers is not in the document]
spenugondaa: I think you are referring this : https://github.com/Azure/SONiC/wiki/PFC-Watchdog-Design#26-events-for-resetting-pfc-wd
I dont see any alarms there. I only see three external events , out of which, counter reset is ignored. 
PortDown marks queues are marked as stormed. Per my understanding, PFC can "raise" an alarm when this happened and can "clear" the alarm when port is up which makes storm is fixed. Developer knows better about it.
Queue reconfiguration and removal - it seems it is one-shot event. If PFC is disabled they do one thing and enabled, they do it other. I am not in a position to make a call on whether it is one-shot event or alarm.
The idea with alarm management is simple and clear : RASE an alarm if the application encountered any issue and CLEAR if the issue is cleared.

doc/event-alarm-framework/event-alarm-framework.md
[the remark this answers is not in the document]
spenugondaa: As mentioned in the 3.1.2, every event causes eventd to send a log message. User acknowledging an alarm causes an event be sent by mgmt-framework and eventd WILL send a syslog message which in turn will be shown in the log.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+   Each of these events is characterized by "state" in addition to "severity".
+   
+   An application *raises* an alarm when it encounters a faulty condition by sending an event with a state: RAISED.
+   After the application recovers from the condition, that alarm is *cleared* by sending an event with a state: CLEARED. 
+   An operator could *acknowledge* an alarm. This indicates that the user is aware of the faulty condition. 
+
+   Overall system health and system LED state is deduced from the severities of alarms.
+   An acknowledged alarm is taken out of consideration from deciding system health and system LED is updated accordingly.
+
+   Operator must acknowledge or investigate alarms with a high severity.
```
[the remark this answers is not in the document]
spenugondaa: correct. Bad choice of words on my part.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+
+
+# Revision
+| Rev |     Date    |       Author       | Change Description                                       |
+|:---:|:-----------:|:------------------:|-----------------------------------                       |
+| 0.1 | 03/20/2021  | Srinadh Penugonda  | Initial Version                                          |
+
+# About this Manual
+This document provides general information on the implementation and functionality of Event and Alarm Framework in SONiC.
+Wherever CLI is specified, it is the CLISH cli that is referred.
```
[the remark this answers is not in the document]
spenugondaa: Done. The scope is **only** to provide the framework. Updating **_any_** application to raise events/alarms is NOT in the scope of this HLD.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+   An operator could *acknowledge* an alarm. This indicates that the user is aware of the faulty condition. 
+
+   Overall system health and system LED state is deduced from the severities of alarms.
+   An acknowledged alarm is taken out of consideration from deciding system health and system LED is updated accordingly.
+
+   Operator must acknowledge or investigate alarms with a high severity.
+
+Both events and alarms get recorded in the DB. 
+
+1.  Event History Table
```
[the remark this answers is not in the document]
spenugondaa: I will be introducing a new DB by name, EVENT_DB to house all the tables of event framework.. I will update the HLD.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: All the tables used in event framework will be housed in new DB called EVENT_DB

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Done

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Done

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: There is no limit on number of events displayed here. The command tries to display **ALL** the records in the table. However user can stop the output at any point of the time as it uses pagination.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Updated

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: We may cross 80-column limit. Will explore.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: This would make the o/p to cross well above the 80-column limit.. The severity is anyway displayed with the details option.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: removed

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Done

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+#### 3.1.4.4 System LED
+The original requirement was to change LED based on severities of the events. But on most of the platforms the system/power/fan LEDs are managed by the BMC. 
+BMC (baseboard management controller) is an embedded system that manages various platform elements like fan, PSU, temperature sensors. 
+There is an API that can be invoked to control LED, but not all platforms will support that API if they are fully controlled by the BMC.
+So, on certain platforms, system LED could not represent events on the system.
+
+Another issue is: Currently pmon controls LED, and as eventd now tries to change the very same LED, this leads to conflicts. 
+A mechanism must exist for one of these to be master.
+
+The proposed solution is to have a system health parameter in the DB populated by eventd and have pmon use this parameter in conjunction with existing logic to update system LED.
```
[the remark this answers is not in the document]
spenugondaa: I am in touch with pmon (Kebo Liu) to decide right approach.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+The gNMI server (gnoi_client.go, gnoi.go, sonic_proto, transl_utils.go) need to be extended to support the RPC to support similar operations for gNMI.
+
+### 3.1.7 Event History Table and Current Alarm Table
+The Event History Table (EVENT) and Current Alarm List Table (ALARM) stored in state db. 
+The size of Event History Table is 40k records or 30 days worth of events which ever hits earlier.
+A manifest file will be created with parameters to specify the number and number of days limits for
+eventd to read and enforce them.
+
+```
+root@sonic:/etc# cat eventd.json
```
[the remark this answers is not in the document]
spenugondaa: will explore.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: All show commands are updated with state names uniform (raised, cleared, acknowledged). The 'show event status' command displays counters corresponding to states hence they are named as Raised, Cleared and Acknowledged.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: show tech will include output of 'show event recent 1hr' and 'show alarm all' commands. I will update the HLD.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+gNMI clients can subscribe to receive events as they are raised. Subscribing through REST is being evaluated.
+
+CLI and REST/gNMI clients can query either table with filters - based on severity, delta based on timestamp, sequence-id etc.,
+
+Application owners need to identify various conditions that would be of interest to the operator and use the framework to raise events/alarms. 
+
+## 1.1 Requirements
+
+
+### 1.1.1 Functional Requirements
```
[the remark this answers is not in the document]
spenugondaa: What we are providing here is a framework. The feature command enables/disables services. Not sure if eventd should be included there.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+| 8.1   | show alarm [ detail \| summary \| severity \| timestamp <from> <to> \| recent <5min\|1hr\|1day> \| sequence-number <from> <to> \| all]          |                    |
+| 8.2   | show event [ detail \| summary \| severity \| timestamp <from> <to> \| recent <5min\|1hr\|1day> \| sequence-number <from> <to>]                 |                     |
+| 8.3   | show event profile                                                              |                     |
+| 8.4   | alarm acknowledge <sequence id>                                                 |                     |
+| 8.5   | logging server <ip> [ log \| event ]                                             | default is 'log'   |
+| 8.6   | event profile   [ default \| name-of-file ]	                                  |                     |
+| 9     | gNMI subscription                                                               |                     |
+| 9.1   | Subscribe to openconfig Event container and Alarm container. All events and alarms published to gNMI subscribed clients. |                    |
+| 10    | Clear all events (Best effort)                                                  |                     |
```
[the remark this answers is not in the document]
spenugondaa: will include the command and udpate the HLD.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+   An user acknowledging a particular alarm will NOT remove that alarm record from this table.
+   
+   In effect, ALARM table contains outstanding alarms that need to be cleared by those components who raised them.
+   This table is NOT persisted and its contents are cleared with a reload.
+   
+In summary, the framework provides both current and historical event status of software and physical entities of the system through ALARM and EVENT tables. 
+
+As mentioned above, each event has an important characteristic: severity. SONiC defines following severities for events.
+
+- CRITICAL : A critical condition exists and requires immediate action. A critical event may trigger if one or more hardware components fail, or one or more hardware components exceed temperature thresholds.
```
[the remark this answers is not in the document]
spenugondaa: will use syslog severities. Will update the HLD.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: The community has comment that the tables need to have separate DB for event/alarm tables. 
Does it mean I need to have a new redis instance or a new DB in one of the existing instances?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+   
+   Events are sent as the condition progresses through each of these states.
+   Each of these events is characterized by "state" in addition to "severity".
+   
+   An application *raises* an alarm when it encounters a faulty condition by sending an event with a state: RAISED.
+   After the application recovers from the condition, that alarm is *cleared* by sending an event with a state: CLEARED. 
+   An operator could *acknowledge* an alarm. This indicates that the user is aware of the faulty condition. 
+
+   Overall system health and system LED state is deduced from the severities of alarms.
+   An acknowledged alarm is taken out of consideration from deciding system health and system LED is updated accordingly.
```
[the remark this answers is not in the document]
spenugondaa: Added a life cycle diagram and an example scenario on how alarms affects system LED

doc/event-alarm-framework/event-alarm-framework.md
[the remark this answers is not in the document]
spenugondaa: The openconfig yang for alarms defined severities as critical, major, minor, warning and unknown.
https://github.com/openconfig/public/blob/master/release/models/system/openconfig-alarm-types.yang#L92
So, we will keep them as defined in the HLD (critical, major, minor, warning and info) and map them internally.

doc/event-alarm-framework/event-alarm-framework.md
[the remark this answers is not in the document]
spenugondaa: A new db, by name EVENT_DB, off redis2 instance is being introduced to host all the tables of the framework.

doc/event-alarm-framework/event-alarm-framework.md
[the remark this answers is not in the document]
spenugondaa: introducing 'clear event history' command to clear event history table and stats. The alarm table is not impacted as records in the alarm table are raised by applications and can only be cleared by respective applications. 
As mentioned operator can ack an alarm.

doc/event-alarm-framework/event-alarm-framework.md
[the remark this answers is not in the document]
spenugondaa: 'clear event history' is being introduced and CLI section is updated.

doc/event-alarm-framework/event-alarm-framework.md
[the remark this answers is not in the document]
spenugondaa: Will support these parameters in config_db.json  in later versions.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: We notice many patches exist in the system. Would like to keep patches.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Updated.

spenugondaa:
@liat-grozovik I have updated the HLD with all the review comments. Please let me know if I missed anything or if there are other comments.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1100 @@
...
+
+## 1.1 Requirements
+
+
+### 1.1.1 Functional Requirements
+
+| ID    | Requirement                                            		                  | Comment             |
+| :---  | :----                                                 		                  | :---                |
+| 1     | Provide API via library for apps to publish events                              |                     |
+| 2     | Provide API via library for apps to publish alarms                   |                |
```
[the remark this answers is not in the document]
spenugondaa: Added it

doc/event-alarm-framework/event-alarm-framework.md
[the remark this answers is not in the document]
spenugondaa: I have reached out to Gouhan and Madhu on the need to use a new DB. So far, i have not heard anything concrete on why cant we use state-db to house these tables from event framework. I have tested this out by sending events every 100 milli seconds with 40k alarms. I have not seen adverse effects. So, using state-db for tables of event framework.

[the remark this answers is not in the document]
spenugondaa: INtroducing new redis instance and a new DB that uses this new instance.

anshuv-mfst:
Hi @lguohan - please assign for review & merge.

bhavini-gada:
@liat-grozovik - Can you please complete review & add your approval ?

spenugondaa:
> @spenugondaa would you please add the code PRs into this PR for tracking purpose? Thanks.

Done

bhavini-gada:
@liat-grozovik - As discussed this morning, please add your approval so we can complete the merge.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+EVENT_STATS Table:
+==============================
+
+Key             : id
+
+id              : key {state}
+events          : Total events raised {uint64}
+raised          : Total alarms raised {uint64}
+cleared         : Total alarms cleared {uint64}
+acked           : Total alarms acknowledged {uint64}
```
liat-grozovik, line 733: Why do we need for event to be acked? I thought only Alarm can be acknowledged

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+Schema for ALARM_STATS table is as below. When an alarm of particular severity is cleared,
+the corresponding severity counter is decremented. 
+```
+ALARM_STATS Table:
+==============================
+
+Key                       : id
+
+id                        : key {state}
+alarms                    : Number of active alarms {uint64}
```
liat-grozovik, line 793: what does active means? cleared/acknowledged?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+    Create a patch for these components by adding libeventnotify library and invoke the API. This however, requires these patches need to be maintained in the code forever.
+
+2.  Listen to syslog messages
+    As many of these components raises syslog messages on an important event, a listener can be implemented to read incoming syslog messages and raise 
+    events based on the message.
+    This however is heavy on performance due to the fact that listener has to parse each syslog message. Also listener need to maintain a map of messages to 
+    event-id and need to be aware of resource and other specific details. It need to be aware of nuances of alarm raising/clearing if the component follows
+    any specific logic. 
+
+Approach 1 is preferred.
```
liat-grozovik, line 835: I think HLD before it is approved should reflect the decision made. I don't see any update on that nor discussion with an agreement.
I don't think approach 1 is the right one as it as it will cause lots of maintenance as SONiC keep updating such components.
@lguohan  i appreciate if you can provide your input.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+
+EVPROFILE table is used by mgmt-framework to communicate name of the custom event profile when configured through NBI. 
+Eventd reads the file name from this table and merges it with its static_event_map.
+
+## 3.3 User Interface
+### 3.3.1 Data Models
+
+The following is SONiC yang for events.
+```
+module: sonic-event
```
liat-grozovik, line 853: For Yang model i think that sub group should lead and approve this HLD.
@zhangyanzhao please assign it to someone from the subgroup.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+       +---w input
+       |  +---w file-name?   string
+       +--ro output
+          +--ro status?   string
+```
+
+openconfig alarms yang is defined at [here](https://github.com/openconfig/public/blob/master/release/models/system/openconfig-alarms.yang)
+
+### 3.3.2 CLI
+#### 3.3.2.1 Exec Commands
```
liat-grozovik, line 988: As previously noted I think the CLI commands should be detailed described in an HLD.
I would like to see the command name, the parameters, the default behaviour and the output of the command.

My assumption is that these commands are provided as native SONiC CLI commands which are also to be shared in the command reference guide. it is recommended to have in the HLD the same format as in the command reference guide and later it can be simple copy/paste.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+sonic(config)# logging server <ip> [log|event] 
+```
+Note: The 'logging server' command is an existing, already supported command. 
+It is only enhanced to take either 'log' or 'event' to indicate either native syslog messages or syslog messages corresponding to events alone are sent to the remote host.
+Support with VRF/source-interface and configuring remote-port are all backward comaptible and will be applicable to either 'log' or 'event' options.
+
+#### 3.3.2.3 Show Commands
+```
+sonic# show event profile 
+--------------------------
```
liat-grozovik, line 1029: Why this output is considered as table? It is a list, right? In this case my suggestion for the output to be

Active Event Profile:
  myProfile.json

Available Event Profiles:
  default.json
  myProfile.json
  userProfile.json

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+default.json
+myProfile.json
+userProfile.json
+
+sonic# show event [ details | summary | severity <sev> | start <from-ts> end <to-ts> | recent <5min|60min|24hr> | id <seq-id> | from <seq-id> to <seq-id> ] 
+
+'show event' commands would display all the records in EVENT table.
+
+sonic# show event
```
liat-grozovik, line 1044: There is an option to general CLIs from yang model in automated way. These PRs are available for that.
https://github.com/Azure/sonic-utilities/pull/1650 
https://github.com/Azure/sonic-utilities/pull/1644
The output is then similar to other CLI commands which i think is a better approach to make is a unified.
If you don't want to use the PRs i suggest you use the table format as in recent commands (like pbh)

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+
+sonic# show event [ details | summary | severity <sev> | start <from-ts> end <to-ts> | recent <5min|60min|24hr> | id <seq-id> | from <seq-id> to <seq-id> ] 
+
+'show event' commands would display all the records in EVENT table.
+
+sonic# show event 
+----------------------------------------------------------------------------------------------------------------------------
+Id           Action          Severity   Name                           Timestamp                   Description                                                  
+----------------------------------------------------------------------------------------------------------------------------
+1            -               WARNING    CUSTOM_EVPROFILE_CHANGE        2021-05-19T21:38:27.455Z    :- handle_custom_evprofile: Custom Event Profile x.json is applied.
```
liat-grozovik, line 1048: IMO Description start with :- is very strange. Why do we have it?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+2            RAISE           CRITICAL   DUMMY_ALARM                    2021-05-19T21:39:31.622Z    :- signalHandler: Raising simulated alarm         
+3            CLEAR           CRITICAL   DUMMY_ALARM                    2021-05-19T21:42:34.371Z    :- signalHandler: Clearing simulated alarm        
+4            RAISE           CRITICAL   DUMMY_ALARM                    2021-05-19T21:46:14.371Z    :- signalHandler: Raising simulated alarm         
+5            ACKNOWLEDGE     CRITICAL   DUMMY_ALARM                    2021-05-19T21:48:05.845Z    Alarm id 4 ACKNOWLEDGE.                           
+6            UNACKNOWLEDGE   CRITICAL   DUMMY_ALARM                    2021-05-19T21:53:24.484Z    Alarm id 4 UNACKNOWLEDGE.                         
+7            CLEAR           CRITICAL   DUMMY_ALARM                    2021-05-19T21:55:54.977Z    :- signalHandler: Clearing simulated alarm        
+
+sonic# show event details 
+----------------------------------------------
```
liat-grozovik, line 1057: the only added info in the details option is the source. Is it because you don't have enough space on the table format for another columns? If so, OK.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+Id:                  2 
+Action:              RAISE 
+Severity:            CRITICAL 
+Type:                DUMMY_ALARM 
+Timestamp            2021-05-19T21:39:31.622Z 
+Description:         :- signalHandler: Raising simulated alarm 
+Source:              simulation 
+
+sonic# show event from 2 to 5
```
liat-grozovik, line 1133: What happens if the request for entries is not available? what should be the CLI command behaviour?
For example 2 to 5 but only 3?
Is there any check for legal range?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+----------------------------------------------------------------------------------------------------------------------------
+Id           Action          Severity   Name                           Timestamp                   Description                                                  
+----------------------------------------------------------------------------------------------------------------------------
+2            RAISE           CRITICAL   DUMMY_ALARM                    2021-05-19T21:39:31.622Z    :- signalHandler: Raising simulated alarm         
+3            CLEAR           CRITICAL   DUMMY_ALARM                    2021-05-19T21:42:34.371Z    :- signalHandler: Clearing simulated alarm        
+4            RAISE           CRITICAL   DUMMY_ALARM                    2021-05-19T21:46:14.371Z    :- signalHandler: Raising simulated alarm         
+5            ACKNOWLEDGE     CRITICAL   DUMMY_ALARM                    2021-05-19T21:48:05.845Z    Alarm id 4 ACKNOWLEDGE.
+
+sonic# show event start 2021-05-19T21:39:31.622Z end 2021-05-19T21:46:14.371Z
```
liat-grozovik, line 1142: What happens if the request for timestamp is not available? what should be the CLI command behaviour?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+'show alarm' command would display all the *active* alarm records in ALARM table. Acknowledged alarms wont be shown here.
+
+sonic# show alarm
+----------------------------------------------------------------------------------------------------------------------------
+Id           Severity   Name                           Timestamp                   Description                                                  
+----------------------------------------------------------------------------------------------------------------------------
+14           WARNING    TEMPERATURE_EXCEEDED           2021-05-20T00:47:52.992Z    :- temperatureCrossedThreshold: Current temperature of sensor/2 is 76 degrees         
+16           WARNING    PSU_FAULT                      2021-05-20T02:16:42.611Z    :- /psu/2 has experienced a fault 
+
+sonic# show alarm all
```
liat-grozovik, line 1159: Not clear what is the difference between show alarm without any additional input and all.
in the example there is only one alarm that is missing but i could not understand why 16 is missing with show alarm is executed. Can you please share the logic? Maybe it is somewhere else in the doc?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+Severity:          CRITICAL
+Source:            /sensor/2
+Name:              TEMPERATURE_EXCEEDED
+Description:       temperatureCrossedThreshold: Current temperature of sensor/2 is 76 degrees
+Raise-time:        Wed Feb 10 18:08:24 2021
+Ack-time:          
+New:               true
+Acknowledged:      false
+
+sonic# show alarm from 14 to 16
```
liat-grozovik, line 1182: Same question for range in event

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+
+# 4 Flow Diagrams
+![Sequence Diagram](event-alarm-framework-seqdiag.png)
+
+# 5 Warm Boot Support
+## 5.1 Application warm boot
+Applications confirming to the warm boot, should have stored their state and compare current values against previous values.
+Such compliant application also "remembers" that it raised an event before for a specific condition.
+They would 
+*  not raise alarms/events for the same condition that it raised pre warm boot
```
liat-grozovik, line 1228: I believe this is hard req for application and logic that should not resides in the application but in the infra. 
@lguohan, @yxieca what do you think?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+When eventd is restarted, events and alarms raised by applications will be waiting in a queue while eventd is coming up.
+When eventd eventually comes back up, it reads those records in the queue.
+
+# 6 Scalability
+In this feature, scalability applies to Event History Table (EVENT). As it is persistent and it records every event generated on the system, to protect
+against it growing indefinitely, user can limit its size through a manifest file.
+By default, the size of Event History Table is set to 40k events or events for 30 days - after which, older records are discarded to make way for new records.
+
+# 7 Showtech support
+The techsupport bundle is upgraded to include output of "show event recent 60min” and “show alarm all”.
```
liat-grozovik, line 1246: The show tech support can also be executed with the option --since.
I believe you should refer to that option in the design and if it is used make sure it is included in the option of the show event recent. Maybe it is even recommended to use the same format of the --since option as it is also taken for logs.

liat-grozovik (commented):
Sorry for the late reply on this HLD. Added few comments.
I would suggest that for the infra to 'proof' is working to have pre defined list of events and alarms used in this 1st delivery.

It would be also very effective to provide timeline of this HLD as it seems not all will be included in the first delivery. If i am mistaken and this HLD will be fully contributed and the questions raised here are all covered, no need.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+2. Alarm Statistics Table
+
+   Statistics on number of alarms per severity are maintained in ALARM_STATS table.
+   When application raises an alarm, the counter corresponding to that alarm's severity is increased by 1.
+   When the alarm is cleared or acknowledged, the corresponding severity counter will be reduced by 1.
+   This table categorizes "active" alarms per severity.
+
+As mentioned above, each event has an important characteristic: severity. SONiC uses following severities for events and alarms as defined in opeconfig alarm yang.
+
+- CRITICAL : Requires immediate action. An critical event may trigger if one or more hardware components fail, or one or more hardware components exceed temperature thresholds.
```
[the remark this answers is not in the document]
spenugondaa: Updated.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+
+In addition to the above tables, the framework maintains various statisitcs.
+
+1. Event Statistics Table
+
+   Statistics on number of events and alarms are maintained in EVENT_STATS table.
+
+2. Alarm Statistics Table
+
+   Statistics on number of alarms per severity are maintained in ALARM_STATS table.
```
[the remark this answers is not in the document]
spenugondaa: Yes. Alarm stats will be cleared as alarms are not persistent. 
The persistent event history contains contains raise/clear/ack/un-ack events corresponding to alarms.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+  ( maps to log-alert )
+- MAJOR : Requires escalation or notification. For example, a major alarm may trigger if an interface failure occurs, such as a port channel being down.
+  ( maps to log-critical )
+- MINOR : If left unchecked, might cause system service interruption or performance degradation. An alarm with minor severity requires monitoring or maintenance.
+  ( maps to log-error )
+- WARNING : It may or may not result in an error condition.
+  ( maps to log-warning )
+- INFORMATIONAL : Does not impact performance. NOT applicable to alarms.
+  ( maps to log-notice )
+
```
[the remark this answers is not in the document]
spenugondaa: This has been answered before. 
The reason is openconfig alarm yang is using above enums and not using syslog log levels.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+| 7.2.2 | Openconfig interface to pull alarm summary information.                         |                     |
+|       | Counters for Total, Critical, Major, Minor, Warning, Acknowledged               |                     |
+| 7.2.3 | Openconfig interface to pull alarms using following filters                     |                     |
+|       | - All (pull all events)                                                         |                     |
+|       | - Severity.                                                                     |                     |
+|       | - Recent alarms (eg., last 5 minutes, one hour, one day).                       |                     |
+|       | - Records between two timestamps, one timestamp and end, and beginning and a timestamp. |                     |
+|       | - All records between two Sequence Numbers (incl end and begin)                 |                     |
+| 7.2.4 | Openconfig interface  to acknowledge an alarm.                                  |                     |
+| 8     | CLI commands                                                                    |                     |
```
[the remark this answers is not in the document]
spenugondaa: This has been emphasized in the very first section: 

Note: Wherever CLI is specified, it is the CLISH cli that is referred - SONiC native (CLICK) CLI is not updated for this feature.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+When operator configures a host with 'event' type, it receives *only* log messages corresponding to events.
+Support for VRF/source-interface/UDP port are all are applicable for 'event' type.
+
+#### 3.1.4.2 REST
+Subcribing through REST to receive event notifications is currently being evaluated.
+
+#### 3.1.4.3 gNMI
+gNMI clients can subscribe to receive event notifications. Subscribed gNMI clients receive event fields as in the DB and 
+there is no customization of these fileds similar to syslog messages.
+
```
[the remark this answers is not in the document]
spenugondaa: Same gNMI interface is used. 
The above comment only mentions that unlike syslog messages which will be formatted, data sent over gNMI is purely what applications sends.

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+TODO: add definitions of protobuf spec
+
+#### 3.1.4.4 System LED
+The original requirement was to change LED based on severities of the events. But on most of the platforms the system/power/fan LEDs are managed by the BMC.
```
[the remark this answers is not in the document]
spenugondaa: The HLD is only for providing infra. The applications that want to add alarms/events and system LED support using this framework can be discussed/added in separate threads.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Only alarms can be acknowledged but when an alarm is acknowledged or cleared or raised, it is an event that is sent by application.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: An Active alarm is the one that is not cleared yet by the application.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Would appreciate a guidance on this one by the community.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: I am confused. Didnt we add description/example output of commands in section 3.2.2 ?

Will update them in the command ref guide format.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Your comment is about these that separates different data?
"--------------------------"

The o/p is displaying two parts: one is active profile and list of profiles.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Will change it.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: yes

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: If the requested info is not there, no o/p is shown just like any other command. 
The range is 1 - 40,000

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: If the requested info is not there, no o/p is shown just like any other command.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: This is mentioned above these commands: "'show alarm' command would display all the *active* alarm records in ALARM table. Acknowledged alarms wont be shown here.",

As mentioned, 'show alarm' wont display acknowledged alarms. 

'show alarm all' displays acknowledged alarms as well.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: There is no range as the sequence-id can grow until 64 bit.

doc/event-alarm-framework/event-alarm-framework.md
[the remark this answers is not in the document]
liat-grozovik: Unless SONiC moves to work only on top of CLISH I don't understand how this path can be taken.
@lguohan I believe we need a clear definition to avoid such comments for new HLDs to come.

As for this HLD I would recommend to have at least the show commands. The configuration if needed can be done via the config db if needed but the feature is kind of limited if the user need to read the redis db to understand the system status. And it can be very simple with the autogenerated CLI executed manually on top of the yang model.

doc/event-alarm-framework/event-alarm-framework.md
liat-grozovik: what i am referring is check for validity such as if the user provides 16 to 14, will you check and provide proper error notification? is 14 to 14 is allowed? Just add a comment what type of validation and behaviour is expected.

doc/event-alarm-framework/event-alarm-framework.md
liat-grozovik: i think it is better to define the CLI in an opposite way
'show alarm' is for all
'show alarm active'

doc/event-alarm-framework/event-alarm-framework.md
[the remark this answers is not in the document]
liat-grozovik: Could you please clarify it in the doc itself?

doc/event-alarm-framework/event-alarm-framework.md
bhavini-gada: @liat-grozovik - HLD should not have to cover each & every validation that will be added. Isn't it ?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+
+Note: Wherever CLI is specified, it is the CLISH cli that is referred - SONiC native (CLICK) CLI is not updated for this feature.
+
+# Scope
+This document describes the high-level design of Event and Alarm Framework.
+It is not in the scope of the framework to update ANY of the applications to raise events and alarms.
+
+# 1 Feature Overview
+
+The Event and Alarm Framework feature provides a centralized framework for applications in SONiC to raise notifications and store them for NBIs to listen and fetch to monitor the device.
```
yozhao101, line 74: Can you please specify what is full name of `NBI`?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+
+Events and Alarms are means to indicate a change in the state of the system that operator may be interested in.
```
yozhao101, line 76: `are means` ---> `means`?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+1.  Event History Table
+
+    All events get recorded in the event history table, by name, "EVENT". EVENT table contains history of all events generated by the system.
+    This table is persisted across system restarts of any kind, including restore to factory defaults and SW upgrades and downgrades.
+
+2. Current Alarm Table
+
+   All events with an action field of *RAISE* get recorded in a table, by name, "ALARM" in addition to getting recorded in Event History Table ( only events corresponding to an alarm has action field ).
+   When an application that raised the alarm clears it ( by sending an event with action *CLEAR* ), the alarm record is removed from ALARM table.
+   An user acknowledging a particular alarm will NOT remove that alarm record from this table; only when application clears it, the alarm is removed from ALARM table.
```
yozhao101, line 115: `An user` ---> `A user`?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+   In effect, ALARM table contains outstanding alarms that need to be cleared by those applications who raised them.
+   This table is NOT persisted and its contents are cleared with a reload.
+
+In summary, the framework provides both current and historical event status of software and physical entities of the system through ALARM and EVENT tables. 
+
+In addition to the above tables, the framework maintains various statisitcs.
+
+1. Event Statistics Table
+
+   Statistics on number of events and alarms are maintained in EVENT_STATS table.
```
yozhao101, line 126: `Event` table containes the statistics of `alarms`?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+
+2. Alarm Statistics Table
+
+   Statistics on number of alarms per severity are maintained in ALARM_STATS table.
+   When application raises an alarm, the counter corresponding to that alarm's severity is increased by 1.
+   When the alarm is cleared or acknowledged, the corresponding severity counter will be reduced by 1.
```
yozhao101, line 132: The severity counter will be deduced by 1 if and only if the corresponding alarm was cleared, right?

doc/event-alarm-framework/event-alarm-framework.md
yozhao101: Do we need two separate tables: `Event` and `Alarm`? Do we consider alarm is a kind of abnormal event?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+   This table categorizes "active" alarms per severity.
+
+As mentioned above, each event has an important characteristic: severity. SONiC uses following severities for events and alarms as defined in opeconfig alarm yang.
```
yozhao101, line 135: SONiC uses following severities for events and alarms ---> SONiC uses following severities for events?

doc/event-alarm-framework/event-alarm-framework.md
```diff
@@ -0,0 +1,1265 @@
...
+Event consumer class in eventd container receives and processes the received event. 
+Event consumer manages received events, updates event history table, current alarm table, event_stats table and alarm_stats tables and invokes logging API, which constructs message and sends it over to syslog. 
+
+Operator can chose to change properties of events with the help of event profile. Default
+event profile is available at */etc/evprofile/default.json*. User can download the default event profile,
+modify and upload it back to the switch to apply it. 
+
+Through event profile, user can change severity of any event and also can enable/disable a event.
+
+Through CLI, REST or gNMI, event history table and current alarm table can be retrieved using various filters.
```
yozhao101, line 281: For simplicity and readability, do we consider rephrasing the `event history table` ---> `event table` and `current alarm table` ---> `alarm table?

spenugondaa: Done

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Done

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Also when the alarm is acknowledged.
Acknowledging an alarm means, user is aware of that particular alarm and decides that alarm is not going to affect the device. So, that particular alarm should not be counted.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Updated.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Updated.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: Updated.
Yes. As mentioned, events are one shot notifications and alarms are dynamic in nature. Alarm table contains those conditions that happened and either need operator attention or may get cleared by those applications. As it is dynamic, and as those conditions gets cleared on a reboot, we need one table to list those alarms. This ultimately can decide system state and system LED.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: updated.

doc/event-alarm-framework/event-alarm-framework.md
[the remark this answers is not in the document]
spenugondaa: Done

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: It is 'show alarm' command that operator uses quite often. 
IMHO operator would want to see "active" alarms and don't want those alarms he/she finds not relevant and acknowledged.
If at all operator wants to see all, 'all' option is there.

doc/event-alarm-framework/event-alarm-framework.md
spenugondaa: @lguohan @yxieca  Please comment and help bring this to a conclusion.

renukamanavalan:
Here some review comments. 
[Sorry about joining late in the game]

```
1) Defaults.json can be part of image, as that helps adding new requirements, transparently. Adding new events is 
clean/easy, but re-defining/removing existing, need to follow some restrictions
    a. Each event in profile has two IDs
        i. Name (e.g. "TEMPERATURE_EXCEEDED")
        ii. ID - A numeric ID which is globally unique across time & space

    b. An image upgrade could like to redefine an event.
    For example, it may like to set temp limit for TEMPERATURE_EXCEEDED to 65 from original 75.
    If so, it updates its default.json, keeps the same name (e.g. TEMPERATURE_EXCEEDED) but updates 
    its numeric ID.

    Most of the consumers/clients are interested in by name only and often indifferent to the re-definition.
    This helps update/re-define of an event to be transparent to most. The few who might be interested,
    could use the numeric ID to distinguish, when messages flow from different sources running different
    releases that adapt different definitions.

    c. The image may delete some old events that are not applicable anymore and add new ones.
        The numeric ID of the deleted ones are never re-used. Hence leave the entries in defauts.json as deprecated.
        This is required as clients may still get these messages from older releases and they might use numeric
         ID for distinction.
        i. Along with enabled/disabled, have additional one as deprecated in Event profile definition
        ii. This would be useful when merging user-custom-profile as any user update for deprecated
            entries will be skipped/ignored.

2) Image upgrade will only carry forward currently configured custom profile -- not all profiles in /etc/evprofile

3)  Add "DEBUG" as one of supported severities.

4) There is a small grey area. The same event, could get raised as alarm or not, as that gets distinguished
    only by non-empty/empty action field. There is no reason/justification for an event to play dual role as
    alarm or not. But an app could fail to send action due to code bug, which can degrade an alarm to an
    event unintentionally.

Q: Can we classify an event as alarm or not in the event profile ?
     This removes the greyness.
     We can fail the event of type alarm w/o action

5) Dynamic message definition
     Often these messages are parsed by tools for required info, like i/f name in an i/f-down message.
     Across different releases the publisher-client could potentially re-phrase the message that can totally fail
     the clients that parse.
     
     So provide some schema, that is flexible yet stable w/o parsing cost.
    a.   A J2 template for human readable message, with named parameters
    b.   This template is part of event profile definition
    c.   Event publisher, only provides the named parameters & values as a JSON string.
    d.   Consumers could just take the named parameters along with Event name & ID and they get the info they need.
    e.   Consumers who need human readable message could use the J2 template to format the string
    f.   The rsyslog will get the formatted string

Pro:
    a.  The J2 template can be updated overtime and it would not impact the consumers.
    b.  The parameters can be added /removed across image releases and consumers could get what 
         they look for from the JSON string. Upgraded consumers would look for new ones and old ones
         will work with previous values and may find some missing.
    c. As publishing apps, only send params & values, the size of the message they send to syslog become smaller.

6) SEQUENCE_ID:
    Instead of 1 to 2^64, we could provide some semantic value to it as <32 bit time_t><5 digit running
    sequence 00000 to 99999>. Being 5 digits, we could use it globally across all events. This avoids the need
    for caching last used value per event. Coin as <time_t><next ID> for all events.

Pro:
    a.  Subscribing clients, when they restart, all they need is the timestamp from last run. With that info,
         they can set the start filter for event-id for query/subscribe.
    b.  Tools that collects events from multiple sources for combined analysis, sorting by Event-ID
          would help layout the logs chronologically.
    For an sample, while debugging two devices, we can see their events as ordered.

7) Manifest file for setting limits on table sizes, could cover all tables, including PUB/SUB.

8) Providing this feature with least code changes.
    Even for non III-party containers, like orchagent, syncd, <85> the adoption would be slow.
    To get over, one more option would be to use syslog filters
    Let the rsyslog parses the message
    Use omfwd to send out filtered messages
    The listener we add will create the events
    Here the listener still needs to parse to get some info out, but it will be limited to only those that are valid events.
  
```

renukamanavalan:
Look at this: [SD-Data in rsyslog](https://datatracker.ietf.org/doc/html/rfc5424#section-6.3)

This provides structured data & as well human readable format.

bhaveshdell:
@yozhao101 Comments addressed. Please review and approve changes.

doc/event-alarm-framework/event-alarm-framework.md
bhavini-gada: This can be considered as an enhancement that can be brought in later to the commands.

renukamanavalan:
@qiluo-msft @zhenggen-xu 
Can you please review and provide your approval?

renukamanavalan (approved):

bhaveshdell (approved):

jeff-yin:
@renukamanavalan @zhangyanzhao -- any way to ping @qiluo-msft @zhenggen-xu internally to review/approve/merge?

renukamanavalan:
> 

Please ping me at @remanava@microsoft.com

bhavini-gada:
@qiluo-msft - Can you please help merge this HLD ?

zhangyanzhao:
close this PR and use https://github.com/sonic-net/SONiC/pull/1064 to continue to track.

PR #762, opened 2021-03-17
Design for handling SAI failures in orchagent

shi-su:
This PR contains the design for SAI failure handling in orchagent.


doc/SAI_failure_handling/SAI_failure_handling.md
```diff
@@ -0,0 +1,100 @@
...
+In the scenario where a specific logic is required in one of the Orchs, this design allows the Orch to inherit the function and include the specific login in the inherited function.
+
+The function also allows an optional input `context`, which allows passing context (e.g., object entry, attribute, etc.) into the function so that it could escalate the information to the ERROR_DB and upper layers.
+
+#### 2.3.2 Possible execution results
+1. Return `task_success` --  No crash, no retry, handled successfully. 
+
+The failure handling function should return `task_success` when the failure is properly handled without the need for another attempt (e.g., the SAI status is `SAI_STATUS_ITEM_NOT_FOUND` in remove operation).
+
+2. Return `task_failed` -- No crash, no retry, not handled successfully.
```
qiluo-msft, line 40: >task_failed [](start = 11, length = 11)

There are seveal causes for task_failed
1. user input is invalid, such as wrong ACL, conflicting IP addresses
2. hardware permanent error
3. internal logic error, but not critical, and we still want to keep dataplane working
3. etc.

doc/SAI_failure_handling/SAI_failure_handling.md
```diff
@@ -0,0 +1,100 @@
...
+
+Some of the failures can be resolved by restarting SwSS.
+In the scenario where such failures happen, the failure handling function in orchagent should exit with `EXIT_FAILURE` and trigger SwSS auto restart.
+
+
+
+### 2.4 DB changes
+An ERROR_DB will be introduced to escalate the failures from orchagent to upper layers such as fpmsyncd.
+
+The schema of ERROR_DB is designed as follows:
```
qiluo-msft, line 58: >schema [](start = 4, length = 6)

If you accumulate them into a table, who will delete them? otherwise it will keep growing in memory.

doc/SAI_failure_handling/SAI_failure_handling.md
```diff
@@ -0,0 +1,100 @@
...
+```
+ERROR_{{SAI_API}}_TABLE|entry
```
qiluo-msft, line 60: >SAI_API [](start = 8, length = 7)

Take route as an example, there are ROUTE SAI API and neighbor SAI API, but upper layer does not care which API fails.

doc/SAI_failure_handling/SAI_failure_handling.md
qiluo-msft: You can mentioned 'no exit' or 'keep running'

---
In reply to: [600051934](https://github.com/Azure/SONiC/pull/762#discussion_r600051934) [](ancestors = 600051934)

shi-su: Added the scenarios of failure and clarified the keep running behavior.

doc/SAI_failure_handling/SAI_failure_handling.md
shi-su: Added a discussion in the design that it is necessary to avoid accumulating failures in ERROR_DB and consuming memory. To make sure all ERROR_DB entries are consumed, the failure handling should only escalate failures when the corresponding handling mechanism is available in the upper layers. Also included a possible implementation of this behavior.

doc/SAI_failure_handling/SAI_failure_handling.md
shi-su: Yes, I agree, neither the SAI type nor the Orch type would be enough to represent the context of the failure. I think a better representation of the entries in ERROR_DB is to make them consistent with the APPL_DB entries where the SAI failure happens. For example, SAI failure could happen in SAI_ROUTE_API or SAI_NEXT_HOP_GROUP_API when conducting operations for APPL_DB entry `ROUTE_TABLE:0.0.0.0/0`, but in either scenario, the corresponding key in ERROR_DB should be `ERROR_ROUTE_TABLE:0.0.0.0/0` so that the upper layer could know the SAI failures happen in which call and do tasks accordingly.

doc/SAI_failure_handling/SAI_failure_handling.md
```diff
@@ -0,0 +1,125 @@
...
+
+The field `status` stores the status of the SAI operation (e.g., SAI_STATUS_NOT_SUPPORTED, SAI_STATUS_FAILURE).
+
+The ERROR_DB entry also includes a list of attributes (comma separated) and the corresponding values (comma separated) that the failed operation tries to set.
+
+The field `counter` stores the number of failures for the same entry. It could be used as a reference for handling the failure.
+
+To avoid accumulating failures in ERROR_DB and consuming memory, it is necessary to ensure that the upper layer properly consumes the entries in ERROR_DB.
+To make sure all ERROR_DB entries are consumed, the failure handling should only escalate failures when the corresponding handling mechanism is available in the upper layers.
+One possible implementation could be escalating failures to ERROR_DB when the input `context` is valid.
```
qiluo-msft, line 98: >One possible implementation could be escalating failures to ERROR_DB when the input `context` is valid [](http://example.com/codeflow?start=0&length=102)

The design purpose of `context` is not for this.

doc/SAI_failure_handling/SAI_failure_handling.md
```diff
@@ -0,0 +1,125 @@
...
+Possible values include `CREATE/SET/DELETE`.
+
+The field `status` stores the status of the SAI operation (e.g., SAI_STATUS_NOT_SUPPORTED, SAI_STATUS_FAILURE).
+
+The ERROR_DB entry also includes a list of attributes (comma separated) and the corresponding values (comma separated) that the failed operation tries to set.
+
+The field `counter` stores the number of failures for the same entry. It could be used as a reference for handling the failure.
+
+To avoid accumulating failures in ERROR_DB and consuming memory, it is necessary to ensure that the upper layer properly consumes the entries in ERROR_DB.
+To make sure all ERROR_DB entries are consumed, the failure handling should only escalate failures when the corresponding handling mechanism is available in the upper layers.
```
qiluo-msft, line 97: >the failure handling should only escalate failures when the corresponding handling mechanism is available in the upper layers [](http://example.com/codeflow?start=48&length=125)

Does "up layers" mean upstreaming processes? If yes, I don't understand the statement.

doc/SAI_failure_handling/SAI_failure_handling.md
shi-su: Agree, removed this part.

doc/SAI_failure_handling/SAI_failure_handling.md
shi-su: Updated the design to clarify that the upstream processes are expected to handle the entries in ERROR_DB and remove them once handled. Clarified that the ERROR_DB should not keep growing with the assumption valid. Also mentioned that the features are not currently available in upstream processes and need to be added for proper failure handling.

qiluo-msft (dismissed):

doc/SAI_failure_handling/SAI_failure_handling.md
```diff
@@ -0,0 +1,123 @@
...
+    "failed_orch": {{failed orch type}}
+    "failed_SAI": {{failed SAI type}}
+    "opcode": {{failed SAI operation type}}
+    "status": {{sai_status}}
+    "attributes": "attr_type0,attr_type1,..."
+    "attr_values": "attr_value0,attr_value1,..."
+    "counter": {{count}}
+```
+
+The table and key in ERROR_DB correspond to the table and key in APPL_DB where SAI failures happen (e.g., SAI failure happens when conducting operations for APPL_DB entry `ROUTE_TABLE:0.0.0.0/0`, the corresponding key in ERROR_DB should be `ERROR_ROUTE_TABLE:0.0.0.0/0`).
```
rck-innovium, line 81: It would be more future proof to prefix the APP_DB name in the key. For the above example:
```
APPL_DB entry `ROUTE_TABLE:0.0.0.0/0`,  key in ERROR_DB should be `ERROR_APPL_DB_ROUTE_TABLE:0.0.0.0/0`
```

This would later allow me to have ERROR_DB entries for other sources like CONFIG_DB

doc/SAI_failure_handling/SAI_failure_handling.md
```diff
@@ -0,0 +1,123 @@
...
+Possible values include `CREATE/SET/DELETE`.
+
+The field `status` stores the status of the SAI operation (e.g., SAI_STATUS_NOT_SUPPORTED, SAI_STATUS_FAILURE).
+
+The ERROR_DB entry also includes a list of attributes (comma separated) and the corresponding values (comma separated) that the failed operation tries to set.
+
+The field `counter` stores the number of failures for the same entry. It could be used as a reference for handling the failure.
+
+The upstream processes are expected to consume the ERROR_DB entries and remove the handled failures from the ERROR_DB.
+Assuming the upstream processes have the proper consumption of ERROR_DB entries and failure handling logic (these are not currently available for upstreams processes and need to be added), the ERROR_DB should not keep accumulating failures in ERROR_DB and consuming memory.
```
rck-innovium, line 97: When an entry is deleted from APPL_DB, any ERROR_DB entries with the same key could be cleared?

doc/SAI_failure_handling/SAI_failure_handling.md
shi-su: I agree, updated in the design doc.

doc/SAI_failure_handling/SAI_failure_handling.md
shi-su: I think this should be true in most scenarios since the ERROR_DB entries are designed to track the failure in applying APPL_DB entries or basically mismatch between APPL_DB and the hardware. When an entry is removed from ERROR_DB, the mismatch between APPL_DB and the hardware should get cleared, and most likely the corresponding ERROR_DB entry is no longer needed. Yet if the upstream process believes that extra operation is needed for the failure, I think the upstream process could have the freedom to hold the entry until the operations are done.

doc/SAI_failure_handling/SAI_failure_handling.md
```diff
@@ -0,0 +1,123 @@
...
+1. Allow different handling for Create/Set/Remove/Get operations.
+1. Allow each Orch to have its specific handling logic.
+1. Adapt handling logic based on SAI API type and SAI status.
+1. Escalate the failure to upper layers when the failure cannot be handled in orchagent.
+
+### 2.2 An overview of the failure handling framework
+
+An illustrative figure of the failure handling framework is shown below.
+The orchagent generates SAI calls according to the information in APPL_DB given by upper layers.
+In the case of SAI failures, the orchagent gets the failure status via the feedback mechanism in synchronous mode.
```
TencentSwitchGroup, line 20: Is it possible for orchagent to get the failure status in asynchronous mode? 
>In the case of SAI failures, the orchagent gets the failure status via the feedback mechanism in synchronous mode.

qiluo-msft (approved):

PR #768, opened 2021-04-01
Add HLD for H/W capabilities fields in platform.json

ArunSaravananBalachandran:
Add H/W capabilities fields in platform.json

### DIFF
--- a/doc/platform_json_enhancement.md
+++ b/doc/platform_json_enhancement.md
@@ -40,9 +40,7 @@ A new set of `capabilities` fields are introduced in platform.json, for providin
 
 For each component's attribute, the defined `capabilities` fields are as follows:
 
-- "controllable" : A boolean, 'true' if the given attribute can be controlled from the NOS, 'false' otherwise.
- 
-**Note:** If 'controllable' field is not specified, the default value is assumed to be 'true'.                   
+- "controllable" : A boolean, 'true' if the given attribute can be controlled from the NOS, 'false' otherwise. Defaults to 'true'.
 - Attribute specific fields:
     - status led - "color" - A list of the supported colors.
     - speed
--- a/doc/platform_json_enhancement.md
+++ b/doc/platform_json_enhancement.md
@@ -1,0 +1,121 @@
+# Platform capability file enhancement #
+
+### Rev 0.1
+
+## Table of Content
+
+## Revision
+
+ | Rev |     Date    |            Author            | Change Description   |
+ |:---:|:-----------:|:----------------------------:|----------------------|
+ | 0.1 |             | Arun Saravanan Balachandran  | Initial version      |
+
+## Scope
+
+This document provides information on the enhancements for platform capability file `platform.json`.
+
+## Definitions/Abbreviations
+
+| Definitions/Abbreviation | Description |
+|--------------------------|-------------|
+| BMC | Baseband Management Controller |
+| NOS | Network Operating System |
+| PSU | Power Supply Unit |
+
+## Overview
+
+Each networking switch has a set of platform components (e.g: Fans, PSUs, LEDs, etc.) and these components in each platform can have different characteristics (like supported colors for a LED). In a given platform, the components could be controlled by a dedicated platform controller (like BMC) or the NOS running on the CPU is required to control it and in the former the control of the components from the NOS could be limited.
+
+In SONiC the platform components' supported attributes are made available via Platform API, but certain platform specific capabilties for the components are not available for the applications.
+
+This document provides the enhancement for `platform.json` to address the above issue.
+
+## Design
+
+Currently, `platform.json` is used for providing the expected structure of the platform components and interface details for supporting dynamic port breakout.
+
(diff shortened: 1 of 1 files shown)


doc/platform_json_enhancement.md
```diff
@@ -0,0 +1,125 @@
...
+
+A new set of `capabilities` fields are introduced in platform.json, for providing platform specific capablities on control and characteristics of the components.
+
+For each component's attribute, the defined `capabilities` fields are as follows:
+
+- "control" : A boolean, 'true' if the given attribute can be controlled from the NOS, 'false' otherwise.
+- Attribute specific fields:
+    - status led - "color" - A list of the supported colors.
+    - speed
+        - "minimum" - Minimum recommended fan speed that can be set.
```
keboliu, line 47: Could you please elaborate on who will be the consumer of this info? Also as platform facts for the test cases?

ArunSaravananBalachandran: Currently, these fields would be used as platform facts for platform API test suite.
But, these can also be consulted by any platform daemon before executing the respective APIs.

doc/platform_json_enhancement.md
```diff
@@ -0,0 +1,125 @@
...
+
+Currently, `platform.json` is used for providing the expected structure of the platform components and interface details for supporting dynamic port breakout.
+
+### Platform capabilities field
+
+A new set of `capabilities` fields are introduced in platform.json, for providing platform specific capablities on control and characteristics of the components.
+
+For each component's attribute, the defined `capabilities` fields are as follows:
+
+- "control" : A boolean, 'true' if the given attribute can be controlled from the NOS, 'false' otherwise.
```
jleveque, line 43: Although it might be a bit longer, maybe this key should be named `controllable` for clarity?

Also, is there a default value assumed if this is omitted? I assume we would default to `true`?

ArunSaravananBalachandran: > Although it might be a bit longer, maybe this key should be named `controllable` for clarity?

As suggested, the name of the key has been changed from `control` to `controllable`

> Also, is there a default value assumed if this is omitted? I assume we would default to `true`?

For the default value, the application/test suite can assume it as `true` to maintain the current behavior

ArunSaravananBalachandran: > Could you add a note about this in the doc, please? If `controllable` is not specified, it is assumed to be true.

Have included the default value for 'controllable'

doc/platform_json_enhancement.md
jleveque: @zzhiyuan: I'm not sure I understand your comment. The main usage of these values is for testing purposes to understand boundaries when testing the device. Can you please elaborate?

zzhiyuan: Ah, I meant more for the latter part, "these can also be consulted by any platform daemon before executing the respective APIs."

Since there is already the fantray abstraction, I believe it is better for platform vendor to control the min/max fan speed which may differ by model, via thermal manager. Having values in platform.json for purpose of testing is fine.

doc/platform_json_enhancement.md
```diff
@@ -0,0 +1,127 @@
...
+
+Currently, `platform.json` is used for providing the expected structure of the platform components and interface details for supporting dynamic port breakout.
+
+### Platform capabilities field
+
+A new set of `capabilities` fields are introduced in platform.json, for providing platform specific capablities on control and characteristics of the components.
+
+For each component's attribute, the defined `capabilities` fields are as follows:
+
+- "controllable" : A boolean, 'true' if the given attribute can be controlled from the NOS, 'false' otherwise.
```
zzhiyuan, line 43: Can you elaborate in this context what controllable from the NOS mean? Most of these attributes would be managed by a daemon, either directly or via vendor code. So if say you change the fan speed, the thermalctld would change it back, and you'd have to stop the container in order to change it. Is this deemed controllable?

zzhiyuan: Can you maybe reword to:

`- "controllable" : Boolean, 'true' if the given attribute can be controlled from the NOS, 'false' if it cannot. Defaults to 'true.'`

So you don't need the note below.

ArunSaravananBalachandran: > Can you elaborate in this context what controllable from the NOS mean? Most of these attributes would be managed by a daemon, either directly or via vendor code. So if say you change the fan speed, the thermalctld would change it back, and you'd have to stop the container in order to change it. Is this deemed controllable?

If a given platform component can be controlled from the NOS (software-controlled), it is deemed to be 'controllable' and 
if a platform component is controlled by a dedicated platform controller (like BMC) and is not controllable from the NOS running on the CPU, it is deemed as 'not controllable'.

jleveque (approved):

doc/platform_json_enhancement.md
jleveque: Ah. Yes, I agree with you. Platform daemons really shouldn't be referring to these values.

PR #773, opened 2021-04-12
Policy Based Hashing HLD

nazariig:
| PR title                                                                  | state                                                                                                             | context                                                                                                                |
| ------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- |
| [Add PBH YANG model](https://github.com/Azure/sonic-buildimage/pull/7461) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [Add PBH DB schema](https://github.com/Azure/sonic-swss-common/pull/495)  | GitHub issue/pull request detail | GitHub pull request check contexts |
| [Add PBH OA](https://github.com/Azure/sonic-swss/pull/1782)               | GitHub issue/pull request detail       | GitHub pull request check contexts       |
| [Add PBH CLI](https://github.com/Azure/sonic-utilities/pull/1701)         | GitHub issue/pull request detail  | GitHub pull request check contexts  |

Signed-off-by: Nazarii Hnydyn nazariig@nvidia.com


doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,758 @@
...
+**The following command shows statistics:**
+```bash
+root@sonic:/home/admin# show pbh statistics
+Table      Rule    Packets Count    Bytes Count
+---------  ------  ---------------  -------------
+pbh_table  nvgre   0                0
+           vxlan   0                0
+```
+
+## 2.7 DPB YANG model
```
venkatmahalingam, line 634: Please remove DPB

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,758 @@
...
+    }
+
+    import sonic-types {
+        prefix stypes;
+        revision-date 2019-07-01;
+    }
+
+    import sonic-extension {
+        prefix ext;
+        revision-date 2019-07-01;
```
venkatmahalingam, line 659: We dont need revision-date any more, please remove it.

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,758 @@
...
+            "hash_list": [
+                "inner_ip_proto",
+                "inner_l4_dst_port",
+                "inner_l4_src_port",
+                "inner_dst_ipv6",
+                "inner_src_ipv6"
+            ],
+            "packet_action": "SET_ECMP_HASH"
+        },
+        "pbh_table|vxlan": {
```
venkatmahalingam, line 457: Upper case table name?

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,758 @@
...
+            "priority": "2",
+            "ip_protocol": "0x11/0xff",
+            "l4_dst_port": "0x12b5/0xffff",
+            "inner_ether_type": "0x0800/0xffff",
+            "hash_list": [
+                "inner_ip_proto",
+                "inner_l4_dst_port",
+                "inner_l4_src_port",
```
venkatmahalingam, line 465: If hash_list is a leaf-list, the values should be ["PBH_HASH|inner_ip_proto", "PBH_HASH|inner_l4_dst_port"...], rather we can have the following, please see if this makes sense, if not, please follow appropriate YANG conventions to represent the references.
1) PBH_TABLE|vxlan 
     hash_list ---leaf-ref---> PBH_HASH_SET|5tuple

2) PBH_HASH_SET|5tuple

3)
PBH_HASH|5tuple|inner_ip_proto
PBH_HASH|5tuple|inner_l4_dst_port
...
....

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,758 @@
...
+key = PBH_TABLE|table_name ; table name. Must be unique
+
+; field     = value
+port_list   = port-list ; ports to which this table is applied. Can be empty
+lag_list    = lag-list  ; portchannels to which this table is applied. Can be empty
+description = *255VCHAR ; table description. Can be empty
+
+; value annotations
+port-name = 1*64VCHAR                         ; name of the port
+port-list = port-name [ 1*( "," port-name ) ] ; list of the ports. Valid values range is platform dependent
```
venkatmahalingam, line 323: We dont need port-name when port-list itself can have a single port-name..correct?

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,758 @@
...
+### 2.4.1 Config DB
+
+#### 2.4.1.1 PBH table
+```abnf
+; defines schema for PBH table configuration attributes
+key = PBH_TABLE|table_name ; table name. Must be unique
+
+; field     = value
+port_list   = port-list ; ports to which this table is applied. Can be empty
+lag_list    = lag-list  ; portchannels to which this table is applied. Can be empty
```
venkatmahalingam, line 318: Can both port_list and lag_list be empty?

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,758 @@
...
+#### 2.4.1.2 PBH rule
+```abnf
+; defines schema for PBH rule configuration attributes
+key = PBH_RULE|table_name|rule_name ; rule name. Must be unique across the table
+
+; field           = value
+priority         = 1*5DIGIT      ; rule priority. Valid values range is platform dependent
+gre_key          = h32 "/" h32   ; GRE key (32 bits)
+ip_protocol      = h8 "/" h8     ; IP protocol (8 bits)
+ipv6_next_header = h8 "/" h8     ; IPv6 Next Header (8 bits)
```
venkatmahalingam, line 339: I assume, ip_protocol and ipv6_next_header fields cant co-exist, right? if no, can we just have one field as IP protocol numbers are unique? also, IP protocol numbers are not bits friendly, not sure how 0xff mask will be helpful, having an example for more than one protocol number would be useful.

IP_PROTOCOL: 0x11/0xff

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,758 @@
...
+
+### 2.6.2 Usage examples
+
+#### 2.6.2.1 Config command group
+
+**The following command adds/updates/removes table:**
+```bash
+config pbh table add 'pbh_table' --port_list 'Ethernet0,Ethernet4' --lag_list 'PortChannel0001,PortChannel0002' \
+--description 'NVGRE and VxLAN'
+config pbh table update 'pbh_table' --port_list 'Ethernet0'
```
venkatmahalingam, line 560: I dont think, we follow "--" for any sonic CLI options today (e.g --port_list in the above command), please check, also, we can avoid table in the command.

doc/pbh/pbh-design.md
venkatmahalingam: Since we use inner header names, maybe having outer name in IP protocol/gre_key..etc field descriptions would be helpful e.g Outer IP header protocol

doc/pbh/pbh-design.md
nazariig: @venkatmahalingam we do follow `--` CLI convention in SONiC, e.g.:
```bash
root@sonic:/home/admin# config vlan member add --help
Usage: config vlan member add [OPTIONS] <vid> port

  Add VLAN member

Options:
  -u, --untagged
  -?, -h, --help  Show this message and exit.

```
As for `config pbh table add 'pbh_table'` we need an identifier (e.g., `table`) to refer to the particular item subgroup.
Not sure why you want to avoid table in the command.

doc/pbh/pbh-design.md
nazariig: @venkatmahalingam we are following the SAI spec, which claims that `sai_acl_field_data_t` should have a mask. The goal is to provided user as much flexibility as possible regardless to what is actually supported by platform:
https://github.com/opencomputeproject/SAI/blob/v1.8.1/inc/saiacl.h#L1829
```
    /**
     * @brief IP Protocol
     *
     * @type sai_acl_field_data_t sai_uint8_t
     * @flags CREATE_AND_SET
     * @default disabled
     */
    SAI_ACL_ENTRY_ATTR_FIELD_IP_PROTOCOL,
```
https://github.com/opencomputeproject/SAI/blob/v1.8.1/inc/saitypes.h#L549
```
/**
 * @brief Defines a single ACL filter
 *
 * @note IPv4 and IPv6 Address expected in Network Byte Order
 *
 * @extraparam const sai_attr_metadata_t *meta
 */
typedef struct _sai_acl_field_data_t
{
    /**
     * @brief Match enable/disable
     */
    bool enable;

    /**
     * @brief Field match mask
     *
     * @note Nothing can be serialized if mask is not needed for data items
     * like object id.
     *
     * @passparam meta
     * @validonly enable == true
     */
    sai_acl_field_data_mask_t mask;

    /**
     * @brief Expected AND result using match mask above with packet field
     * value where applicable.
     *
     * @passparam meta
     * @validonly enable == true
     */
    sai_acl_field_data_data_t data;
} sai_acl_field_data_t;
```
https://github.com/Azure/sonic-swss/blob/master/orchagent/aclorch.cpp#L323
```
else if (attr_name == MATCH_IP_PROTOCOL || attr_name == MATCH_NEXT_HEADER)
{
    value.aclfield.data.u8 = to_uint<uint8_t>(attr_value);
    value.aclfield.mask.u8 = 0xFF;
}
```
The IP proto and IPv6 Next Header can't co-exist at the same level (e.g., outer/inner frame). These are two different fields.
For the attributes description (e.g., Outer IP header protocol) we have `2.2 SAI API` section. Also to distinguish between outer/inner area we are following SAI name convention:
```
SAI_ACL_ENTRY_ATTR_FIELD_IP_PROTOCOL - outer field
SAI_ACL_ENTRY_ATTR_FIELD_INNER_ETHER_TYPE - inner field
```
If you still believe that we need to add it to the schema description explicitly - no issues i can do it.

doc/pbh/pbh-design.md
venkatmahalingam: ok, that's fine, it's just a suggestion.

doc/pbh/pbh-design.md
venkatmahalingam: ok, it is possible to configure more than one IP protocol with mask?

nazariig: @venkatmahalingam no. The match/action fields should be unique. this means that `ip_protocol` can appear only once.

doc/pbh/pbh-design.md
nazariig: @venkatmahalingam we still use `revision-date` in latest master, so can't remove it so far.
Also, according to RFC 7950 it's a bad idea to completely remove/abandon it:

https://tools.ietf.org/html/rfc7950#section-7.1.5
```
When the optional "revision-date" substatement is present, any
typedef, grouping, extension, feature, and identity referenced by
definitions in the local module are taken from the specified revision
of the imported module.  It is an error if the specified revision of
the imported module does not exist.  If no "revision-date"
substatement is present, it is undefined from which revision of the
module they are taken.
```

doc/pbh/pbh-design.md
nazariig: @venkatmahalingam `also, we can avoid table in the command`
Please elaborate on that

doc/pbh/pbh-design.md
nazariig: @venkatmahalingam done

doc/pbh/pbh-design.md
venkatmahalingam: There is an effort in progress to remove the revision date, please refer to the new YANG models in review for better clarity.

doc/pbh/pbh-design.md
venkatmahalingam: ok.

doc/pbh/pbh-design.md
nazariig: @venkatmahalingam done

doc/pbh/pbh-design.md
nazariig: @venkatmahalingam according to the SWSS DB schema (https://github.com/Azure/sonic-swss/blob/master/doc/swss-schema.md) we have a concept of key reference. Today in SONiC it is being used in few places (qos, buffer, pg, queue, etc.):

**Example:**
```
    "PORT_QOS_MAP": {
        "Ethernet8": {
            "tc_to_queue_map": "[TC_TO_QUEUE_MAP|AZURE]",
            "dscp_to_tc_map": "[DSCP_TO_TC_MAP|AZURE]",
            "pfc_enable": "3,4",
            "pfc_to_queue_map": "[MAP_PFC_PRIORITY_TO_QUEUE|AZURE]",
            "pfc_to_pg_map": "[PFC_PRIORITY_TO_PRIORITY_GROUP_MAP|AZURE]",
            "tc_to_pg_map": "[TC_TO_PRIORITY_GROUP_MAP|AZURE]"
        },

    "BUFFER_PORT_INGRESS_PROFILE_LIST": {
        "Ethernet8": {
            "profile_list": "[BUFFER_PROFILE|ingress_lossless_profile]"
        },

    "BUFFER_PORT_EGRESS_PROFILE_LIST": {
        "Ethernet8": {
            "profile_list": "[BUFFER_PROFILE|egress_lossless_profile],[BUFFER_PROFILE|egress_lossy_profile]"
        },

    "BUFFER_PG": {
        "Ethernet14|0": {
            "profile": "[BUFFER_PROFILE|ingress_lossy_profile]"
        },

    "BUFFER_QUEUE": {
        "Ethernet12|5-6": {
            "profile": "[BUFFER_PROFILE|q_lossy_profile]"
        },

    "QUEUE": {
        "Ethernet20|5": {
            "scheduler": "[SCHEDULER|scheduler.0]"
        },
        "Ethernet48|5": {
            "scheduler": "[SCHEDULER|scheduler.0]"
        },
        "Ethernet48|4": {
            "wred_profile": "[WRED_PROFILE|AZURE_LOSSLESS]",
            "scheduler": "[SCHEDULER|scheduler.1]"
        },
```

With a new YANG validation concept, it seem to be redundant to use this style, as it makes sense only for ABNF notation.
Please have a look at this YANG model: https://github.com/Azure/sonic-buildimage/pull/7461.

It was written following the existing ACL `ports` leaf-list concept:
https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-acl.yang#L257
```yang
leaf-list ports {
	/* union of leafref is allowed in YANG 1.1 */
	type union {
		type leafref {
			path /port:sonic-port/port:PORT/port:PORT_LIST/port:name;
		}
		type leafref {
			path /lag:sonic-portchannel/lag:PORTCHANNEL/lag:PORTCHANNEL_LIST/lag:name;
		}
		type string {
			pattern "";
		}
	}
	/* Today in SONiC, we do not delete the list once
	 * created, instead we set to empty list. Due to that
	 * below default values are needed.
	 */
	default "";
}
```

And this is how the ACL/PORT/LAG configuration looks like in `config_db.json`:
```json
"ACL_TABLE": {
    "EVERFLOW": {
        "ports": [
            "PortChannel0001",
            "PortChannel0002",
            "PortChannel0003",
            "PortChannel0004",
            "Ethernet110",
            "Ethernet108",
            "Ethernet24",
            "Ethernet26",
            "Ethernet12",
            "Ethernet14",
            "Ethernet62",
            "Ethernet60",
            "Ethernet46",
            "Ethernet44",
            "Ethernet78",
            "Ethernet76",
            "Ethernet48",
            "Ethernet28",
            "Ethernet116",
            "Ethernet74",
            "Ethernet72",
            "Ethernet30",
            "Ethernet80",
            "Ethernet82",
            "Ethernet120",
            "Ethernet18",
            "Ethernet16",
            "Ethernet122",
            "Ethernet58",
            "Ethernet56",
            "Ethernet50",
            "Ethernet96",
            "Ethernet112",
            "Ethernet52",
            "Ethernet54",
            "Ethernet84",
            "Ethernet124",
            "Ethernet22",
            "Ethernet20",
            "Ethernet64"
        ],
        "type": "MIRROR",
        "policy_desc": "EVERFLOW",
        "stage": "ingress"
    }
}

"PORT": {
    "Ethernet80": {
        "index": "21",
        "lanes": "80,81",
        "description": "Servers61:eth0",
        "pfc_asym": "off",
        "mtu": "9100",
        "alias": "etp21a",
        "admin_status": "up",
        "speed": "50000"
    },
    "Ethernet224": {
        "index": "57",
        "lanes": "224,225",
        "description": "etp57a",
        "mtu": "9100",
        "alias": "etp57a",
        "pfc_asym": "off",
        "speed": "50000"
    }
}

"PORTCHANNEL": {
    "PortChannel0001": {
        "admin_status": "up",
        "min_links": "2",
        "members": [
            "Ethernet0",
            "Ethernet2"
        ],
        "mtu": "9100"
    },
    "PortChannel0003": {
        "admin_status": "up",
        "min_links": "2",
        "members": [
            "Ethernet32",
            "Ethernet34"
        ],
        "mtu": "9100"
    }
}
```

As you may see, it doesn't use `["PORT|Ethernet80", "PORTCHANNEL|PortChannel0001", ...]` notation.

Note: PBH YANG model was tested with DPB feature - no issues observed

venkatmahalingam: ok, thanks.

doc/pbh/pbh-design.md
venkatmahalingam: My suggestion was to have the command as follows, but I'm okay even if you keep the current command format.

config pbh add 'pbh_table' --port_list 'Ethernet0,Ethernet4' --lag_list 'PortChannel0001,PortChannel0002'

nazariig: @venkatmahalingam thanks for the clarification. Now your point is clear. Still, i am tend to keep the original CLI format. The strong argument here is: we need to keep things as simple as possible and also make sure everything is self explanatory.

Having `config pbh add 'pbh_table'` rather then `config pbh table add 'pbh_table'` is less user friendly and adds some confusion: it is not clear what object type you actually adding/updating/removing.

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,770 @@
...
+description = *255VCHAR ; table description. Can be empty
+
+; value annotations
+port-name = 1*64VCHAR                         ; name of the port
+port-list = port-name [ 1*( "," port-name ) ] ; list of the ports. Valid values range is platform dependent
+lag-name  = "PortChannel" 1*4DIGIT            ; name of the portchannel
+lag-list  = lag-name [ 1*( "," lag-name ) ]   ; list of the portchannels. Valid values range is platform dependent
+```
+
+**Note:** at least one member of _port_list_ or _lag_list_ is required
```
vadymhlushko-mlnx, line 340: Above said `can be empty`

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,770 @@
...
+
+## 1.1 Feature overview
+
+PBH is a feature which allows user to configure a custom hashing for different packet types.  
+Under the hood is uses ACL rules to match the specific types of frames and calculates hash  
+based on user-defined rules.
+
+For flexible hash calculation a new SAI FG Hash API is used.  
+It allows user not only to configure which fields should be used,  
+but also to specify a mask for IPv4/IPv6 addresses and sequence ID.
```
madhupalu, line 110: what is the significance of sequence Id, list out few use cases where does it use it?

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,770 @@
...
+**The following command adds/updates/removes hash:**
+```bash
+config pbh hash add 'inner_dst_ipv6' --field 'INNER_DST_IPV6' --mask 'FFFF::' --sequence 4
+config pbh hash update 'inner_dst_ipv6' --mask 'FFFF:FFFF::'
+config pbh hash remove 'inner_dst_ipv6'
+```
+
+#### 2.6.2.2 Show command group
+
+**The following command shows table configuration:**
```
madhupalu, line 595: Is the desgin handles config validations? 
How about the config didn;t go through the ASIC? does it handle failures?

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,770 @@
...
+--hash_list 'inner_ip_proto,inner_l4_dst_port,inner_l4_src_port,inner_dst_ipv6,inner_src_ipv6' \
+--action set_ecmp_hash --counter enabled
+config pbh rule update 'vxlan' 'pbh_table' --counter disabled
+config pbh rule remove 'vxlan' 'pbh_table'
+```
+
+**The following command adds/updates/removes hash:**
+```bash
+config pbh hash add 'inner_dst_ipv6' --field 'INNER_DST_IPV6' --mask 'FFFF::' --sequence 4
+config pbh hash update 'inner_dst_ipv6' --mask 'FFFF:FFFF::'
```
madhupalu, line 589: is the same cli command used for ipv4? list down an example.

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,770 @@
...
+
+### 2.5.1 PBH add
+
+![PBH add flow](images/pbh_add_flow.svg "Figure 3: PBH add flow")
+
+###### Figure 3: PBH add flow
+
+### 2.5.2 PBH remove
+
+![PBH remove flow](images/pbh_remove_flow.svg "Figure 4: PBH remove flow")
```
madhupalu, line 510: I see the workflow, there is no dependency which is good . However, shouldn't it be delete Rules before the Pbh table get deleted? How does it work if table get deleted and rules failed to deleted. 
How do we handling the error scenarios?

madhupalu (commented):
Provided few comments.

doc/pbh/pbh-design.md
nazariig: @madhupalu please have a look at:
https://github.com/Azure/SONiC/blob/6af2959ee5829f801409fc833389e78802d5258a/doc/pbh/pbh-design.md#242-configuration-sample

```json
"inner_l4_dst_port": {
    "hash_field": "INNER_L4_DST_PORT",
    "sequence_id": "2"
},
"inner_l4_src_port": {
    "hash_field": "INNER_L4_SRC_PORT",
    "sequence_id": "2"
},
```

So here we have dst/src L4 port configured as associative fields.
Which means that if values are swapped, the resulting hash will be the same.

doc/pbh/pbh-design.md
nazariig: @madhupalu In SONiC there is no response concept for CLI mechanism. Which means that we are limited in validation and sometimes the only way to get to know if the configuration was successful is to check the logs.

doc/pbh/pbh-design.md
nazariig: @madhupalu yes, the same is used for both IPv4/IPv6:
```bash
config pbh hash-field add 'inner_dst_ipv4' --field 'INNER_DST_IPV4' --mask '255.0.0.0' --sequence 3
config pbh hash-field add 'inner_src_ipv4' --field 'INNER_SRC_IPV4' --mask '0.0.0.255' --sequence 3

config pbh hash-field add 'inner_dst_ipv6' --field 'INNER_DST_IPV6' --mask 'FFFF::' --sequence 4
config pbh hash-field add 'inner_src_ipv6' --field 'INNER_SRC_IPV6' --mask '::FFFF' --sequence 4
```
P.S: HLD is currently outdated due to some DB schema changes, expect the new revision later next week

doc/pbh/pbh-design.md
nazariig: @madhupalu the implementation assumes strict dependencies for all PBH objects: there is no way to remove PBH table if it's still being referenced by any PBH rule

doc/pbh/pbh-design.md
madhupalu: sounds good, thanks

doc/pbh/pbh-design.md
madhupalu: sounds good

doc/pbh/pbh-design.md
madhupalu: You are right for CLI validations, however, there is a SAI Error handling HLD - https://github.com/Azure/SONiC/blob/master/doc/error-handling/error_handling_design_spec.md
you can refer this for SAI error handling for PBH SAI object failiures. It would be advice to list down what are the applications should listen/react to PBH failures?

liat-grozovik:
I don't think this is available yet
If I remember correct in the HLD it was mentioned that the infra will be provided and later on decide which to listen
I would wait first to see the infra in place and available at the time the PBH is upstream and then it will be added.

Get Outlook for iOS<https://aka.ms/o0ukef>

msreddypaluru (dismissed):
Thanks for the responses.

liat-grozovik:
@anish-n can you please approve the HLD? code is ready and aligned with all the feedback provided offline

doc/pbh/pbh-design.md
```diff
@@ -0,0 +1,941 @@
...
+
+**Note:** at least one member of _interface_list_ is required
+
+#### 2.4.1.2 PBH rule
+```abnf
+; defines schema for PBH rule configuration attributes
+key = PBH_RULE|table_name|rule_name ; rule name. Must be unique across the table
+
+; field           = value
+priority         = 1*5DIGIT      ; rule priority. Valid values range is platform dependent
```
anish-n, line 393: Could we list order of evaluation of priorities, eg: higher value priority will evaluate before lower value priority in hw?

anish-n (approved):

doc/pbh/pbh-design.md
nazariig: @anish-n done

nazariig:
@anish-n if no more objections, can we merge this?

anish-n (approved):

anish-n:
@nazariig Looks good to me, I don't have merge permission so please request someone with merge permission for the merge.

nazariig:
@qiluo-msft / @yxieca please help to merge

nazariig:
@prsunny can you please help to merge?

nazariig:
@qiluo-msft / @yxieca / @prsunny please review & merge

nazariig:
@qiluo-msft / @yxieca / @prsunny please help to merge

nazariig:
@qiluo-msft / @yxieca / @prsunny please help to merge

liat-grozovik (approved):

PR #796, opened 2021-06-08
Class Based Forwarding HLD

abanu-ms:
This is the spec for the Class Based Forwarding feature.

Code PRs:
swss-common: https://github.com/Azure/sonic-swss-common/pull/525
sairedis: https://github.com/Azure/sonic-sairedis/pull/909
swss: https://github.com/Azure/sonic-swss/pull/1963
buildimage: https://github.com/Azure/sonic-buildimage/pull/8689
utilites: https://github.com/Azure/sonic-utilities/pull/1799

### DIFF
--- a/doc/cbf/cbf_hld.md
+++ b/doc/cbf/cbf_hld.md
@@ -45,11 +45,12 @@ NHG - Next Hop Group
 # 1 Introduction
 Class Based Forwarding allows the routed traffic according to the IP/MPLS decision rules to be forwarded on different paths for the same destination depending on the Forwarding Class (different from the Traffic Class), which is determined by a mapping from the DSCP/EXP value of the packet to the Forwarding Class value. A packet coming in with a DSCP/EXP value of X will receive a Forwarding Class (FC) value of Y according to the mapping table provided at the start-of-day. This packet will then be routed, as mentioned earlier, using the traditional IP/MPLS lookup. If the chosen route uses Class Based Forwarding, the next hop will be chosen based on the Forwarding Class value. You can find a flow diagram describing this below:
 
+```
 Packet is received with     A lookup is performed      FC value X is      IP routing decision     Routing lookup returns      The next hop group Z is       Packet is forwarded
 DSCP/EXP value of W for --> in the DSCP/EXP to FC --> assigned to the --> lookup is performed --> next hop group Y, which --> selected from the members --> via group Z to the
    destination D               map table for W            packet           for destination D          is a CBF group            of Y based on the FC           destination D
-                                                                                                                                    value X
-
+                                                                                                                                       value X
+```
 This feature enables opeartors, among other things, to send the important (foreground) traffic through the shortest path, while sending the background traffic through longer paths to still give it some bandwidth instead of using QoS queues which may block background traffic from getting bandwitdh.
 
 These new class based next hop groups are allowed thanks to the changes in https://github.com/opencomputeproject/SAI/pull/1193, which allow a next hop group object to also have other next hop group objects as members of the group along with the next hop objects. The way such a next hop group works is that a packet which has a Forwarding Class value of X will be matched against an appropriate member of this group, selected based on the Forwarding Class value thanks to the "class_map" property of the group. As an example, given the CBF group with members Nhg1, Nhg2 and Nhg3 and a class map of FC 0 -> Nhg1, FC 1 -> Nhg2 and FC 3 -> Nhg3, a packet which has an FC value of 0 will be forwarded using Nhg1. Note that multiple FC values can point to the same member, but a single FC value can't be mapped to more than one member.
--- a/doc/cbf/cbf_hld.md
+++ b/doc/cbf/cbf_hld.md
@@ -1,0 +1,214 @@
+
+# Class Based Forwarding Enhancement
+#### Rev 0.1
+
+# Table of Contents
+  * [Revision](#revision)
+  * [About This Manual](#about-this-manual)
+  * [1. Introduction](#1-introduction)
+  * [2. Requirements Overview](#2-requirement-overview)
+      * [2.1 Functional Requirements](#21-functional-requirements)
+      * [2.2 Configuration and Management Requirements](#22-configuration-and-management-requirements)
+      * [2.3 Scalability Requirements](#23-scalability-requirements)
+      * [2.4 Warm Boot Requirements](#24-warm-boot-requirements)
+      * [2.4 Restrictions](#25-restrictions)
+  * [3. Design](#3-design)
+      * [3.1 Overview](#31-overview)
+      * [3.2 DB Changes](#32-db-changes)
+          * [3.2.1 APPL DB](#321-appl-db)
+          * [3.2.2 CONFIG DB](#322-config-db)
+      * [3.3 Switch State Service Design](#33-switch-state-service-design)
+          * [3.3.1 Orchestration Agent](#331-orchestration-agent)
+      * [3.4 sairedis](#34-sairedis)
+      * [3.5 SAI](#35-sai)
+      * [3.6 CLI](#36-cli)
+  * [4. Warm Boot Support](#4-warm-boot-support)
+      * [4.1 Warm Upgrade](#31-warm-upgrade)
+  * [5. Unit Test](#5-unit-test)
+
+
+# Revision
+| Rev |     Date    |       Author            | Change Description                         |
+|:---:|:-----------:|:-----------------------:|--------------------------------------------|
+| 0.1 | 03/06/2021  |     Alexandru Banu      | Initial version                            |
+
+
+# About this Manual
+This document provides general information about Class Based Forwarding which allows traffic to be steered through the network by policy, adding a layer of traffic engineering based on a Forwarding Class value which allows custom paths to be configured for a destination based on this value.
+
+Along this document the following abbreviations might be used:
+
+FC - Forwarding Class
+CBF - Class Based Forwarding
+NHG - Next Hop Group
+
+# 1 Introduction
+Class Based Forwarding allows the routed traffic according to the IP/MPLS decision rules to be forwarded on different paths for the same destination depending on the Forwarding Class (different from the Traffic Class), which is determined by a mapping from the DSCP/EXP value of the packet to the Forwarding Class value. A packet coming in with a DSCP/EXP value of X will receive a Forwarding Class (FC) value of Y according to the mapping table provided at the start-of-day. This packet will then be routed, as mentioned earlier, using the traditional IP/MPLS lookup. If the chosen route uses Class Based Forwarding, the next hop will be chosen based on the Forwarding Class value. You can find a flow diagram describing this below:
+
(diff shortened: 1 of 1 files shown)


anshuv-mfst:
hi @ashutosh-agrawal - please review the feature, thanks.

doc/cbf/cbf_hld.md
```diff
@@ -0,0 +1,215 @@
...
+## 3.2 DB Changes
+### 3.2.1 APPL DB
+Based on the next hop group split (https://github.com/Azure/SONiC/pull/712) on which this HLD is based on, a new CLASS_BASED_NEXT_HOP_GROUP table will be added to the APPL_DB with the following format:
+```
+### CLASS_BASED_NEXT_HOP_GROUP_TABLE
+    ;Stores a list of FC-aware next hop groups.
+    ;Status: Mandatory
+    key           = CLASS_BASED_NEXT_HOP_GROUP_TABLE:string ; arbitrary string identifying the class based next hop group, as determined by the programming application.
+    members       = NEXT_HOP_GROUP_TABLE.key,    ; one or more indexes within NEXT_HOP_GROUP_TABLE, separated by “,”
+    class_map     = number:number,   ; one or more mapping from Forwarding Class to index in "members" field to use as NHG, separated by ","
```
venkatmahalingam, line 95: Why can't we have DSCP/EXP configs from App and use a config-DB-based mapping for DSCP/EXP to FC before programming the HW? This way, App need not be aware of internal FC mapping.

doc/cbf/cbf_hld.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+Example:
+    127.0.0.1:6379[1]> hgetall  "CLASS_BASED_NEXT_HOP_GROUP:CbfNhg1"
+    1) "members"
+    2) "Nhg1,Nhg2,Nhg3,Nhg4"
+    3) "class_map"
+    4) "0:0,1:0,2:1,3:1,4:2,5:2,6:3,7:3"
```
venkatmahalingam, line 103: Is it possible to define one more table for class (or DSCP/EXP) to NH mapping and refer it from this table?

doc/cbf/cbf_hld.md
```diff
@@ -0,0 +1,215 @@
...
+### 3.2.2 CONFIG_DB
+In order to store the DSCP/EXP to FC mappings, 2 new CONFIG_DB tables will be added:
+
+```
+### DSCP_TO_FC_MAP
+    ;Stores a mapping between DSCP values and FC values. qos_map object with SAI_QOS_MAP_ATTR_TYPE == sai_qos_map_type_t::SAI_QOS_MAP_DSCP_TO_FC
+    ;Status: Mandatory
+    key        = DSCP_TO_FC_MAP_TABLE:string ; arbitrary string identifying the name of the map.
+    dscp_value = 1*DIGIT
+    fc_value   = 1*DIGIT
```
venkatmahalingam, line 138: Please add SONiC YANG changes for config-DB tables.

doc/cbf/cbf_hld.md
rck-innovium: > Is it possible to define one more table for class (or DSCP/EXP) to NH mapping and refer it from this table to make the mapping clean?

I too second that. As discussed in the community meeting consider having a separate table for DSCP_to_FC, and refer to those table names in CLASS_BASED_NEXT_HOP_GROUP_TABLE.


```
### FC_TO_NH_INDEX_MAP_TABLE
; FC to Next hop group index map
key                    = "FC_TO_NH_INDEX_MAP_TABLE:"name
;field
fc_num = 1*DIGIT
;value
nh_index  = 1*DIGIT;  index of NH inside NH group

### CLASS_BASED_NEXT_HOP_GROUP_TABLE
    ;Stores a list of FC-aware next hop groups.
    ;Status: Mandatory
    key           = CLASS_BASED_NEXT_HOP_GROUP_TABLE:string ; arbitrary string identifying the class based next hop group, as determined by the programming application.
    members       = NEXT_HOP_GROUP_TABLE.key,    ; one or more indexes within NEXT_HOP_GROUP_TABLE, separated by “,”
    map_forwarding_class_to_index    = ref_hash_key_reference   ; refer
```

doc/cbf/cbf_hld.md
abanu-ms: Could you please point me to the YANG for `DSCP_TO_TC_MAP_TABLE` to have an example? I'm not sure where to look for it.

doc/cbf/cbf_hld.md
TACappleman: That would limit flexibility in a couple of ways, so I think the current proposal is better in general.
- There is SAI support for setting the FC via ACL rules as well. We're not adding support for that to SONiC at this time, but shouldn't prevent it by making this map go all the way from DSCP/EXP to NH in one go
- The DSCP/EXP to FC maps can potentially be configured per-port, which also wouldn't work with this proposed map.

doc/cbf/cbf_hld.md
TACappleman: Could you explain the reasoning behind preferring the separate map? I see that it matches the style of other similar maps, but would like to understand why those were done that way originally.

TACappleman:
Hi @venkatmahalingam thanks for the feedback so far. We've got a few queries on some of these - could you have a look at them this week please?

doc/cbf/cbf_hld.md
```diff
@@ -0,0 +1,215 @@
...
+The ROUTE_TABLE is updated to allow the "nexthop_group" to allow both keys from NEXT_HOP_GROUP_TABLE and from the new CLASS_BASED_NEXT_HOP_GROUP_TABLE.
+```
+### ROUTE_TABLE
+    ;Stores a list of routes
+    ;Status: Mandatory
+    key           = ROUTE_TABLE:prefix
+    nexthop       = *prefix, ;IP addresses separated “,” (empty indicates no gateway)
+    ifname        = *PORT_TABLE.key,   ; zero or more separated by “,” (zero indicates no interface)
+    blackhole     = BIT ; Set to 1 if this route is a blackhole (or null0)
+    nexthop_group = NEXT_HOP_GROUP_TABLE.key or CLASS_BASED_NEXT_HOP_GROUP_TABLE.key ; index within the NEXT_HOP_GROUP_TABLE or CLASS_BASED_NEXT_HOP_GROUP_TABLE, optionally used instead of nexthop and intf fields
```
kperumalbfn, line 114: @TACappleman Could you also add the changes to Sonic orchagent(if any) when some or all members of CBF NHG in ROUTE_TABLE are not resolved?

doc/cbf/cbf_hld.md
```diff
@@ -0,0 +1,215 @@
...
+
+# 2 Requirement Overview
+## 2.1 Functional Requirements
+
+Allow traffic to be forwarded through the network based on their DSCP/EXP values following these rules:
+- If a packet is not matched against an FC value and the route for its destination does not reference a CBF NHG, the packet will use the route's NH
+- If a packet is not matched against an FC value and the route for its destination references a CBF NHG, the packet will be dropped
+- If a packet is matched against an FC value and the route for its destination does not reference a CBF NHG, the packet will use the route's NH
+- If a packet is matched against an FC value and the route for its destination references a CBF NHG which maps the packet's FC value, the packet will use the mapped NHG
+- If a packet is matched against an FC value and the route for its destination references a CBF NHG which doesn't map the packet's FC value, the packet will be dropped
```
kperumalbfn, line 68: Any CBF specific drop reason for the above drop cases?

doc/cbf/cbf_hld.md
```diff
@@ -0,0 +1,215 @@
...
+     1) "3" ;exp
+     2) "3" ;fc
+     3) "6"
+     4) "5"
+     5) "7"
+     6) "5"
+     7) "8"
+     8) "7"
+     9) "9"
+    10) "8"
```
kperumalbfn, line 174: Does orchagent checks for valid FC based on SAI_SWITCH_ATTR_MAX_NUMBER_OF_FORWARDING_CLASSES?

doc/cbf/cbf_hld.md
```diff
@@ -0,0 +1,215 @@
...
+```
+This feature enables opeartors, among other things, to send the important (foreground) traffic through the shortest path, while sending the background traffic through longer paths to still give it some bandwidth instead of using QoS queues which may block background traffic from getting bandwitdh.
+
+These new class based next hop groups are allowed thanks to the changes in https://github.com/opencomputeproject/SAI/pull/1193, which allow a next hop group object to also have other next hop group objects as members of the group along with the next hop objects. The way such a next hop group works is that a packet which has a Forwarding Class value of X will be matched against an appropriate member of this group, selected based on the Forwarding Class value thanks to the "class_map" property of the group. As an example, given the CBF group with members Nhg1, Nhg2 and Nhg3 and a class map of FC 0 -> Nhg1, FC 1 -> Nhg2 and FC 3 -> Nhg3, a packet which has an FC value of 0 will be forwarded using Nhg1. Note that multiple FC values can point to the same member, but a single FC value can't be mapped to more than one member.
```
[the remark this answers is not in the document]
abanu-ms: We're not going to offer this kind of support for now.

doc/cbf/cbf_hld.md
abanu-ms: Yes, it will.

doc/cbf/cbf_hld.md
abanu-ms: Both drop cases have the reasoning that a CBF NHG needs to forward a packet based on its FC value. If that FC value doesn't exist or the FC value isn't mapped by the CBF NHG to a specific NH, we can't make any assumptions on which NH we should choose. The user user application should make sure to provide a correct mapping for each CBF NHG and to have the correct CBF maps configured.

doc/cbf/cbf_hld.md
abanu-ms: If a `ROUTE_TABLE` entry references a CBF NHG which hasn't been added yet by the user application, the behavior is the same as for a normal NHG - the route entry will remain in the Consumer's queue for the event of the NHG being created at some point which will allow the route entry to be created.

If a `ROUTE_TABLE` entry references a CBF NHG which has been added by the user application, but doesn't exist in the `ASIC_DB` (because the NH objects weren't created yet in the ASIC_DB perhaps), the same thing as in the first case will happen. `RouteOrch` will keep the entry in the queue, trying to create it. Beyond that, the CBF NHG handler will also keep trying to create the CBF NHG. This behavior is covered by the following statement from the HLD:
`otherwise the task will be kept in the process queue for the event of the missing member(s) being created which would allow the class based next hop group to be created.`

TACappleman:
@zhangyanzhao can this be merged?

zhangyanzhao:
@TACappleman   https://github.com/Azure/sonic-swss/pull/1963 #1963 is marked as a code PR but not merged yet.  What is the plan for that PR?

TACappleman:
We have a review from Intel which we've almost finished dealing with, and are still awaiting an internal review. Expect to get it merged this week

zhangyanzhao:
All code PRs are merged, HLD is reviewed in community. Merge it.

PR #802, opened 2021-06-16
[VOQ LAG HLD] Updates for show commands

vganesan-nokia:
Show commands section is updated to include syntax and sample output

### DIFF
--- a/doc/voq/lag_hld.md
+++ b/doc/voq/lag_hld.md
@@ -574,10 +574,10 @@ Usage: show chassis system-lags [OPTIONS] [SYSTEMLAGNAME]
   Show VOQ system lags information
 
 Options:
-  -x, --asicname TEXT   Asic name
-  -t, --hostname TEXT   Host or Linecard name
-  --verbose             Enable verbose output
-  -?, -h, --help        Show this message and exit.
+  -n, --asicname     TEXT   Asic name
+  -l, --linecardname TEXT   Linecard or Host name
+  --verbose                 Enable verbose output
+  -?, -h, --help            Show this message and exit.
 admin@Linecard2:~$ 
 ```
 ### Sample output
--- a/doc/voq/lag_hld.md
+++ b/doc/voq/lag_hld.md
@@ -1,6 +1,6 @@
 # LAG on Distributed VOQ System
 # High Level Design Document
-### Rev 1.0
+### Rev 1.1
 
 # Table of Contents
   * [List of Tables](#list-of-tables)
@@ -42,6 +42,7 @@
 | Rev |     Date    |       Author                                                                              | Change Description                |
 |:---:|:-----------:|:-----------------------------------------------------------------------------------------:|-----------------------------------|
 | 1.0 |             |     Srikanth Keesara, Vedavinayagam Ganesan, Sureshkannan Duraisamy (Nokia's Sonic Team)  | Initial version                   |
+| 1.1 |             |     Vedavinayagam Ganesan (Nokia's Sonic Team)                                            | Updates for show commands         |
 
 # About this Manual
 This document describes changes needed for the correct operation of LAG in a distributed VOQ system.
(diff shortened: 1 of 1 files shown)


anshuv-mfst:
@ ysmanman - could you please review , thanks

ysmanman (approved):

doc/voq/lag_hld.md
```diff
@@ -545,15 +546,50 @@ Shown below is the new attribute of SAI_OBJECT_TYPE_LAG object that is used for
...
-    2  6            PortChannel1  SystemLAG  Slot2|Asic0|Ethernet1(S)  Slot2|Asic0|Ethernet2(S)
+### Syntax
 ```
+admin@Linecard2:~$ show chassis system-lags -h
+Usage: show chassis system-lags [OPTIONS] [SYSTEMLAGNAME]
+
+  Show VOQ system lags information
+
+Options:
+  -d, --display [all]  Show internal interfaces  [default: all]
```
arlakshm, line 577: The help string is incorrect

arlakshm, line 577: Since we are displaying internal lag information, I think we dont need the `-d` option.

doc/voq/lag_hld.md
```diff
@@ -545,15 +546,50 @@ Shown below is the new attribute of SAI_OBJECT_TYPE_LAG object that is used for
...
+  -n, --namespace []   Namespace name or all
```
arlakshm, line 578: Can you elabarote what will the difference in output between the commands `show chassis system-lags` and `show chassis system-lags -n asic0` ?

arlakshm (commented):
as comments

doc/voq/lag_hld.md
vganesan-nokia: I'll fix this.

vganesan-nokia: Yes. We do not need the -d option. I'll remove this.

doc/voq/lag_hld.md
vganesan-nokia: The intention was to use this to display the lags from a specific asic. The system lag name includes the DEVICE_METADATA['localhost']['asic_name'] which is expected to be same as the namespace name. If not, this will not work. 
Do you think we can remove this option?

doc/voq/lag_hld.md
vganesan-nokia: Fixed. Removed "-d" option

vganesan-nokia: Fixed.

doc/voq/lag_hld.md
vganesan-nokia: Fixed. Since the system lag info is retrieved form chassis app db, which in global name space in supervisor card (i.e., it is namespace agnostic), instead of using namespace option, an option is introduced to take the asic_name (this need not be namespace name) which is used for filtering the output for the given specific asic.

doc/voq/lag_hld.md
```diff
@@ -545,15 +546,49 @@ Shown below is the new attribute of SAI_OBJECT_TYPE_LAG object that is used for
...
-    2  6            PortChannel1  SystemLAG  Slot2|Asic0|Ethernet1(S)  Slot2|Asic0|Ethernet2(S)
+### Syntax
 ```
+admin@Linecard2:~$ show chassis system-lags -h
+Usage: show chassis system-lags [OPTIONS] [SYSTEMLAGNAME]
+
+  Show VOQ system lags information
+
+Options:
+  -x, --asicname TEXT   Asic name
```
arlakshm, line 577: why not use `-n` here ? similar to `show chassis system_ports`

vganesan-nokia: "-n" stands for "namespace". Only for system ports we get the system ports config info from databases from specific namespace. The "-x" option stands for "switch". This is getting info form chassis data bases in the supervisor card which is not in any specific namespace. The "-x" option is used for getting entries that have the given switch name in their keys.

arlakshm: @vganesan-nokia. Thanks for the explanation. 
The option is little confusing for me. For example the output for the command `show chassis system-lags -x asic0` will display lags which have `asic0` in the key from all linecards. Is this correct ?

Would it simple if we provide option to filter on linecard iso of asic ?

vganesan-nokia: > @vganesan-nokia. Thanks for the explanation.
> The option is little confusing for me. For example the output for the command `show chassis system-lags -x asic0` will display lags which have `asic0` in the key from all linecards. Is this correct ?

**[@vganesan-nokia]** correct
> 
> Would it simple if we provide option to filter on linecard iso of asic ?

**[@vganesan-nokia ]**  Even though we filter on specific asic, the entries will be shown with the full key, which includes the 'hostname'.  So we can identify which asic is from which linecard (assuming 'hostname' has the string that identifies the linceard)

We can replace the "asicname" option. by filtering on 'hostname'. The asic name is the value of the mandatory attribute DEVICE_METADATA['localhost']['asic_name'].  So we have this filter on asic name. For 'linecard' we do not have such thing. However, in the multi slot chassis system we have a convention (not a written rule but kind of best practice) of using the mandatory attribute DEVICE_METADATA['localhost']['hostname'] to identify linecard/slot. This 'hostname' must be unique within the chassis. This is following the model of non chassis multi-asic box where the 'hostname' holds the name of the box. In chassis systems since each card acts as a multi asic box, this convention is followed. 

I feel that instead of replacing 'asicname' by 'hostname', we can have 'hostname' as additional option so that we can filter on entries for specific asic from a specific line card or entries from all asics of a specific line card. If you feel that this is too much filtering, I'll replace 'asicname' by 'hostname'.

vganesan-nokia: Fixed. Added option for filtering on host or linecard name.

PR #813, opened 2021-07-09
Add TACACS+ protocol requirement document.

liuh-80:
Add initial version of TACACS+ protocol requirement document.

### DIFF
--- a/doc/aaa/TACACS+ Requirement.md
+++ b/doc/aaa/TACACS+ Requirement.md
@@ -1,0 +1,161 @@
+# SONiC TACACS+ Protocol Requirement
+
+# Table of Contents
+- [Table of Contents](#table-of-contents)
+- [About this Manual](#about-this-manual)
+- [1 Functional Requirements](#1-functional-requirement)
+  * [1.1 Authentication](#11-authentication)
+  * [1.2 Authorization](#12-authorization)
+  * [1.3 Accounting](#13-accounting)
+  * [1.4 User script](#14-user-script)
+  * [1.5 Docker support](#15-docker-support)
+  * [1.6 Multiple TACAS server](#16-multiple-tacas-server)
+- [2 Configuration and Management Requirements](#2-configuration-and-management-requirements)
+  * [2.1 SONiC CLI](#21-sonic-cli)
+  * [2.2 Config DB](#22-config-db)
+  * [2.3 Counter](#23-counter)
+  * [2.4 System log](#24-system-log)
+- [3 Limitation](#limitation)
+  * [3.1 Command size](#31-command-size)
+  * [3.2 Server count](#32-server-count)
+
+# About this Manual
+This document provides a detailed description on the requirement of TACACS+ protocol support.
+
+# 1 Functional Requirement
+## 1.1 Authentication
+- Authentication when user login to SONiC host.
+
+
+## 1.2 Authorization
+- Authorization when:
+	- User login to SONiC host.
+	- User run command on SONiC host.
+		- User can only run commands in whitelist.
+		- For recursive command/script, only the top level command have authorization.
+
+- Only command in whitelist  visible to user:
+	- All privilege level share same whitelist to simplify design.
+	- Different privilege level have different permission to run these command.
+	- All commands in sudoers will add to the whitelist. and sudoers config file still need for RO users, this is because when remote TACACS server not avaliable, we need use local group permission for failover.
+
+- Disable user behavior in shell:
+	- Changing directories with the cd builtin.
+		- User still can access files in other folder, and use 'ls' command to list content in other folder, for example:
+        ```
+            test@testhost:~$ ls -l /etc/
+            total 880
+            drwxr-xr-x 3 test test       4096 Aug 22  2020 NetworkManager
+            drwxr-xr-x 7 test test       4096 Aug 22  2020 X11
+            drwxr-xr-x 3 test test       4096 Aug 22  2020 acpi
+        ```
+	- Setting or unsetting the values of the SHELL, PATH, HISTFILE, ENV, or BASH_ENV variables.
+	- Specifying command names containing slashes.
+	- Importing function definitions from the shell environment at startup.
+	- Parsing the value of SHELLOPTS from the shell environment at startup.
+	- Redirecting output using the ‘>’, ‘>|’, ‘<>’, ‘>&’, ‘&>’, and ‘>>’ redirection operators.
+	- Builtin commands:
+
+| **Command** | **Behavior**                |
+| -------- | -------------------------- |
+| . | Specifying a filename containing a slash as an argument.     |
+| history     | Specifying a filename containing a slash as an argument. |
+| hash     | Specifying a filename containing a slash as an argument to the -p option. |
+| exec     | Specifying a filename containing a slash as an argument to the -p option. |
+| deleting/adding     | Use the -f and -d options to the enable builtin. |
+| enable     | Using the 'enable' builtin command to enable disabled shell builtins. |
+| command     | Specifying the -p option to the 'command' builtin command. |
+| set     | Turning off restricted mode with ‘set +r’ or ‘set +o restricted’. |
+
+	- All these behavior disabled only for user input, command in script will not be affected, see here for more details: https://www.gnu.org/software/bash/manual/html_node/The-Restricted-Shell.html
+
+
+- Supported Authorization types:
+	- EXEC: user session authorization support. this happen when user login.
+	- Command: user run command in shell.
+
+- Failover:
+	- Authorization will happen before execute, if remote TACACS server not available, use local group based authorization as failover.
+
+
+## 1.3 Accounting
+ - Accounting when:
+ 	- User login to SONiC host.
+ 	- User run command on host: 
+		- Command start run.
+		- Command finish.
+		- For recursive command/script, only the top level command have Accounting.
+- Failover:
+	- Use syslog as backup when remote TACACS not avaliable.
+
+
+## 1.4 User script
+ - Any script in whitelist can run with Authorization and Accounting. 
+ - If user create a script, admin user can use config command add script to whitelist.
+
+## 1.5 Docker support
+ - Docker exec command will be covered by Authorization and Accounting. 
+ - SONiC AAA can't cover any command user run inside a docker.
+
+## 1.6 Multiple TACACS server
+ - Support config multiple TACACS server.
+ - When a server not available, will try next server as backup.
+ - When all server not available, use native failover solution.
+
+# 2 Configuration and Management Requirements
+## 2.1 SONiC CLI
+ - Enable/Disable TACACS Authorization/Accounting command
+```
+    config tacacs authorization enable
+    config tacacs authorization disable
+    config tacacs accounting enable
+    config tacacs accounting disable
+```
+
+ - Whitelist management command
+```
+    show config tacacs restrict command whitelist
+    Config tacacs restrict command whitelist remove <command>
+    Config tacacs restrict command whitelist add <command>
+```
+
+ - Counter command
+```
+    show tacacs counter
+    clear tacacs counter
+```
+
+## 2.2 Config DB
+ - TACACS AAA are fully configable by config DB.
+
+## 2.3 Counter
+ - Support  AAA counter:
+```
+    show tacacs counter
+    
+        server1: 10.1.1.45
+        Messages sent: 24
+        Messages received: 20
+        Requests accepted: 14
+        Requests rejected: 8
+        Requests timeout: 2
+        Requests retransmitted: 1
+        Bad responses: 1
+```
+
+## 2.4 System log
+- Generate system log when Authentication/Authorization/Accounting.
+- When remote TACACS server not avaliable, use system log for accounting.
+
+
+# 3 Limitation
+## 3.1 Command size
+ - TACACS protocol limittation: command + parameter size should smaller than 240 byte. The longer than 240 bytes parts will be drop.
+ 	- This limitation is a protocol level, all TACACS implementation have this limittation, include CISCO, ARISTA and Cumulus.
+ 	- Both Authorization and Accounting have this limitation.
+ 	- When user user a command longer than 240 bytes, only commands within 240 bytes will send to TACACS server. which means Accounting may lost some user input. and Authorization check can only partly check user input.
+
+
+## 3.2 Server count
+ - Max TACACS server  count was hardcoded, default count is 8.
+


doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+	- User run command on SONiC host.
+		- User can only run commands in whitelist.
+		- For recursive command/script, only the top level command have authorization.
+
+- Only command in whitelist  visible to user:
+	- All privilege level share same whitelist to simplify design.
+	- Different privilege level have different permission to run these command.
+	- All commands in sudoers will add to the whitelist. and sudoers config file still need for RO users, this is because when remote TACACS server not avaliable, we need use local group permission for failover.
+
+- Disable user behavior in shell:
```
lguohan, line 42: why rbash is a must for command authorization?

liuh-80: The reason is because for command authorization, we create a symbol link for every command we want enable authorization. 
Without rbash user can run any command, include those command we not create symbol link. 
Then they can bypass the authorization very easy.

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+# Table of Contents
+- [Table of Contents](#table-of-contents)
+- [About this Manual](#about-this-manual)
+- [1 Functional Requirements](#1-functional-requirement)
+  * [1.1 Authentication](#11-authentication)
+  * [1.2 Authorization](#12-authorization)
+  * [1.3 Accounting](#13-accounting)
+  * [1.4 User script](#14-user-script)
+  * [1.5 Docker support](#15-docker-support)
+  * [1.6 Multiple TACAS server](#16-multiple-tacas-server)
```
qiluo-msft, line 12: >TACAS [](http://example.com/codeflow?start=18&length=5)

typo

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+# 1 Functional Requirement
+## 1.1 Authentication
+- Authentication when user login to SONiC host.
+
+
+## 1.2 Authorization
+- Authorization when:
+	- User login to SONiC host.
+	- User run command on SONiC host.
+		- User can only run commands in whitelist.
```
qiluo-msft, line 34: >whitelist [](http://example.com/codeflow?start=34&length=9)

Is the whitelist per host or per user?

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+		- For recursive command/script, only the top level command have authorization.
+
+- Only command in whitelist  visible to user:
+	- All privilege level share same whitelist to simplify design.
```
qiluo-msft, line 38: >All privilege level share same whitelist to simplify design [](http://example.com/codeflow?start=3&length=59)

What is the technical difficulty to support multiple whitelists?

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+	- Different privilege level have different permission to run these command.
+	- All commands in sudoers will add to the whitelist. and sudoers config file still need for RO users, this is because when remote TACACS server not avaliable, we need use local group permission for failover.
```
qiluo-msft, line 40: >avaliable [](http://example.com/codeflow?start=149&length=9)

Accessible from sonic?

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+        ```
+            test@testhost:~$ ls -l /etc/
+            total 880
+            drwxr-xr-x 3 test test       4096 Aug 22  2020 NetworkManager
+            drwxr-xr-x 7 test test       4096 Aug 22  2020 X11
+            drwxr-xr-x 3 test test       4096 Aug 22  2020 acpi
+        ```
+	- Setting or unsetting the values of the SHELL, PATH, HISTFILE, ENV, or BASH_ENV variables.
+	- Specifying command names containing slashes.
+	- Importing function definitions from the shell environment at startup.
```
qiluo-msft, line 54: >Importing function [](http://example.com/codeflow?start=3&length=18)

Is the bash prompt still working? By default it include the  hostname and the current working directory.

doc/aaa/TACACS+ Requirement.md
liuh-80: Every command in whitelist will create a symbol link, so if we support multiple whitelist, there will be 2 solution:
1. we need create different symbol links when user login. 
2. Or we setup 2 set of symbol links, one for RW user groups which contains full set of command. and one for RO user groups which contains RO commands.

I think the solution 2 seems better, I will try it during POC, and update this document if it can work.

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+		- User still can access files in other folder, and use 'ls' command to list content in other folder, for example:
+        ```
+            test@testhost:~$ ls -l /etc/
+            total 880
+            drwxr-xr-x 3 test test       4096 Aug 22  2020 NetworkManager
+            drwxr-xr-x 7 test test       4096 Aug 22  2020 X11
+            drwxr-xr-x 3 test test       4096 Aug 22  2020 acpi
+        ```
+	- Setting or unsetting the values of the SHELL, PATH, HISTFILE, ENV, or BASH_ENV variables.
+	- Specifying command names containing slashes.
```
qiluo-msft, line 53: >command names containing slashes [](http://example.com/codeflow?start=14&length=32)

Do you mean `/bin/diff` is not working?

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+	- Importing function definitions from the shell environment at startup.
+	- Parsing the value of SHELLOPTS from the shell environment at startup.
+	- Redirecting output using the ‘>’, ‘>|’, ‘<>’, ‘>&’, ‘&>’, and ‘>>’ redirection operators.
```
qiluo-msft, line 56: >Redirecting [](http://example.com/codeflow?start=3&length=11)

What are the alternatives for user to achieve similar goal?

doc/aaa/TACACS+ Requirement.md
qiluo-msft, line 42: >Disable [](http://example.com/codeflow?start=2&length=7)

Disable for RO users are understandable. Is it possible to keep RW users capability at the same time of AAA?

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+
+- Supported Authorization types:
+	- EXEC: user session authorization support. this happen when user login.
+	- Command: user run command in shell.
+
+- Failover:
+	- Authorization will happen before execute, if remote TACACS server not available, use local group based authorization as failover.
+
+
+## 1.3 Accounting
```
qiluo-msft, line 81: >Accounting [](http://example.com/codeflow?start=7&length=10)

Could you explain what is accounting and what will be accounted?

doc/aaa/TACACS+ Requirement.md
liuh-80: The bash prompt still working, the rbash just not allow user create their bash function and import it to bash environment. usually those function works as some hook.

doc/aaa/TACACS+ Requirement.md
liuh-80: If user run /bin/diff , then it will not work, but if we add 'bin' to path, then user still can run diff command, here is example: 

liuh@MININT-6C763K9:~$ /etc/date
rbash: /etc/date: restricted: cannot specify `/' in command names

liuh@MININT-6C763K9:~$ date
Fri Jul  9 15:15:42 CST 2021

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+		- Command start run.
+		- Command finish.
+		- For recursive command/script, only the top level command have Accounting.
+- Failover:
+	- Use syslog as backup when remote TACACS not avaliable.
+
+
+## 1.4 User script
+ - Any script in whitelist can run with Authorization and Accounting. 
+ - If user create a script, admin user can use config command add script to whitelist.
```
qiluo-msft, line 94: >whitelist [](http://example.com/codeflow?start=76&length=9)

Is it true that tacacs server owns this whitelist? How can a sonic change the server's whitelist?

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+```
+    config tacacs authorization enable
+    config tacacs authorization disable
+    config tacacs accounting enable
+    config tacacs accounting disable
+```
+
+ - Whitelist management command
+```
+    show config tacacs restrict command whitelist
```
qiluo-msft, line 117: >show config [](http://example.com/codeflow?start=4&length=11)

`show config` -> `show` ?

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+    Config tacacs restrict command whitelist remove <command>
```
qiluo-msft, line 118: >Config tacacs restrict command whitelist [](http://example.com/codeflow?start=4&length=40)

The command is super long.

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+    Config tacacs restrict command whitelist add <command>
+```
+
+ - Counter command
+```
+    show tacacs counter
+    clear tacacs counter
```
qiluo-msft, line 125: >clear [](http://example.com/codeflow?start=4&length=5)

`clear` is part of `ncurses-bin`. We'd better not to override.

doc/aaa/TACACS+ Requirement.md
liuh-80: In that case, we need consider some other solution. I will check other solution in POC and update this document.

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,161 @@
...
+        Requests accepted: 14
+        Requests rejected: 8
+        Requests timeout: 2
+        Requests retransmitted: 1
+        Bad responses: 1
+```
+
+## 2.4 System log
+- Generate system log when Authentication/Authorization/Accounting.
+- When remote TACACS server not avaliable, use system log for accounting.
```
qiluo-msft, line 148: >When remote TACACS server not avaliable [](http://example.com/codeflow?start=2&length=39)

When tacacs server is accessible, do we also use local syslog to account?

doc/aaa/TACACS+ Requirement.md
qiluo-msft: I see the difficult if we must use symlink. You can keep only one.

doc/aaa/TACACS+ Requirement.md
liuh-80: Will update document, accounting basically record what user doing, and following event will accounted:
user login/logout, user start run a command, and user command finish.

doc/aaa/TACACS+ Requirement.md
liuh-80: Will add example to document.

Tacacs server not have this whitelist, and sonic can't change the tacacs server setting.
However, for user script, if the script name follow some naming convention, and tacacs server side allow run those script which follow the naming convention, user can create their own script, for example:

1. Tacacs service allow RW user run any script named as 'user_script_*'
2. RW user create a new script on sonic host, script name is 'user_script_collect_information.sh'
3. Then user can add this script to whitelist and run it.

But if tacacs service not have any config, then user can't run their script because Authorization will block it.

doc/aaa/TACACS+ Requirement.md
liuh-80: Improved, will push update later.

doc/aaa/TACACS+ Requirement.md
liuh-80: I think it's not necessary.
If tacacs server is accessible, we only put trace log to syslog for metering and debugging.

doc/aaa/TACACS+ Requirement.md
liuh-80: Fixed

doc/aaa/TACACS+ Requirement.md
liuh-80: Fixed.

doc/aaa/TACACS+ Requirement.md
liuh-80: Fixed.

doc/aaa/TACACS+ Requirement.md
liuh-80: Fixed.

doc/aaa/TACACS+ Requirement.md
```diff
@@ -0,0 +1,178 @@
...
+    config tacacs authorization enable
+    config tacacs authorization disable
+    config tacacs accounting enable
+    config tacacs accounting disable
+```
+
+ - Whitelist management command
+```
+    show tacacs whitelist
+    Config tacacs whitelist remove <command>
```
qiluo-msft, line 135: >Config [](http://example.com/codeflow?start=4&length=6)

`Config` -> `config`

doc/aaa/TACACS+ Requirement.md
qiluo-msft: We have a similar command `sonic-clear`. ref: [https://github.com/Azure/sonic-utilities/blob/888701b67fd4f1cc5b9da534a360048f93f263f4/setup.py#L157](https://github.com/Azure/sonic-utilities/blob/888701b67fd4f1cc5b9da534a360048f93f263f4/setup.py#L157)

doc/aaa/TACACS+ Requirement.md
qiluo-msft: I guess your `system log` == `syslog`.
Then if the networking is not stable, you will have some random log lines locally, does it help anything?

For login authentication, the log messages always go to local syslog, no matter remote tacacs accessible or not.

doc/aaa/TACACS+ Requirement.md
liuh-80: Fixed with sonic-clear command, will push update later.

doc/aaa/TACACS+ Requirement.md
liuh-80: Document updated.

What I mean here is when the remote TACACS service not accessible, for example TACACS server down after user login, but networking still healthy, then the Authorization and Accounting information will write to syslog. Then the syslog will upload via network later, and those information will not lost.

doc/aaa/TACACS+ Requirement.md
liuh-80: Fixed, will push update later.

doc/aaa/TACACS+ Requirement.md
liuh-80: I check some other solution and they are based on shell function hook, those solution have a common limitation: user can bypass restriction by ingest new shell hook.

Then after check the rbash restriction, I think we have 2 solution:
1. Patch bash to allow redirection in restrict mode, and accept most other restriction because those are necessary for use tacplus-auth: those restriction we accept will make both RO/RW user can't access commands not in our white list, and make user can't switch to other shell. 

2.Patch bash, so bash can do Authorization before execute any user command.
       With this solution, there is no any restriction in router machine side, in TACACS server side we just need disable user switch to other shell.

I will try a POC for solution 2, it's seems will be a minor change.

doc/aaa/TACACS+ Requirement.md
liuh-80: By patch bash, we can allow user use redirect, will remove this from document.

qiluo-msft: Do you have a link to the patch? I am wondering if redirect could be configurable. For example, RW users could use redirect as normal, but RO users are banned to use redirect.

liuh-80: It's a very simple code change, remove following code in  redir.c line:836

	#if defined (RESTRICTED_SHELL)
	      if (restricted && (WRITE_REDIRECT (ri)))
	{
	  free (redirectee_word);
	  return (RESTRICTED_REDIRECT);
	}
	#endif /* RESTRICTED_SHELL */

So for RW allow and RO disable, we can:
1. create a variable isReadonlyUser
2. When bash startup, check current user group and set this variable.
3. If any behavior we want to disable for RO only we can change code like this:

	#if defined (RESTRICTED_SHELL)
	      if (restricted && **isReadonlyUser** && (WRITE_REDIRECT (ri)))
	{
	  free (redirectee_word);
	  return (RESTRICTED_REDIRECT);
	}
	#endif /* RESTRICTED_SHELL */

doc/aaa/TACACS+ Requirement.md
liuh-80: The POC for solution 2 is ready, which will authorize any user 'disk' command, so if solution 2 is acceptable, will update the document.

For 'disk' command, in bash, if a command is a file stored in disk, it's a disk command. built-in command and bash function are not disk command,  executable file and script are disk command.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,521 @@
...
+## 3.1 Command size
+ - TACACS protocol limitation: command + parameter size should smaller than 240 byte. The longer than 240 bytes parts will be drop.
+ 	- This limitation is a protocol level, all TACACS implementation have this limittation, include CISCO, ARISTA and Cumulus.
+ 	- Both Authorization and Accounting have this limitation.
+ 	- When user user a command longer than 240 bytes, only commands within 240 bytes will send to TACACS server. which means Accounting may lost some user input. and Authorization check can only partly check user input.
+
+## 3.2 Server count
+ - Max TACACS server  count was hardcoded, default count is 8.
+
+## 3.1 Local authorization
```
qiluo-msft, line 154: >3.1 [](http://example.com/codeflow?start=3&length=3)

Section numbers are wrong.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,521 @@
...
+ - Operation system limitation: SONiC based on linux system, so permission to execute local command are managed by Linux file permission control. This means TACACS+ authorization can't config to disable 'local' authorization, and local authorization must be last authorization in authorization method list.
```
qiluo-msft, line 155: >can't config to disable 'local' authorization [](http://example.com/codeflow?start=179&length=45)

If tacacs server is not reachable during authorization, we have option to fail the authorization, so local auth is disabled.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,521 @@
...
+    // accounting with TACACS+ server
+    config aaa accounting tacacs
+
+    // accounting with local syslog
+    config aaa accounting local
+```
+
+ - When config AAA authorization with "no" prefix, SONiC will use local authorization, so following commands have same effect
+```
+    no config aaa authorization tacacs local
```
qiluo-msft, line 334: >no config [](http://example.com/codeflow?start=4&length=9)

It is difficult to implement `no config`, can we just config it as `local` as the last resort?

liuh-80: 'no config' have same effect with local.
I will update document.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,521 @@
...
+ - When config AAA accounting with "no" prefix, SONiC will use stop accounting, following command have same effect.
+```
+    no config aaa authorization tacacs local
+    no config aaa authorization tacacs
+    no config aaa authorization local
+```
+# 5 Error handling
+ - Bash plugin for authorization will return error code [Bash](#bash). and patched Bash will:
+   - Output error log to syslog.
+   - Output error message to stdout.
```
qiluo-msft, line 348: >stdout [](http://example.com/codeflow?start=29&length=6)

stderr?

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,521 @@
...
+```
+    Verify TACACS+ authorization passed.
+```
+
+ - [audisp-tacplus](#audisp-tacplus) test, all TACACS+ server accessable.
+```
+    Verify TACACS+ accounting succeeded.
+```
+
+ - [audisp-tacplus](#audisp-tacplus) test, all TACACS+ server not accessable.
```
qiluo-msft, line 389: >all TACACS+ server not accessable [](http://example.com/codeflow?start=43&length=33)

Add a test case which tacacs is available for a while, user could login, run some comands, and suddently tacacs is not accessible, and test for the expected behavior.

liu-hua-joe-80: Test added:

config AAA authorization with TACACS+ only:

- when user login server are accessible.
- user run some command in whitelist and server are accessible.
- then all  server not accessible, and run some command

doc/aaa/TACACS+ Design.md
liu-hua-joe-80: Fixed.

doc/aaa/TACACS+ Design.md
liu-hua-joe-80: Fixed.

doc/aaa/TACACS+ Design.md
liu-hua-joe-80: Fixed.

doc/aaa/TACACS+ Design.md
qiluo-msft: My points is that you have 2 options
1. tacacs + local
2. tacacs only, no local
And this is not a "Operation system limitation".

liu-hua-joe-80: Here is a example of this issue:
           user name: domainuser
           When user login, tacacs return privilege level 0 so in SONiC this user is a RO user, then this user on sonic only have Linux permission to run RO commands. 
           But in tacacs server side, the 'config' command was config as allow user run.

Then when user trying to run the 'config' command, tacacs authorization succeeded, but user still can't run this command because user not have Linux permission.

In EOS system, the behavior is different:
          We add domainuser as local RO user.
          In tacacs server config this user can run RW command.
         Then we user run a RW command, user can run it.

The reason is the EOS system can bypass the local permission check, but SONiC is a Linux system, and currently we using Linux permission for 'local authorization', so tacacs authorization can't bypass the Linux permission check.

qiluo-msft: I see your point!
Actually I am considering another config "tacacs only, no fallback to local", which means tacacs server unreachable, will stop the already-login user to run anything except `logout` or `exit`. Then the user need to re-login with local account, local authentication and local authorization.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,526 @@
...
+
+
+## 1.2 Authorization
+- Authorization when:
+	- User login to SONiC host.
+	- User run any executable file or script on SONiC host.
+      - The full path and parameters will be send to TACACS+ server side for authorization.
+      - For recursive command/script, only the top level command have authorization.
+        - No authorization for  bash built-in command and bash function, but if a bash function call any executable file or script, those executable file or script will have authorization.
+      - Commands entered through the console not have authorization by default.
```
qiluo-msft, line 59: >Commands entered through the console not have authorization by default. [](http://example.com/codeflow?start=8&length=71)

No need to diffentiate sonsole access.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,544 @@
...
+   - Use regex in /etc/sudoers PASSWD_CMDS to identify user secrets.
+   - User secret will be replaced with ***.
+
+## 1.3 User script
+ - User can create and run their own script. 
+ - If user create a script but TACACS+ service side not have configuration to allow user run this script, user script will be blocked by authorization.
+
+## 1.4 Docker support
+ - Docker exec command will be covered by Authorization and Accounting. 
+ - Any command run inside docker container will not covered by Authorization and Accounting.
```
jeff-yin, line 99: Should move/add this to the limitations section. 
A user can run commands inside a container to evade TACACS authorization and accounting. 

It may be good to add a recommendation to administrators to disable access to 'docker exec' for most users.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,544 @@
...
+
+## 1.5 Multiple TACACS server
+ - Support config multiple TACACS server.
+ - When a server not accessible, will try next server as backup.
```
dgsudharsan, line 103: Is there a way to specify which server is primary in the list of servers?

seiferteric:
To expand on my comments during the review meeting, here is a list of my concerns, but I don't know how you implemented this patch in bash so some of these concerns may be invalid, so please excuse/ignore them if you have already covered these cases.

Concerns with tacacs+ command authorization:

Cannot limit built-in commands:
  - You mention that you cannot limit shell built-ins, but bash has a lot of built-in commands that may allow you to circumvent controls:
    - exec: exec is a built-in command that will execute another process and replace it's own process with the new one. This alone could bypass the authorization.
    - eval: similar to exec
    - alias: can I fool authorization by aliasing an allowed command to a disallowed one?
    - functions: Can a user create a shell function that calls a disallowed command? How about in the users .bashrc, .bash_profile?
    - command: built-in to call another command like: "command /bin/disallowed_command"
    - source or .: sources (runs) a script, will this block disallowed commands?
    - trap: The commands in arg are to be read and executed when the shell receives signal sigspec. 

I am not certain how you limit what bash can execute, is it only the top level command, or all sub-commands that a command might run?
  - If a user can run a script that uses /bin/sh, can a user not directly run /bin/sh? Or "bash -c /bin/sh"?
  - Can a user run the loader directly? Like: /lib/x86_64-linux-gnu/ld-linux-x86-64.so.2 /bin/sh ?
  - Need to disable chroot (should not be a problem) but could be used to trick authorization

Commands that can run other commands:
  - Some commands allow the user to run other commands:
    - find: You may not suspect find command but it has an -exec option to run a command on found files. Not sure if this spawns a sub-shell that will honor the authorization, or forks/exec directly which will not honor authorization.


LD_LIBRARY_PATH LD_PRELOAD:
  - These ENV vars can be set to circumvent the use or the installed shared libraries. It could be possible for a user to copy over their own version of libc or any other library that could execute arbitrary commands.

example:
This is just an example using libfakeroot which is harmless, but can demonstrate that you can change library calls with custom loaded libraries. This shows the "whoami" command being tricked to think we are root:

    eric.e.seifert@xenlogin-eqx-04:~:$ LD_PRELOAD="/usr/lib/x86_64-linux-gnu/libfakeroot/libfakeroot-0.so /usr/lib/x86_64-linux-gnu/libfakeroot/libfakeroot-sysv.so"  whoami
    root
    eric.e.seifert@xenlogin-eqx-04:~:$


If you are going to go this route, it might be safer/easier to path glibc execve directly to check user authorization...

seiferteric:
I wonder if SELinux would be better suited to restrict user commands? Here is some examples: https://access.redhat.com/solutions/65822

doc/aaa/TACACS+ Design.md
liuh-80: There is no special flag to specify primary server, the first server in the TACACS+ config will will be the primary server.

liuh-80:
@seiferteric, Thanks for your comments, about your concern:

1. The reason we doesn't limit built-in commands: in bash, when user run any command stored in disk, the code will go to following method:
               execute_cmd.c

                       /* Call execve (), handling interpreting shell scripts, and handling
                 exec failures. */
              int
              shell_execve (command, args, env)
                   char *command;
                   char **args, **env;

    And our patch will work inside this method to check TACACS+ authorization before command been executed.
    Then user cannot use built-in command/function/variable to bypass TACACS+ authorization. So following way to bypass authorization will not work:
           1. Use built-in command to bypass authorization.
           2. Use shell function to bypass authorization.
           3. Use shell variable to bypass authorization.

2. For the bash script authorization, we will only check the script name,  and will not check the sub commands inside the script, this may cause potential security issue, so for RO user, administrator should disable user from run any user defined script. and for Admin user, because they already have permission to do anything, so it's not necessary to block them from create/run their scripts.

3. Start another harmless process and run command from it will be a security risk, so administrator should disable RO user from run following commands:
           1. Run another shell.
           2. Run another interpreter command.
           3. Run loader.
           4. Any command can run other commands, for example: find, VI.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
+# SONiC TACACS+ improve
```
qiluo-msft, line 1: >improve [](http://example.com/codeflow?start=16&length=7)

improvement?

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+   - User command authorization with local permission.
+
+ - New features:
+   - User command authorization with TACACS+ server.
+   - User command accounting with TACACS+ server.
+
+# 1 Functional Requirement
+## 1.1 User command authorization
+ - Authorization when user run any executable file or script on SONiC host.
+   - The full path and parameters will be send to TACACS+ server side for authorization.
```
qiluo-msft, line 59: >send [](http://example.com/codeflow?start=42&length=4)

sent

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+   - For recursive command/script, only the top level command have authorization.
+   - No authorization for  bash built-in command and bash function, but if a bash function call any executable file or script, those executable file or script will have authorization.
```
qiluo-msft, line 61: >  [](http://example.com/codeflow?start=25&length=1)

One extra blank

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+     - When all TACACS+ server not accessible, user can login with this method.
+## 1.2 User command accounting
+ - Accounting is the action of recording what a user is doing, and/or has done.
+
+ - Following event will be accounted:
+   - Command start event.
+   - Command finish event.
+
+ - User command in Docker will not be accounted.
+   - User command in docker actually run by docker service, so we can't identify if command are run by user or system service.
```
qiluo-msft, line 85: >docker [](http://example.com/codeflow?start=21&length=6)

I guess you mean shell environment inside docker container. On the other hand, the `docker exec <container> <process>` itself will be accounted.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+ - Max TACACS server  count was hardcoded, default count is 8.
+
+## 3.3 Local authorization
+ - Operation system limitation: SONiC based on Linux system, so permission to execute local command are managed by Linux file permission control. This means when enable both TACACS+ authorization and local authorization, local authorization will always happen after TACACS+ authorization.
+
+## 3.4 Docker support
+ - Any command run inside docker container will not covered by Authorization and Accounting.
+ - Docker exec command will be covered by Authorization and Accounting. 
+ - Administrator may setup TACACS+ rules to block docker exec command for RO user:
+   - user can start a interactive shell on the docker container, then run command inside container to evade TACACS+ authorization and accounting.
```
qiluo-msft, line 163: >a [](http://example.com/codeflow?start=20&length=1)

an

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+ 	- When user user a command longer than 240 bytes, only commands within 240 bytes will send to TACACS server. which means Accounting may lost some user input. and Authorization check can only partly check user input.
+
+## 3.2 Server count
+ - Max TACACS server  count was hardcoded, default count is 8.
+
+## 3.3 Local authorization
+ - Operation system limitation: SONiC based on Linux system, so permission to execute local command are managed by Linux file permission control. This means when enable both TACACS+ authorization and local authorization, local authorization will always happen after TACACS+ authorization.
+
+## 3.4 Docker support
+ - Any command run inside docker container will not covered by Authorization and Accounting.
```
qiluo-msft, line 160: >inside [](http://example.com/codeflow?start=19&length=6)

inside a shell in a docker container

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+ - Docker exec command will be covered by Authorization and Accounting. 
+ - Administrator may setup TACACS+ rules to block docker exec command for RO user:
+   - user can start a interactive shell on the docker container, then run command inside container to evade TACACS+ authorization and accounting.
+
+## 3.5 Recursive commands
+  - Many linux command allow user start a harmless process and run another command from it,  administrator may setup TACACS+ rules from server side to block user from:
+    - Run another shell.
+    - Run interpreter, for example python.
+    - Run loader, for example /lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
+    - Run find/VI command which can run other commands inside it.
```
qiluo-msft, line 170: >find/VI [](http://example.com/codeflow?start=10&length=7)

I can understand vim. `find` is similiar to other normal command and we can use regex to limit its power, right?

liuh-80: Yes, we can use regex to limit user run 'find' with -exec parameter.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+ - Support config multiple TACACS server.
+ - First server in the list will be primary server.
+ - When a server not accessible, will try next server as backup.
+ - When all server not accessible from SONiC, use native failover solution.
+
+# 2 Configuration and Management Requirements
+## 2.1 SONiC CLI
+ - Enable/Disable TACACS Authorization/Accounting command
+```
+	config aaa authorization {local | tacacs+}
```
qiluo-msft, line 113: >	 [](http://example.com/codeflow?start=0&length=1)

Use spaces instread of tab.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+    config aaa accounting {local | tacacs+ | disable}
```
qiluo-msft, line 114: >tacacs+ [](http://example.com/codeflow?start=35&length=7)

In below section, you use `tacacs` without `+`.

doc/aaa/TACACS+ Design.md
qiluo-msft, line 113: >local | tacacs [](http://example.com/codeflow?start=27&length=14)

In below sectin, you support combined `local | tacacs`, which is different from this syntax.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+
+- config AAA authorization with TACACS+ only:
+```
+    Verify TACACS+ user run command in server side whitelist:
+        If command have local permission, user can run command.
+        If command not have local permission, user can't run command.
+    Verify TACACS+ user can't run command not in server side whitelist.
+```
+
+- config AAA authorization with TACACS+ only:
```
qiluo-msft, line 428: >config AAA authorization with TACACS+ only [](http://example.com/codeflow?start=2&length=42)

This a very practical work flow. Normally use will continue to login with local account, and run command as local authorization, and even tacacs server recover, it does not impact local user.

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+    - when user login server are accessible.
+    - user run some command in whitelist and server are accessible.
+    - then all  server not accessible, and run some command
```
qiluo-msft, line 431: >  [](http://example.com/codeflow?start=15&length=1)

Remove extra blank

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,537 @@
...
+           +---------+                      +-------+--------+
+           |         |                      |                |
+           |   CLI   +---------------------->    ConifgDB    |
+           |         |                      |                |
+           +---------+                      +----------------+
+```
+
+
+## 4.3 ConfigDB Schema
+ - Existing tables, for more detail please check [TACACS+ Authentication](#TACPLUS-Authentication)
```
qiluo-msft, line 309: >[TACACS+ Authentication](#TACPLUS-Authentication) [](http://example.com/codeflow?start=49&length=49)

Is it a wrong link?
I guess [https://github.com/Azure/SONiC/blob/master/doc/aaa/TACACS%2B%20Authentication.md#aaa-table-schema](https://github.com/Azure/SONiC/blob/master/doc/aaa/TACACS%2B%20Authentication.md#aaa-table-schema)

doc/aaa/TACACS+ Design.md
liuh-80: Fixed

doc/aaa/TACACS+ Design.md
liuh-80: Fixed.

doc/aaa/TACACS+ Design.md
liuh-80: Fixed.

doc/aaa/TACACS+ Design.md
liuh-80: Fixed.

doc/aaa/TACACS+ Design.md
liuh-80: Fixed.

doc/aaa/TACACS+ Design.md
liuh-80: Fixed.

doc/aaa/TACACS+ Design.md
liuh-80: Verified with POC code, we can block -exec parameter.

doc/aaa/TACACS+ Design.md
liuh-80: Fixed.

doc/aaa/TACACS+ Design.md
liuh-80: Fixed, all changed to tacacs+

doc/aaa/TACACS+ Design.md
liuh-80: Fixed.

doc/aaa/TACACS+ Design.md
liuh-80: Fixed.

doc/aaa/TACACS+ Design.md
liuh-80: Fixed.

doc/aaa/TACACS+ Design.md
liuh-80: Will change to config aaa authorization {local | tacacs+ | local tacacs+}

doc/aaa/TACACS+ Design.md
liuh-80: config AAA authorization with TACACS+ and local, but server not accessible:

Add this step: when remote tacacs accessible, local user still can run command. doesn't impact local user authorization.

Check if we have PR cover this: when tacacs accessible, local can't login.

qiluo-msft: Discuss offline, and you will add some step in another test case below.

liuh-80: Fixed.

doc/aaa/TACACS+ Design.md
liuh-80: Fixed.

zhangyanzhao:
Please link the code PR by referring to this example https://github.com/Azure/SONiC/pull/806

doc/aaa/TACACS+ Design.md
```diff
@@ -0,0 +1,559 @@
...
+    Verify TACACS+ user can't run command not in server side whitelist.
+    Verify Local user can't login.
+```
+
+- config AAA authorization with TACACS+ and local, but server not accessible:
+```
+    Verify TACACS+ user can run command not in server side whitelist but have permission in local.
+    Verify TACACS+ user can't run command in server side whitelist but not have permission in local.
+    Verify Local user can login, and run command with local permission.
+    Verify after Local user login, then server accessible, Local user still can run command with local permission.
```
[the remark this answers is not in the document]
liuh-80: Fixed. also update AAA table schema according to current SONiC yang model.

zhangyanzhao:
@liuh-80  please update this PR to add the code PRs with example of https://github.com/Azure/SONiC/pull/806. Thanks.

PR #840, opened 2021-08-13
Initial draft of P4RT App DB Schema

kishanps:
P4RT App Db schema

### DIFF
--- a/P4RT_App_DB_Schema.md
+++ b/P4RT_App_DB_Schema.md
@@ -1,0 +1,485 @@
+# P4RT APPL DB Schema
+
+_Rev v0.1_
+
+
+## Revision
+
+Rev  | RevDate    | Author(s)   | Change Description
+---- | ---------- | ----------- | ------------------
+v0.1 | 07/19/2021 | Google, ONF | Initial Version
+
+
+#
+Scope
+
+This document captures the schema changes done for supporting the P4RT app in Sonic. 
+
+
+# 
+Overview
+
+As part of the PINS effort, a new P4RT table is introduced into the APPL_DB. The P4RT application and P4Orch will communicate through this newly introduced table. The new table organises entries into sub-tables based on the functionality the table caters to, based on the P4 program.
+
+As part of the PINS effort there are also extensions made to the existing SWITCH_TABLE and a new HASH_TABLE added for the hashing parameters.
+
+
+# 
+P4RT Tables
+
+P4RT application introduces new tables with the prefix P4RT in the APPL_DB that holds the definitions and entries for the various types of entities that it programs in the DB. It follows the convention of **P4RT:&lt;TableType>&lt;TableName>** where TableType is of **FIXED** or configurable (only **ACL** for now) and TableName is the specific table in the SAI pipeline specification like router interface, neighbor, nexthop, IPV4/IPV6 tables etc. 
+
+The key and value entries in the table are generally formatted according to the P4 program of the corresponding table. The table entries are uniformly represented in both the fixed SAI and configurable SAI aspects (ACL tables) or P4 extensions. The table format is compatible with other tables in SONiC and follows a human readable format to aid debugging. 
+
+
+# 
+Common definitions
+
+Every table entry is stored using a key made up of the table name, all match keys and the priority (if there is one). Then the value in the Redis DB is a hash map, mapping `action` to the name of the action and mapping all parameter names to the parameter values. Every key and the value in the Redis DB is represented as a JSON object that makes it consistent with the ASIC_DB representation and permitting the use of any char with variable length. 
+
+**Match Fields**
+
+Match fields may be formatted as one of three types:
+
+
+
+* **Exact** (`"<value>"`)
+    * Available to all tables.
+    * Example: `"match/ether_type": "0x8000"`
+* **Ternary** (`"<value> & <mask>"`)
+    * Available for ACL tables.
+    * Example: `"match/ether_type": "0x8000 & 0xFFFF"`
+* **LPM** (`"<value>/<prefix_length>"`)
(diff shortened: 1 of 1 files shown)


P4RT_App_DB_Schema.md
```diff
@@ -0,0 +1,396 @@
...
+6) "my metadata"
+
+
+```
+
+
+
+##### 3. FIXED_* Tables
+
+FIXED_* Tables are similar to ACL_* Tables. However, FIXED_* Tables have a predefined set of matches & actions. This prevents the need for a DEFINITION table for these tables.
```
lguohan, line 266: for most of the FIXED table, there are already existing orch handling the those tables, for example, the neighorch handles the neighbor table, the routeroch handles the route table. with this design, in the orchagent, we now have two orch handling the same sai objects. It appears that current design introduce redundancy to handle same sai objects, and is causing architecture divergence in orchagent. It also means P4RT orch could not leverage the investment of sonic community on existing orchagent. what is the reason to have separate table instead of reusing existing app db, like ROUTE_TABLE, NEXTHOP_TABLE, and etc. on? What is the north star here?

P4RT_App_DB_Schema.md
```diff
@@ -0,0 +1,396 @@
...
+| match/<name> | String / Integer | Describes a match parameter for the rule. The match/<name> comes from the DESCRIPTION table. Matches may be exact or ternary values. See the Match Fields formatting section for details. Certain types may only appear as exact matches: SAI_ACL_TABLE_ATTR_FIELD_ACL_IP_TYPE/* SAI_ACL_TABLE_ATTR_FIELD_IN_PORT SAI_ACL_TABLE_ATTR_FIELD_OUT_PORT The value format depends on the match type.  |
+|     priority | Integer          | The priority of the ACL rule relative to other rules in the ACL table.                                                                                                                                                                                                                                                                                                                           |
+
+
+```
+127.0.0.1:6379> keys P4RT:ACL_ACL_INGRESS*
+ 1) "P4RT:ACL_ACL_INGRESS_TABLE:{\"match/dst_mac\":\"33:33:00:00:00:02&ff:ff:ff:ff:ff:ff\",\"match/icmpv6_type\":\"0x87&0xff\",\"match/ip_protocol\":\"0x3a&0xff\",\"match/is_ipv6\":\"0x1\",\"priority\":2070}"
```
venkatmahalingam, line 212: We can be aligned with the existing ACL_TABLE/ACL_RULE tables with unique key names, this would help for easy migration in the future with legacy tables.

P4RT_App_DB_Schema.md
```diff
@@ -0,0 +1,396 @@
...
+; defines the P4RT DEFINITION table used for defining a dynamic table.
+
+| key                     = | P4RT:DEFINITION:table_name | ; Dynamic table name |
+|---------------------------|----------------------------|----------------------|
+
+
+
+```
+127.0.0.1:6379> keys P4RT*
+1) "P4RT:DEFINITION:ACL_ACL_PRE_INGRESS_TABLE"
```
venkatmahalingam, line 133: Did we consider separate DB instance e.g P4RT_DB for P4 table definitions? this would help to avoid any accidental flush of the P4RT configs in APPL_DB during warm restart, swss restart..etc.

P4RT_App_DB_Schema.md
```diff
@@ -0,0 +1,396 @@
...
+
+## Schema definition
+
+
+### APP_DB
+
+
+#### P4RT Table
+
+All data in the Application Database (APPL DB) published by the P4RT application resides in  the P4RT table. Logically, the P4RT table is divided into **subtables**:
```
venkatmahalingam, line 110: If we're planning to use generic SAI programmable objects (no P4 name in the objects) at the SAI layer, can we also use generic table names here? instead of P4RT prefix?

P4RT_App_DB_Schema.md
bhagatyj: The existing orchagent code does not meet some of the requirements of the remote controller path such as response path, application ownership tags etc. Multiple writers(local and remote applications) to the same table without tags would cause additional issues. The new path can be merged to the existing path after these functionalities are built in the existing path. The north star is a single path in the future which meets the requirements for local and remote paths.

P4RT_App_DB_Schema.md
bhagatyj: P4RT tables share very similar requirements during warm restart, swss restart etc. In some cases, local applications require more time to learn the new conditions following a restart. However, the remote controller is not affected by a local restart and can immediately push its intent as soon as the local restart is complete. There is more similarity than differences and the longer term approach is merging with the local tables.

P4RT_App_DB_Schema.md
bhagatyj: In the MVP version, the P4 orchagent processes the P4 tables. In the future as the main path builds response, ownership capabilities, the P4 tables will merge into the local tables.

P4RT_App_DB_Schema.md
kishanps: We chose json format so that it allows a wider choice of chars and also so that it doesnt conflict with existing chars in the key like ':','"' etc.

venkatmahalingam: I understand but this is definitely a deviation w.r.t APP_DB definition, it's better to be aligned with current APPL_DB tables so that we can easily migrate to common tables for SONiC regular and P4RT flows in the future.

kishanps: Sure, we will definitely look at moving these definitions closer to the regular SONiC table format as we get near to the unified tables approach.

zhangyanzhao:
@kishanps can you please sign the EasyCLA which is required to merge this PR? Thanks.

prsunny (approved):

PR #846, opened 2021-08-18
Initial version of APPL STATE DB & Response Path HLD

mint570:


doc/pins/appl_state_db_response_path_hld.md
```diff
@@ -0,0 +1,271 @@
...
+   </td>
+   <td>SAI_STATUS_INSUFFICIENT_RESOURCES
+<p>
+SAI_STATUS_TABLE_FULL
+   </td>
+  </tr>
+  <tr>
+   <td>SWSS_RC_IN_USE
+   </td>
+   <td>SAI_STATUS_OBJECT_IN_USE
```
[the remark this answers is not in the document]
StephenWangGoogle: To be more specific, the P4 programming is synchronous, that's what Runming is describing the behavior of P4Orch using the response path. For existing orch that does retry, and if the applications want to know an request is being retry or not, we can extend the design to add a "status" attribute (pending/done...etc). But that is currently not in the scope of the initial MVP

mint570:
@lguohan 
PTAL
Thanks

doc/pins/appl_state_db_response_path_hld.md
```diff
@@ -0,0 +1,218 @@
...
+
+![drawing](images/appl_state_db_response_path_hld_example_batch.png)
+
+## SAI API
+
+N/A
+
+## Configuration and management
+
+A new configuration file will be added to enable the response channel and the APPL STATE DB for the selected APPL DB tables. There are two configurations: response channel and APPL STATE DB. They are defined in separate section of the configuration file. The configuration will contain a list of APPL DB table names. Only the tables in the list will have the feature enabled.
```
ravi861, line 181: Can you please list the location of this new configuration file?

mint570: @donNewtonAlpha

We did not implement this approach.
We used a different approach to configure it in the request entry.
I have updated this section.

ravi861: Thanks for the update. It looks like some events in P4RT depend on updates from PORT_TABLE in APPL_STATE_DB. How will they now work?
A better question is how will non P4RT created objects get added to APPL_STATE_DB?

mint570: In the MVP release, the response path support only applies to P4Orch and VrfOrch.
Non-p4rt objects are not in APPL STATE DB yet. They will need code change in each orch.

P4RT will wait for all ports ready before it starts.

zhangyanzhao:
@mint570  can you please sign the EasyCLA which is required to merge your code?  Thanks.

zhangyanzhao:
@mint570  can you please sign the EasyCLA which is required to merge this PR? Thanks.

mint570:
My account has signed EasyCLA already. Just need to rebase the PR.

PR #853, opened 2021-09-01
[HLD] [SFlow] local config precedence rule updated

vivekreddynv:
This rule is not documented in the HLD.

Fix: 
Added this rule to avoid confusion


dgsudharsan (approved):

dgsudharsan:
@prsunny  Can you please review.

prsunny (approved):

PR #854, opened 2021-09-04
Update fgnhg CLI format

kktheballer:
Updating fgnhg CLI format as per CLI format changes in sonic-utilities.

Note: the file appears to have been completely deleted and erased but this was not the case.
I used diffchecker to make sure only the intended changes are there


liat-grozovik:
@kktheballer looks like the entire file changed. Could you please check so i can review the CLI changes as this is the main reason for the HLD update?

anish-n (approved):

PR #858, opened 2021-09-07
[flow counters] High level design for trap flow counters support

Junchao-Mellanox:
High level design for flow counters support.

Flow counters are usually used for debugging, troubleshooting and performance enhancement processes. Flow counters could cover cases like:

- Host interface traps (number of received traps per Trap ID)
- Routes matching the configured prefix pattern (number of hits and number of bytes)
- FDB entries matching the configured VXLAN tunnel or using the VLAN ID as pattern
- Next-Hop/Next-Hop Group/Next-Hop Group Member

This document focus on host interface traps counter.

Related PRs:

| [[sonic-buildimage] Add trap flow counter support](https://github.com/Azure/sonic-buildimage/pull/8940) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-swss] [orchagent] Add trap flow counter support](https://github.com/Azure/sonic-swss/pull/1951) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-mgmt] Add test case for trap flow counter feature](https://github.com/Azure/sonic-mgmt/pull/4456) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-utilities] Add trap flow counter support](https://github.com/Azure/sonic-utilities/pull/1868) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-sairedis] Add trap flow counter support](https://github.com/Azure/sonic-sairedis/pull/954) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-swss-common] Add trap flow counter support](https://github.com/Azure/sonic-swss-common/pull/534) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-utilities] [Command Reference] Add command reference for trap flow counters](https://github.com/Azure/sonic-utilities/pull/1876) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-buildimage] [YANG] Add trap flow counter to yang model](https://github.com/Azure/sonic-buildimage/pull/9353) | GitHub issue/pull request detail | GitHub pull request check contexts |


doc/flow_counters/flow_counters.md
```diff
@@ -0,0 +1,767 @@
...
+### Requirements
+
+- Generic Counters shall be used as Flow Counters introduced by the feature
+- CLI shall be used for configuration, showing and clearing of statistics
+- Statistics of traps shall be enabled and disabled using the CLI commands (counterpoll)
+- Polling interval shall be configured using the CLI commands (counterpoll)
+- Statistics shall be enabled or disabled for all configured traps (not per-trap type)
+- Statistics for traps shall be provided as a number of messages delivered from ASIC to host over the host interface per trap type
+- Statistics shall be available for clearing (for all traps together) using the CLI command
+- Statistic shall support number of trap packets/bytes and number of trap PPS
```
lguohan, line 61: in the community meeting, zhenghui raise a question that we should add drop counter for the trap since trap policer is mainly used for dos scenario. In this case, it is important to know what packets are dropped. in order to count the drop counters, we need to use the sai policer_state to get drop packets/bytes. https://github.com/opencomputeproject/SAI/blob/master/inc/saipolicer.h#L219

caizhenghui-juniper: In this case, flow counter is somewhat duplicated with policer stats, we probably want to use the existing policer objects even for received packets/bytes?.

lguohan: agree, instead of attaching counter to the policer, we can just attach policer_stat then we can get dropped and non-dropped counters.

Junchao-Mellanox: Hi @lguohan and @caizhenghui-juniper , thanks for the comments. 

The policer object is usually bound to a trap group, and a trap group usually contains multiple traps. For example, trap group "queue4_group1" could have trap like "bgp,bgpv6,lacp". Based on that, _sai_policer_stat_t can only provide the stats for a group like "queue4_group1", and there is no detail stat for trap like "bgp,bgpv6,lacp".  
I would say that flow counter is not a duplicated with policer stats because it provides more detailed information, and it can be used to debug the policer if it does not work as expect. 

In addition, for DDOS case, trap flow counter is still useful, as you can see that there are too many packet traps to CPU via command `show flowcnt-trap stats`, and user can get the exact trap type easily, and user might realize that he/she forgot to set the policer for the trap group. 

So, my view is that "trap flow counter" is the counter that provides a per trap basis statistic; policer stat is the counter that provides a per policer basis statistic, and policer now can only be bound to a trap group. Policer stat is not the scope of this feature and can be covered in future feature.

Junchao-Mellanox: Hi @lguohan and @caizhenghui-juniper ,

Policer object is usually bound to a trap group, and a trap group usually contains multiple traps. For example, trap group "queue4_group1" could have trap like "bgp,bgpv6,lacp". Based on that, sai_policer_stats can only provide the stats for a group like "queue4_group1", and there is no detail stat for trap like "bgp,bgpv6,lacp". 
I would say that flow counter is not a duplicated with policer stats because it provides different use case and more detailed information, and it can be used to debug the policer if it does not work as expect. 

For example, for DDOS case, trap flow counter is still useful, as user can see that there are too many packet traps to CPU via command `show flowcnt-trap stats`, and user can get the exact trap type easily, and user might realize that he/she forgot to set the policer for the trap group.

So, my view is that "trap flow counter" is the counter that provides a per trap basis statistic; policer stat is the counter that provides a per policer basis statistic, and policer now can only be bound to a trap group. These are different use-cases and these counters (per trap or per policer object) provide different granularity. Policer stats should be supported in another feature.

keboliu: @caizhenghui-juniper @lguohan would you please have a look at Junchao's feedback?

Junchao-Mellanox: @caizhenghui-juniper Thanks for your comment. For the CoPP test case change, you can check PR https://github.com/Azure/sonic-mgmt/pull/4456. For drop counter for a trap type, I don't see a SAI attribute support it, I will discuss with Arch and get back to you.

Junchao-Mellanox: Hi @caizhenghui-juniper and @lguohan , I agree that we should have drop counter per trap to enable better debug capability. As you may know, the existing generic counter framework does not provide such an SAI attribute, and it is probably not suitable to add such attribute to generic counter. We are working on a new SAI attribute to cover that part, and it will be a new feature (need SAI/SDK/SONiC design and development).

liat-grozovik:
@prsunny could you please help to review as well?
Code changes are already under review and it is better to have the HLD approved a head of the review completion.

prsunny (approved):

prsunny:
@caizhenghui-juniper , could you please sign-off?

caizhenghui-juniper (approved):
Please add drop counter capability.

PR #860, opened 2021-09-08
Add proprietary code guidelines

TACappleman:


zhangyanzhao:
Are you trying to add "proprietary code" to SONiC or just referring to proprietary code in SONiC?

TACappleman:
@zhangyanzhao The second - this would be things like build support and changes to show commands that would work with the proprietary coe

PR #866, opened 2021-09-15
correct the link for doc and image

richardyu-ms:
correct the link for doc and image in system-warmboot.md


qiluo-msft:
The png does still not display in the markdown. Is it expected?

richardyu-ms:
move files to the local repo and changed file links.

qiluo-msft (approved):

PR #873, opened 2021-09-29
Updating new files to create sonic latest build images link

kannankvs:
Updating new files to create sonic latest build images link


kannankvs:
Note that the alert coming from "lgtm-com" (bot) w.r.t. "11 for Missing variable declaration" is not applicable for the HTML file. This HTML file has been tested and it works fine. Hence, the alerts can be ignored.

PR #874, opened 2021-09-30
Password Hardening HLD

davidpil2002:


liuh-80 (commented):
I have following concern and questions:
1. What's the responsbility of PASSWH? If this daemon only for read hardening setting from config DB and update PAM config file, then there already hostcfgd for this, not necessary add a new daemon. And if this daemon also will handle the password history to config DB storage, I think there will be security risk.

2. According to the design, password history will store in DB, if that means config DB, then because currently there is no ACL for config DB, so password history can be access by anyone, this will be a security risk, but if the password history stored by pam_pwhistory it self, It think it's OK.

davidpil2002:
Hi,
In continue of your concerns and questions, 

regarding question 1:
The feature should run in the host, we had a discussion if is better to integrate the feature to the current hostcfg daemon or to create a daemon to manage this feature only. For now, we think it's better to have it separate, that because the feature can be enabled by a compilation flag, meaning that the feature is not always exposed to the user, but we can discuss about it in the meeting. (in addition, the feature will use the STATE_DB as well).

Regarding concern 2, the "old passwords" (passwords history), the meaning of saving the password history is to save only have many old passwords the feature can save, not to save the passwords itself. The passwords will be saved like you commented by pw_history in /etc/security/opasswd or similar file with just root permission access.
Thanks
David

doc/passw_hardening/hld_password_hardening.md
```diff
@@ -0,0 +1,815 @@
...
+| :---: | :-----: | :--------------: | ------------------ |
+|  0.1  | 11/2020 |   David Pilnik   | Phase 1 Design     |
+
+###  1.2. <a name='Scope'></a>Scope
+
+This password hardening hld doc described the requirements, architecture and configuration details of password hardening feature in switches Sonic OS based.
+
+###  1.3. <a name='DefinitionsAbbreviations'></a>Definitions/Abbreviations
+	PW - password
+	SPASSWD - secure password daemon
```
liuh-80, line 57: Please remove this line, since SPASSWD been removed.

doc/passw_hardening/hld_password_hardening.md
```diff
@@ -0,0 +1,815 @@
...
+| PW constraints | The PW class represents the possible characters comprising the PW. There are 4 character classes that can make up the PW:small characters, big characters, numbers and special chars presented on a standard keyboard <br /><br />The account name and PW need to be different <br /><br />The new PW is to be different from current PW|The user should be able to define which characters classes are required for the PW out of the 4 provided<br /><br /><br /><br />The user will be able to enable or disable this feature|Legal special characters `~!@#$%^&*()-_=+\|[{}];:',<.>/? and white space<br /><br /><br /><br />An error will be shown to the user if it tries to set the same PW as the username<br /><br /><br /><br />default - enable, default remember last 10 password
+
+###  1.6. <a name='ArchitectureDesign'></a>Architecture Design
+Arc design diagram\
+![passh_arc_sonic](passh_arc_sonic.jpg)
```
liuh-80, line 74: This diagram need update, because how the 'password daemon' change to hostcfgd.

doc/passw_hardening/hld_password_hardening.md
```diff
@@ -0,0 +1,815 @@
...
+	* 1.5. [Requirements](#Requirements)
+	* 1.6. [Architecture Design](#ArchitectureDesign)
+	* 1.7. [High-Level Design](#High-LevelDesign)
+		* 1.7.1. [Flow description:](#Flowdescription:)
+		* 1.7.2. [Password Hardening Constrains](#PasswordHardeningConstrains)
+	* 1.8. [Init Flow](#InitFlow)
+		* 1.8.1. [Compilation](#Compilation)
+		* 1.8.2. [dependencies](#dependencies)
+		* 1.8.3. [Feature default](#Featuredefault)
+		* 1.8.4. [How the daemon works internally:](#Howthedaemonworksinternally:)
```
liuh-80, line 20: Section 1.8.4 been removed.

liuh-80 (dismissed):
Found some minor issue in latest document.

liuh-80 (approved):

doc/passw_hardening/sonic-passwh.yang
```diff
@@ -0,0 +1,91 @@
...
+    }
+    typedef feature_state {
+        type enumeration {
+			enum enabled;
+			enum disabled;
+		}
+	}
+
+    container sonic-passwh {
+		container PASSWH {
```
venkatmahalingam, line 20: Suggest to have a table name "PASSWD"

doc/passw_hardening/sonic-passwh.yang
```diff
@@ -0,0 +1,91 @@
...
+			description "PASSWORD HARDENING part of config_db.json";
+			container POLICIES {
```
venkatmahalingam, line 22: Not sure why do we need nested containers... we can have the table name PASSWD_POLICY for the configurations.

doc/passw_hardening/sonic-passwh.yang
```diff
@@ -0,0 +1,91 @@
...
+				leaf state {
+					description "state of the feature";
```
venkatmahalingam, line 24: Not sure why do we need this field, can we have this entry itself only if we desire the password hardening?

doc/passw_hardening/sonic-passwh.yang
```diff
@@ -0,0 +1,91 @@
...
+					}
+				}
+				leaf expiration_warning {
+					description "expiration warning time (days unit)";
+					default 15;
+					type uint8 {
+						range 1..30;
+					}
+				}
+				leaf history {
```
venkatmahalingam, line 42: history_cnt?

doc/passw_hardening/sonic-passwh.yang
davidpil2002: PASSWH, meaning "password hardening"

doc/passw_hardening/sonic-passwh.yang
davidpil2002: state - is for enabling or disabling the feature.
basically is meaning that the admin user can choose if he wants to use the password hardening feature or not.

doc/passw_hardening/sonic-passwh.yang
davidpil2002: thanks, I will update this one

davidpil2002:
Hi @liuh-80,

How are you doing?
I answered the questions above.
In addition, I had an HLD review with the community and added the changes suggested in the commit: cbe88dd
can you approve this HLD, or pls let me know if you have any more comments.

Thanks,
David

doc/passw_hardening/sonic-passwh.yang
```diff
@@ -0,0 +1,91 @@
...
+    namespace "http://github.com/Azure/sonic-passwh";
+	prefix passwh;
+
+    description "PASSWORD HARDENING YANG Module for SONiC OS";
+
+	revision 2021-10-12 {
+        description
+            "First Revision";
+    }
+    typedef feature_state {
```
liuh-80, line 12: Please follow sonic yang model guideline, every definition should inside top level container, in this case, should inside sonic-passwh

liuh-80 (approved):

davidpil2002:
@liuh-80 @liat-grozovik 
can you help to merge to master?

liat-grozovik:
@venkatmahalingam can you please approve from your side? you had comments and i wish to merge once we agree they were handled.

doc/passw_hardening/sonic-passwh.yang
venkatmahalingam: Not sure when we'll configure this table with state=disabled....we could very well not have this entry itself..right?

davidpil2002:
> 
Hi @venkatmahalingam,

this is a kind reminder to reply to the Liat Q in the comment above.
Thanks,

venkatmahalingam (approved):

aidan-gallagher:
Hi @davidpil2002 @venkatmahalingam @liuh-80 

Are you aware that when the operator uses `sudo` to configure the password then it bypasses the hardening checks?

For example I can't set a 1 letter password using `passwd` however I can do it using `sudo passwd`. Is this the expected behaviour?

PR #875, opened 2021-10-01
SYSTEM READY

sg893052:
Support for System Ready

<html>
<body>

Repo | PR title | State
-- | -- | --
sonic-buildimage | <a href="https://github.com/Azure/sonic-buildimage/pull/8889" data-hovercard-type="pull_request" data-hovercard-url="/Azure/sonic-buildimage/pull/8889/hovercard">System Ready</a>  | 
sonic-utilities | <a href="https://github.com/Azure/sonic-utilities/pull/1851" data-hovercard-type="pull_request" data-hovercard-url="/Azure/sonic-utilities/pull/1851/hovercard">Show commands for System Ready</a> | 

</body>
</html>

### DIFF
--- a/doc/system-ready/system-ready-HLD.md
+++ b/doc/system-ready/system-ready-HLD.md
@@ -1,0 +1,546 @@
+
+
+# System Ready HLD
+
+#### Rev 0.2
+
+# Table of Contents
+
+- [List of Tables](#list-of-tables)
+- [Revision](#revision)
+- [Definition/Abbreviation](#definitionabbreviation)
+- [About This Manual](#about-this-manual)
+- [1 Introduction and Scope](#1-introduction-and-scope)
+  - [1.1 Limitation of Existing tools](#11-existingtools-limitation)
+  - [1.2 Benefits of this feature](#12-benefits-of-this-feature)
+- [2 Feature Requirements](#2-feature-requirements)
+  - [2.1 Functional Requirements](#21-functional-requirements)
+  - [2.2 Configuration and Management Requirements](#22-configuration-and-management-requirements)
+  - [2.3 Scalability Requirements](#23-scalability-requirements)
+  - [2.4 Warm Boot Requirements](#24-warm-boot-requirements)
+- [3 Feature Description](#3-feature-description)
+- [4 Feature Design](#4-feature-design)
+  - [4.1 Overview](#41-design-overview)
+  - [4.2 Sysmonitor](#42-db-changes)
+    - [4.2.1 Subtasks in Sysmonitor](#421-subtasks-in-sysmonitor)
+  - [4.3 Service Identification and Categorization](#43-service-categorization)
+  - [4.4 System ready Framework business logic](#44-systemready-fremework-logic)
+  - [4.5 Provision for apps to mark closest UP status](#45-provision-for-apps-to-mark-UP)
+    - [4.5.1 CONFIG_DB Changes](#451-config-db-changes)
+    - [4.5.2 STATE_DB Changes](#452-state-db-changes)
+    - [4.5.3 Feature yang Changes](#453-feature-yang-changes)
+  - [4.6 Syslogs](#46-syslogs)
+- [5 CLI](#5-cli)
+  - [5.1 Output Format](#51-cli-output-format)
+  - [5.2 show system status core](#52-system-status-core)
+  - [5.3 show system status all](#53-system-status-all)
+  - [5.4 show system status all --brief](#54-system-status-all-brief)
+  - [5.5 show system status all --detail](#55-system-status-all-detail)
+- [6 Serviceability and Debug](#6-serviceability-and-debug)
+- [7 Warm reboot Support](#7-warm-reboot-support)
+- [8 Unit Test Cases ](#8-unit-test-cases)
+- [9 References ](#9-references)
+
+# List of Tables
+
+[Table 1: Abbreviations](#table-1-abbreviations)
+
+# Revision
+| Rev  |    Date    |       Author        | Change Description                                           |
+|:--:|:--------:|:-----------------:|:------------------------------------------------------------:|
+| 0.1  |  | Senthil Kumar Guruswamy   | Initial version                                              |
+| 0.2  |  | Senthil Kumar Guruswamy   | Update as per review comments                                              |
+
+
+# Definition/Abbreviation
+
+### Table 1: Abbreviations
+
+| **Term** | **Meaning**                               |
+| -------- | ----------------------------------------- |
+| FEATURE  | Docker/Service                            |
+| App      | Docker/Service                            |
+
+
+# About this Manual
+
+This document provides general information about the System Ready feature implementation in SONiC. 
+
+
+# 1 Introduction and Scope
+
+This document describes the Functionality and High level design of the System Ready feature.
+
+At present, there is no mechanism to know that the system is up with all the essential sonic services and also, all the docker apps are ready along with port ready status to start the network traffic.
+With the asynchronous architecture of SONiC, we will not be able to verify if the config has been applied all the way down to the hardware. 
+However, if we could get the closest up status of each docker app considering their config receive ready and port ready, the system readiness could be arrived.
+
+A new python based System monitor framework is introduced to monitor all the essential system host services including docker wrapper services on an event based model and declare the system is ready.
+This framework gives provision for docker and host apps to notify its closest up status.
+CLIs are provided to fetch the current system status and also service running status and its app ready status along with failure reason if any.
+
+
+## 1.1 Limitation of Existing tools:
+ - Monit tool is a poll based approach which monitors the configured services for every 1 minute.
+ - Monit tools feature of critical process monitoring was deprecated as supervisor does the job. Hence system-health tool which depends on monit does not work.
+ - Container_checker in monit checks only for running status of expected containers.
+ - Monits custom script execution can only run a logic to take some action but it is yet again a poll based approach.
+
+
+## 1.2 Benefits of this feature:
+ - Event based model - which does not hog cpu unlike poll based approach.
+ - Know the overall system status through syslog and as well through CLIs
+ - CLIs just connect to the backend daemon to fetch the details. No redundant codes.
+ - Combatibility with application extension framework.
+    SONiC package installation process will register new feature in CONFIG DB.
+    Third party dockers(signature verified) gets integrated into sonic os and runs similar to the existing dockers accessing db etc.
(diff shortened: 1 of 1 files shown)


sg893052:
Code PR: 
https://github.com/Azure/sonic-buildimage/pull/8889
https://github.com/Azure/sonic-utilities/pull/1851

doc/system-ready/system-ready-HLD.md
```diff
@@ -0,0 +1,482 @@
...
+
+            }
+           },
+
+- Example:
+  "FEATURE|bgp": {
+    "type": "hash",
+    "value": {
+      "FAIL_REASON": "",
+      "TIME": "20211005 04:19:30",
```
venkatmahalingam, line 264: Can we use consistent lower case field names like other fields we have today?

doc/system-ready/system-ready-HLD.md
```diff
@@ -0,0 +1,482 @@
...
+
+
+
+In addition to this, sysmonitor posts the core system status to STATE_DB as below.
+
+```
+  "SYSTEM_READY|SYSTEM_STATE": {
+    "type": "hash",
+    "value": {
+      "Status": "UP"
```
venkatmahalingam, line 280: Can we use "status"?

doc/system-ready/system-ready-HLD.md
```diff
@@ -0,0 +1,482 @@
...
+```
+- Schema in /etc/sonic/init_cfg.json
+  This json file will be fed to FEATURE table of CONFIG_DB during bootup.
+  This json file will be factory default and no config command will be provided for "check_up_status" entry to be updated in CONFIG_DB later.
+           {
+              "FEATURE": {
+                 "<dockername>": {
+                     ...
+                     "state": "enabled",
+                     "check_up_status": "True"
```
dgsudharsan, line 221: Feature table is for docker based features. Can we introduce a new table for host services like caclmgrd, hostcfgd?

doc/system-ready/system-ready-HLD.md
```diff
@@ -0,0 +1,482 @@
...
+
+
+STATE_DB:
+```
+- sonic-db-cli STATE_DB HSET "FEATURE|<dockername>" UP_STATUS True
+- sonic-db-cli STATE_DB HSET "FEATURE|<dockername>" FAIL_REASON "<some reason in string format>" / ""
+- sonic-db-cli STATE_DB HSET "FEATURE|<dockername>" TIME "<timestamp in <datetime.now().strftime('%Y%m%d %H:%M:%S')> format >"
+
+- Schema in STATE_DB
+  sonic-db-dump -n STATE_DB output
```
dgsudharsan, line 248: The host services don't belong to FEATURE table and we need a different table for host services like hostcfgd

dgsudharsan (changes requested):

doc/system-ready/system-ready-HLD.md
dgsudharsan, line 221: Please also update the feature yang on adding the new field.

dgsudharsan (changes requested):

doc/system-ready/system-ready-HLD.md
sg893052: Agreed. Will update as per suggestion.

doc/system-ready/system-ready-HLD.md
```diff
@@ -0,0 +1,482 @@
...
+
+## 1.2 Benefits of this feature:
+ - Event based model - which does not hog cpu unlike poll based approach.
+ - Know the overall system status through syslog and as well through CLIs
+ - CLIs just connect to the backend daemon to fetch the details. No redundant codes.
+ - Combatibility with application extension framework.
+    SONiC package installation process will register new feature in CONFIG DB.
+    Third party dockers(signature verified) gets integrated into sonic os and runs similar to the existing dockers accessing db etc.
+    Now, once the feature is enabled, it becomes part of either sonic.target or multi-user.target and when it starts, it automatically comes under the system monitor framework watchlist.
+    However, app ready status for those dockers cant be tracked unless they comply with the framework logic.
```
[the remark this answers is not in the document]
sg893052: Sure. Will add the following for application extension: 
App ready status for those dockers cant be tracked unless they comply with the framework logic. Hence any third party docker needs to follow the framework logic by including "check_up_status" field while registering itself in CONFIG_DB and also make use of the provision given to docker apps to mark its closest up status in STATE_DB.

doc/system-ready/system-ready-HLD.md
sg893052: Agreed, Will introduce a new table in CONFIG_DB called HOST_FEATURE for this purpose.

doc/system-ready/system-ready-HLD.md
sg893052: Agreed.

doc/system-ready/system-ready-HLD.md
dgsudharsan: https://github.com/Azure/sonic-buildimage/pull/7955

sg893052:
> How does it monitor hardware status? Like fan/PSU/ASIC

System ready framework monitors all the sonic services(docker+host) along with its app readiness to declare the system is ready for network traffic. It does not monitor hardware status.

venkatmahalingam (approved):

sg893052:
> > Support for System Ready
> > Repo	PR title	State
> > sonic-buildimage	[System Ready](https://github.com/Azure/sonic-buildimage/pull/8889)	 Open
> > sonic-utilities	[Show commands for System Ready](https://github.com/Azure/sonic-utilities/pull/1851)	 Open
> 
> Can you update this table with template #806 which display the status with better view? Thanks

Have updated the table as suggested.

lguohan:
this pr is reverted, need sign off from @dgsudharsan

dgsudharsan:
Looks good to me. @Junchao-Mellanox Can you please review and signoff?

Junchao-Mellanox:
I don't have comment to the design itself. There are many good point in the PR (like using the event driven way to monit service status), but I am not sure that we need another service to do similar job like system-health. See what we have:

1. We have container_checker runs on top of monit which monitors the containers status
2. We have supervisord runs on each container which monitors the critical processes status
3. We also have system-health which monitors containers/critical processes/hardware status. The function of system-health was broken by other PR, but there is a PR to fix it (https://github.com/Azure/sonic-buildimage/pull/9068)

So, my opinion is like this: how about we extend system health instead of creating another new service?

liat-grozovik:
@zhangyanzhao i don't understand why this PR has been merged while there is outstanding feedback.
the feature originally was contributed by Nvidia and we have provided valuable feedback to this design.
@sg893052 please note that the HLD is now reverted. Please refer to the comments provided and we will be happy to discuss offline if this is needed.

sg893052:
> I don't have comment to the design itself. There are many good point in the PR (like using the event driven way to monit service status), but I am not sure that we need another service to do similar job like system-health. See what we have:
> 
> 1. We have container_checker runs on top of monit which monitors the containers status
> 2. We have supervisord runs on each container which monitors the critical processes status
> 3. We also have system-health which monitors containers/critical processes/hardware status. The function of system-health was broken by other PR, but there is a PR to fix it ([[system-health] No longer check critical process/service status via monit sonic-buildimage#9068](https://github.com/Azure/sonic-buildimage/pull/9068))
> 
> So, my opinion is like this: how about we extend system health instead of creating another new service?


SONiC runtime environment is made up of systemd services, dockers and processes. A systemd service encompasses docker or processes running on host, and thus monitoring the status of the services provide a snapshot of the overall system health. Also a real time feedback is thus helpful to know when a particular service / docker or a critical process went down. Thus System ready feature emphasises the following design philosophy:

1. Service Monitoring over Process Monitoring
2. Event driven monitoring over Polling for real time feedback

In SONIC, we have service restart policies based on critical processes. So it is essential that we need to monitor all the services. We already have supervisord monitoring the critical processes within a container. Supervisord's proc-exit-listener basically flags which critical process exited and then because of which when the service restarts, our system ready feature will catch that the particular service is down instantly. When it comes up, it will flag that as up without any delay. Service monitoring will help us know that some critical processes died, and we will get a notification that one of the services went down. This also helps with container monitoring indirectly as docker wait is a typical service criteria. Once docker goes down, service goes down. The existing container_checker in monit is poll based and does not provide real time feedback. So, we really don't need anything from outside to monitor critical processes of dockers once we have the service monitoring in place. 

So System ready is all on top of the existing infrastructure and with that point#3 mentioned by Junchao-Mellanox for critical processes monitoring is not required. I understand that the new fix(Azure/sonic-buildimage#9068) is to only show the status of the critical process through docker exec and what if the docker dies? And since everything(monit/container_checker/newone in 9068) is polling based and it is polled every minute, it is not realtime and so likely to miss critical service restarts. 


Benefits of system ready:
System ready is event driven where the feedback is immediate.
Since system ready monitors systemd services, docker monitoring is achieved indirectly. No need for polling based container_checker utility in monit.
It monitors all the sonic host systemd services.
In addition, System ready feature also brings in the concept of application readiness to allow each application/service/docker to declare themselves as ready based on different application specific criteria.


Agree, it makes sense to put the entire System ready feature implementation into System health monitor code, and also leverage the system ready feature to check for service readiness inside the system health monitor. This can definitely be considered for the next release, as we don’t have much time for 202111 release.

Junchao-Mellanox:
Hi @sg893052 , like I said, this design brings some good points, I like that part especially for the event driven idea. But it is still something similar to system-health ([Azure/sonic-buildimage#9068](https://github.com/Azure/sonic-buildimage/pull/9068) also support monitoring services). I am not sure that adding a new service "SYSTEM READY" to 202111 and then merge it to system health in future is a good idea. We cannot handle the backward compatible for user.

Junchao-Mellanox:
If we want to merge this feature to system health, some thought here:
1. The subprocess to listen service status can be easily integrated to system health, and it is still a subprocess. I would call it "LS" process
2. There is a queue between main process and "LS" process, "LS" process notify service change event via the queue. This is the same as what "SYSTEM READY" is doing.
3. Currently, there is a class ServiceChecker to monitor services/critical processes status. It shall no longer monitor services because "LS" process will do it. But it shall still monitor the critical processes as we don't want any "function degradation".
4. The concept of service readiness can be also integrated into system health, the integration would not be difficult as it is also a subprocess in current design.
5. System health CLI can be extended if there is more information provided to user.

zhangyanzhao:
> @zhangyanzhao i don't understand why this PR has been merged while there is outstanding feedback. the feature originally was contributed by Nvidia and we have provided valuable feedback to this design. @sg893052 please note that the HLD is now reverted. Please refer to the comments provided and we will be happy to discuss offline if this is needed.

I am ok to revert my merge. Please go ahead

PR #877, opened 2021-10-05
Static route subscribe path

zhaozhenhong:


zhaozhenhong:
Added by mistake, should be deleted

PR #884, opened 2021-10-20
Update sonic_image_links.json

kannankvs:
Updated json file for sonic build image


kannankvs:
As per the new solution, it is planned to keep the json file in a separate branch. Hence, this PR is not required. Closing it.

PR #885, opened 2021-10-20
Update sonic_image_links_create_workflow.yml

kannankvs:
Updated workflow file sonic_image_links_create_workflow.yml  for permission as suggested.

### DIFF
--- a/.github/workflows/sonic_image_links_create_workflow.yml
+++ b/.github/workflows/sonic_image_links_create_workflow.yml
@@ -6,6 +6,9 @@ name: Run the shell script to generate sonic image links in sonic_image_links.js
 on:
   schedule:
     - cron:  '* * * * *'
+    
+  permissions:
+    actions: read|write
 
 jobs:
   build:


.github/workflows/sonic_image_links_create_workflow.yml
```diff
@@ -6,6 +6,9 @@ name: Run the shell script to generate sonic image links in sonic_image_links.js
 on:
   schedule:
     - cron:  '* * * * *'
+    
+  permissions:
+    actions: read|write
```
xumia, line 11: The permissions should be aligned with on and jobs, see https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions
Need to change:
```
  permissions:
    actions: read|write
```
TO:
```
permissions:
  actions: read|write
```

xumia: And the permissions are granted to GITHUB_TOKEN, I guess we need to use the token in the script to commit the change.

xumia: You can test in your repo to use the GITHUB_TOKEN.

PR #896, opened 2021-11-08
Order ECMP HLD.

abdosi:
HLD for Order ECMP Feature. Nexthop in ECMP will be sorted/order and the order will be preserved when programming into ASIC via SAI. 

Signed-off-by: Abhishek Dosi <abdosi@microsoft.com>


prsunny (approved):

zhangyanzhao:
@abdosi would you please add the code PRs as what we asked?

abdosi:
> @abdosi would you please add the code PRs as what we asked?

@zhangyanzhao Updated.

PR #903, opened 2021-11-19
Copp manager redesign test plan

JibinBao:
Add test plan for CoPP Manager redesign feature 

1. Copp Manager redesign HLD:
2. Implementation PRs:  https://github.com/Azure/sonic-mgmt/pull/4891

### DIFF
--- a/doc/copp/Copp_Manager_Redesign_test_plan.md
+++ b/doc/copp/Copp_Manager_Redesign_test_plan.md
@@ -4,7 +4,7 @@
 
 | **Document Name** | **Link** |
 |-------------------|----------|
-| CoPP Manager Redesign HLD | [https://github.com/noaOrMlnx/sonic-swss/blob/4865a8eb37444c2c070ef49cc69690e4295ce07d/doc/copp_redisgn.md](https://github.com/noaOrMlnx/sonic-swss/blob/4865a8eb37444c2c070ef49cc69690e4295ce07d/doc/copp_redisgn.md)|
+| CoPP Manager Redesign HLD | []|
 
 
 ## Overview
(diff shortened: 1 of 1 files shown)


liat-grozovik:
@yxieca and @prsunny could you please assign someone to review?

liat-grozovik:
@dgsudharsan and @prsunny  could you please approve this change?

doc/copp/Copp_Manager_Redesign_test_plan.md
```diff
@@ -0,0 +1,134 @@
...
+1. Set always_enabled of bgp to false (refer to test #2)
+2. Enable bgp in feature table: config feature state bgp enabled    
+3. Verify the tested trap is installed by sending traffic
+4. Disable bgp in feature table: config feature state bgp disabled 
+5. Verify the tested trap is uninstalled by sending traffic
+6. Recover the config
+
+
+   
+### Test cases #5 - Verify trap configuration is saved or not after reboot(reboot, fast-reboot, warm-reboot)
```
prsunny, line 126: I think, one more case to verify is the sflow example where the entry is present in copp.json, but not in feature table. It should not be installed until the feature is enabled

dgsudharsan: @prsunny I believe Test case 4 should cover the scenario and our copp swss UT handles it. The reason we cannot test sflow in sonic mgmt is if we disable the feature, sampling itself will be disabled and so we cannot assume that packets are not lifted because of trap removal or sampling itself is disabled. That's why we use BGP to test this usecase.

prsunny: I meant sflow as an example. Just want to make sure the sequence is covered. But as you mentioned, Test 4 seems to cover

prsunny (approved):

doc/copp/Copp_Manager_Redesign_test_plan.md
```diff
@@ -0,0 +1,134 @@
...
+## Related documents
+
+| **Document Name** | **Link** |
+|-------------------|----------|
+| CoPP Manager Redesign HLD | []|
+
+
+## Overview
+
+In the current design of Copp Manager, before installing a trap, CoppMgr checks if one of the following requirements is fulfilled:
```
dgsudharsan, line 12: Can we modify 'current design' to 'previous design'?

JibinBao: Fixed

prsunny (approved):

dgsudharsan (approved):

liat-grozovik:
@JibinBao once the PR is ready please add it to the PR description for reference.

PR #912, opened 2021-12-14
Switch Port Modes and VLAN CLI Enhancement

ham-xa:
We have created an HLD for Switchport modes and enhancements to VLAN CLI.

Related PRs:
| Repo  | PR title | State |
| ----------------- | ---------------- | ------------|
| sonic-utilities | [Switch Port Modes and VLAN CLI Enhancement](https://github.com/sonic-net/sonic-utilities/pull/2419) | GitHub issue/pull request detail |
| sonic-buildimage | [Switchport Modes Port & Port Channel Yang Model Configurations](https://github.com/sonic-net/sonic-buildimage/pull/13580) | GitHub issue/pull request detail |
| sonic-mgmt | [updated vlan creation in '/spytest/apis/switching/vlan.py' with switchport mode assignment](https://github.com/sonic-net/sonic-mgmt/pull/7625) | GitHub issue/pull request detail |

HLD signoff people:
@gechiang ,  @venkatmahalingam 

Yang mode code PR:
@ganglyu


doc/L2_802.1q_Tunneling_Support_HLD/L2 dot1q HLD.markdown
```diff
@@ -0,0 +1,460 @@
...
+
+## Introduction
+
+IEEE 802.1Q is a networking standard which defines the VLANs tagging on an ethernet network. The Ethernet frame contains a TAG with fields that specify VLAN membership and user priority. As seen in the below Ethernet frame, the TAG is inserted between the Source MAC address and Type/Length. The TAG is of 16-bit length. The 802.1Q tags are inserted into the frames before  transmitting out of the interface and similarly, the tags are removed from frames received from the interface. The frames are transmitted and received from the interfaces that are configured with VLAN and they are tagged with VID.
```
venkatmahalingam, line 71: I believe, there is a HLD for VLAN stacking (https://github.com/Azure/SONiC/pull/915/files), this HLD seems duplicate, please clarify.

ahmadarsalan: To the best of our understanding, our HLD and HLD 915 (VLAN Stacking), submitted about a week after our HLD, are very different. Our HLD covers the 802.1Q trunking encapsulation implementation within a client network. On the contrary, HLD 915 covers the VLAN staking Q-in-Q which is related to the ISP network. So, there is no duplication in both the HLDs.

venkatmahalingam: Per my understanding, the outer tag TPID is 0x8100 instead of 0x88a8 in the HLD 915, and I think, your encapsulation also does the same thing, please read the HLD 915 and combine the HLDs if required.

Mati86: In this HLD we have implemented 802.1Q standard with respect to client networks which performs frame tagging for carrying different VLANs traffic on the trunk link within a client network. Only one tag is added within the client network (No stacking). On the other hand, HLD # 915 has implemented 802.1ad which deals with the forwarding of VLAN IDs (VIDs) between geographically separated different sites of the clients connected via ISPs. ISPs use 802.1ad for stacking VID from different clients and add an ISP tag on top of those stacked VIDs. So, 802.1ad deals with the tagging inside an ISPs network and 802.1Q deals with the tagging inside clients networks. There is another type of tagging according to 802.1ah standard which deals with another layer of stacking of ISP's tags inside Backbone networks. All of these three Vlan tagging standards are independent and have a completely disjoint scope of operation. These standards are identified by switches from different values of TPID field inside the tags, 81-00 for customer VLAN Tag, 88-a8 for Service Provider or Backbone VLAN Tags and 88-e7 for Backbone Service Instance Tag. Please refer to table 9-1 of the IEEE 802.1Q-2018 standards guide for the details of technical differences in these standards. The IEEE 802.1Q-2018 standards guide is available at: https://standards.ieee.org/standard/802_1Q-2018.html. Thus, The scope of this HLD and HLD#915 are very different and are implementing related but different and disjoint standards. In SONiC roadmap "L2 Dot1Q tunneling support" is mentioned so we started implementation from the 802.1Q baseline standard and we have had plans for implementing 802.1ad and 802.1ah in separate HLDs.

zhangyanzhao: This feature is not in community release tracking list.

ridahanif96:
We have updated our HLD along with the code. @venkatmahalingam pls review.

zhangyanzhao:
@venkatmahalingam can you please review this PR and approve if you are ok? Thanks.

zhangyanzhao:
@venkatmahalingam  can you please help to review this HLD and corresponding code PR? Thanks.

zhangyanzhao:
@venkatmahalingam  can you please approve this PR if all your comments have been addressed? Otherwise, can you please help to repaste your question? I am not quite clear what is open now.

venkatmahalingam:
@prsunny @dgsudharsan Please review this HLD.

ridahanif96:
@prsunny @dgsudharsan , Please review this HLD.

venkatmahalingam:
@zhangyanzhao I think, we should quickly update the community in one of the Tuesday's call about this change as it affects the way we configure the VLAN configs on port/port-channel.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,390 @@
...
+
+In trunk mode a port possesses the following properties
+* Receives untagged traffic for a single Vlan
+* Forwards untagged traffic for a single Vlan
+* Receives tagged traffic for one or more Vlan(s)
+* Forwards tagged traffic for one or more Vlan(s)
+
+### Hybrid Mode
+
+In hybrid mode a port possesses the following properties
```
prsunny, line 79: Is this planned to be added as a CLI option? I don't see it in example below

ridahanif96: Hybrid mode is the future scope of this HLD. This HLD has two switchport modes (access, trunk). 
For hybrid mode, we have a separate [HLD](https://github.com/sonic-net/SONiC/issues/1136) which is in the roadmap of 202305 and we will present it later in one of the future community reviews.

zhangyanzhao:
re-discussed in sonic community meeting.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,388 @@
...
+
+Switch ports are Layer 2-only interfaces associated with a physical port. A switch port can have three modes: access port, a trunk port, or a hybrid port. We can manually configure a port as an access port, trunk port or hybrid port depending on requirement . Switch ports are used for managing the physical interface and associated Layer 2 protocols and do not handle routing.
+
+### Access Mode
+
+In access mode a port possesses the following properties
+* Receives untagged/tagged traffic for a particular Vlan
+* Forwards untagged/tagged traffic for a particular Vlan
+
+### Trunk Mode
```
madhupalu, line 69: - What is the use case adding new switch port mode? 
- List down a use case scenario for existing switch modes (access, trunk and routed port)
- what is the default switch port mode? 
- what is transition graph from one switch port mode to other?

rck-innovium:
The CONFIG\_DB changes are missing in this doc.

https://github.com/sonic-net/SONiC/blob/master/doc/hld_template.md#config-db-enhancements

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,388 @@
...
+### Add Multiple VLANs
+ 
+![MultipleVlans](https://user-images.githubusercontent.com/61490193/208155973-801f99e5-bc79-4e4d-9db4-be13decf6637.png)
+
+
+__*Figure 3: Sequence Diagram for adding Multiple VLANs*__
+
+**1.** The SONiC CLI will call the multiple vlan list parser which will parse multiple vlan(s) in a list and return.
+
+**2.** Check if the VLAN is in the range (2-4094) and whether the vid already exists or not.
```
madhupalu, line 139: Is this HLD introducing vlan range? if it is do you see any performance challenges programming bulk vlans/range to ASIC?  Please include a test case to validate bulk vlans?

zhangyanzhao:
Target 202305 release.

lguohan:
feedbacks:
1. state transition graph between access/trunk mode
2. db migrator to address backward compatibility
3. default vlan mode should be trunk
4. sonic-mgmt tests needs to modified accordingly

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,388 @@
...
+
+
+## High-level Design
+
+In this section we will explain sequence diagrams for the implemented features.
+
+### Switch Port Mode
+
+![SwitchPortMode-PortPortChannel](https://user-images.githubusercontent.com/61490193/207826139-3ff0c41a-69af-4018-bf19-c083e1307bf7.png)
+
```
rck-innovium, line 106: This sequence diagram doesn't clearly show the separation between the processing done in the CLI vs that done in the clicommon. For example, the same "if" checks are done in both the components.

zhangyanzhao:
community re-reviewed this HLD on 2/7/2023 and recording is https://zoom.us/rec/share/1RIEGIIRn2O3ltEUDUVmyrze-7QSKDORTz1CL9JZlijh_sShvIOzkoIKXXIx6J0T.JZgQNK8ky0a0HhFd

zhangyanzhao:
@ham-xa  can you please help to address the review comments?  @venkatmahalingam  @gechiang are the registered reviewers for this PR.

ridahanif96:
@zhangyanzhao we are on it and updating HLD as per reviewers' suggestions/comments along with sonic-mgmt test cases.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Updated HLD as per the suggestions/concerns.

ridahanif96:
> The CONFIG_DB changes are missing in this doc.
> 
> https://github.com/sonic-net/SONiC/blob/master/doc/hld_template.md#config-db-enhancements

Done

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Yes, we have introduced the vlan range. No performance challenges will be faced using this. Please, see HLD for test cases. Thanks

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Added state transition diagram for Switchport modes in HLD for more clarity. Thanks

ridahanif96:
> feedbacks:
> 
> 1. state transition graph between access/trunk mode
> 2. db migrator to address backward compatibility
> 3. default vlan mode should be trunk
> 4. sonic-mgmt tests needs to modified accordingly

Updated as per suggestions and modified sonic-mgmt testcases as well.

ridahanif96:
We have updated our HLD as per suggestions/concerns. @venkatmahalingam @gechiang Please Review. Thanks

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+## Use Caess for Switchport modes
+### Access Mode
+Here are some common use cases for switch port mode access:
+
+**1.** Connecting devices to a network: When users need to connect devices to a switch, such as computers, printers, or IP phones, you can configure the switch ports to operate in access mode. This allows the devices to communicate with other devices on the network.
+
+**2.** Restricting VLANs: Users can use switch port mode access to restrict the VLANs that a device can access. This can help control access to network resources and improve network security.
+
+**3.** Improving network performance: By using switch port mode access, users can reduce the amount of unnecessary broadcast traffic on the network. When a switch port is in access mode, it only sends traffic to the device connected to that port, rather than broadcasting traffic to all devices on the network.
```
gechiang, line 117: What was stated was not correct... it only reduced flooding to the VLAN that this port belongs to which is one single VLAN.  if you had all devices connected to this same VLAN, it does not really reduce the flooding traffic.  you may want to re-format this...

ridahanif96: Ok, will re-format it

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+
+
+__*Figure 4: Sequence Diagram for deleting Multiple VLANs*__
+
+**1.** The SONiC CLI will call the multiple vlan list parser which will parse multiple vlan(s) in a list.
+
+**2.** Check if the VLAN is in the range .
+
+**3.** Check if vlan exists or not. If vlan does not exist it will show an error message and truncate. Examples for such cases have been mentioned in the examples section as well.
+
```
gechiang, line 189: If the VLAN still has member ports, you should not allow delete of the VLAN right?  Or do you implicitly remove the port membership to the VLAN being deleted?

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+
+__*Figure 5: Sequence Diagram for Adding VLAN member on Trunk Port*__
+
+
+**1.**  The SONiC CLI will Call vlan_member_input_parser to add multiple vlan member(s).
+
+**2.**  Check for vlan_range.
+
+**3.**  Check port mode (access or trunk) for adding vlan members (untagged or tagged) on a port from the PORT table and return.
```
gechiang, line 209: what if the mode is trunk and user is adding an untagged vlan that the trunk already had a different untagged VLAN ID? do you allow this operation by implicitly remover the port from previous Untagged VLAN membership and then add it to the new untagged VLAN membership? or you will reject the command?

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+
+**4.**  Add member vlan(s) in multiple commas or range list, all flag & except flag can be used as well.
```
gechiang, line 211: What if part of the requested VLAN is already part of this Trunk port? do you ignore them or do you flag that as an error and reject the request?

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+### Delete Multiple VLAN member on Trunk Port
+
+![Delete-Vlan-Member](https://user-images.githubusercontent.com/61490193/223199172-0ae75308-6e93-4176-9db6-3bcfbd02b948.png)
+
+
+__*Figure 6: Sequence Diagram for deleting VLAN member on Trunk Port*__
+
+**1.** The SONiC CLI will call the vlan_member_input_parser which will parse multiple vlan(s) in a list.
+
+**2.** Check if the members exist in the range .
```
gechiang, line 228: what happens if the requested vlan to be deleted is not part of the port's vlan membership? Do you reject the request?

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+            "mode": "trunk",
+            "mtu": "9100"
+        }
+      }
+ **Assumption: PortChannel0001 has been configured as trunk by user
+
+
+## DB Migrator Enhancements
+As there is a change in schema for PORT & PORT CHANNEL Table in config DB . We have added support to migrate the entries to the new schema in db_migrator.py. A new “mode” field is added whose default value is “trunk” to map it with the old configurations for backward compatibility.
+
```
gechiang, line 299: You also need to check if there are any VLAN membership.  If there are no vlan membership the default should be routed instead of Trunk?

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: We did not change the current functionality of SONiC with regards to VLAN and Member port. If the VLAN has member assigned to it and user tries to delete the VLAN, CLI will give an error. Please refer to examples section, we have already included example of this behavior.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: It will not allow to add more than one untagged members to the Trunk VLAN. Since the trunk mode can have only one untagged member, it will not accept another untagged vlan member when it already has one untagged member assigned to it. An error will be generated saying the PORT already contains a untagged member, which is normal SONiC VLAN behavior.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: By using multiple VLAN command, CLI will add the VLANs to the port in ascending order and explicitly print on the CLI the VLANs which have been added to the PORT and when it reaches the VLAN which already exists it will give an error. Please refer to the examples section, we have included example of this behavior as well.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Yes, it will give an error saying the VLAN doesn't exist as a member of PORT and the command will abort.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: The mode attribute will appear only if VLAN is configured in the old configuration.  For db_migrator, if there is no VLAN membership it does not explicitly put the mode attribute in the PORT Table, and if we run the command "show interfaces status" it will show it as routed. The default mode for old vlan configurations is set as "trunk" as per the community suggestion/comments. But if you want, we can change the db_migrator behavior to set mode as routed in case of no VLAN membership.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+    ***This command only work for trunk ports***
+
+  ```
+  
+  ![Portchannel-except](https://user-images.githubusercontent.com/61490193/223192175-396b487b-fe75-43f6-8598-d9e74773dd58.png)
+
+
+#### Multiple VLans command truncate:
+
```
venkatmahalingam, line 526: Looks like we print lot of message e.g 
VLAN10 has been added
VLAN11 has been added
.....................

If possible, please print only the failures.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+
+###  Examples/Usage of VLAN CLI Enchancemnts Command
+
+ #### Add multiple Vlan(s) using a single command:
+
+```
+  admin@sonic:~$ sudo config vlan add -m 100,200,300
+
+  Usage:  This command will add multiple comma separated vlan in a list
```
venkatmahalingam, line 468: If you show the errors being shown for each error conditions would be helpful.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+![Architecture](https://user-images.githubusercontent.com/61490193/217232727-811b98a8-ce9c-4227-a94a-1c83a4825a62.png)
+
+
+## High-level Design
+
+In this section we will explain sequence diagrams for the implemented features.
+
+### State Transition Diagram of Switchport modes for PORT or PORT CHANNEL
+
+The following state transition depicts the behavior of Switchport  modes access and trunk for adding port modes on a Port or PortChannel. Switching between these modes is also depicted here. Default port mode is “routed”.
```
venkatmahalingam, line 149: image

venkatmahalingam: Not sure what's PortChannel001/002 under VLAN column, please do the self review.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Sure we will update it.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+**3.** Read the existing vlan table and delete vlans(s).
+
+
+__*NOTE: This works same for Delete Multiple VLANs members from a PortChannel*__
+
+## CLI & YANG Model Configuration Commands & Usage 
+
+### CLI Configuration Commands
+
+**1.** config switchport mode <routed|access|trunk> <member_portname>/<member_portchannel>
```
gechiang, line 239: Shouldn't a port without switchport mode is meant to be "routed"?  switchport mode is a L2 concept where only layer 2 ports should have switchport mode configured.  L3 port in industry typically do not have switchport mode configured.  and I havenot seen any vendors use "switch port routed" for L3 interfaces...

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
gechiang: If no VLAN is configured, no "mode" should be set.  that is the desired default  DB migrator behavior.  Your previous sections seems to indicate every interface will have a "mode" regardless of being L2 or L3... which I disagree...

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+
+![Switch-Access-Trunk](https://user-images.githubusercontent.com/61490193/223190991-aeef8fd6-e71a-4bbf-9862-42ea4613002a.png)
+
+To verify the port mode user can make of “show interfaces status” command
+
+
+#### Switching Ports & PortChannels mode from Trunk to Routed/Access
+
+#### Switch from trunk to routed
+Switch from trunk to routed user need to remove the untagged/tagged member(s) first otherwise, the command will give an error:
```
gechiang, line 449: If there are no VLAN membership what would happen? will the "mode" be removed from the interface configuration?

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+
+![Switch-Trunk-Routed](https://user-images.githubusercontent.com/61490193/223191099-c037754d-c221-43b6-874e-6dbcd4eae639.png)
+
+
+#### Switch from trunk to access
+Switch from trunk to access, user need to remove the tagged members firstotherwise, the command will give an error:
```
gechiang, line 455: What happens to the Trunk Untagged membership when switch from Trunk to Access?  Will it become the Access port's untagged VLAN membership?

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+  ```
+  
+![Vlan-MultipleAdd](https://user-images.githubusercontent.com/61490193/223199652-01b5ed35-15c0-480c-ad78-c78566db62df.png)
+
+
+ ####  Add all Vlan(s) as Vlan Member(s) excluding a range of vlans using a single command:
+
+  ```
+
+  admin@sonic:~$ sudo config vlan member add -m -e 12-17 PortChannel0003
```
gechiang, line 512: Is this correct? I thought -m is for range of VLANs... should it be add all -e 12-17 instead?

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,608 @@
...
+## SAI API
+
+No SAI API change or addition is needed for this HLD. We will be using the existing SAI API.
+
+## Restrictions/Limitations
+
+This HLD did not consider hybrid switch port mode. We will be adding a separate HLD for hybrid mode. 
+
+### Testing Requirements/Design
+
```
gechiang, line 560: You should also mention a new show command that will display the siwtchport mode of the interface.  You should not be changing the output format of the existing "show interface status" output as it will break many existing applications, parsers, and testcases.  So it is best to disclose what is the new show command that you will be implementing here as well.

Also, I was looking for a section in this doc to clearly spell out all the possible transitions foe the switchport mode and what are allowed and not allowed.  Also, things as  if there are any default Access VLANs or Trunk untagged  (Native) VLAN when an interface is configured to switchport mode trunk.  It would be better to have a section dedicated to spell out all these for completeness and easy of review so that there are nothing unspecified and became ambiguous...

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: The untagged member will be retained and become member of Access Port

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: We will add a new command for switchport modes and will not change existing "show int status" output format.
Also we will add more specific examples for Transition of port mode to avoid ambiguity.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Yes, this is correct. We will add more examples in this regard to avoid confusion.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Agree on this, we will update it.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Initially, when no VLAN membership is configured , there will be no "mode" but soon after a port is configured as access and then it is switched from access/trunk to routed, mode entry will remain in DB and mode will be "routed".

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Consider, if a port is in access/trunk mode and no vlan member is configured then this situation is called "blackhole". So, a port can be in access/trunk mode without having any member.
We have given switchport mode routed here because if a user is coming back from access/trunk then he needs to specifically configure the port back to routed. This mode is actually equivalent of the negation command like "no switchport mode access" available in Cisco. As we do not have this command in SONiC so, we need routed mode to exit from access/trunk mode.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
venkatmahalingam: switchport mode access/trunk config cant be removed as long as the port is associated with VLAN(s).
Even if all the VLANs are removed from the port, the user has to explicitly remove the mode configuration, there should not be any automatic mode field removal from PORT/PORTCHANNEL table.

ridahanif96: Yes, this is correct. There is no automatic mode removal from PORT/PORTCHANNEL. The user will explicitly remove the mode (access/trunk)configuration.

ridahanif96:
> defer to 202305 release to continue

Part of 202305, we are already in HLD & code review with the reviewers, will finalize this by tomorrow.

ridahanif96:
We have added detailed examples for Switchport Modes as per suggestion/concerns. @gechiang @venkatmahalingam please review.

venkatmahalingam (approved):

ridahanif96:
@gechiang can you please review and merge this PR. Target is 202305

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Updated as suggested.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,760 @@
...
+## High-level Design
+
+In this section we will explain sequence diagrams for the implemented features.
+
+### State Transition Diagram of Switchport modes for PORT or PORT CHANNEL
+
+The following state transition depicts the behavior of Switchport  modes access and trunk for adding port modes on a Port or PortChannel. Switching between these modes is also depicted here. Default port mode is “routed”.
+
+![State-Transition-Diagram](https://user-images.githubusercontent.com/61490193/223195784-05cc4e98-8883-4911-bca3-fded6a87d202.png)
```
gechiang, line 152: This diagram conflicts with what you have states under this section:

Examples/Usage of Switchport Modes Command
Following example shows usage of switchport modes “access” and “trunk” and switching of port mode from routed to access/trunk. Before switching mode from routed to access/trunk, the user must remove IP first. After the mode has been switched to access/trunk there will be no untagged/tagged vlan member(s) assigned

So when you move from trunk to access or vice versa, do youy keep the "untagged" membership or that is also removed?  the diagram seems to say it retains the "untagged" membership.  I actually like retaining the untagged membership and only force the user to have to remove the  tagged memebrship when moving from trunk to access...

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,760 @@
...
+**7.** IP Assignment on Access Port
+
+Ethernet0 is in Access mode, IP assignment on the access port is not allowed. Configuring IP Assignment on Access Port will show following error:
+
+<img src="https://user-images.githubusercontent.com/61490193/236170194-2423f230-483a-42b7-936c-5939bf0e84ae.png" width="65%" height="30%"> 
+
+**8.** Change Mode from Access to Routed
+
+Ethernet0 is in Access mode, switching an access port to routed is not possible until it has an untagged member. Changing  mode from access to routed will show following error:
+
```
gechiang, line 551: When changing from aCcess to routed, what is in the confiogDB for the corresponding port?  will it remove the "mode" completely or have "mode: routed"?  Can you clarify on this?  I would prefer the "mode" be removed completely...

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,760 @@
...
+**7.** Change Mode from Trunk to Routed 
+
+Ethernet4 is in Trunk mode, Changing Trunk port to routed is not possible until it has an untagged and tagged members. Changing Trunk to routed will show following error:
+
+
+<img src="https://user-images.githubusercontent.com/61490193/236533178-bba78b3f-f429-4042-8de9-0decbe938f36.png" width="65%" height="40%">
+ 
+**8.** Change Mode from Trunk to Access
+
+Ethernet4 is in Trunk mode, Changing Trunk port to access is possible and  its  untagged members wll retain. Changing Trunk to access will show following:
```
gechiang, line 628: What about from Access to Trunk does it also retain the untagged vlan membership?

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
```diff
@@ -0,0 +1,760 @@
...
+Ethernet4 is in Trunk mode, it can have only 1 untagged member. Configuring More than 1 untagged member on trunk Port will show following error:
+
+
+<img src="https://user-images.githubusercontent.com/61490193/236170542-587403ed-3cf0-465a-9e58-fff7c8130508.png" width="65%" height="30%">
+
+
+**5.** Tagged Vlan Member Assignment on Trunk Port
+
+Ethernet4 is in trunk mode, it can have multiple tagged members. Configuring tagged member on trunk Port will show following
+
```
gechiang, line 608: What if you attempt to add a tagged vlan to the port that is alerady part of the untagged vlan? what is its behavior?

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: when changing from access to routed, the mode of the corresponding port in config_db will be changed from "access" to "routed". It cannot be removed completely as this "mode" attribute is saving state of the port that is either routed , access or trunk. 
This was discussed in our internal discussion and it was decided that this attribute will not be removed completely for changing from access to routed or from trunk to routed.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: Yes, untagged member will be retained when moving from access to trunk or vice versa. Modified this section, as per your concern.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
gechiang: I see... Please spell this out in your doc as I have asked this before and your previous response is yes you will modify to remove the "mode" when transition from "access" or "trunk" to "routed". But now you are reversing what you have previously agreed to... So please specify this clearly in your doc so there is no ambiguity.

gechiang: So an interface in mode "routed" means it has no VLAN membership. it may not have ip address configured either and still qualify as mode "routed" correct?   If you agree this is correct statement, can you add this in the doc as well?

ridahanif96: I have updated this concern and explicitly spelled it out to avoid any ambiguity.

ridahanif96: Yes this is the correct statement, an interface in "routed" mode implies that it does not have VLAN membership. It's possible for an interface in "routed" mode not to have an IP address configured. This statement is added to provide further clarity on the behavior of interfaces in "routed" mode. Please see HLD example section **point 8** under "**Configuring Port & PortChannel from Routed to Trunk"** & **point 8** under **"Configuring Port & PortChannel from Routed to Access"** for specific example on this behavior.

doc/vlan/switchport-mode-support/Switchport Mode and VLAN CLI Enhancement.md
ridahanif96: A Port in access mode can have only one untagged Vlan member. if a user tries to add a tagged vlan to such a port that already has an untagged vlan it will through an error. Please see the above section, **point 4,6.** under **"Configuring Port & PortChannel from Routed to Access"**  for a specific example of this behavior.

gechiang (approved):

ridahanif96:
@zhangyanzhao can you please merge this PR, Thanks

PR #918, opened 2022-01-04
Add counters enabling redesign document

noaOrMlnx:
enable_counters.py script sleeps for 3 minutes before enabling the counters. 
This design purpose is to change the sleep mechanism to be event driven. 

After the change, the script logic will be in flexCounterOrch and we will wait for the system to be stable and just then enable the counters.


doc/counters_enabling_redesign.md
```diff
@@ -0,0 +1,66 @@
...
+## Requirements
+
+Change enable_counters.py script to be event driven from config_db instead of using sleeps.
+
+
+## High-Level Design
+
+### Ports Daemon
+
+enable_counters.py script will be refactored to be running as a daemon.
```
lguohan, line 34: this should be merged into the orchagent

lguohan: please also add config db schame and yang schema for this feature.

bingwang-ms:
Will this change increase the time cost of reconciliation after fast/warm reboot since the counter is expected to be enabled earlier after this change?

doc/counters_enabling_redesign.md
noaOrMlnx: @lguohan 
Updated the design according to your request. 
regarding the config db schema & yang model - why do we need to add config db schema? we don't add anything new to DB - it's all internal. 
If you mean the delay_status attributes, we already have schema & yang for them. 
https://github.com/Azure/sonic-swss-common/blob/master/common/schema.h#L225
https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-flex_counter.yang#L49

lguohan: i think in the config db, we need to add something to set the delay time, for example 300 seconds which allow user to specify how long they want to delay?

yxieca:
@vaibhavhd please review from the fast/warm reboot perspective.

doc/counters_enabling_redesign.md
noaOrMlnx: @lguohan 
This PR is related to the fact we are moving from sleep mechanism to be event driven. 
I agree that adding an option of how much delay the user wants is good, but I don't think it's related to this design. 

If you want,  our team will take this feature, but in another design and plan. 
what do you think?

liat-grozovik:
@vaibhavhd any feedback on warm/fast boot? we wish to close the coding part and would like to get your feedback soon please.
@lguohan HLD was updated following your comments. Could you please take further look?

vaibhavhd:
For fast/warm reboot -
I don't see concerns for warmboot.
For fastboot - do we have any prototype to be evaluated for the design? It will be interesting to measure CPU consumption at the time when counters are enabled much earlier than 3 minutes after reboot.
AFAIK, high CPU consumption due to enabled counters was the impacting factor in fastboot reconciliation. This caused the dataplane downtime to be > 30s (which is not accepted).

vaibhavhd:
More for fast-reboot:

@shlomibitton with this new enable_counters re-design, do we still need `FLEX_COUNTER_DELAY_STATUS` and its related changes?
`FLEX_COUNTER_DELAY_STATUS` was added to prevent `FlexCountersOrch` from enabling counters until enable_counters script is called.

Since enable_counters.py script is now not needed, and the delay mechanism is local to `FlexCountersOrch`, we probably not need the changes that were added to support new flag - `FLEX_COUNTER_DELAY_STATUS`.:

https://github.com/Azure/sonic-buildimage/pull/8500
https://github.com/Azure/sonic-swss-common/pull/523
https://github.com/Azure/sonic-swss/pull/1877
https://github.com/Azure/sonic-utilities/pull/1768

shlomibitton:
@vaibhavhd Yes, we don't need it anymore.
But the new change need to handle this internally.
Means, ignore all events from Config DB and enable after all ports are up or the timeout reached.

doc/counters_enabling_redesign.md
```diff
@@ -0,0 +1,66 @@
...
+Eventually, when the list is empty, enable the counters.
+
+- The daemon will also wait for a timer in order to be able to enable counters even if one of the ports is not stable.
+    
+    If after 3 minutes (180 seconds), the counters were not enabled yet, enable counters.
+
+    The daemon will create a SelectableTimer and wait for the it to poke.
+
+    When doTask(SelectableTimer*) function will be triggered, it will check if the counters are already enabled.
+    If not, enable them.
```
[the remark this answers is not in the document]
noaOrMlnx: > Chances of race-condition here? What happens if counters enabling is in progress and the `SelectableTimer` is also triggered?

If counters enabling is in progress, it means the daemon first changed a boolean to True. so when the timer will expire, it will check the boolean and will not enable the counters as well.

> Also, after the enabling of counters, what happens to the new data structure defined in `FlexCountersOrch`? There can be two cases to be covered here - event-driven mechanism (new design) and timeout (old design).

Not sure I understand the concern here. we are using the event driven mechanism, but also don't want to be blocked by one flipping port for example, so we have timer to cover this case.

doc/counters_enabling_redesign.md
```diff
@@ -0,0 +1,66 @@
...
+    - If a port state has been changed, change it in the internal data structure.
+    - If a port/portchannel was added, add it to the internal data structure with the proper admin status.
+    - If a port/portchannel was deleted, remove it from the internal data structure.
+    - If a delete operation was performed to one of the ports inside the internal data structure - remove the port.
+
+Eventually, when the list is empty, enable the counters.
+
+- The daemon will also wait for a timer in order to be able to enable counters even if one of the ports is not stable.
+    
+    If after 3 minutes (180 seconds), the counters were not enabled yet, enable counters.
```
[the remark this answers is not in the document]
noaOrMlnx: Fixed

doc/counters_enabling_redesign.md
```diff
@@ -0,0 +1,66 @@
...
+
+In the current design of counter enabling, when enable_counters.py script is triggered, if the switch uptime is less then 5 minutes, it sleeps for 3 minutes. This is done in order to make sure the system is up and synchronized.
+
+The purpose of counters redesign is to change the 'sleep' mechanisem to be event driven and to have the exact time when the switch is ready to enable counters.
+In addition, enable_counters.py script will be removed, and its logic will now be in flexCountersOrch.
+
+## Requirements
+
+Remove enable_counters.py script and merge the logic into FlexCountersOrch.
+FlexCountersOrch will wait for system to be up using events from APP DB, then enable the counters using the logic written in enable_counters.py.
```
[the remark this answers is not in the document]
noaOrMlnx: Done

noaOrMlnx:
> @vaibhavhd Yes, we don't need it anymore. But the new change need to handle this internally. Means, ignore all events from Config DB and enable after all ports are up or the timeout reached.


@vaibhavhd  @shlomibitton 
In current design, `FLEX_COUNTER_DELAY_STATUS` is an optional flag that is being set only when we do fast-boot. 
If I will implement your suggestion, we will delay counters enabling after all boot types. 
do you think this is the right behavior?

doc/counters_enabling_redesign.md
```diff
@@ -0,0 +1,143 @@
...
+    
+    As enable_counters.py script, if the system is after reboot (uptime < 5 minutes), the timer will be set to 3 minutes. 
+    Else, the timer will be set to 1 minute. 
+    
+    If after the timer has expired, the counters were not enabled yet, enable counters.
+
+    When doTask(SelectableTimer*) function will be triggered, it will check if the counters are already enabled.
+    If not, enable them.
+
+
```
kperumalbfn, line 125: Please include a section for fast/cold/warm reboot and dynamic port breakout scenarios.

liat-grozovik:
@noaOrMlnx 
Please align the HLD according to the HLD template. In additional 
1. Please add a clear table for each type of reboot what will be the max time. please include: reboot, fastboot, warmboot, config reload, swss service restart, etc. It will be helpful to understand the system behaviour
2. Please add a section for testing. Lets cover the entire flow including the time expired as well as counters start before and timer is stopped.  
3.Following  the community review feedback need to find a simpler way to declare 'fastboot' completed. based on that we can start counters/stop the max timer.

liat-grozovik:
Following offline discussion we will improve fastboot flow and wont move forward with this HLD. thus it is closed.
Please refer to https://github.com/Azure/SONiC/pull/980 for more information

PR #932, opened 2022-01-24
Put port operational speed to STATE DB

Junchao-Mellanox:
**Why I did this?**

Fix issue https://github.com/Azure/sonic-buildimage/issues/9595. 

**What I did?**

1. Move operational speed to STATE DB
2. `show interface status` shall show the speed from STATE DB first, and if speed in STATE DB is not present, it shows the interface speed from APPL DB

**Related PRs:**

| [[sonic-swss] Try get port operational speed from STATE DB](https://github.com/Azure/sonic-swss/pull/2119) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-utilities] Try get port operational speed from STATE DB](https://github.com/Azure/sonic-utilities/pull/2030) | GitHub issue/pull request detail | GitHub pull request check contexts |

### DIFF
--- a/doc/port_auto_neg/port-auto-negotiation-design.md
+++ b/doc/port_auto_neg/port-auto-negotiation-design.md
@@ -139,7 +139,7 @@ The related port attributes are listed below:
 
 Please note that `SAI_PORT_ATTR_ADVERTISED_INTERFACE_TYPE` is a new attribute introduced in SAI 1.7.1. Vendors need to implement this attribute in their SAI implementation.
 
-### Configuration and management 
+### Configuration and management
 
 #### CLI Enhancements
 
@@ -183,7 +183,7 @@ Return:
   error message if interface_name or speed_list is invalid otherwise empty
 
 Note:
-  speed_list value "all" means all supported speeds 
+  speed_list value "all" means all supported speeds
 ```
 
 This command always replace the advertised speeds instead of append. For example, say the current advertised speeds value are "10000,25000", if user configure it with `config interface advertised-speeds Ethernet0 40000,100000`, the advertised speeds value will be changed to "40000,100000".
@@ -227,7 +227,7 @@ Return:
   error message if interface_name or interface_type_list is invalid otherwise empty
 
 Note:
-  interface_type_list value "all" means all supported interface type 
+  interface_type_list value "all" means all supported interface type
 ```
 
 This command always replace the advertised interface types instead of append. For example, say the current advertised interface types value are "KR4,SR4", if user configure it with `config interface advertised-types Ethernet0 CR4`, the advertised interface types value will be changed to "CR4".
@@ -346,7 +346,7 @@ The advantage here is that user can get a list of valid interface type from CLI.
 
 For speed and adv_speeds, there is a SAI API to get the supported speed list for a given port. The idea here is to query supported speed after orchagent creating port object, and the supported speed list will be save to STATE_DB for CLI to validate. A new field **supported_speeds** will be added to **PORT_TABLE**. If this field is present, CLI will use this field to validate the input speed and adv_speeds argument, otherwise, no validation will be performed on CLI side. The STATE_DB change will be described in [State DB Enhancements](#state-db-enhancements).
 
-#### Config DB Enhancements  
+#### Config DB Enhancements
 
 SONiC already defined two fields related to port speed setting: **speed**, **autoneg**. 3 new fields **adv_speeds**, **interface_type**, **adv_interface_types** will be added to **PORT** table:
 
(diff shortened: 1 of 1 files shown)


doc/port_auto_neg/port-auto-negotiation-design.md
```diff
@@ -396,9 +396,23 @@ To support validate interface speed on CLI side, a new field **supported_speeds*
...
+
+1. Configure port speed to 100G
+2. Configure advertised speed to 50G and enable auto negotiation
+3. Port operational speed turns to 50G
+4. Configure any port attribute, e.g mtu, portsyncd would put all port attributes to APP DB
+5. `show interface status` displays port speed as 100G which is incorrect
+
+To overcome this issue, following changes are required:
+
+1. Put port operational speed to STATE DB PORT_TABLE
```
prsunny, line 413: Agree to put oper state to STATE_DB, but i think as long as PORT table exists in APP_DB, we should put oper state in the APP_DB as well for consistency with previous model

prgeor: @prsunny is it because you still suspect there could be places where we may be still getting the "operational" speed from APP_DB?

prsunny: Thats right.

Junchao-Mellanox: Hi @prsunny and @prgeor , the problem is that we cannot guarantee that the oper speed value in APP_DB is always consistent with the actual oper speed in PHY. Please see issue https://github.com/Azure/sonic-swss/pull/1769. And I paste the issue description here:

The field "speed" in PORT_TABLE of APPL_DB has two meanings:
1. The configured port speed
2. The actual port speed

This is all right when SONiC has no auto negotiation feature because the actual speed must equal to the configured speed or the operational status is down. However, with port auto negotiation feature, if autoneg mode is enabled, the actual speed could be a value of adv_speeds which might not equal to "speed" in CONFIG DB. Now port auto negotiation feature handle it this way: 
1. if autoneg mode  is disabled, everything goes to the old way, the actual speed is always equal to field "speed"; 
2. if autoneg mode is enabled, and if operational status is up, get actual speed from SAI and update the field "speed" in APPL_DB.

For case#2, there is an issue. Think about following flow:
1. Set Ethernet0 speed=400000
2. Set Ethernet0 autoneg=on, adv_speeds=100000, and Ethernet0 is operational up. Now the speed of Ethernet0 in "show interfaces status" is 100G because orchagent get the actual speed from SAI. This is right.
3. Issue command: "config interface type Ethernet0 CR". (or configure any other port attribute)

Step#3 won't affect the auto negotiation status because the autoneg mode is enabled, force interface type will not be applied to SAI. But it triggers PORT_TABLE change and portsyncd will push all Ethernet0 related configuration to APPL_DB. So, speed of Ethernet0 in "show interface status" changed to 400G. And 400G is not the actual speed in PHY.

liat-grozovik:
@prsunny @prgeor could you please update if you agree to the above statement and what if we can approve this HLD?

doc/port_auto_neg/port-auto-negotiation-design.md
```diff
@@ -396,9 +396,23 @@ To support validate interface speed on CLI side, a new field **supported_speeds*
...
+  speed                   = STRING                         ; operational speed
 
 An example value of supported_speeds could be "10000,25000,40000,100000".
 
+Before this feature, port speed in APP DB indicates both the configured speed and the operational speed. It is OK without this feature because port operational speed must be configured speed or port operational status is down. However, this is not true with this feature. Consider following flow:
+
+1. Configure port speed to 100G
+2. Configure advertised speed to 50G and enable auto negotiation
+3. Port operational speed turns to 50G
+4. Configure any port attribute, e.g mtu, portsyncd would put all port attributes to APP DB
```
qiluo-msft, line 408: >portsyncd [](http://example.com/codeflow?start=42&length=9)

This statement is wrong. portmgrd is the right process.

qiluo-msft, line 408: >all port attributes [](http://example.com/codeflow?start=62&length=19)

The right process should capture the changed attributes and only put the difference into APP DB. This could be achieved either by ConsumerStateTable or by the application.

Junchao-Mellanox: Hi Qi, portsyncd is also putting port configuration to APP DB. Please see https://github.com/Azure/sonic-swss/blob/45bdd1928e7d2e61d10fd1ce0da99abe1418bd13/portsyncd/portsyncd.cpp#L229

Junchao-Mellanox: I agree. Should we put this logic to portmgrd? I suppose both portmgrd and portsyncd put port configuration to APP DB. I suppose portmgrd is newer and should be the right one, but I am not sure why portsyncd is still doing the job. Maybe is some kind of workaround before?

qiluo-msft: portmgrd should subscribe ConfigDB port table and put the incremental attribute into AppDB.

doc/port_auto_neg/port-auto-negotiation-design.md
prgeor: i had offline discussion with @prsunny, we agree to keep configured speed in ConfigDB/AppDB and operational speed in StateDB

doc/port_auto_neg/port-auto-negotiation-design.md
prgeor: @Junchao-Mellanox agreed to fix this with another PR

prgeor: @Junchao-Mellanox has agreed to raise PR to bring portmgrd instead of portsyncd for updating the port attributes in the AppDB

prgeor (approved):

prsunny (approved):

PR #939, opened 2022-02-04
[auto-techsupport] support techsupport generation on potential memory…

stepanblyschak:
… leaks


doc/auto_techsupport_and_coredump_mgmt.md
```diff
@@ -68,14 +76,59 @@ The naming format and compression is governed by the script `/usr/local/bin/core
...
+
+```${container}_mem_usage```  used memory by $container ("docker stats --no-stream --format {{.MemUsage}}" $container),
+
+```${container}_mem_usage_threshold```  configured memory threshold for $container, (100 - ${container}available_mem_threshold), 
+
+```mem_free``` is the total minus mem usage, ```mem_free_threshold``` - mem free threshold.
+
+the SONiC techsupport is automatically generated.
+
+```mem_free_threshold``` is there to invoke dump when there is quite small amount of memory left that is needed to successfully execute "show techsupport". This is going to be 200 Mb by default, as at least 80-90 Mb takes "show techsupport" execution.
```
qiluo-msft, line 97: >mem_free_threshold [](http://example.com/codeflow?start=3&length=18)

If suddenly mem_free_threshold < 80 MB, `show techsupport` will always fail. This invocation will make situation worse.

doc/auto_techsupport_and_coredump_mgmt.md
```diff
@@ -68,14 +76,59 @@ The naming format and compression is governed by the script `/usr/local/bin/core
...
+
+The check will be implemented as a script that is ran by monit periodically:
+
+```
+check program mem_checker with path "/usr/bin/mem_threshold_check"
+    if status != 0 for 10 times within 20 cycles then exec /usr/local/bin/mem_threshold_check_handler"
+```
+
+The action is going to be ran only once the mem_check script detects memory usage above threshold.
```
qiluo-msft, line 106: >once [](http://example.com/codeflow?start=35&length=4)

Do you mean once after each bootup? Considering some fluctuations and finally a memory exhausting, you will keep only the first fluctuation "show techsupport" file?

doc/auto_techsupport_and_coredump_mgmt.md
qiluo-msft, line 97: >Mb [](http://example.com/codeflow?start=213&length=2)

Mb -> MB

doc/auto_techsupport_and_coredump_mgmt.md
```diff
@@ -68,14 +76,59 @@ The naming format and compression is governed by the script `/usr/local/bin/core
...
+## 5. Schema Additions
 
 ### Config DB
 
 #### AUTO_TECHSUPPORT Table
 ```
 key                    = "AUTO_TECHSUPPORT|global"  
 state                  = "enabled" / "disabled"    ; Enable this to make the Techsupport Invocation event driven based on core-dump generation
+available_mem_threshold = 1*2DIGIT                 ; Memory threshold; 0 to disable techsupport invocation on mem leak.
+min-available-mem      = 1*5DIGIT                  ; Minimum free memory amount in Kb when techsupport will be executed.
```
qiluo-msft, line 131: >Kb [](http://example.com/codeflow?start=83&length=2)

Kb -> KB

qiluo-msft, line 131: >- [](http://example.com/codeflow?start=13&length=1)

Use underscore instead of dash?

qiluo-msft, line 131: >Kb [](http://example.com/codeflow?start=83&length=2)

The same

qiluo-msft: If always > 1MB, use MB as unit?

doc/auto_techsupport_and_coredump_mgmt.md
```diff
@@ -226,27 +298,43 @@ module sonic-auto_techsupport {
 #### AUTO_TECHSUPPORT_DUMP_INFO Table
```
qiluo-msft, line 298: >Table [](http://example.com/codeflow?start=32&length=5)

Is there an entry number limit in this table?

doc/auto_techsupport_and_coredump_mgmt.md
```diff
@@ -36,13 +37,20 @@ Currently, techsupport is run by invoking `show techsupport` either by orchestra
 
 However if the techsupport invocation can be made event-driven based on core dump generation, that would definitely improve the debuggability. That is the overall idea behind this HLD. All the high-level requirements are summarized in the next section
 
+Another use case is to gather more information about the system in case there is a memory usage threshold crossed.
+SONiC dump generated after system reboots due to out of memory is not enough for debugging the issue
+as all the information about processes and their mem usage, smaps (/proc/PID/smaps) is lost.
```
qiluo-msft, line 42: >lost [](http://example.com/codeflow?start=87&length=4)

Instead of invoking a full `show techsupport`, can we just save volatile information before reboot? For example, it makes no sense to save syslog during the memory pressure time. We could collect them after reboot.

doc/auto_techsupport_and_coredump_mgmt.md
qiluo-msft, line 106: >threshold [](http://example.com/codeflow?start=88&length=9)

If auto invoking too many times, do we have file rotation? Do we only rotate the auto dump files and keep user manual dump?

doc/auto_techsupport_and_coredump_mgmt.md
stepanblyschak: @qiluo-msft If user configures too low mem_free_thrshold the "show techsupport" will fail. Currently there is no upper limit of "show techsupport" peak memory consumption, which will make it hard to verify the user does not configure this threshold too low.

> This invocation will make situation worse.

It is assumed that the situation is already critical and we try to take the dump as best effort.

BTW, the swap partition would make it at least possible to execute some commands, however, it is not configured in SONiC.

doc/auto_techsupport_and_coredump_mgmt.md
stepanblyschak: I mean that the techsupport will run once if the script returns non zero status code for 10 times within 20 cycles. If memory usage is constantly too high for e.g 100 cycles, only 1 dump will be taken. If lets say the memory usage fluctuates, in example, 10 cycles too high in a row and then goes down for 10 cycles and then again up - there is a protection against too frequent auto tech dumps already present in the auto tech infrastructure.

doc/auto_techsupport_and_coredump_mgmt.md
stepanblyschak: @qiluo-msft updated in next patch

doc/auto_techsupport_and_coredump_mgmt.md
stepanblyschak: @qiluo-msft updated in next patch

stepanblyschak: @qiluo-msft Right, updated in next patch

stepanblyschak: @qiluo-msft Agree, updated in next patch

doc/auto_techsupport_and_coredump_mgmt.md
stepanblyschak: @qiluo-msft There is a rate limiting mechanism already available in auto techsupport infrastructure that limits the number of dumps that can be generated.

doc/auto_techsupport_and_coredump_mgmt.md
stepanblyschak: @qiluo-msft The user can configure the "since" parameter in CFG_DB and achive that result. If "since" is set to "0 sec" that would mean that no logs, sai, swss recs are recorded.

doc/auto_techsupport_and_coredump_mgmt.md
qiluo-msft: Sorry for misleading. I am asking if the free memory suddenly drop to below 80 MB, show techsupport will always fail. This invocation will make situation worse.

doc/auto_techsupport_and_coredump_mgmt.md
qiluo-msft: I see. Is auto rotation covered by design?

doc/auto_techsupport_and_coredump_mgmt.md
stepanblyschak: @qiluo-msft Do you mean that if suddenly an application perfroms a very big allocation so that we won't be able to perform "show techsupport". A similar situation might happen when the process leaks memory very quickly so that we don't have enough memory to inoke "show techsupport". In this design the dump collection is best effort, so we don't guaranty that the dump will be actually generated because of that. It is also assumed that if the system is running with that low free memory the situation is already critical (very likelly that reboot will happen) so that additional "show tech" does not help or make it worth, it is making the reboot happen earlier. If the user does want the system to stay with a very low free memory than the user has to disable this feature. Otherwise I don't see a way to handle such a situation.

doc/auto_techsupport_and_coredump_mgmt.md
stepanblyschak: Yes

qiluo-msft (approved):

liat-grozovik:
@stepanblyschak please update PR description with the PRs list covering this feature.

liat-grozovik (approved):

PR #945, opened 2022-02-23
SONiC Chassis Platform Requirements and Enhancements Analysis

abdosi:
Capture Platform specific Requirements for Chassis Platforms.


doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
+Section 1 Requirements that are needed by default:-
+        1. On LC the reboot command should power-cycle the entire LC . Expectation is Peer node should detect link down when reboot is given on LC
```
abdosi, line 2: Sonic Reboot command

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		2. On RP the reboot command should reboot the entire system (RP and LC). . Expectation is Peer node should detect link down when reboot is given on RP
```
abdosi, line 3: Graceful reboot for LC vs power down for LC ? Open question.

doc/pmon/pmon-chassis-requirements.md
abdosi: sonic reboot command that invokes platform plugin to reboot

rlhui:
Can we rename this PR to better reflect this doc?

abdosi:
cc @shyam77git

abdosi:
cc @mprabhu-nokia

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		3. Config shut/unshut of LC will be supported as per the Chassis-d design.
+		4. Generate syslog for all the critical events and share the threshold (for appropriate/needed components)  in documents and recommended for given threshold range.  Expectation is we will bind syslog to our Alert Orchestration system and perform recommnded action based on the documents.
+		5. PCI-e issue of not able to detect FC ASIC’s and LC ASIC’s and syslog for same.
+		Integrate with pcied process in PMON[sonic-platform-daemons/pcied at master · Azure/sonic-platform-daemons (github.com)]. Note: Current PCI daemon polling for pci devices is 60sec which is large poll interval. Does it need optimization ?
```
abdosi, line 7: should we make this change generic https://github.com/Azure/sonic-buildimage/blob/master/files/scripts/swss.sh#L193 ?

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
+Section 1 Requirements that are needed by default:-
+        1. On LC the reboot command should power-cycle the entire LC . Expectation is Peer node should detect link down when reboot is given on LC
+		2. On RP the reboot command should reboot the entire system (RP and LC). . Expectation is Peer node should detect link down when reboot is given on RP
+		3. Config shut/unshut of LC will be supported as per the Chassis-d design.
```
abdosi, line 4: not all platforms can support LC shut as there might not be power control on LC

abdosi: @abdosi will update the point `from supervisor`.

doc/pmon/pmon-chassis-requirements.md
abdosi: @abdosi remove `power-cycle` in the point

doc/pmon/pmon-chassis-requirements.md
abdosi: shut: power shut for card (platform dependent)
unshut: bring power back (platform dependent)

abdosi: reboot-cause shows what on LC when above option are invoke on supervisor.

abdosi: `config chassis startup <module-name>`
`config chassis shutdown <module-name>`
`reboot <module-name>`

module-name here are are LC. FC are not in discussion here.

shyam77git: > reboot <module-name>
similar to this, there should be 'shutdown <module-name> CLI to shutdown specified module (LC/FC).

>config chassis startup <module-name>
>config chassis shutdown <module-name>
intent/goal of these commands  is to do config shutdown or startup (reload/bring-up) of specified module.
wondering as to why 'chassis' keyword is there in these commands?

doc/pmon/pmon-chassis-requirements.md
shyam77git: > if Line-card can not detect sup going down then sup after comin-up broadcast all LC's to self-reboot

This might have an issue! There are chassis/platform whereby LC comes up on their own the moment power is turned-on to the chassis. So, Supervisor asking for its boot-up may interrupt already booting-up LC in a chassis power-on/reload scenarios. Suggest to have Supervisor 'booting up' workflow to be same regardless of SUP-only reload/boot-up or chassis reload.

doc/pmon/pmon-chassis-requirements.md
gechiang: > 

In my opinion when a FC goes bad/down, expect syslog be generated.  For system that has extra FC redundancy I think losing one FC does not impact the overall operation of the chassis. So in this case the syslog should cause an alarm to allow schedule maintenance service to replace the FC. If loosing another FC or chassis has no FC redundancy then the lost of FC syslog should clearly indicate it is running in degraded mode with expected traffic impact.  This syslog should cause alarm service to detect and trigger mitigation steps (whether be Admin user intervention or automation to start isolate it). The chassis itself can not tell if there are redundancy built into the involved network and if trigger self isolation it might causes more customer impact...  This is just my own opinion...

doc/pmon/pmon-chassis-requirements.md
abdosi: Summary based of internal discussion:
1. If Supervisor goes down unplanned/ungracefully then there is no need to reboot LC or any other action.
2. In above scenario LC should be generating syslog complaining about Supervisor not being reachable (Eg: PMON trying to access Chassis DB to push the data)
3. Above syslog can be used by Alerting logic and necessary action can be taken form LC/Chassis perspective like doing `config reload` on LC.

doc/pmon/pmon-chassis-requirements.md
abdosi: this can cause of delay of overall SW initialization.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		4. Generate syslog for all the critical events and share the threshold (for appropriate/needed components)  in documents and recommended for given threshold range.  Expectation is we will bind syslog to our Alert Orchestration system and perform recommnded action based on the documents.
+		5. PCI-e issue of not able to detect FC ASIC’s and LC ASIC’s and syslog for same.
+		Integrate with pcied process in PMON[sonic-platform-daemons/pcied at master · Azure/sonic-platform-daemons (github.com)]. Note: Current PCI daemon polling for pci devices is 60sec which is large poll interval. Does it need optimization ?
+		6. Boot-up failure Handling. Need to see the SONiC behaviour from system perspective/docker status/syslog getting generated with required/correct information
+		7. HW-Watchdog adhering to current SONiC behavior. Start before reboot and explicitly disabled post reboot by SONiC (This means SONiC is booted up and Services are fine)
```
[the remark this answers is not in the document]
abdosi: Watchdog Scheme will need enhancement in case if we want to detect CPU getting stuck in running state and given platform/vendor has HW-watchdog that can take recovery action in such cases.

doc/pmon/pmon-chassis-requirements.md
shyam77git: > Generate syslog for all the critical events and share the threshold (for appropriate/needed components)  in documents and recommended for given threshold range.  Expectation is we will bind syslog to our Alert Orchestration system and perform recommnded action based on the documents.

Per today's chassis workgroup sync-up, can we enhance this point to highlight following:
a) For now (near-term solution): External controller would take the recommended action (based on document) in real-time
b) Enhancing it to have a system-driven solution: I suggested having a platform-supplied policy (look-up) file of events and actions. On receiving an event (syslog), SONiC (LC, RP) or Ext Controller to perform a lookup on this policy file and take recommended action.

doc/pmon/pmon-chassis-requirements.md
kartik-arista: Regarding (1),  "there is no need to reboot LC or any other action"
There are a number of concerns with letting linecards run headless while a supervisor reboots
- There will be a long window of time where no state exchange between linecards will be possible due to lack of chassis_db. 
- The software on the linecard now has to deal with live disconnection/reconnection of connectivity  to chassis_db or other supervisor
- There's no real stated benefit motivating this propsal. A strong working assumption behind the chassis architecture conception for the current set of use cases is that there is enough redundancy in the network to easily tolerate the entire chassis going down and coming back, especially when it is a rare event like an unplanned supervisor rebbot.
- 
 
In general, it is best to keep failure handling simple and predictable, and avoid divergent flows across different scenarios. So unless there is a specific concrete problem that is solved by keeping the linecards, it would be strongly preferable to always reboot linecards when a supervisor reboots.

At the very least, the headless operation should be made optional and not mandated.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		8. chassisd daemon support on both LC and RP with all fields of table "CHASSIS_MODULE_TABLE|xxxx” correctly populated
+		9. chassisd daemon support populating fields in table "CHASSIS_ASIC_TABLE|xxx", this is used to start swss/syncd in SUP when FABRIC ASIC is ready.
+        10. Slot Nummber in "CHASSIS_MODULE_TABLE|xxxx” need not be unique ? Slot Number is based on physcial layout (Ex: LC can be back facing and can have 0..n and  FC can be Front facing and be 0.n). chassisd can support this model ?
```
abdosi, line 12: 1. Physcial slot Id need to be unique and can use sticker/label name based on given platform vendor. Use Case: Technician to identify the given Card based on Visual Inspection

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		10. psud power algorithm on supervisor as specified in chassis design document
+		11. PSU LED Status  in the show command of supervisor
```
abdosi, line 14: Ref for Point10: https://github.com/Azure/SONiC/blob/master/doc/pmon/pmon-chassis-design.md  please check if there is Platform API for setting Master LED for PSU.  API is there: https://github.com/Azure/sonic-platform-common/blob/master/sonic_platform_base/psu_base.py#L226

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		12. TEMPERATURE_INFO table update into Chassis State DB from both Supervisor and LC. Local TEMPERATURE_INFO is also available in LC STATE_DB.
+		13. Fan speed algorithm on supervior as specified in chassis design document
+		14. FAN LED Status in the show command of supervisor
```
abdosi, line 17: fan tray led display enhancement.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		15. reboot-cause reason and history is working fine for both RP and LC
+		16. show commands for mid-plane switch as per Chassis Design Document. Add namespace parameter support for "show chassis midplane-status" command.
```
abdosi, line 19: Need to check. `show ip interface -n <asic> -d all` should be displaying it.

doc/pmon/pmon-chassis-requirements.md
abdosi: Option 1 (Preferable)
If Supervisor goes down (headless operation) then LC should go down

Option 2
IF LC are running in headless mode the LC should be able to send syslog asap (mid-plane connectivity should get restored) so that LC can communicate the error to External Management.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		10. psud power algorithm on supervisor as specified in chassis design document
+		11. PSU LED Status  in the show command of supervisor
+		12. TEMPERATURE_INFO table update into Chassis State DB from both Supervisor and LC. Local TEMPERATURE_INFO is also available in LC STATE_DB.
+		13. Fan speed algorithm on supervior as specified in chassis design document
+		14. FAN LED Status in the show command of supervisor
+		15. reboot-cause reason and history is working fine for both RP and LC
+		16. show commands for mid-plane switch as per Chassis Design Document. Add namespace parameter support for "show chassis midplane-status" command.  
+	 
+2. Section2: General Chassis Enhancements that are needed:-
+		1. LC/FC Fabric Link down Handling
```
abdosi, line 22: Data-path component. Not in scope of platform

doc/pmon/pmon-chassis-requirements.md
abdosi: Document can also provide how many FC'S ASIC should be there to support X FC and Y LC Scenario. Each  Platform vendor need to provide this matrix.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		2. Module/Chassis/Board LED’s .  Need general infra enhancement of led daemon and show commands
+		3. LC/FC  operation status detection quicker using (get_change_event() notification handling to detect async card up/down events) rather than using current Polling Interval of 10 sec
+		4. Generic console for LC using . Possible using this: https://github.com/Azure/SONiC/blob/master/doc/console/SONiC-Console-Switch-High-Level-Design.md ?
```
abdosi, line 25: need more analysis from platform vendor's and will need revisit.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		5. Process for RMA the card (Fabric/LC). This is just a discussion to document correct process for doing so.
```
abdosi, line 26: Platform vendor recommendation/guidance needed here.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		6. Monit check on the supervisor to check if the LCs are  reachable. This is to alert if the linecard is down. Do we need Monit here or use above 10 sec polling ?
```
abdosi, line 27: need sonic enhancement for monit. From Platform side requirement is to have API to provide midplane internal IP's which can be consumed by monit (eg: by doing ping to check reachability). @anamehra can you provide API link

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		7. Handling of parallel reboot of linecard and supervisor. This should not result in the chassis/linecard to go down or unreachable. (Mention by Arvind) . If we follow Section 1 Point 2 this should           be handled ?
```
abdosi, line 28: need to add more test case to cover different scenarios here.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+	 
+2. Section2: General Chassis Enhancements that are needed:-
+		1. LC/FC Fabric Link down Handling
+		2. Module/Chassis/Board LED’s .  Need general infra enhancement of led daemon and show commands
+		3. LC/FC  operation status detection quicker using (get_change_event() notification handling to detect async card up/down events) rather than using current Polling Interval of 10 sec
+		4. Generic console for LC using . Possible using this: https://github.com/Azure/SONiC/blob/master/doc/console/SONiC-Console-Switch-High-Level-Design.md ?
+		5. Process for RMA the card (Fabric/LC). This is just a discussion to document correct process for doing so.
+		6. Monit check on the supervisor to check if the LCs are  reachable. This is to alert if the linecard is down. Do we need Monit here or use above 10 sec polling ?
+		7. Handling of parallel reboot of linecard and supervisor. This should not result in the chassis/linecard to go down or unreachable. (Mention by Arvind) . If we follow Section 1 Point 2 this should           be handled ? 
+		8. Mechanism to recover an down/unreachable linecard without power-cycle or reboot of the whole chassis.
```
abdosi, line 29: config chassis startup <module-name>  ==> Power cycle if Platform can do
config chassis shutdown <module-name> ===> Power Cycle  if Platform can do
reboot <module-name> ===> Platform that can do power-cycle of LC and CPU reset/reboot.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		9. Enhance "Show chassis module status" command for linecard should display hostname iso of generic names like LINECARD1
+		10. Support "show system-health detail/monitor-list/summary" commands in RP/LC
```
abdosi, line 31: Ref: https://github.com/Azure/SONiC/blob/e1744f1ff05916d2b61b429f67e569fb3e29d0a6/doc/system_health_monitoring/system-health-HLD.md

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+	 
+3. Section3 : Enhancements based on Significat Design Changes 
+		1. Auto Handling by Platfrom SW to reboot/shutdown the HW Component when detecting the critical Fault’s.
+		2. Temperature Measuring Category Enhancements. More Granular and Increase Polling Interval for same. Also show command optimize not dump all sesors and filter based on location
```
abdosi, line 35: Need Sonic Design Document.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+		3. Move Voltage and Current sensors support from existing sensorsd/libsensors model to PMON/ thermalCtld model This provide Ability/mechanism in SONiC NOS to poll for board’s Voltage and Current sensors (from platform) for power alogorithm.
```
abdosi, line 36: Need SONiC Design Document. Cisco can propose something on this.

doc/pmon/pmon-chassis-requirements.md
```diff
@@ -0,0 +1,37 @@
...
+        4. Midplane Switch Counters (Debugging) /Modifying QOS Properties if needed (Performance)
```
abdosi, line 37: each platform can provide some document to debug the midplane drop and any optimization that we need to do.

doc/pmon/pmon-chassis-requirements.md
shyam77git: For clarity to platforms/vendors, though these commands are under "config" but they are only executed (and not saved) until config save is issued.

doc/pmon/pmon-chassis-requirements.md
sanmalho-git: FYI - this change requires changing the get_slot (in module.py) and get_supervisor_slot and get_my_slot (in chassis.py) PMON API's to return a string instead of an int.

However, in platform api sonic-mgmt tests this field is expected to be an int (example: https://github.com/Azure/sonic-mgmt/blob/master/tests/platform_tests/api/test_module.py#L334)

Do we modify the tests to be either int or str for now to allow backward compatibility till all vendors implement this enhancement?

rlhui (approved):

PR #949, opened 2022-02-25
Installation and Testing of SONiC on Edgecore AS7716-24SC

ham-xa:
We have successfully installed and configured IP addresses on Edgecore AS7716-24SC aka Cassini.


lihuay:
@ham-xa : can you please use the LFX EasyCLA system instead of the pdf doc attached.

ham-xa:
> @ham-xa : can you please use the LFX EasyCLA system instead of the pdf doc attached.

PR #954, opened 2022-03-01
Events producer

renukamanavalan:
SONiC to stream events occurring in SONiC switch (_e.g BGP flap, process not running, ..._) in real time via gNMI subscription model. 
The events are defined using versioned YANG schema.
For more details, please refer the doc.


doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+- In this mode, there will be plugin processes running in host & in containers as one instance per process.
+- Provide the regex patterns to use for matching events as i/p to the plugin (*list of patterns for a process*).
+- Each plugin scans messasges **only** from a *single* process and matches only against patterns for that process.
+- For messages that match a pattern, retrieve parameters of interest and fire event using event generator.
+- The rsyslog plugin binary, which does the parsing & reporting is a single binary in host, shared/used by all plugins.
+- The rsyslog plugin binary being under host control, ensures a single/unified behavior across all.
+- The unit tests can use hardcoded log messages to validate regex.
+
+
+![image](https://user-images.githubusercontent.com/47282725/157343412-6c4a6519-c27b-459b-896b-7875d8f952b8.png)
```
zbud-msft, line 144: Hi Renuka,

On 121 and 124, you mention a watcher and a scanner. Does this just refer to the part of the plugin that is running in the host or container that will use regex parsing on the syslogs to create/match an event, as it seems that the plugin will also be responsible for sending the event to the UDP server?

renukamanavalan: Yes watcher/scanner is part of the plugin that looks for every message and match. I just updated the doc s/scanner/watcher.
This plugin, upon match will call the reporter, which will do the send.
In cases, where we can update code, the code could directly call the reporter.

zbud-msft (dismissed):

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+The external tools that monitor system health often use syslog messages to look for events that need alert raised.
+The syslog messages being text string could potentially **change** across releases. Some log messages could get split into multiple different ones.
+This poses a challenge to the external tools as they are forced to adapt for multiple different versions and parse a message based on the OS version of the producer.
+This would become even more challenging, when we upgrade individual container images via kubernetes/app-package-manager.
+
+The tools that use syslog messages face higher latency, as they have to wait for syslog messages to arrive to an external repository. 
+This latency could run in the order of minutes.
+
+
+![image](https://user-images.githubusercontent.com/47282725/156947460-66d08b3d-c981-4413-b0d5-232643dfba01.png)
```
qiluo-msft, line 23: >image [](http://example.com/codeflow?start=2&length=5)

In this diagram, what is "rsyslog server"? Is it the one running inside SONiC host?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+
+
+![image](https://user-images.githubusercontent.com/47282725/159061158-30ff3c8a-5fc0-4af2-8822-6bfc67b2329c.png)
+
+
+
+## Requirements
+### Events
+1. Events are defined with schema.
+2. Every event is identified by tag, which is unique within a process with zero or more event specific parameters.
```
qiluo-msft, line 33: >tag, which is unique [](http://example.com/codeflow?start=32&length=20)

In the above sample, "admin_down" as a tag seems not a unique string. Do you mean (tag, program) pair as unique?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+This would become even more challenging, when we upgrade individual container images via kubernetes/app-package-manager.
+
+The tools that use syslog messages face higher latency, as they have to wait for syslog messages to arrive to an external repository. 
+This latency could run in the order of minutes.
+
+
+![image](https://user-images.githubusercontent.com/47282725/156947460-66d08b3d-c981-4413-b0d5-232643dfba01.png)
+
+
+![image](https://user-images.githubusercontent.com/47282725/159061158-30ff3c8a-5fc0-4af2-8822-6bfc67b2329c.png)
```
qiluo-msft, line 26: >image [](http://example.com/codeflow?start=2&length=5)

In the example, I think timestamp is missing in each event json string.

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+
+
+
+## Requirements
+### Events
+1. Events are defined with schema.
```
qiluo-msft, line 32: >schema [](http://example.com/codeflow?start=27&length=6)

Could you give concrete design of the schema?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+2. Every event is identified by tag, which is unique within a process with zero or more event specific parameters.
+3. Events are static (*don't change*) across releases, but can be deprecated in newer releases.
+4. Every event is described in YANG schema.
+5. YANG schema files for all events are available in a single location for NB clients to refer in Switch.
```
qiluo-msft, line 36: >all events [](http://example.com/codeflow?start=25&length=10)

It's not easy to implement for "all events". Do we have a minimal events which will plan to implement first?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Event detection
+1. The method of detection can be any.
+2. This can vary across events.
+3. Syslog messages might be used to detect an event or some custom queries or code-update to report events directly or ...
```
qiluo-msft, line 41: >code-update to report events directly [](http://example.com/codeflow?start=78&length=37)

Could you explain "code-update to report events directly"?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+4. There can be multiple event detectors running under different scopes, concurrently.
```
qiluo-msft, line 42: >scopes [](http://example.com/codeflow?start=65&length=6)

What is the definition of "scope"?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Event reporting
+1. Event detectors stream the events with structured data.
+2. The streaming supports one or more local listeners to receive streams from multiple detectors (many-to-many).
+3. The structured data is per YANG definition.
+4. RFE: A listener to update redis/any other persistence destination with current event status.
+
+### Event exporter
+1. Telemetry container receive all the events reported from all the event detectors.
```
qiluo-msft, line 51: >receive [](http://example.com/codeflow?start=23&length=7)

How does telemetry receive? What is the communication channel? How do you make it working with containers?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Event reliability
+There are two kinds of reliability.
+1. Events are not modified across releases (*except deprecation*). We may allow addition of new params, as long as they don't affect existing params.
+2. Events are verified to fire as expected in every image release. 
+3. Ensure the perf goals are met.
+
+#### Protection:
+
+1. The unit tests are required to ensure the immutability of event definition across releases. This would help block a PR that mutates a definition.
```
qiluo-msft, line 63: >unit tests [](http://example.com/codeflow?start=7&length=10)

Which component's unit test?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+4. A separate stress test is required to ensure the performance rate of 10K events/sec and 99.5% of reliability end-to-end.
+</br>
+
+## Design
+
+### YANG schema
+1. Schema defines the description of the event, its unique tag and all the possible parameters.
+2. Schema can be maintained in multiple files, preferably one per process/continer/host.
+3. All the schema files are copied into one single location (e.g. /usr/shared/sonic/events).
+4. The schema for processes running in host are copied into this location at image creation/install time.
```
qiluo-msft, line 75: >creation/install [](http://example.com/codeflow?start=83&length=16)

When did them copied? compilation time or installation time?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+#### A sample Defintion
+A sample:
+```
+module events-bgp {
+    . . .
+    container bgp_status {
+        list event_list {
+            key "type ip"
+            
+            leaf type {
```
qiluo-msft, line 88: >type [](http://example.com/codeflow?start=17&length=4)

There is no 'tag' in the example.

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+
+Here you have code that sends the log and a watcher who has the regex pattern for that log message to match. Anytime the log messsage is changed the pattern has to be updated for the event to fire consistently across releases.
+
+Though this sounds like a redundant/roundabout way, this helps as below.
+- For III party code, not owned by SONiC, this is an acceptable solution
+- For code that SONiC own, a code update would take in the order of months to reach thousands of switches, but this approach can be applied instantly to every switch
+- A new event addition could be as simple as copying couple of small files to a switch and a rsyslog/container restart.
+
+##### Design at high level
+- Configure a rsyslog plugin as per process with rsyslog.d.
```
qiluo-msft, line 132: >per process [](http://example.com/codeflow?start=32&length=11)

How about one plugin per container? If 'per process', the syslog will be duplicated for multiple plugins.

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+- For logs raised by host processes, configure this plugin at host.
+- For logs raised by processes inside the container, configure for rsyslog.d running inside the container.
+- In this mode, there will be plugin processes running in host & in containers as one instance per process.
+- Provide the regex patterns to use for matching events as i/p to the plugin (*list of patterns for a process*).
+- Each plugin scans messasges **only** from a *single* process and matches only against patterns for that process.
+- For messages that match a pattern, retrieve parameters of interest and fire event using event reporter.
+- The rsyslog plugin binary, which does the parsing & reporting is a single binary in host, shared/used by all plugins.
+- The rsyslog plugin binary being under host control, ensures a single/unified behavior across all.
```
qiluo-msft, line 140: >under host control [](http://example.com/codeflow?start=34&length=18)

Could you explain "under host control" ?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+![image](https://user-images.githubusercontent.com/47282725/157343412-6c4a6519-c27b-459b-896b-7875d8f952b8.png)
+
+##### Pros & Cons
+
+###### Pro
+1) This support is external to app, hence adapts well with III party applications, bgp, teamd, ... 
+2) This feature can be added to released builds too, as all it takes is to copy two files into each container and restart rsyslogd in the container
+3) The regex for parsing being local to container, it supports any container upgrade transaparently.
+4) The rsyslog plugin binary is maintained by host, hence provide a overall control across all containers.
+5) The message parsing load is distributed as per container. Within a container parsing is done at the granularity of per process with no extra cost as rsyslogd already pre-parsed it per-process, *always*
```
qiluo-msft, line 153: >rsyslogd already pre-parsed it per-process [](http://example.com/codeflow?start=152&length=42)

Do you have reference link to explain "rsyslogd already pre-parsed it per-process"?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+
+###### con:
+1) Two step process for devs, as for each new/updated log message ***for an event***, add/update regex as needed. Updating code directly to raise the event will help avoid this.
+2) The rsyslogd is *required*. It should be treated as critical process in each container.
```
qiluo-msft, line 157: >2 [](http://example.com/codeflow?start=0&length=1)

Another CON is: it is not possible to cross reference between syslog from different containers. For example, we want to alert if container A has syslog A, and immediately container B has syslog B.

qiluo-msft, line 157: >critical process [](http://example.com/codeflow?start=55&length=16)

This is a good point.
How to make sure it is "critical process"?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+Here you have code that sends the log and a watcher who has the regex pattern for that log message to match. Anytime the log messsage is changed the pattern has to be updated for the event to fire consistently across releases.
+
+Though this sounds like a redundant/roundabout way, this helps as below.
+- For III party code, not owned by SONiC, this is an acceptable solution
+- For code that SONiC own, a code update would take in the order of months to reach thousands of switches, but this approach can be applied instantly to every switch
+- A new event addition could be as simple as copying couple of small files to a switch and a rsyslog/container restart.
+
+##### Design at high level
+- Configure a rsyslog plugin as per process with rsyslog.d.
+- For logs raised by host processes, configure this plugin at host.
```
qiluo-msft, line 133: >logs raised by host processes [](http://example.com/codeflow?start=6&length=29)

The host rsyslog is special because it collects the host process syslog message, and also collects relayed syslog message from every container. If you have a plugin on host, it actually parsed all the SONiC syslog. It is expected? If this is true, maybe a centralized parsing is better?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+- The event detectors could be running in host and/or some/all containers.
+- Support multiple local clients to be able to receive the events concurrently.
+- Each local client should be able to receive updates from all event detectors w/o being aware of all of the sources.
+- The local clients could be operating at different speeds.
+- The clients may come and go.
+- There may not be any local client to receive the events.
+- Events reporting should not be blocked by any back pressure from any local client.
+
+#### Design
+- Event detectors send UDP messages to a multicast group
```
qiluo-msft, line 172: >multicast group [](http://example.com/codeflow?start=41&length=15)

Could you explain "send UDP messages to a multicast group" in details?

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+
+##### con
+- Messages could get lost, if the client is slow. It is the client's responsibility to ensure no loss.
+- Clients that are slow by design (*like redis-DB updater*), could have a dedicated thread to receive all events and cache the latest. The main thread could be slow, it could miss updates, but it would use/record the latest.
+
+
+### Event exporting
+
+#### requirements
+- Telemetryt container receives all locally raised events.
```
qiluo-msft, line 189: >Telemetryt [](http://example.com/codeflow?start=2&length=10)

typo

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+- Telemetry container supports exporting all the locally raised events to one or more external clients.
+- RFE: When restarted, ensure to provide the latest on all events that were missed during downtime.
```
qiluo-msft, line 191: >RFE [](http://example.com/codeflow?start=2&length=3)

Are you going to resolve all 'RFE' lines in this design PR?

qiluo-msft, line 191: >ensure [](http://example.com/codeflow?start=23&length=6)

How to ensure? I guess you need a storage.

doc/event-alarm-framework/events-producer.md
qiluo-msft, line 26: >image [](http://example.com/codeflow?start=2&length=5)

In the 'Desired' diagram, I don't see a complete design. Such as Event detection, Event reporting, Event exporter, YANG schema are not covered.

doc/event-alarm-framework/events-producer.md
renukamanavalan: Yes. Updated image to be explicit.

doc/event-alarm-framework/events-producer.md
renukamanavalan: Current telemetry code that does streaming from redis,  puts its own timestamp before sending out. As receiver is working at a high speed, this would suit and as well cut down the payload size from event reporter to receiver.

But if your telemetry client design warrants, the event reporter could send the timestamp.

renukamanavalan: This diagram attempts to depict the base concept at high level.

Processes exporting structured data to receivers directly, which gets streamed out.

Event detection is not by any single mode. Orchagent may directly send out, whereas in case of BGP rsylog-plugin may do it, all the monit scripts will send directly  and ...  The detection is per process and it could vary, which is implementation details.

Reporting would be an API provided, embedded into the process. Not any separate component to depict.

Event exporter -- This is telemetry client -- The HLD needs to be updated with this detail.

doc/event-alarm-framework/events-producer.md
renukamanavalan: The design: YANG schema
Provided a full schema for an event.

doc/event-alarm-framework/events-producer.md
renukamanavalan: Yes

doc/event-alarm-framework/events-producer.md
renukamanavalan: We do. But that is implementation details. HLD is generic for any event. Sets the guidelines of how it is defined, reported, exported, ....

doc/event-alarm-framework/events-producer.md
renukamanavalan: The code we own, like orchagent, syncd, all the monit scripts and similar, could call the API/macro directly to report the event.

doc/event-alarm-framework/events-producer.md
renukamanavalan: Host & within container

doc/event-alarm-framework/events-producer.md
renukamanavalan: Telemetry client design is a TODO. Apparently Yong has it. That has to be added to this HLD, to make it complete. The design of the client would define the communication channel. The design would also spell out, how the telemetry container restart will handle the events missed during its downtime.

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+</br>
+
+## Design
+
+### YANG schema
+1. Schema defines the description of the event, its unique tag and all the possible parameters.
+2. Schema can be maintained in multiple files, preferably one per process/continer/host.
+3. All the schema files are copied into one single location (e.g. /usr/shared/sonic/events).
+4. The schema for processes running in host are copied into this location at image creation/install time.
+5. The schema for processes running inside the containers are held inside the containers and copied into the shared location on the first run. This allows for independent container upgrade scenarios.
```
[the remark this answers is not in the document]
renukamanavalan: The container run, will copy its schema files onto the shared location in the host.

doc/event-alarm-framework/events-producer.md
renukamanavalan: Corrected.

doc/event-alarm-framework/events-producer.md
```diff
@@ -0,0 +1,212 @@
...
+                type inet::date-and-time;
+                description "time of the event"
+            }
+        }
+    }
+``` 
+
+### Event detection
+The event detection could happen in many ways
+- Update the code to directly invoke Event reporter, which will stream it out.
```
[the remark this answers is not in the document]
renukamanavalan: The API design depends on telemetry client design.

doc/event-alarm-framework/events-producer.md
renukamanavalan: No. The rsyslog.d  will spawn one plugin per process and pass on only messages from that process to that plugin.

doc/event-alarm-framework/events-producer.md
renukamanavalan: No. We configure host rsyslog.d only for host processes and rest are not passed to any at host.

doc/event-alarm-framework/events-producer.md
renukamanavalan: The rsyslog-plugin binary is shared by all rsyslog based detectors. The detectors are running in host and multiple containers. If we do want a behavioral change, it is easy to accomplish.

doc/event-alarm-framework/events-producer.md
renukamanavalan: It is one of the properties for use as provided by rsyslog.d in its .conf.

doc/event-alarm-framework/events-producer.md
renukamanavalan: It is possible.
If your detection is based off of multiple processes from one/more containers, you would run the detector at host.

renukamanavalan: Implementation details.

doc/event-alarm-framework/events-producer.md
renukamanavalan: many to many. -- UDP multicast concept.
Again, if the telemetry client is designed to accept UDP stream.

doc/event-alarm-framework/events-producer.md
renukamanavalan: done

doc/event-alarm-framework/events-producer.md
renukamanavalan: This will be covered by telemetry client design.

renukamanavalan: This will be covered by telemetry client design.

renukamanavalan (commented):
Pending: Telemetry client design from Yong.

doc/event-alarm-framework/events-producer.md
zbud-msft: Can you clarify more on how this works? What changes would we have to make to create these plugins?

doc/event-alarm-framework/events-producer.md
zbud-msft: I'm a little confused about the plugin binary. How do the plugins communicate with this binary, is there anything that we would have to implement? If this binary does all the parsing & reporting does that mean this binary does the regex logic of creating events and sending them to the UDP listener? How would this binary talk to the UDP listener running in the telemetry container?

doc/event-alarm-framework/events-producer.md
zbud-msft: Can we add telemetry client design to the HLD?

doc/event-alarm-framework/events-producer.md
zbud-msft: Is the current scope of the project to only have one sender being the plugin binary in the host and 2 listeners being the telemetry container and the redis db container?

qiluo-msft: The timestamp ground truth is from syslog message, which is very important. Telemetry timestamp is different.

doc/event-alarm-framework/events-producer.md
qiluo-msft: From you answer of next comment, seem you mean container first run time, not "image creation/install time".

doc/event-alarm-framework/events-producer.md
qiluo-msft: This is not along with current situation. If you need configuring this way, please specify how to achieve in this design doc.

venkatmahalingam:
I was thinking about 2 approaches in the meeting today.

**1) Introduce event specific container and reference to the list used for the config-DB table.**
For example, in the port table YANG file, we can do the following,

https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-port.yang

           container PORT {    -> This is 1:1 representation of config-DB schema
			description "PORT part of config_db.json";

			list PORT_LIST {
                           key "name";
                           leaf name {
					type string {
						length 1..128;
					}
		           } 
                        ..............
                        }
            }
            **container PORT_STATE {  -> This container can be used for events, PORT_STATE table name is not good, we can use 
                                                         better name.
                config false;
                 list PORT_LIST {
                     key "name";
                   leaf name {
                                 type leafref {
                                      path "../../../PORT/PORT_LIST/name;
				}
                   leaf oper_status { .. }
           }**
  }

**2) Enhance the container e.g PORT defined already for the config-DB schema to include state information e.g port_state_info as well.**

   grouping port_state_info {
        <<<<< read-only attributes>>>>>>
        leaf oper_status { .. }
   }
   container PORT { 
			description "PORT part of config_db.json";

			list PORT_LIST {
                           key "name";
                           leaf name {
					type string {
						length 1..128;
					}
		           } 
                        ..............
                          uses port_state_info;
                        }
            }

renukamanavalan:
> I was thinking about 2 approaches in the meeting today. @renukamanavalan @bandaru-viswanath, @dgsudharsan @praveen-li we can discuss the pros and cons in the next meeting.
> 
> **1) Introduce event specific container and reference to the list used for the config-DB table.** For example, in the port table YANG file, we can do the following,
> 
> https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-port.yang
> 
> ```
>        container PORT {    -> This is 1:1 representation of config-DB schema
> 		description "PORT part of config_db.json";
> 
> 		list PORT_LIST {
>                        key "name";
>                        leaf name {
> 				type string {
> 					length 1..128;
> 				}
> 	           } 
>                     ..............
>                     }
>         }
>         **container PORT_STATE {  -> This container can be used for events, if PORT_STATE table name is not good, we can use 
>                                                      better name.
>             config false;
>              list PORT_LIST {
>                  key "name";
>                leaf name {
>                              type leafref {
>                                   path "../../../PORT/PORT_LIST/name;
> 			}
>                leaf oper_status { .. }
>        }**
> ```
> 
> }
> 
> **2) Enhance the container e.g PORT defined already for the config-DB schema to include state information e.g port_state_info as well.**
> 
> grouping port_state_info { <<<<< read-only attributes>>>>>> leaf oper_status { .. } } container PORT { description "PORT part of config_db.json";
> 
> ```
> 		list PORT_LIST {
>                        key "name";
>                        leaf name {
> 				type string {
> 					length 1..128;
> 				}
> 	           } 
>                     ..............
>                       uses port_state_info;
>                     }
>         }
> ```

Thanks Venkat!
Here is a 3rd option. 

```
module sonic-events-swss {
…
        container if-state { 
               leaf ifname {
                     type string;
                     description "Interface name";
               }
               type leafref {
                      path "/port:sonic-port/port:PORT/port:PORT_LIST/port:name";
               }

```

The following are the reasons behind.
1) The consumer of these events are not running inside the switch, but external and have no access to config.
2) The state reported by these events are *not* saved inside the switch and hence never accessed from within the switch.
3) As events are from switches running multiple different versions, we put a restriction that YANG events module ***should not*** break backward compatibility, but config schema does ***not*** have such restriction and no need either.
4) The consumer are looking for events across all switches for correlating & reporting and count of switches is often few tens of thousands in numbers. Also, the config among the switches could be using multiple different YANG versions, potentially.

renukamanavalan:
@venkatmahalingam, can we conclude on this discussion above?

venkatmahalingam:
> @venkatmahalingam, can we conclude on this discussion above?

A gNMI client could subscribe for events with optional filter on event source in streaming mode. Below shows the command & o/p for subscribing all, and receiving BGP events.

gnmic --target events --path "/events/" --mode STREAM --stream-mode ON_CHANGE

The instance data would indicate YANG module path & revision that is required for validation.

o/p
{
    "sonic-events-bgp:bgp-state": {
      "timestamp": "2022-08-17T02:39:21.286611Z",
      "ip": "100.126.188.90",
      "status": "down"
    }
}

Instead of doing the above generic events subscription, if we can follow either approach 1 or 2, it will be cleaner to subscribe particular table state information, right? in the future, we can enhance the subscription for config changes as well.

I agree that if it's just an event, we dont need to store it in the DB, we can use another event YANG but what we use is the state information (i.e interface admin/oper status, BGP session status..etc), also, we should support get operation as well on the state information, right?

renukamanavalan:
> ta would indica

1) An explicit decision is made to support only one gNMI client - Please refer the section "Event exporting".
    At the rate of 10k events/second, with offline cache support & missed count tracking, supporting multiple can be expensive.
    On the other hand, looking at the consumer requirements/design model, a client receiving at 10K/sec should be light weighted and end consumers are often looking at events across thousands of switches. Hence the receiving client, which  is 1:1 with SONiC switch is expected to dump the events into an external storage and the events consumers query/watch for updates at the storage. Often the consumers, need to collect info from multiple connected switches to make a call.

Hence gNMI client can only subscribe for /events which is all.

The "state" is not the *only* event reported on an i/f, e.g. PFC-Storm. With the absolute path provided as leafref, helps consumer identify object where needed.

venkatmahalingam:
> > ta would indica
> 
> 1. An explicit decision is made to support only one gNMI client - Please refer the section "Event exporting".
>    At the rate of 10k events/second, with offline cache support & missed count tracking, supporting multiple can be expensive.
>    On the other hand, looking at the consumer requirements/design model, a client receiving at 10K/sec should be light weighted and end consumers are often looking at events across thousands of switches. Hence the receiving client, which  is 1:1 with SONiC switch is expected to dump the events into an external storage and the events consumers query/watch for updates at the storage. Often the consumers, need to collect info from multiple connected switches to make a call.
> 
> Hence gNMI client can only subscribe for /events which is all.
> 
> The "state" is not the _only_ event reported on an i/f, e.g. PFC-Storm. With the absolute path provided as leafref, helps consumer identify object where needed.

I understand that state is not the only event being reported, I think, we can report non-state information (absolutely doesnt make sense to have it in DB) in event YANG model.

What's the design to handle both the cases 1) interface oper. status changes (stored in the STATE_DB) 2) PFC storm (not in DB)?

For the first case, enhance the existing YANG to have state information as well and for the second case, introduce new event specific YANGs and leafref to existing tables if applicable.

@bandaru-viswanath, @dgsudharsan any concerns?

renukamanavalan:
> > > ta would indica
> > 
> > 
> > 
> > 1. An explicit decision is made to support only one gNMI client - Please refer the section "Event exporting".
> >    At the rate of 10k events/second, with offline cache support & missed count tracking, supporting multiple can be expensive.
> >    On the other hand, looking at the consumer requirements/design model, a client receiving at 10K/sec should be light weighted and end consumers are often looking at events across thousands of switches. Hence the receiving client, which  is 1:1 with SONiC switch is expected to dump the events into an external storage and the events consumers query/watch for updates at the storage. Often the consumers, need to collect info from multiple connected switches to make a call.
> > 
> > Hence gNMI client can only subscribe for /events which is all.
> > The "state" is not the _only_ event reported on an i/f, e.g. PFC-Storm. With the absolute path provided as leafref, helps consumer identify object where needed.
> 
> I understand that state is not the only event being reported, I think, we can report non-state information (absolutely doesnt make sense to have it in DB) in event YANG model.
> 
> What's the design to handle both the cases 1) interface oper. status changes (stored in the STATE_DB) 2) PFC storm (not in DB)?
> 
> For the first case, enhance the existing YANG to have state information as well and for the second case, introduce new event specific YANGs and leafref to existing tables if applicable.
> 
> @renukamanavalan @bandaru-viswanath, @dgsudharsan any concerns on the above design?

We persist neither in DB.

bandaru-viswanath:
> > > ta would indica
> > 
> > 
> > 
> > 1. An explicit decision is made to support only one gNMI client - Please refer the section "Event exporting".
> >    At the rate of 10k events/second, with offline cache support & missed count tracking, supporting multiple can be expensive.
> >    On the other hand, looking at the consumer requirements/design model, a client receiving at 10K/sec should be light weighted and end consumers are often looking at events across thousands of switches. Hence the receiving client, which  is 1:1 with SONiC switch is expected to dump the events into an external storage and the events consumers query/watch for updates at the storage. Often the consumers, need to collect info from multiple connected switches to make a call.
> > 
> > Hence gNMI client can only subscribe for /events which is all.
> > The "state" is not the _only_ event reported on an i/f, e.g. PFC-Storm. With the absolute path provided as leafref, helps consumer identify object where needed.
> 
> I understand that state is not the only event being reported, I think, we can report non-state information (absolutely doesnt make sense to have it in DB) in event YANG model.
> 
> What's the design to handle both the cases 1) interface oper. status changes (stored in the STATE_DB) 2) PFC storm (not in DB)?
> 
> For the first case, enhance the existing YANG to have state information as well and for the second case, introduce new event specific YANGs and leafref to existing tables if applicable.
> 
> @renukamanavalan @bandaru-viswanath, @dgsudharsan any concerns on the above design?

Looks good to me @venkatmahalingam

zbud-msft (approved):

zbud-msft:
| Repo  | PR title | State |
| sonic-buildimage| [Event libswsscommon deps](https://github.com/sonic-net/sonic-buildimage/pull/11618)| GitHub issue/pull request detail | 
| sonic-buildimage| [Event libswsscommon deps](https://github.com/sonic-net/sonic-buildimage/pull/11639)| GitHub issue/pull request detail | 
| sonic-buildimage| [Streaming structured events implementation](https://github.com/sonic-net/sonic-buildimage/pull/11848)| GitHub issue/pull request detail | 
| sonic-buildimage| [Fix PR build failure](https://github.com/sonic-net/sonic-buildimage/pull/11973)| GitHub issue/pull request detail | 
| sonic-buildimage| [Add events to host and create rsyslog_plugin deb pkg](https://github.com/sonic-net/sonic-buildimage/pull/12059)| GitHub issue/pull request detail | 
| sonic-buildimage| [Add Structured Events w/ YANG Models](https://github.com/sonic-net/sonic-buildimage/pull/12270)| GitHub issue/pull request detail | 
| sonic-buildimage| [Add Yang Models for structured events](https://github.com/sonic-net/sonic-buildimage/pull/12290)| GitHub issue/pull request detail | 
| sonic-buildimage| [Add YANG model and unit tests for additional structured events](https://github.com/sonic-net/sonic-buildimage/pull/12554)| GitHub issue/pull request detail | 
| sonic-buildimage| [Publish additional events](https://github.com/sonic-net/sonic-buildimage/pull/12563)| GitHub issue/pull request detail | 
| sonic-buildimage| [Add rsyslog plugin regex for select operation failure](https://github.com/sonic-net/sonic-buildimage/pull/12659)| GitHub issue/pull request detail | 
| sonic-buildimage| [Add YANG model for alpm parity error](https://github.com/sonic-net/sonic-buildimage/pull/12687)| GitHub issue/pull request detail | 
| sonic-swss | [Publish identified events via structured-events channel](https://github.com/sonic-net/sonic-swss/pull/2446) | GitHub issue/pull request detail |
| sonic-gnmi | [Telemetry support for streaming events](https://github.com/sonic-net/sonic-gnmi/pull/13)  | GitHub issue/pull request detail |
| sonic-gnmi | [Events client: Ensure all go routines exit upon client disconnect.](https://github.com/sonic-net/sonic-gnmi/pull/41)  | GitHub issue/pull request detail |
| sonic-gnmi | [Structured events: Publish as JsonIetfVal instead of StringVal](https://github.com/sonic-net/sonic-gnmi/pull/43)  | GitHub issue/pull request detail |
| sonic-gnmi | [gnmi_cli - Tool update](https://github.com/sonic-net/sonic-gnmi/pull/44)  | GitHub issue/pull request detail |
| sonic-gnmi | [Streaming events URL support "not to use cache"](https://github.com/sonic-net/sonic-gnmi/pull/45)  | GitHub issue/pull request detail |
| sonic-utilities | [Event Counters CLI](https://github.com/sonic-net/sonic-utilities/pull/2449) | GitHub issue/pull request detail
| sonic-swss-common| [APIs to support streaming structured events](https://github.com/sonic-net/sonic-swss-common/pull/667) | GitHub issue/pull request detail |
| sonic-swss-common| [Events: APIs to set/get global options](https://github.com/sonic-net/sonic-swss-common/pull/672) | GitHub issue/pull request detail |

praveen-li:
Hi Zain

Nice report for all PR  under event producer.


Thanks,

Regards
Praveen

PR #982, opened 2022-04-14
[HLD] Add `SYSTEM_DEFAULTS` table in `config_db` to control the behavior of SONiC

bingwang-ms:
This PR is to introduce a new table `FLAGS` into `config_db` to control the behavior of SONiC.

Signed-off-by: bingwang <bingwang@microsoft.com>

### DIFF
--- a/doc/sonic-flags/control-sonic-behaviors-with-sonic-flags.md
+++ b/doc/sonic-flags/control-sonic-behaviors-with-sonic-flags.md
@@ -1,0 +1,108 @@
+# Control SONiC behaviors with FLAGS table
+
+## 1 Table of Content ###
+
+- [Revision](#11-revision)
+- [Scope](#2-scope)
+- [Definitions/Abbreviations](#3-definitionsabbreviations)
+- [Overview](#4-overview)
+- [Design](#5-design)
+- [Change required](#6-change-required)
+- [Test requirement](#7-test-requirement)
+
+
+### 1.1 Revision ###
+| Rev |     Date    |       Author       | Change Description                |
+|:---:|:-----------:|:------------------:|-----------------------------------|
+| 0.1 |             | Bing Wang   | Initial version                   |
+
+
+## 2 Scope ##
+
+This document covers high level design of `FLAGS` table in SONiC.
+
+## 3 Definitions/Abbreviations ##
+
+
+| Term | Meaning |
+|:--------:|:---------------------------------------------:|
+|  |  |
+
+
+## 4 Overview
+
+A number of flags are required to turn on/off certain feature or control the behaviors of various features in SONiC. Currently, these flags are put into `DEVICE_METADATA` table.
+
+```
+ "DEVICE_METADATA": {
+        "localhost": {
+            "default_bgp_status": "down",
+            "default_pfcwd_status": "enable",
+            "synchronous_mode": "enable",
+            "dhcp_server": "enable"
+        }
+    }
+```
+As a result, the `DEVICE_METADATA` table is inflating rapidly as we are having more and more flags, although these flags seem not to be categorized into `DEVICE_METADATA` 
+
+To have a better management of the flags, a new table `FLAGS` is introduced in this design. 
+
+## 5 Design ##
+
+### 5.1 DB Schema
+
+A new table `FLAGS` is added into config_db.
+```
+    key             = FLAGS
+
+    ;field          = value
+    FLAG_NAME       = 1*255VCHAR ; FLAG_NAME must be unique, the value is a string, which can be 'enable'/'disable', 'down'/'up' or any string. 
+```
+Below is a sample of `FLAGS` table
+
+```
+"FLAGS": {
+        "default_bgp_status": "down",
+        "default_pfcwd_status": "enable",
+        "synchronous_mode": "enable",
+        "dhcp_server": "enable"
+    }
+```
+
+### 5.2 How to update flags in `FLAGS` table
+
+#### 5.2.1 Set default value with `init_cfg.json`
+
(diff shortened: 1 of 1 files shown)


doc/sonic-flags/control-sonic-behaviors-with-sonic-flags.md
```diff
@@ -0,0 +1,108 @@
...
+### 6.1 Template update
+1. Templates that generate default values in `DEVICE_METADATA|localhost` table are required to be updated. The generated flags will be put into `FLAGS` table now.
+2. Templates that depend on `DEVICE_METADATA|localhost` table are required to be updated.
+
+### 6.2 Yang model update
+A new Yang model is to be added to restrict the valid flags in `FLAGS` table. The existing entries for flags in current [sonic-device_metadata.yang](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-device_metadata.yang) are to be removed.
+
+### 6.3 Code change
+1. Update `db_migrator.py` to migrate flags from `DEVICE_METADATA|localhost` table to `FLAGS` table.
+2. Update the code of component to subscribe `FLAGS` table to watch the update.
```
yxieca, line 103: db migrator change would be required for warm reboot upgrade.

doc/sonic-flags/control-sonic-behaviors-with-sonic-flags.md
```diff
@@ -0,0 +1,108 @@
...
+#### 5.2.1 Set default value with `init_cfg.json`
+
+The default vaule of flags in `FLAGS` table can be set in `init_cfg.json` and loaded into db at system startup. These flags are usually set at image being build, and are unlikely to change at runtime.
+
+#### 5.2.2 Parse from `minigraph.xml` when loading minigraph
+
+For the flags that can be changed by reconfiguration, we can update entries in `minigraph.xml`, and parse the new values in to config_db with minigraph parser at reloading minigraph.
+
+#### 5.2.3 Update value directly in db memory
+For some behavior change, we may don't have to interrupt dataplane. To support controlling SONiC behavior on-the-fly, we can update the value of flags in memory with tools like `sonic-cfggen`, `configlet` or `config apply-patch`.
```
prsunny, line 83: Does init_cfg overwrites the one that is already saved and restored from config_db by the user? How is it handled? Can you provide some details for this case?

doc/sonic-flags/control-sonic-behaviors-with-sonic-flags.md
bingwang-ms: That's a good catch. I think same migration operation is supposed to be added to `finalize-warmboot.sh`. Is that a correct change?

doc/sonic-flags/control-sonic-behaviors-with-sonic-flags.md
bingwang-ms: The value in `config_db.json` will not be rewritten by `init_cfg.json` as `config_db.json` is loaded after `init_cfg.json`. I added clarification in the doc, please help have a review. Thanks

doc/sonic-flags/control-sonic-behaviors-with-sonic-flags.md
```diff
@@ -0,0 +1,120 @@
...
+### 6.1 Template update
+1. Templates that generate default values in `DEVICE_METADATA|localhost` table are required to be updated. The generated flags will be put into `FLAGS` table now.
+2. Templates that depend on `DEVICE_METADATA|localhost` table are required to be updated.
+
+### 6.2 Yang model update
+A new Yang model is to be added to restrict the valid flags in `FLAGS` table. The existing entries for flags in current [sonic-device_metadata.yang](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-device_metadata.yang) are to be removed.
+
+### 6.3 Code change
+1. Update `db_migrator.py` to migrate flags from `DEVICE_METADATA|localhost` table to `FLAGS` table.
+2. Update the code of component to subscribe `FLAGS` table to watch the update.
```
neethajohn, line 115: What are the current flags and what components need to be updated to subscribe to this new table? Can you provide more details

bingwang-ms: Updated. Thanks

doc/sonic-flags/control-sonic-behaviors-with-sonic-flags.md
```diff
@@ -0,0 +1,129 @@
...
+```
+    key             = FLAGS
+
+    ;field          = value
+    FLAG_NAME       = 1*255VCHAR ; FLAG_NAME must be unique, the value is a string, which can be 'enable'/'disable', 'down'/'up' or any string. 
+```
+Below is a sample of `FLAGS` table
+
+```
+"FLAGS": {
```
prsunny, line 64: FLAG seems to be too generic. May I suggest a rename to `SYSTEM_DEFAULTS`.

bingwang-ms: Actually, it's not `SYSTEM_DEFAULTS` sometimes. For example, the FLAGS for pcbb is `enable` by default, but somehow we may need to change the value to `disable` for some reason, then the table name `SYSTEM_DEFAULTS` seems wired. 
Any more suggestions for a more beautiful table name? I also think the name `FLAGS` is too generic.

prsunny: why PCBB is enabled by default? In any case, this is loaded during the bootup and probably changed via feature or CLI etc. I see this as an extension of init_cfg.json and DEVICE_METADATA. Prefer to convey the meaning as SYSTEM_DEFAULT.

prsunny (approved):

PR #989, opened 2022-04-22
[POC] swsscommon default value from yang model

liuh-80:


stephenxs:
I was just wondering whether it's possible to introduce a way to "shadow" the fields that existed in `STATIC_CONFIG_DB` but are removed from `CONFIG_DB` by the user. By doing so, we do not have to rule out a table from `STATIC_CONFIG_DB` only because the table can be modified by a user.
I think it's better to decide whether a table is eligible to be in `STATIC_CONFIG_DB` based on the meaning of the table instead of the possibility to be modified.

liuh-80:
> I was just wondering whether it's possible to introduce a way to "shadow" the fields that existed in `STATIC_CONFIG_DB` but are removed from `CONFIG_DB` by the user. Maybe we can refer to how it is handled in `APPL_DB`: introduce a set to store the items, and fields that are deleted by the user. By doing so, we do not have to rule out a table from `STATIC_CONFIG_DB` only because the table can be modified by a user. I think it's better to decide whether a table is eligible to be in `STATIC_CONFIG_DB` based on the meaning of the table instead of the possibility to be modified.

Currently there are 2 scenario about delete:
1. Xoff can be deleted/reset by user.
2. Mellanox may have multiple pool config in J2 template, and will delete pool config.

For 1, Delete means change database schema, maybe we should use Xoff = 0 or Xoff=-1 instead of delete.

For 2, still not quite undertsand the scenario.

Will setup a offline discussion for this.

stephenxs:
> > I was just wondering whether it's possible to introduce a way to "shadow" the fields that existed in `STATIC_CONFIG_DB` but are removed from `CONFIG_DB` by the user. Maybe we can refer to how it is handled in `APPL_DB`: introduce a set to store the items, and fields that are deleted by the user. By doing so, we do not have to rule out a table from `STATIC_CONFIG_DB` only because the table can be modified by a user. I think it's better to decide whether a table is eligible to be in `STATIC_CONFIG_DB` based on the meaning of the table instead of the possibility to be modified.
> 
> Currently there are 2 scenario about delete:
> 
> 1. Xoff can be deleted/reset by user.
> 2. Mellanox may have multiple pool config in J2 template, and will delete pool config.
> 
> For 1, Delete means change database schema, maybe we should use Xoff = 0 or Xoff=-1 instead of delete.
> 
> For 2, still not quite undertsand the scenario.
> 
> Will setup a offline discussion for this.

For `xoff`, it's OK to disable it via setting it to `0`.
For deleting a pool, a possible example is to remove the lossless pools when a device is deployed in a lossy-traffic-only scenario.

Anyway, now that we are introducing a new mechanism to support multiple overlay DB, I believe it should not be restricted in the buffer area. Probably the requirement is also applicable to other features.

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
+# Multiple overlay ConfigDB
```
qiluo-msft, line 1: >Multiple overlay ConfigDB [](http://example.com/codeflow?start=2&length=25)

Overlay ConfigDB with Yang Default Values and SONiC Image Default Profiles

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Table of Contents
+
+- [Table of Contents](#table-of-contents)
+- [About this Manual](#about-this-manual)
+  + [SONiC issue solved by this feature](#sonic-issue-solved-by-this-feature)
```
qiluo-msft, line 7: >SONiC issue solved by this feature [](http://example.com/codeflow?start=5&length=34)

Problem Statement

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+- Yang model default value overlay.
+- Static config overlay.
+- swss-common API change.
+
+## SONiC issue solved by this feature
+
+- Still using old default value and config from j2 template after OS upgrade:
+  - Following config may update after OS upgrade:
+    - Default value: defined in yang model.
+    - Static config: defined in j2 template.
```
qiluo-msft, line 40: >Static config: defined in j2 template [](http://example.com/codeflow?start=6&length=37)

Before the problem statement, let's explain the terminologies: profiles, static config, templates. Better with some examples.

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+  - Currently all config stored in config DB, so config above can't update after OS upgrade.
```
qiluo-msft, line 41: >so config above can't update after OS upgrade [](http://example.com/codeflow?start=46&length=45)

You are assuming that config_db.json is carried over when OS upgrade. This is not a problem if we uses minigraph right now.

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+- Potential risk, Yang model default value conflict with hardcoded value:
+  - Default value hardcoded in source code.
+  - Yang model default value not used.
+- SONiC utilities not support get user config and full config:
```
qiluo-msft, line 45: >user config and full config [](http://example.com/codeflow?start=34&length=27)

let's explain the terminologies: user config, all config. Better with some examples.

Full config is opposite to incremental config. All config is aligned with `show running all`.

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+
+- Return default value is optional.
+  - Application can read config without default value, also can read config with default value.
+- Backward compatibility with existed code and applications.
+
+## 1.2 swss-common return static config from static config overlay
+
+- Static buffer config stored in static config tables.
+- Return static config is optional.
+  - Application can decide read config with static config or not.
```
qiluo-msft, line 63: >config with static config or not [](http://example.com/codeflow?start=32&length=32)

Could you explain 2 options in detailed. Do you mean static config possible overridden by user config if existing?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+- Backward compatibility with existed code and applications.
```
qiluo-msft, line 64: >existed [](http://example.com/codeflow?start=30&length=7)

typo: existing

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+  - For backward compatibility when initialize buffer config from graph, config will be write to both config tables and static config tables.
```
qiluo-msft, line 65: >graph [](http://example.com/codeflow?start=66&length=5)

minigraph

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+  - After code migrate to use static buffer config, static config will only write to static config tables.
+  - Static config support delete/revert operation:
+    - In some user scenario, user need delete a static config, and also may add config back later.
+    - For delete operation, data in static config table will not be delete.
+    - Static overlay use STATIC_CONFIG_DELETE table to handle dlete/revert:
```
qiluo-msft, line 70: >dlete [](http://example.com/codeflow?start=62&length=5)

typo: deleted

qiluo-msft, line 70: Combine the 2 points into one.

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+      - When delete a static config item, static overlay will add the item key to STATIC_CONFIG_DELETE table.
+      - When read static config, all item in STATIC_CONFIG_DELETE will not include in result.
+
+# 2 Design
+
+- Layered config DB design diagram:
```
qiluo-msft, line 76: >Layered [](http://example.com/codeflow?start=2&length=7)

When reading the design diagram, I feel "Yang Model Underlay" may be a better term.

How about renaming "Static Config Overlay" -> "Profile Underlay"?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+
+<img src="./images/swss-common-default-value.png"  />
+
+## 2.1 Considerations
+
+### How to get default value
+
+|                                                               | Pros                                              | Cons                                                                                                                                              |
+| ------------------------------------------------------------- | ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Get default value from Yang model in read API.                | Redis config DB keeps no change.                  | 3 MB memory per-process because need load Yang model and reference libyang.<br>50ms to load yang model.<br>8ms to read default value 10000 times. |
```
qiluo-msft, line 90: >8ms to read default value 10000 times [](http://example.com/codeflow?start=225&length=37)

Since there is performance penalty, make the loading and mapping as lazy as possible?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+
+- API design diagram:
+
+<img src="./images/swss-common-default-value.png"  />
+
+## 2.1 Considerations
+
+### How to get default value
+
+|                                                               | Pros                                              | Cons                                                                                                                                              |
```
qiluo-msft, line 88: >| [](http://example.com/codeflow?start=0&length=1)

It's good to provide table. However not clear which option do you choose

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+| ------------------------------------------------------------- | ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Get default value from Yang model in read API.                | Redis config DB keeps no change.                  | 3 MB memory per-process because need load Yang model and reference libyang.<br>50ms to load yang model.<br>8ms to read default value 10000 times. |
+| Write default value to default value DB when write config DB. | Better read performance, Less memory consumption. | Need add new Redis DB for default value.                                                                                                          |
+
+### How to get static config
+
+- Static config will stored in a new redis database 'STATIC_CONFIG_DB'
```
qiluo-msft, line 95: >database [](http://example.com/codeflow?start=43&length=8)

What is the database index you will use?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+- Static config DB tables will have exactly same name and schema with config DB tables.
```
qiluo-msft, line 96: >Static config DB tables [](http://example.com/codeflow?start=2&length=23)

If there are only a few tables, just list them as example?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+- Data will read form static config DB with swsscommon API.
+- Static config DB will save and persist for warm-reboot and fast-reboot.
```
qiluo-msft, line 98: >warm-reboot and fast-reboot [](http://example.com/codeflow?start=45&length=27)

We need to consider more scenarios like cold-reboot, and old schema->new schema upgrade with db migration. In general, each HLD should have a section to cover warm-reboot/fast-reboot.

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+| Existed API keeps no change.                          | When update existed code, can cleanup code to remove hard coded default value.    | All apps need code update.                                                                                                               |
+
+### Current design:
+
+- Existed read API keeps no change. 
+- Add decorator API to return default value and static config data.
+
+## 2.2 New class
+
+- YangModelLoader class
```
qiluo-msft, line 114: >YangModelLoader class [](http://example.com/codeflow?start=2&length=21)

Is it a new class or just extension of existing ConfigMgmt class?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+  
+  - Find default value information by table name and config DB key
+  - Merge default value to API result.
+
+- StaticConfigProvider class
+  
+  - Read static config from static config DB.
+  - Merge static config to API result.
+
+- OverlayConfigDBConnectorDecorator python class
```
qiluo-msft, line 128: >OverlayConfigDBConnectorDecorator [](http://example.com/codeflow?start=2&length=33)

Name is too long. How about "YangDefaultUnderlay"? One class per underlay? I think some application only use one of the underlay decorator.

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+
+- OverlayConfigTable c++ class
+
+- OverlaySubscriberStateTable c++ class
```
qiluo-msft, line 132: >OverlaySubscriberStateTable [](http://example.com/codeflow?start=2&length=27)

To make names consistent, OverlayConfigSubscriberStateTable, or OverlayConfigTable->OverlayTable. Need to think about overlay vs underlay

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+
+## 2.3 Other code change
+
+- Add new methods to TableEntryEnumerable  interface:
```
qiluo-msft, line 136: >Add new metho [](http://example.com/codeflow?start=2&length=13)

What is the reason?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+  - virtual bool hget(const std::string &key, const std::string &field, std::string &value) = 0;
+
+## 2.4 Database Schema
+
+- All static config table will have exactly same schema with existed ConfigDB tables.
+- STATIC_CONFIG_DELETE Table:
```
qiluo-msft, line 142: >STATIC_CONFIG_DELETE [](http://example.com/codeflow?start=2&length=20)

One DELETE table for all tables? Is there performance issue?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+  
+   db = DBConnector("CONFIG_DB", 0)
+   # table = Table(db, 'VLAN_INTERFACE')
+   table = OverlayConfigTable(db, 'VLAN_INTERFACE')
+   table.get("Vlan1000")
+  ```
+
+## 2.6 Other solutions for Yang model default value
+
+|                                                                                                                                                                                                        | Pros                                                                                                                                                                                   | Cons                                                                                                                                                                                                                                                                                                                                 |
```
qiluo-msft, line 177: >| [](http://example.com/codeflow?start=0&length=1)

Should be in Considerations section?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+
+- Static config will stored in a new redis database 'STATIC_CONFIG_DB'
+- Static config DB tables will have exactly same name and schema with config DB tables.
+- Data will read form static config DB with swsscommon API.
+- Static config DB will save and persist for warm-reboot and fast-reboot.
+
+### API compatibility
+
+|                                                       | Pros                                                                              | Cons                                                                                                                                     |
+| ----------------------------------------------------- | --------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
```
qiluo-msft, line 103: which option do you choose

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+| 1. Write API change: when write data to config DB, also write default value to 'Default_value_DB'.<br/>2. Read API change: read default value from 'Default_value_DB' and merge with config DB result. | 1. Less memory consumption and better performance when only call read API: read API no need to load yang model.<br/>2. Less code change, all app will get default value automatically. | Hardcoded default value code still need cleanup.                                                                                                                                                                                                                                                                                     |
+
+# 3 Error handling
+
+- Load yang model: throw exception when found yang model data issue.
+- swss-common API: if not found Yang model schema data for a given table name, write warning message to syslog.
```
qiluo-msft, line 185: >warning message [](http://example.com/codeflow?start=85&length=15)

In some use cases, the application may expect a default value if not user config, then treat no default value as error?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+
+# 4 Serviceability and Debug
+
+- Debug version will write debug log to syslog.
+
+# 5 Unit Test
+
+- All new code will 100% covered by gtest test case.
```
qiluo-msft, line 193: >gtest [](http://example.com/codeflow?start=36&length=5)

gtest is only for C++

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+- Add E2E test case for all new APIs.
```
qiluo-msft, line 194: >E2E [](http://example.com/codeflow?start=6&length=3)

Is it possible to unit test all the testcase?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,234 @@
...
+
+## 1.2 swss-common return static config from static config overlay
+
+- Static buffer config stored in static config tables.
+- Return static config is optional.
+  - Application can decide read config with static config or not.
+- Backward compatibility with existed code and applications.
+  - For backward compatibility when initialize buffer config from graph, config will be write to both config tables and static config tables.
+  - After code migrate to use static buffer config, static config will only write to static config tables.
+  - Static config support delete/revert operation:
```
qiluo-msft, line 67: >Static config support delete/revert operation [](http://example.com/codeflow?start=4&length=45)

This point is complex. need more details and examples.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Yes, currently the load and mapping only happened when first time read default values.
Will update design document with this.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: It's a new C++ class, and it's a internal class, not visible to user.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: The OverlayConfigTable need decorate this method.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: No default value can't treat as a error, because most config item does not has default value.
The swss-common code impossible to know if some item should has a default value or not.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Yes, I will remove this part.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed, add new section, also there will be a slides to cover more example.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed, add example.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed, added more example.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed, move the current design part to above this section.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed, currently we using lazy load.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Index will be 15

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed, added a reboot section.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed, move the current design section to above.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: So when we use minigraph, after OS upgradem the entire config_db will be re-generate. and all user config not in minigraph will be lost.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed, rename to UnderlayTable

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: There is no performance issue because the table only contains item key.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed

doc/mgmt/SONiC_overlay_ConfigDB.md
qiluo-msft: profiles == static config?

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,338 @@
...
+
+- Backward compatibility with existing code and applications.
+  - For backward compatibility when initialize buffer config from minigraph, config will be write to both config tables and static config tables.
+  - After code migrate to use static buffer config, static config will only write to static config tables.
+  - Static config support delete/revert operation:
+    - In some user scenario, user need delete a static config, and also may add config back later.
+    - For delete operation, data in static config table will not be delete, profile underlay use PROFILE_DELETE table to handle delete/revert:
+      - When delete a static config item, profile underlay will add the item key to PROFILE_DELETE table.
+      - When read static config, any key in PROFILE_DELETE will not exist in result.
+      - When user set deleted item back, the item key will be remove from PROFILE_DELETE table.
```
qiluo-msft: >remove [](http://example.com/codeflow?start=62&length=6)

typo: removed

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,338 @@
...
+            "xoff": "6291456"
+        },
+        "egress_lossless_pool": {
+            "size": "32822528",
+            "type": "egress",
+            "mode": "static"
+        }
+    }
+  ```
+        - On mellanox devices, when set buffer mode to 'dynamic' mode, buffer manager will use swsscommon API to delete buffer pool to reclam unused buffer resource.
```
qiluo-msft: >reclam [](http://example.com/codeflow?start=135&length=6)

typo

doc/mgmt/SONiC_overlay_ConfigDB.md
```diff
@@ -0,0 +1,338 @@
...
+
+|                                                               | Pros                                              | Cons                                                                                                                                              |
+| ------------------------------------------------------------- | ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Get default value from Yang model in read API.                | Redis config DB keeps no change.                  | 3 MB memory per-process because need load Yang model and reference libyang.<br>50ms to load yang model.<br>8ms to read default value 10000 times. |
+| Write default value to default value DB when write config DB. | Better read performance, Less memory consumption. | Need add new Redis DB for default value.                                                                                                          |
+
+### How to get static config
+
+- Static config will stored in a new redis database 'PROFILE_DB', database index is 15.
```
qiluo-msft: >15 [](http://example.com/codeflow?start=84&length=2)

Some UT is using 15 as TEST_DB. Please be aware for any conflict risk.

doc/mgmt/SONiC_overlay_ConfigDB.md
qiluo-msft: I realize it may be a re-implementation of ConfigMgmt  in C++. People will ask for a single implementation, and wrapper in various other programming languages.

liuh-80:
Fix EasyCLA issue by merge all change to one and do a force push.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Will test this issue with this PR: https://github.com/sonic-net/sonic-swss-common/pull/625
Will update here after all UT passed.

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed

doc/mgmt/SONiC_overlay_ConfigDB.md
liuh-80: Fixed

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
```diff
@@ -0,0 +1,392 @@
...
+            "xoff": "6291456"
+          },
+          "egress_lossless_pool": {
+            "size": "32822528",
+            "type": "egress",
+            "mode": "static"
+          }
+          }
+          ```
+        - On mellanox devices, when set buffer mode to 'dynamic' mode, buffer manager will use swsscommon API to delete buffer pool to reclaim unused buffer resource.
```
stephenxs, line 176: Just FYI, on Mellanox devices, no buffer pool or buffer profile will be removed to reclaim unused buffers.
But it's possible that a user decides to remove some buffer pool, eg.
- to migrate from double-ingress-pool mode to single-ingress-pool mode
- or to execute `config qos clear`

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
```diff
@@ -0,0 +1,392 @@
...
+
+## Problem Statement
+
+- SONiC still using old default value and config from j2 template after OS upgrade:
+  - Following config may update after OS upgrade:
+    - Default value: defined in yang model.
+    - Profile: defined in j2 template.
+  - Currently all config stored in config DB, so above configs can't be update after OS upgrade.
+- Potential risk, Yang model default value conflict with hardcoded value:
+  - Default value hardcoded in source code.
```
stephenxs, line 113: Just to double check, did you mean the hardcoded value in the templates?

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
```diff
@@ -0,0 +1,392 @@
...
+
+- Fast-reboot will reload minigraph, profile DB will re-render in reload minigraph.
+- sonic-cfggen and sonic-utility will update to support generate profile DB during reload minigraph.
+
+## 3.4 Schema upgrade and DB migration
+
+- DB migrator will re-render profile DB to handle schema change.
+  - Profile DB will generate by sonic-cfggen command.
+- DB migrator code will improve by deprate hardcoded config and complex upgrade logic.
+- Some profile configuration change are disruptive, for example buffer config change,those change need fast-reboot to make sure change refelected on ASIC.
```
stephenxs, line 321: IMO it should also work to do a warm reboot to upgrade the schema

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
```diff
@@ -0,0 +1,392 @@
...
+
+- API design diagram:
+
+<img src="./images/swss-common-default-value.png"  />
+
+### Current design:
+
+- Existing API keeps no change. 
+- Add decorator API to return default value and profile data.
+- Load yang model as lazy as possible.
```
stephenxs, line 194: Maybe we can cache the parsed yang model information somewhere globally so that we just need to parse the yang model once and can use them many times.
But I'm not sure whether it is worth doing so since caching/storing it also takes time.

doc/mgmt/sonic-profile-delete.yang
```diff
@@ -0,0 +1,41 @@
...
+	import ietf-yang-types {
+		prefix yang;
+	}
+
+	import sonic-common {
+		prefix scommon;
+	}
+
+	organization
+		"BRCM";
```
venkatmahalingam, line 15: Change BRCM to SONiC.

doc/mgmt/sonic-profile-delete.yang
```diff
@@ -0,0 +1,41 @@
...
+			"Initial revision.";
+	}
+
+	container sonic-profile-delete {
+			scommon:db-name "CONFIG_DB";
+
+		list PROFILE_DELETE {
+			key "itemkey";
+			scommon:key-delim "|";
+			scommon:key-pattern "PROFILE_DELETE_TABLE|{itemkey}";
```
venkatmahalingam, line 34: We dont need key-delim and key-pattern, right?

doc/mgmt/sonic-profile-delete.yang
```diff
@@ -0,0 +1,41 @@
...
+	description
+		"SONIC PROFILE DELETE";
+
+	revision 2022-07-22 {
+		description
+			"Initial revision.";
+	}
+
+	container sonic-profile-delete {
+			scommon:db-name "CONFIG_DB";
```
venkatmahalingam, line 29: Config-DB is by default, we don't need it here.

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
liuh-80: No, it's ins the source code, for example in hostcfgd, when read config from config DB, if some item does not exist in config DB, hostcfgd will use default values hardcoded in it's code.

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
```diff
@@ -0,0 +1,392 @@
...
+| ------------------------------------------------------------- | ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Get default value from Yang model in read API.                | Redis config DB keeps no change.                  | 3 MB memory per-process because need load Yang model and reference libyang.<br>50ms to load yang model.<br>8ms to read default value 10000 times. |
+| Write default value to default value DB when write config DB. | Better read performance, Less memory consumption. | Need add new Redis DB for default value.                                                                                                          |
+
+### How to get profile
+
+- Profile will stored in a new redis database 'PROFILE_DB', database index is 15.
+- Profile DB tables will have exactly same name and schema with config DB tables.
+- Data will read form profile DB with swsscommon API.
+- Profile DB will save and persist for warm-reboot and fast-reboot.
```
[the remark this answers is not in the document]
liuh-80: When upgrade from old version of SONiC which does not support this feature to new version:
1. The config DB will keep no change, all data in config DB profile table will be treat as user config.
2. Profile DB will be created, and DB migrator will initialize Profile DB.

So, from user aspect:
1. all existing buffer pool/profile keeps no change.
2. There maybe new buffer pool/profile exist, which introduced by the new version of OS.

I will update this to reboot part.

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
```diff
@@ -0,0 +1,392 @@
...
+          "egress_lossless_pool": {
+            "size": "32822528",
+            "type": "egress",
+            "mode": "static"
+          }
+          }
+          ```
+        - On mellanox devices, when set buffer mode to 'dynamic' mode, buffer manager will use swsscommon API to delete buffer pool to reclaim unused buffer resource.
+        - When delete a profile item 'ingress_lossless_pool', the key of this item will be write to PROFILE_DELETE table. and swsscommon read API will not return this item anymore.
+        - Revert operation: when user set buffer mode back, buffer manager will write the deleted item back, when this happen, the key of the item will be removed from PROFILE_DELETE table. and swsscommon read API will return this item.
```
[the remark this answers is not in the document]
liuh-80: It's not correct, when the content writes back is **same** with item with profile table, following happen:
1. Removing the key from PROFILE_DELETE table 
2. Write content to config table, this is because the 'SET' operation is from user, so it will be a user config, even it exactly same with the data in profile DB.

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
```diff
@@ -0,0 +1,392 @@
...
+            },
+    
+            ......
+    
+        {%- endmacro %}
+    ```
+
+- Profile
+  
+  - Profile is part of OS image, should upgrade with OS upgrade.
```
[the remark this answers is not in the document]
liuh-80: The 'profile' is a common concept, profile DB will not only be used by buffer config.

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
```diff
@@ -0,0 +1,392 @@
...
+      - If user not overwrite the profile, profile will be return.
+      - If user 'delete' the profile, will return nothing.
+    - When application not read profile, API will only return user config.
+      - If user overwrite the profile, user config will be return.
+      - If user not overwrite the profile, will return nothing.
+      - If user 'delete' the profile, then there also will no user config exist, will return nothing.
+
+- Backward compatibility with existing code and applications.
+  
+  - For backward compatibility when initialize buffer config from minigraph, config will be write to both config tables and profile tables.
```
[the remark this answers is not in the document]
liuh-80: Fixed, add more detail and change this to config DB and profile DB.

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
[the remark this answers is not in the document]
liuh-80: Fixed

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
```diff
@@ -0,0 +1,392 @@
...
+- sonic-cfggen and sonic-utility will update to support generate profile DB during reload minigraph.
+
+## 3.2 Warm-reboot
+
+- Profile DB will follow SONiC warm-reboot process, warm-reboot will save whole Redis DB to /host/warmboot/dump.rdb, please refer to [SONiC YANG MODEL GUIDELINES](#7-1-sonic-yang-model-guidelines)
+- DB migrator will run after warm reboot, will handle OS upgrade in DB migrator.
+
+## 3.3 Fast-reboot
+
+- Fast-reboot will reload minigraph, profile DB will re-render in reload minigraph.
```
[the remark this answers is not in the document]
liuh-80: Thanks, fixed, I check the reboot code and fix the mistake here.

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
liuh-80: Fixed, my mistake, the buffermgrd are deleting items in appldb and statedb. config DB keeps no change.

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
liuh-80: Fixed, the code/warm/fast reboot will use DB migrator to upgrade schema.

doc/mgmt/SONiC_ConfigDB_support_Yang_default_values_and_profiles.md
liuh-80: Fixed by add more detail, current design is to only load once and cache it.

doc/mgmt/sonic-profile-delete.yang
liuh-80: Fixed.

doc/mgmt/sonic-profile-delete.yang
liuh-80: Fixed, 2 lines removed.

doc/mgmt/sonic-profile-delete.yang
liuh-80: Fixed.

liuh-80:
@stephenxs @venkatmahalingam , thanks for your comments, I have updated the HLD and fix it, please check.

PR #994, opened 2022-05-01
[Password Hardening]Modify compilation flow chapter in HLD

davidpil2002:
The Password Hardening compilation chapter modify by removing the possibility of not compiling the feature.
Meaning that the feature is always compiled when building the image, and is disabled by default.

This modification was done because in general the user always wants to have the possibility to enable/disable the feature and if the user for some reason does not compile the feature, this will enforce him to burn the image again.

In addition, the code of the feature exists in hostcfgd and this daemon is coded in python so it does not really influence the compilation time because python code is an interpreter.


davidpil2002:
Hi @liuh-80 @qiluo-msft, 
I wrote here the small modification in the HLD,
pls can you review

liat-grozovik:
@qiluo-msft  kindly rmeinder to review and signoff

PR #996, opened 2022-05-06
[HLD] SONiC GNMI Server Interface Design

ganglyu:
Signed-off-by: Gang Lv ganglv@microsoft.com

PR:
| [[sonic-gnmi] Use new repo sonic-gmni](https://github.com/sonic-net/sonic-gnmi/pull/3) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-buildimage] Create GNMI server](https://github.com/Azure/sonic-buildimage/pull/10948) | GitHub issue/pull request detail | GitHub pull request check contexts |


ganglyu:
@qiluo-msft

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,514 @@
...
+
+* Set and get RPCs must be supported. Customer will use get RPC to retrieve configurations including FRR and VNET route, and use set PRC to apply new configurations.
+* SetRequest message must support to incrementally update configuration and fully update configurations.
+* Data models can be SONiC Yang models or CONFIG_DB schema.
+* Configurations must be verified using YANG models even it is in CONFIG_DB schema.
+* Need to configure huge VNET route entries to ApplDB, with high speed.
+* Configurations must be persisted after a device restart.
+* Need to support multi-ASIC platform.
+* Support authentication and authorization with TACACS+ server.
+* [low-priority] FRR Yang models will also be supported.
```
prsunny, line 105: These can be classified as phase 1, phase 2 requirements instead of marking low-priority.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,514 @@
...
+
+<img src="images/telemetry.png" alt="telemetry" width="500px"/>
+
+sonic-telemetry provides gNMI/gNOI server interface. And sonic-telemetry provides only telemetry via gRPC, and the gap is configuration.
+
+## 1.1 Requirements
+
+* Set and get RPCs must be supported. Customer will use get RPC to retrieve configurations including FRR and VNET route, and use set PRC to apply new configurations.
+* SetRequest message must support to incrementally update configuration and fully update configurations.
+* Data models can be SONiC Yang models or CONFIG_DB schema.
```
prsunny, line 99: Suggest also add "Ability to configure/read different DBs - APP_DB, CONFIG_DB, STATE_DB, COUNTERS_DB etc"

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,514 @@
...
+* Configurations must be verified using YANG models even it is in CONFIG_DB schema.
+* Need to configure huge VNET route entries to ApplDB, with high speed.
+* Configurations must be persisted after a device restart.
```
prsunny, line 102: This is optional based on target DB. APP_DB entries need not be persisted and requires clients to reprogram during a reboot.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,514 @@
...
+* Need to support multi-ASIC platform.
+* Support authentication and authorization with TACACS+ server.
+* [low-priority] FRR Yang models will also be supported.
+* [low-priority] Upgrade operation will be provided with GNOI.
+* Must follow gnmi-specification.
+    * All changes to the state of the target that are included in an individual SetRequest message are considered part of a transaction.
+    * Within an individual transaction (SetRequest) the order of operations is 'delete', 'replace', 'update'.
+    * If a single path is in multiple operations in one SetRequest (i.e., within 'update' or 'replace'), then the state of the target MUST reflect the application of all the operations in order, even if they overwrite each other.
+    * The session between the client and server MUST be encrypted using TLS.
+    * New connections are mutually authenticated.
```
prsunny, line 112: I also suggest adding a section for heart-beat mechanism and reset-status of the gnmi container. For e.g, if the gnmi container restarts, does clients have to reprogram or not? Full system reload vs just gnmi container reload.

qiluo-msft:
@anand-kumar-subramanian @kannankvs @mikelazar @kerry-meyer
@praveen-li @dgsudharsan @rathnasabapathyv @venkatmahalingam
Could you help review this design doc?

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: done

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: done

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: updated.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: Heart-beat mechanism is used to detect reload?
If gnmi container restarts, client doesn't need to reprogram, but for system reload, client needs to reprogram.
I think client can use heart-beat to detect reload, and then use gnmi get to check if needs to reprogram.

ganglyu: Updated.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,525 @@
...
+* SONIC CLI commands
+
+<img src="images/cli.png" alt="cli" width="500px"/>
+
+1. Users could use SONiC CLI command "show runningconfig all" to read current configuration. 
+    - The blue arrow indicates this process.
+2. User could use SONiC CLI command "config apply-patch" to incrementally update configurations systematically and validate target configuration using SONiC Yang models. With generic_config_updater, all operations in a 'config apply-patch' command are processed in a single transaction that will either succeed or fail as a whole.
+    - "config apply-patch" supports ConfigDB, and Yang models.
+    - The green arrow indicates this process, and dotted arrow indicates Yang validation.
+3. SONiC CLI commands can use "config reload" to fully update configurations and validate target configuration using SONiC Yang models.
```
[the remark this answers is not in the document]
ganglyu: this is config reload

admin@vlab-01:~$ sudo config reload -h
Usage: config reload [OPTIONS] [FILENAME]

  Clear current configuration and import a previous saved config DB dump
  file. <filename> : Names of configuration file(s) to load, separated by
  comma with no spaces in between

Options:
  -y, --yes
  -l, --load-sysinfo              load system default information (mac,
                                  portmap etc) first.
  -n, --no_service_restart        Do not restart docker services
  -d, --disable_arp_cache         Do not cache ARP table before reloading
                                  (applies to dual ToR systems only)
  -f, --force                     Force config reload without system checks
  -t, --file_format [config_yang|config_db]
                                  specify the file format  [default:
                                  config_db]
  -h, -?, --help                  Show this message and exit

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,530 @@
...
+| gNOI     | gRPC Network Operations Interface |
+| JSON     | JavaScript Object Notation |
+| Yang     | Yet Another Next Generation, it is a modular language representing data structures in an XML tree format |
+| Generic Config Updater | SONiC configuration generic update and rollback feature |
+| ConfigDB | Configuration Database |
+| ApplDB | Application Database |
+
+# 1 Feature Overview
+
+SONiC should provide a gNMI/gNOI server interface where the client can communicate with SONiC (server) through the interface using their respective gRPC libraries based on ConfigDB/ApplDB/StateDB/CountersDB schema or SONiC Yang schema.
```
madhupalu, line 61: Looks like you are opening up gNMI to write directly to APPDB (any DB) without any application - this may result in data consistency issues? for security concerns, someone bypasses some checks and directly operates the DB. what is the plan to mitigate?

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,530 @@
...
+
+<img src="images/full_rpc_gnoi.png" alt="full-gnmi-gnoi" width="500px"/>
+
+The full configuration request will be overwritten by subsequent full configuration request or incremental configuration request.
+
+<img src="images/mixed requests.png" alt="overwritten-config" width="500px"/>
+
+### 1.2.2 Container
+
+All the introduced features will be part of the sonic-gnmi package installed in sonic-gnmi container.
```
madhupalu, line 443: Looks like you are introducing new container - Why introduce a new container instead of reusing telemetry? Telemetry container has been used for config purposes by some applications today, can telemetry container be renamed and continue to be used? If a new container is required, what is the backward compatible/migration plan?

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,530 @@
...
+
+# 2 Error Handling
+supervisord will autorestart gnmi service if it exits after it has successfully started up.
```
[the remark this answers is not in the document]
ganglyu: GNMI server will report config apply error in GNMI response.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,530 @@
...
+<img src="images/full_rpc_gnoi.png" alt="full-gnmi-gnoi" width="500px"/>
+
+The full configuration request will be overwritten by subsequent full configuration request or incremental configuration request.
+
+<img src="images/mixed requests.png" alt="overwritten-config" width="500px"/>
+
+### 1.2.2 Container
+
+All the introduced features will be part of the sonic-gnmi package installed in sonic-gnmi container.
+
```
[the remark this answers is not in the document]
ganglyu: not now, any suggestion?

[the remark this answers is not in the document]
ganglyu: > Is this feature controlled via build time flags?

Not now, any suggestion?

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
venkatmahalingam: There needs to be an application to handle SAI errors for the configs from controller via gNMI server, what's the plan handles this use-case?

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
venkatmahalingam: We did not have an consensus to ahead with the two containers approach one for config and another one for get and subscription in the community call 6/28, please revisit the design and explain why cant we combine all the gNMI operations in one container (today it is telemetry container but can be renamed as gNMI container), this helps to avoid code duplication.

@bhagatyj mentioned in the call today that there is a need for telemetry container restart for new paths addition, if you could explain why cant we do this on-the-fly without container restart, would be helpful to understand the rationale better.
@prvattem Please add if I have missed to capture anything.

bhagatyj:
Hi Venkat.
  Thank you for capturing the minutes. Please see inline..

On Tue, Jun 28, 2022 at 11:44 PM Venkatesan Mahalingam <
***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
> <https://github.com/sonic-net/SONiC/pull/996#discussion_r909254234>:
>
> > +For full configuration, there’re 2 possible solutions:
>
> +* gNMI server needs to send response at first, and then invoke 'config reload'.
>
> +
>
> +full-gnmi
>
> +
>
> +* Use gNMI request and gNOI request together to implement full configuration update.
>
> +
>
> +full-gnmi-gnoi
>
> +
>
> +The full configuration request will be overwritten by subsequent full configuration request or incremental configuration request.
>
> +
>
> +overwritten-config
>
> +
>
> +### 1.2.2 Container
>
> +
>
> +All the introduced features will be part of the sonic-gnmi package installed in sonic-gnmi container.
>
>
> We did not have an consensus to ahead with the two containers approach one
> for config and another one for get and subscription in the community call
> 6/28, please revisit the design and explain why cant we combine all the
> gNMI operations in one container (today it is telemetry container but can
> be renamed as gNMI container), this helps to avoid code duplication.
>
> @bhagatyj <https://github.com/bhagatyj> mentioned in the call today that
> there is a need for telemetry container restart for new paths addition, if
> you could explain why cant we do this on-the-fly without container restart,
> would be helpful to understand the rationale better.
>
I mentioned that I can relate with a need for a separate telemetry
container for new paths addition. If there are other methods to perform new
paths addition without disrupting the config/gNOI operations, that works
too. One disadvantage of separating the containers is that was mentioned in
the call was duplication of multiple files. While this could be partially
addressed by overlay functionality, this duplication is a concern that
needs to be thought about. In short, we still need to weigh the pros/cons
of telemetry container separation.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
adyeung: Agree with the above comments, we should keep a single gnmi server in SONiC, maintaining two gnmi servers in the same NOS unnecessarily divide the code and future contributions. If there are enhancement or capability that needs to be addressed, it should be directed towards the current gnmi server. The gnmi server in telemetry container was introduced 2018, there are active PR contributions in github

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,530 @@
...
+sonic-telemetry provides gNMI/gNOI server interface. And sonic-telemetry provides only telemetry via gRPC, and the gap is configuration.
+
+## 1.1 Requirements
+
+### 1.1.1 Phase 1 Requirements
+* Set and get RPCs must be supported. Customer will use get RPC to retrieve configurations including FRR and VNET route, and use set RPC to apply new configurations.
+* SetRequest message must support to incrementally update configuration and fully update configurations.
+* Data models can be SONiC Yang models or CONFIG_DB schema.
+* Ability to configure/read different DBs - ApplDB, ConfigDB, StateDB, CountersDB etc.
```
bandaru-viswanath, line 101: Please see the comment above, writing directly to APPL_DB/STATE_DB/COUNTERS_DB is against the general norm for SONIC management model.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
bandaru-viswanath, line 443: There is no need for a new GNMI server. There is already one GNMI container which runs the server. Please enhance the existing container instead of creating a new one.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,530 @@
...
+| Replace (update) | If a particular path-value is specified in the client-supplied data, it is replaced with the client-specified value.<br>If the path specified does not exist, the target MUST create the data tree element and populate it with the data in the Update message. | Add |
+
+Below table provides an example to translate from SetRequest to JsonPatch.
+
+### Table 3: Translate Example
+| SetRequest Format | JsonPatch Format |
+|  ----  | ----  |
+| replace {<br>&ensp;path {<br>&ensp;&ensp;elem {name: "CONFIG_DB"} elem {name: "DEVICE_NEIGHBOR"} elem {name: "Ethernet96"}<br>&ensp;}<br>}<br>replace {<br>&ensp;path {<br>&ensp;&ensp;elem {name: "CONFIG_DB"} elem {name: "DEVICE_NEIGHBOR"} elem {name: "Ethernet8"} elem {name: "port"}<br>&ensp;}<br>&ensp;val { json_ietf_val: "eth1" }<br>} | [<br>&ensp;{<br>&ensp;&ensp;"op": "remove", "path": "/DEVICE_NEIGHBOR/Ethernet96"}, <br>&ensp;&ensp;{ "op": "add", "path": "/DEVICE_NEIGHBOR/Ethernet8/port", "value": "eth1"}<br>] |
+
+#### 1.2.1.5 Full Configurations
```
bandaru-viswanath, line 361: Hows the simultaneous configuration from CLI and GNMI  protected/serialized ?

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,530 @@
...
+<img src="images/telemetry.png" alt="telemetry" width="500px"/>
+
+sonic-telemetry provides gNMI/gNOI server interface. And sonic-telemetry provides only telemetry via gRPC, and the gap is configuration.
+
+## 1.1 Requirements
+
+### 1.1.1 Phase 1 Requirements
+* Set and get RPCs must be supported. Customer will use get RPC to retrieve configurations including FRR and VNET route, and use set RPC to apply new configurations.
+* SetRequest message must support to incrementally update configuration and fully update configurations.
+* Data models can be SONiC Yang models or CONFIG_DB schema.
```
bandaru-viswanath, line 100: Current SONIC-Yang models are flat and are not meant for Northbound configurations. How are the dependencies resolved in flat-models ?

bandaru-viswanath (changes requested):

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
prsunny: This was discussed in the meeting. There are some usecases such as route add/deletes (for e.g vnet routes) which shall be programmed by an external controller. These routes are not intended to be persistent across reboots. It is essentially extending the existing RestAPI model to gNMI.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
qiluo-msft: Could you explain a little bit more on "SONIC-Yang models are flat and are not meant for Northbound configurations"? What is "flat"?

This gNMI server depends on "Generic Config Updater" design, which could figure out the dependencies from SONiC Yang models. ref: https://github.com/sonic-net/SONiC/blob/master/doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md#stage-2-json-patch-ordering

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
ganglyu: For now, gNMI can't handle SAI errors.
GCU has implemented a framework to take care of manual service restart and verify services have absorbed change correctly. So gNMI server with GCU has better error handling than SONiC CLI.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: Totally agree. Simultaneous configuration from CLI and GNMI should be protected.
We will update CLI implementation, and it's not covered in this design.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: gNMI can only update table in ApplDB with Yang models.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,530 @@
...
+| JSON     | JavaScript Object Notation |
+| Yang     | Yet Another Next Generation, it is a modular language representing data structures in an XML tree format |
+| Generic Config Updater | SONiC configuration generic update and rollback feature |
+| ConfigDB | Configuration Database |
+| ApplDB | Application Database |
+
+# 1 Feature Overview
+
+SONiC should provide a gNMI/gNOI server interface where the client can communicate with SONiC (server) through the interface using their respective gRPC libraries based on ConfigDB/ApplDB/StateDB/CountersDB schema or SONiC Yang schema.
+
```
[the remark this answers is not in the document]
ganglyu: If there's no Yang models for a table, this table can't be changed.
And we can also follow gNMI spec to use "config false" tag, please refer to [link](https://datatracker.ietf.org/doc/html/rfc7950#section-4.2.3).

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
venkatmahalingam: @prsunny We need to selectively allow (e.g VNET routes, DASH objects in APPL_DB) to control via REST/gNMI server, exposing DB objects via north bound interfaces would be a big security concern. I like to see some design around this in this HLD or in a separate HLD.

ganglyu:
@madhupalu @venkatmahalingam @adyeung @bandaru-viswanath @bhagatyj 

SONiC telemetry is using prefix target to identify target database, and we can use the same method to support backward compatibility.

* If prefix target is ConfigDB or ApplDB, gNMI server should follow current design for set request.
* If prefix target is empty, gNMI server should invoke translib API for set request.

Now we have two gNMI server, one for read and one for configure. When SONiC telemetry and SONiC gNMI are stable in the future, we can use this prefix target solution to merge them to a single container.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
bandaru-viswanath: @prsunny I don't think we can make infrastructure changes to allow specific need for a select-isolated deployment. Tomorrow someone might comeup with a need to directly write to ASIC_DB for effecting shutdown of all port-channels quickly. I don't think these things should be allowed. I don't agree that 'some external entity' wants to write to internal database tables of SONIC' is a requirements that needs to be fulfilled.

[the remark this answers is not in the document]
prsunny: Agree with Venkat that it shall be only selectively allowed. @bandaru-viswanath, i tend to disagree that this is infrastructure change. For your comment, if there is an external orchagent, may be it is a valid scenario of writing directly to ASIC_DB. But currently that's not the case. As mentioned, this is not a new design as APP_DB means application DB and depending on where the application sits (like BGP etc), it can act as the producer. Why should we restrict that application must be localized. Also this is not a persitent config. So config_db is out of scope.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
mm0930: The gnmi container (Telemetry) already has support for telemetry, config set and config get today. Is there a reason why the existing gnmi server does not fit / cannot be extended for the intended functions ? 
Introducing duplicate containers in the architecture comes with a lot of overhead the community will pay in terms of maintenance.
- The developer community will be divided on which one to contribute to. Does the developer need to ensure both work ? This is an overhead on multiple contributions.
- If it's not mandated that both should work, there will be wide divergence on what features will be supported through which container
- There will be code duplication. Why should the same NOS have two GNMI servers?
If there is something deficient in the existing container kindly list it clearly, and articulate why the existing container cannot be enhanced (if required) to meet the requirements.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
bandaru-viswanath: Thank you, please enhance the HLD to add appropriate protection.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
bandaru-viswanath: @qiluo-msft Please see the hierarchical yang models here. https://www.openconfig.net/projects/models/ . It shows how data is defined in a hierarchy and also cross referenced. They are useful as NB yang models because of the hierarchy and referencing. 

SONIC yang models are flat, and are really meant only for validating config data.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
prsunny: Please suggest how it can be selectively enabled, without hardcoding specific details in the container. The only place this can be done is in the GNMI container, and if we do so, we will be hardcoding some specific usecase/deployment requirements into UI.

As suggested above, this can be done using Yang. 

Yes, APPL_DB is the right place. Im just saying config_db is out of scope for programming something that is not intended to be persistent. I'm not sure on how positioning of APPL database in architecture warrants that only config_db can be modified for write operations.  My point is APPL_DB is for application data (irrespective of whether it is local or remote). Agree with you on ASIC_DB.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,564 @@
...
+* SONIC CLI commands
+
+<img src="images/cli.svg" alt="cli" width="600px"/>
+
+1. Users could use SONiC CLI command "show runningconfig all" to read current configuration. 
+    - The blue arrow indicates this process.
+2. User could use SONiC CLI command "config apply-patch" to incrementally update configurations systematically and validate target configuration using SONiC Yang models. With generic_config_updater, all operations in a 'config apply-patch' command are processed in a single transaction that will either succeed or fail as a whole.
+    - "config apply-patch" supports ConfigDB, and Yang models.
+    - The green arrow indicates this process, and dotted arrow indicates Yang validation.
+3. SONiC CLI commands can use "config reload" to fully update configurations and validate target configuration using SONiC Yang models.
```
qiluo-msft, line 74: >update [](http://example.com/codeflow?start=55&length=6)

To be more accurate:
fully update configurations -> erase original and fully reload configurations from a file

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
qiluo-msft: @madhupalu The requirement comes from DASH project. The gNMI server will only write to limited tables in ApplDB, and these tables must have Yang model defined. In this use case, gNMI play the same role as a RestAPI server.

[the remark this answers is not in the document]
qiluo-msft: Just a simple clarification: there is no requirement to write directly to STATE_DB, in this HLD, the gNMI server will read from StateDB, like a telemetry server.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
qiluo-msft: @bandaru-viswanath this is a great question! Currently the HLD of gNMI server does not protect against confliction with CLI, similar to one CLI does not protect against confliction with another CLI today.

In the future vision, there should be a centralized service inside SONiC responsible for all kinds of configuration. The requests could come from CLI/gNMI/RestAPI or SONiC applications. And the request should be serialized if there is confliction. The gNMI server or a sub service under gNMI server may be a candidate for such centralized service.

qiluo-msft (approved):

prsunny (approved):

venkatmahalingam (approved):

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
madhupalu: sounds good.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
madhupalu: good with it.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
madhupalu: you can add a build time flag similar to other features ex: sonic_mgmt_framework

madhupalu (approved):

zhangyanzhao:
@ganglyu  can this PR be merged?

ganglyu:
> @ganglyu can this PR be merged?

we are still waiting for community review

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,568 @@
...
+
+## 1.1 Requirements
+
+### 1.1.1 Phase 1 Requirements
+* Set and get RPCs must be supported. Customer will use get RPC to retrieve configurations including FRR and VNET route, and use set RPC to apply new configurations.
+* SetRequest message must support to incrementally update configuration and fully update configurations.
+* Data models can be SONiC Yang models or CONFIG_DB schema.
+* Ability to read different DBs - ApplDB, ConfigDB, StateDB, CountersDB etc.
+* Ability to write different DBs - ApplDB, ConfigDB.
+* Configurations must be verified using YANG models even it is in CONFIG_DB schema.
```
anand-kumar-subramanian, line 105: How are we planning to achieve this? Are we planning to integrate with the python validator(In which case validation of the bulking configurations will become time-consuming due to loading and unloading of pythong files)?

ganglyu: For CONFIG_DB, we can accept time-consuming configurations. Validation is more important to us.

anand-kumar-subramanian: Validation can be done by other means(go lang based validator that will not be time consuming) why integrate with python validator knowing that it is going to be time-consuming.

ganglyu: We need to use generic config updater to support incremental configuration update.
For now, generic config updater is implemented with Python and it's time consuming, and we can improve generic config updater in the future.

qiluo-msft: @anand-kumar-subramanian Could you provide some sample of "bulking configurations will become time-consuming" in your mind. If the "bulking configurations" is in one SetRequest, the "loading and unloading of python files" will happen only once and it will not be a significant performance issue. If "bulking configurations" is in multiple SetRequest, it is better to ask client to combine them into one.

qiluo-msft (approved):

qiluo-msft:
@adyeung @bandaru-viswanath @bhagatyj Could you review again? We believe already resolved all your comments.

anand-kumar-subramanian (approved):

adyeung (approved):

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -0,0 +1,568 @@
...
+      LOGLEVEL_DB     = 3;
+      CONFIG_DB       = 4;
+      // PFC_WD_DB shares the the same db number with FLEX_COUNTER_DB
+      PFC_WD_DB       = 5;
+      FLEX_COUNTER_DB = 5;
+      STATE_DB        = 6;
+      // For none-DB data
+      OTHERS          = 100;
+      // For mixed schema
+      MIXED_SCHEMA    = 101;
```
qiluo-msft, line 475: >MIXED_SCHEMA [](http://example.com/codeflow?start=6&length=12)

The `MIXED_SCHEMA` is not necessary. The most industry keep Target as empty. We also need to refine the "SetRequest flow chart"

zhangyanzhao:
@qiluo-msft  can you please double check if this feature has been included in 202211 branch? Thanks.

qiluo-msft:
Yes, this feature has been included in 202211 branch.

PR #1009, opened 2022-05-26
SONiC bulk counter support

Junchao-Mellanox:
PR https://github.com/opencomputeproject/SAI/pull/1352/files introduced new SAI APIs that supports bulk stats:

- sai_bulk_object_get_stats
- sai_bulk_object_clear_stats

SONiC flex counter infrastructure shall utilize bulk stats API to gain better performance. This document discusses how to integrate these two new APIs to SONiC. 

Related PRs:
TBD


doc/bulk_counter/bulk_counter.md
```diff
@@ -0,0 +1,144 @@
...
+- sai_bulk_object_get_stats
+- sai_bulk_object_clear_stats
+
+SONiC flex counter infrastructure shall utilize bulk stats API to gain better performance. This document discusses how to integrate these two new APIs to SONiC. 
+
+### Requirements
+
+- Syncd shall use bulk stats APIs based on object type. E.g. for a counter group that queries queue and pg stats, queue stats support bulk while pg stats does not, in that case queue stats shall use bulk API, pg stats shall use non bulk API
+- For a certain object type in a counter group, it shall use bulk stats only if:
+  - The stats capability for each counter IDs shall include SAI_STATS_MODE_BULK_READ, SAI_STATS_MODE_BULK_CLEAR, SAI_STATS_MODE_BULK_READ_AND_CLEAR (avoid handling stats mode change)
```
jimmyzhai, line 32: Must include all three, or some (for example BULK_READ and BULK_CLEAR) ? Even if it only includes BULK_READ, can it do bulk get and do individual clear?

doc/bulk_counter/bulk_counter.md
```diff
@@ -0,0 +1,144 @@
...
+  - Each object queries exactly the same counter IDs. (Requirement from function signature of sai_bulk_object_get_stats and sai_bulk_object_clear_stats)
+- Syncd shall automatically fall back to old way if bulk stats APIs are not supported
+- Syncd shall utilize API sai_query_stats_capability to query bulk capability. Syncd shall treat counter as no bulk capability if API sai_query_stats_capability return error except SAI_STATUS_BUFFER_OVERFLOW (SAI_STATUS_BUFFER_OVERFLOW requires a retry with larger buffer)
+- Syncd shall call bulk stats API in flex counter thread and avoid calling it in main thread to make sure main thread only handles short and high priority tasks. (This is the default behavior in flex counter infrastructure)
+- In phase 1, the change is limited to syncd only, no CLI/swss change. Syncd shall use bulk stats API by querying the statistic capability. Syncd shall deduce the bulk stats mode according to the stats mode defined in FLEX DB:
+  - SAI_STATS_MODE_READ -> SAI_STATS_MODE_BULK_READ
+  - SAI_STATS_MODE_READ_AND_CLEAR -> SAI_STATS_MODE_BULK_READ_AND_CLEAR
+
+### Architecture Design
```
jimmyzhai, line 41: Here `bulk get` shall be a block API and counter thread calls it to talk with asic/npu in sync mode. Is it possible to support async IO, i.e `bulk get` is non-block and provides callback, asic/npu will call callback to finish rest things after counter collection? Since bulk get collets a lot of data, async IO is much efficient. With async IO, another benefit is we do not need many counter threads.

doc/bulk_counter/bulk_counter.md
```diff
@@ -0,0 +1,144 @@
...
+No extra logic on SONiC side is needed to handle warmboot/fastboot.
+
+- As fastboot dealys all counters querying, this feature does not affect fastboot. 
+- For warmboot, it is vendor SAI implementation's responsible to make sure that there must be no error if warmboot starts while bulk API is called.
+
+### Restrictions/Limitations
+
+- Bulk collect attribute value is not supported
+- Buffer pool stats is not support for bulk because different buffer pool may have different stats mode. E.g. pool1 has mode SAI_STATS_MODE_READ, pool2 has mode SAI_STATS_MODE_READ_AND_CLEAR. The new SAI bulk API only allows specify one stats mode.
+
```
jimmyzhai, line 136: Any limitation on maximum object number at one bulk get?

doc/bulk_counter/bulk_counter.md
Junchao-Mellanox: Yes, `bulk get` is a block API. Async IO is a great idea, and it will surely improve the performance. It requires  several new SAI API first and might be another feature.

doc/bulk_counter/bulk_counter.md
Junchao-Mellanox: It depends on vendor implementation. I will add an item to this chapter.

doc/bulk_counter/bulk_counter.md
```diff
@@ -0,0 +1,144 @@
...
+- object_keys: objects that participate the bulk call. E.g. for port, SAI object id value shall be put into sai_object_key_t structure.
+- counter_ids: SAI statistic IDs that will be queried/cleared by the bulk call.
+- get_statuses: SAI bulk get return value for each object.
+- clear_statuses: SAI bulk clear return value for each object. Only effective if stats mode is read and clear.
+- counters: counter values that will be fill by vendor SAI.
+- stats_capas: stats capability for each statitstic IDs for current object type.
+
+The flow of how to updating bulk context will be discussed in following section.
+
+An instance of BulkStatsContext shall be added to FlexCounter class for each object type.
```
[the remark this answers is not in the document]
Junchao-Mellanox: I will fix the sentence here.

doc/bulk_counter/bulk_counter.md
```diff
@@ -0,0 +1,145 @@
...
+
+- Meet the signature of sai_bulk_object_get_stats and sai_bulk_object_clear_stats
+- Avoid constructing these information each time collecting statistic. The bulk context shall only be updated under below cases:
+  - New object join counter group. E.g. adding a new port object.
+  - Existing object leave counter group. E.g removing an existing port object.
+
+```cpp
+struct BulkStatsContext
+{
+    bool support = true;
```
jimmyzhai, line 71: Why needs `support` field? The `BulkStatsContext` itself has proved support.

Junchao-Mellanox: Will fix

doc/bulk_counter/bulk_counter.md
```diff
@@ -0,0 +1,145 @@
...
+    std::vector<sai_object_id_t> object_vids;
```
jimmyzhai, line 72: Field `object_vids` looks needless, since having `object_keys`. In addition, `object_type` should be added for self-awareness.

doc/bulk_counter/bulk_counter.md
```diff
@@ -0,0 +1,145 @@
...
+    std::vector<sai_status_t> clear_statuses;
+    std::vector<uint64_t> counters;
+    std::shared_ptr<sai_stat_capability_list_t> stats_capas;
+};
+```
+- support: indicate if bulk is supported for this object type.
+- object_keys: objects that participate the bulk call. E.g. for port, SAI object id value shall be put into sai_object_key_t structure.
+- counter_ids: SAI statistic IDs that will be queried/cleared by the bulk call.
+- get_statuses: SAI bulk get return value for each object.
+- clear_statuses: SAI bulk clear return value for each object. Only effective if stats mode is read and clear.
```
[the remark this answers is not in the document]
Junchao-Mellanox: Will fix

doc/bulk_counter/bulk_counter.md
```diff
@@ -0,0 +1,145 @@
...
+...
+
+```
+
+##### Update Bulk Context
+
+1. New object join counter group.
+
+![Add Object Flow](/doc/bulk_counter/object_join_counter_group.svg).
+
```
jimmyzhai, line 107: In the figure:

1. Why needs to check 'Stats capability available?' Should always 'Query stats capability' after 'Extract object ID...'
2. If success and 'All counter IDs support bulk', this object is a candidate for bulk get; otherwise bulk not supported.
3. If it is a candidate, check bulkStatsContext list to find the matched one according to stats capabilities.
4. If found, add it to the matched bulkStatsContext
5. If not found, add it to a new bulkStatsContext, then add the context to context list

doc/bulk_counter/bulk_counter.md
```diff
@@ -0,0 +1,145 @@
...
+2. Existing object leave counter group, related data shall be removed from bulk context.
+
+##### Statistic Collect
+
+![Collect Counter Flow](/doc/bulk_counter/counter_collect.svg).
```
jimmyzhai, line 112: 1. If mode SAI_STATS_MODE_BULK_READ_AND_CLEAR is supported, call sai_bulk_object_get_stats is enough. I think it is equative to sai_bulk_object_get_stats(mode=BULK_READ) and then sai_bulk_object_clear_stats(mode=BULK_CLEAR).
2. If not sucess after sai_bulk_object_get_stats or sai_bulk_object_clear_stats, why go to 'Old way'?

doc/bulk_counter/bulk_counter.md
Junchao-Mellanox:
```
Why needs to check 'Stats capability available?' Should always 'Query stats capability' after 'Extract object ID' 
```
Stat capability is usually static data stored in SAI, so if we got it once and save it somewhere, we don't need to get it anymore.

For 2~5, I think you suggested to put bulkStatsContext to a list so that we don't need to manage them for each type. That is a great idea. However, it requires a refactor for class FlexCounter. As you can see in FlexCounter, each object type has similar logic:
```
setXXXCounterList
removeXXX
queryXXXSupportedCounters
getXXXSupportedCounters
collectXXXCounters
```
There are many reasons that we need a refactor:
1. The SAI bulk API call for each object is exactly the same, however, the code to save the stats data to COUNTERS DB is not exactly the same for each object.
2. If SAI bulk API call fails, it shall fall back to non-bulk API to keep backward compatible, and those non-bulk API is encapsulated into collectPortCounters, collectQueueCounters...
3. so on

I am working on a proposal to refactor the FlexCounter, it will be ready within this week.

doc/bulk_counter/bulk_counter.md
Junchao-Mellanox: 1. yes, you are right
2. we need keep backward compatible

doc/bulk_counter/bulk_counter.md
jimmyzhai: 1. IMO, stats capability is per object query, should not be cached in syncd
2. Even each type may have a bulkStatsContext list

jimmyzhai:
For performance improvement, do you have any raw data in your prototype to estimate how much the gain is for **bulk get** vs **get**?

jimmyzhai:
@kcudnik please review it.

doc/bulk_counter/bulk_counter.md
Junchao-Mellanox: I will fix: stats capability should match flex counter group mode.

doc/bulk_counter/bulk_counter.md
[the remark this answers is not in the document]
Junchao-Mellanox: Each object type can have a list of BulkStatsContext to handle case like: same object type has different counter/counter capability.

doc/bulk_counter/bulk_counter.md
Junchao-Mellanox: Shall we fall back to old way?

Junchao-Mellanox:
> For performance improvement, do you have any raw data in your prototype to estimate how much the gain is for **bulk get** vs **get**?

Add a performance section to describe the performance improvement.

doc/bulk_counter/bulk_counter.md
```diff
@@ -0,0 +1,145 @@
...
+  - The stats capability for each counter IDs shall include SAI_STATS_MODE_BULK_READ, SAI_STATS_MODE_BULK_CLEAR, SAI_STATS_MODE_BULK_READ_AND_CLEAR (avoid handling stats mode change)
+  - Each object queries exactly the same counter IDs. (Requirement from function signature of sai_bulk_object_get_stats and sai_bulk_object_clear_stats)
+- Syncd shall automatically fall back to old way if bulk stats APIs are not supported
+- Syncd shall utilize API sai_query_stats_capability to query bulk capability. Syncd shall treat counter as no bulk capability if API sai_query_stats_capability return error except SAI_STATUS_BUFFER_OVERFLOW (SAI_STATUS_BUFFER_OVERFLOW requires a retry with larger buffer)
+- Syncd shall call bulk stats API in flex counter thread and avoid calling it in main thread to make sure main thread only handles short and high priority tasks. (This is the default behavior in flex counter infrastructure)
+- In phase 1, the change is limited to syncd only, no CLI/swss change. Syncd shall use bulk stats API by querying the statistic capability. Syncd shall deduce the bulk stats mode according to the stats mode defined in FLEX DB:
```
kcudnik, line 37: why syncd would need to deduce stats mode here?

Junchao-Mellanox: Maybe the description is not accurate here. Currently, there are two possible stats mode in FLEX DB:
```
STATS_MODE_READ
STATS_MODE_READ_AND_CLEAR
```
This requirement item means that: we will not define new stats mode in FLEX DB, syncd shall automatically use bulk API according to the existing stats mode. E.g. the current stats mode is `STATS_MODE_READ` and the current object type supports bulk API, then syncd will call SAI bulk API with mode `SAI_STATS_MODE_BULK_READ`

jimmyzhai (approved):

kcudnik (approved):

PR #1015, opened 2022-06-09
Updated Innovium to Marvel

kannankvs:
Have updated Innovium to Marvel for supported platform files.


xinliu-seattle (approved):

bluecmd:
@kannankvs This seems to have caused the download links for these platforms to point to Marvell's Prestera firmware instead of the correct Innovium builds. There are no successful Innovium builds for master so it might not be a big deal, but right now the download link is for the wrong firmware (sonic-marvell-armhf.bin vs sonic-innovium-dbg.bin) and CPU architecture (armhf vs amd64)

PR #1027, opened 2022-07-04
[password-hardening]Add small modification in compilation chapter in …

davidpil2002:
The Password Hardening compilation chapter is updated to include recent decisions taken during the code review.
The feature is always compiled when building the image, and is disabled by default.

This modification was done because in general the user always wants to have the possibility to enable/disable the feature and if the user for some reason does not compile the feature, this will enforce him to burn the image again.

In addition, the code of the feature exists in hostcfgd and this daemon is coded in python so it does not really influence the compilation time because python code is an interpreter.ng HLD

### DIFF
--- a/doc/passw_hardening/hld_password_hardening.md
+++ b/doc/passw_hardening/hld_password_hardening.md
@@ -49,7 +49,7 @@
 
 ###  1.2. <a name='Scope'></a>Scope
 
-This password hardening hld doc described the requirements, architecture and configuration details of password hardening feature in switches Sonic OS based.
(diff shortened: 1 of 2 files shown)


davidpil2002:
Hi @liuh-80 @qiluo-msft,
I wrote here the small modification done compilation time in the HLD,
pls can you review it?

liuh-80 (dismissed):

liat-grozovik (approved):

liat-grozovik:
changes in doc are now aligned with the code approved.

PR #1041, opened 2022-07-31
Persistent loglevel HLD

EdenGri:
Adding HLD for persistent loglevel


dprital (dismissed):

dprital:
@qiluo-msft - The comments you raised after the online review were addressed. 
Eden added the "db-migrator" in case of an warm upgrade and also analyzed the use of JINJA2_CACHE and it seems that it is not required so the entire LOGLEVEL_DB can be removed. 

Can you please review and approve this HLD ?

Thanks,
Dror.

doc/logging/persistent_logger/persistent_loglevel.md
```diff
@@ -0,0 +1,460 @@
...
+# 3 Persistent log level design
+## 3.1 High-level design
+
+To make the loglevel persistent to reboot, we will move the Logger's tables in LOGLEVEL DB to CONFIG DB. Since the Config DB is already persistent, the log level will also be persistent to reboot. The log level will be saved using the "config save" CLI command.
+
+We will split this design into two phases to keep the pr with a small content:
+
+Phase 1: make the loglevel persistent by moving it to the CONFIG DB.
+
+Phase 2: Removing LOGLEVEL DB and the jinja2 cache.
```
qiluo-msft, line 130: >jinja2 cache [](http://example.com/codeflow?start=38&length=12)

@stepanblyschak Please check whether it is possible to remove jinja2 cache.

doc/logging/persistent_logger/persistent_loglevel.md
```diff
@@ -0,0 +1,460 @@
...
+  With the current implementation, we don't flush the CONFIG DB before warm-reboot, which means that if the user configures some loglevel (for example, debug), after warm-reboot, the system startup with the same configurable loglevel (debug).
+
+## 6.1 Support warm upgrade
+
+ Since we are not flushing the LOGLEVEL DB when we perform a warm upgrade, we need to update the db migrator.
+  - Add LOGLEVEL DB connector in the db migrator.
+  - In case of a warm upgrade, move Logger tables from LOGLEVEL DB to CONFIG DB (with a minor change in the key) and delete LOGLEVEL DB.
+  - Exposing the "del" function from the swss-common/sonicv2-connector.h to the db migrator. 
+
+(Downgrade will not be supported).
```
qiluo-msft, line 349: >Downgrade will not be supported [](http://example.com/codeflow?start=1&length=31)

@yxieca Can we skip downgrade in design?

doc/logging/persistent_logger/persistent_loglevel.md
stepanblyschak: @qiluo-msft This cache was added to speed up template rendering in start scripts. There were a lot of them rendered during system start. This caused a delay in warm boot LAG restore time. According to the test did by @EdenGri with and without the cache we don't see any difference in this timing now. It is probably due to a lot of other optimizations done to sonic-cfggen. Since there is no noticible improvement made by j2 cache now it is safe to remove it.

doc/logging/persistent_logger/persistent_loglevel.md
yxieca: Good catch. I don't think the design has to work for downgrade. However, downgrade is an important operation. It is important to state what would be the behavior for downgrade.

qiluo-msft: Thanks @yxieca! I am mostly worried the process:
1. running with an image with this feature
2. downgrade to older image without this feature. By default, the new ConfigDB table is kept and no processes consume it
3. upgrade again to an image with this feature: since the config_db.json is carried over to new image, will it be override by "Logger tables from LOGLEVEL DB"?

EdenGri: In this case, after step 1: we will have the Logger tables in the CONFIG DB and the LOGLEVEL DB will be empty.
after step 2: the CONFIG DB will contain the old Logger tables and the LOGLEVEL DB will contain new Logger tables with default values.
after step 3: in case of warm-upgrade the content of the LOGLEVEL DB will replace the content from the CONFIG DB.

qiluo-msft: That's nice design consideration. I am looking forward to have this in the design doc itself.

EdenGri: I just updated it, thanks!

qiluo-msft (approved):

yxieca (approved):

zhangyanzhao:
@EdenGri  are all code PRs captured in this HLD now? I see new code PRs came recently. Thanks.

EdenGri:
@zhangyanzhao all code PRs captured in the HLD.

PR #1044, opened 2022-08-02
Added sonic-front-pannel-port-prefix-regex.md

itamar-talmon:
In order to support distinguishing between different types of front panel ports in a maintainable fashion.
Specifically, we are planning to bring up a system with 'service' ports (in addition to the regular ethernet data ports) - these
are lower speed ports that used for connection to accelerators, internal loopbacks and more.


lihuay:
@itamar-talmon : can you please sign the EasyCLA so that this can be accepted.

PR #1045, opened 2022-08-03
SONiC NPU MDIO access support and gbsyncd docker enhancement HLD

jiahua-wang:
SONiC NPU MDIO access support and gbsyncd docker enhancement HLD
| Repo | PR title | State |
|------|----------|-------|
| SAI  | [Add switch api for clause 22 mdio access](https://github.com/opencomputeproject/SAI/pull/1507) |    |
| sonic-buildimage | [Add sai\_mdio\_access\_clause22=1 in td3x2-a720dt-48s-flex.config.bcm](https://github.com/Azure/sonic-buildimage/pull/11303) |   |
| sonic-sairedis | [Add support of mdio IPC server class using sai switch api and unix socket](https://github.com/sonic-net/sonic-sairedis/pull/1080) |   |


doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,246 @@
...
+
+	"phys": [
+	{   
+	  "phy_id": 1,
+	  ...
+	  "lib_name": "/usr/lib/libpai_layer.so",
+	  ...
+	  "phy_access": "mdio",
+	+ "phy_access_lib_name": "/usr/lib/libgbsyncdaccess.so",
+	+ "mdio_cl22_only": true,
```
jimmyzhai, line 238: Instead of a boolean key, I prefer to explicitly specify mdio access method, like

`phy_access_sw ::= "ipc_mdio_cl22" | "ipc_mdio_cl45"`

doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,246 @@
...
+
+It depends on the switch hardware design that the external PHY could be connected to a FPGA or CPLD based MDIO controller or a switch NPU MDIO bus. The FPGA or CPLD based MDIO controller often has linux kernel driver and provides linux sysfs programming interface. The switch NPU MDIO bus uses SAI library hence an Inter-Process-Communication (IPC) mechanism is required between the syncd daemon and gbsyncd daemon.
+![libPAI MDIO access](images/PAI-MDIO.png)
+
+There are 2 mdio access modes: clause 45 and clause 22. Some external PHY uses clause 45 mdio while other external PHY uses clause 22 mdio. The switch NPU sai switch api should distinguish the 2 modes.
+
+When a configured platform target is built, there is only one syncd docker as the SAI library for the configured platform will cover all the device with switch NPU in the same configured platform class. Although the gbsyncd docker is very much similar to the syncd docker, there could be many gbsyncd docker required as there are different PAI library for different PHY and MDIO access method. Our goal is to build a single gbsyncd docker to cover all PHY and MDIO access method.
```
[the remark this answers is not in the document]
jiahua-wang: Not necessary true. Two counterexamples comes to mind: Dell 3248pxe and Accton as4630-54NPE.

doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Architecture Design 
+
+There are many choices for the IPC mechanism between the syncd daemon and gbsyncd daemon. One performance requirement is that it should finsh firmware download within a reasonable time. Our design choice is to use the Unix socket as the IPC mechanism. Our design has the MDIO IPC server in the syncd daemon with its own thread. A new syncd class MdioIpcServer is added to start a new thread, to create an unix socket, to listen on the socket, to accept connection and to read/reply IPC messages.
+
+There is a corresponding MDIO access IPC client code in the form of dynamic link library which provides the flexiblity to load the library at runtime. Assuming the MDIO access library for sysfs is also in the form of dynamic library, gbsyncd can select the MDIO access library at runtime based on some configuration in the gearbox\_config.json file.
+
+The same gearbox\_config.json file already has the information of the PAI library name. The information can be used to dynamically load the PAI library at runtime.
```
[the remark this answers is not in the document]
jiahua-wang: Our design is to have a single gbsyncd container. When new PAI/external PHY type is added, only the new library is added to the gbsyncd container, but not to create another gbsyncd container.

doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,246 @@
...
+
+The informationn of PHY device using clause 22 mdio can also be added to the gearbox\_config.json file. It will help to decide the mdio access mode at runtime.
+
+![syncd gbsyncd MDIO IPC](images/MDIO-IPC.png)
+
+The VendorSai class constructor is overloaded to handle the dynamic linking of the libraries. The  syncd class VendorSai currently only has a constructor with no argument. The linked SAI library path is determined at syncd build time. New constructors with arguments for the VendorSai class is introduced as the overloading constructors. The new constructors can either have the SAI/PAI library path as the argument or have PHY id as one of the arguments. When the constructor with PHY id as the argument is used, the syncd can index into the "phys" section of gearbox configuration file "gearbox\_config.json" to get the PAI library path and MDIO access library path at run time.
+
```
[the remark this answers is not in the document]
jiahua-wang: The reason is the same as above.

doc/gearbox/gearbox_mdio-HLD.md
jiahua-wang: Because MDIO clause 22 could access clause 45 register but not vice versa. The MDIO with clause 22 register is only referred as clause 22 access only.

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
jimmyzhai: Are the PHYs from one Vendor? If yes, can be driven by one PAI libary?

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
jimmyzhai: Each PAI/PHY type may also need specific psai.profile for initialization.

doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,246 @@
...
+	- There is a new MDIO IPC client library is added.
+	- External PHY firmware download time always has performance requirements/impact.
+	- For debugging, beside the syslog logging, the Unix socket IPC mechanism can be simulated by socat.
+	- The current change is only limited to the Broadcom switch NPU platform.
+	- SAI switch api has added the mido clause 22 access functions.
+	- The SaiInterface class has added the mdio clause 45 read/write and mdio clause 22 read/write virtual functions.
+	- The VendoSai class overrides the mdio clause 45 read/write and mdio clause 22 read/write virtual functions.
+	- New keys are added to the gearbox configuration file "gearbox\_config.json".
+
+When the VendorSai class constructor with PHY id argument is used in syncd.cpp, we can assume a simple case where syncd context number is the same as the PHY id.
```
jimmyzhai, line 108: Now one syncd process associates with one context entry in context_config.json. Will it launch multiple syncd processes in new design?

doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,246 @@
...
+	        (int)commandLineOptions->m_globalContext);
+	        ...
+	    } else {   
+	        auto vendorSai = std::make_shared<VendorSai>();
+	        ...
+	    }
+	    ...
+	}
+
+For the case that multiple PHY ids aggregate in a single syncd context, it is a new separate topic. We don't discuss the topic in this HLD.
```
jimmyzhai, line 125: Currently we are in the case of "multiple PHY ids aggregate in a single syncd context" by default. Need consider it for backwards compatibility. Even different cases could coexist.

doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,246 @@
...
+
+When a configured platform target is built, there is only one syncd docker as the SAI library for the configured platform will cover all the device with switch NPU in the same configured platform class. Although the gbsyncd docker is very much similar to the syncd docker, there could be many gbsyncd docker required as there are different PAI library for different PHY and MDIO access method. Our goal is to build a single gbsyncd docker to cover all PHY and MDIO access method.
+
+### Architecture Design 
+
+There are many choices for the IPC mechanism between the syncd daemon and gbsyncd daemon. One performance requirement is that it should finsh firmware download within a reasonable time. Our design choice is to use the Unix socket as the IPC mechanism. Our design has the MDIO IPC server in the syncd daemon with its own thread. A new syncd class MdioIpcServer is added to start a new thread, to create an unix socket, to listen on the socket, to accept connection and to read/reply IPC messages.
```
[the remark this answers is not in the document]
jimmyzhai: Unix socket with SOCK_DGRAM could do same with epoll/select. Clients are able to be identified by client addresses.

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
jiahua-wang: Dell 3248pxe has PHYs from 2 different vendors.

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
jiahua-wang: It can be done by either socket types. It is easier to track clients using connection.

doc/gearbox/gearbox_mdio-HLD.md
jiahua-wang: This design does not aggregate PHYs. It also does not block other software components/changes to manage multiple PHYs in a single context.

doc/gearbox/gearbox_mdio-HLD.md
jiahua-wang: This design does not block other software components/changes to manage multiple PHYs in a single context.

doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,298 @@
...
+	    ...
+	    m_context = context;
+	    m_sai_dll_handle = dlopen (gbsyncd_get_pai_lib_name().c_str(), RTLD_LAZY);
+	    *(void**)(&dll_sai_api_initialize) = dlsym(m_sai_dll_handle, "sai_api_initialize");
+	    *(void**)(&dll_sai_api_uninitialize) = dlsym(m_sai_dll_handle, "sai_api_uninitialize");
+	    *(void**)(&dll_sai_api_query) = dlsym(m_sai_dll_handle, "sai_api_query");
+	    ...
+	    m_access_lib_handle = dlopen (gbsyncd_get_phy_access_lib_name().c_str(), RTLD_LAZY);
+	    m_mdio_cl22_only = gbsyncd_get_mdio_cl22_only();
+	    ...
```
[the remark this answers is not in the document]
jiahua-wang: Using Broncos/Broadcom PAI as an example, the MDIO IPC client is part of the PAI. We think the MDIO IPC client and PAI should be separate libraries. With a separate PAI library which does not include the MDIO access library, we need another mechanism to specify what kind of MDIO access method to use, e.g. IPC vs. SYSFS. dlopen()/dlsym() is one of the ways to provide the mechanism.

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
jiahua-wang: When a new platform with external PHY is added to the SONiC image. The sonic device data directories will have the new init PHY config file, gearbox config file and PHY config file in the new "platform" and "hwsku" directories. The new  PAI/PHY driver libraries need to be added to the 'gbsyncd' docker when the SONiC image is built. dlopen()/dlsym() do not require LD_LIBRARY_PATH or ldconfig setup. As long as the new library has a distinguish path/name, it can be open and loaded at runtime.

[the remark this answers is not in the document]
jimmyzhai: I see dlopen()/dlsym() do not require LD_LIBRARY_PATH.  My point is the library might need its runtime context setup in gbsyncd docker. How to dynamically make it ready?

- In docker init, some shell scripts could be run to help PAI
- How to let each library know the value of SAI_INIT_CONFIG_FILE
- If the library depends on other libs, how to load them automatically in dlopen? Looks they must be under standard lib directory or LD_LIBRARY_PATH?
- The library name, as well as the related script, needs distinguished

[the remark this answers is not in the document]
jiahua-wang: I understand that the syncd needs a profile name passed by -p option, e.g.
command=/usr/bin/syncd -p /etc/sai.d/pai.profile -x /usr/share/sonic/hwsku/context_config.json -g 1
in the supervisord.conf. The profile has the environment SAI_INIT_CONFIG_FILE which is PHY specific. There are many different ways to solve the issue. Because the changed VendorSai class will parse the gearbox_config.json file and the key "sai_init_config_file": exists in the file, the VendorSai class will override the value of SAI_INIT_CONFIG_FILE from the profile.
The PAI library is packaged as a debian file and the gbsyncd docker build uses the debian package, the PAI library debian package has the dependence built in at package build time. When it is installed in the gbsyncd docker, all dependence package will be installed too.
The PAI library name is in the  gearbox_config.json file  with the key "lib_name": . We can put any related script under /usr/share/sonic/hwsku or platform module which are platform specific.

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
jimmyzhai: Like MDIO IPC server in syncd, MDIO IPC client could be a part of gbsyncd. This is common and mdio read/write is provided to PAI with switch attr SAI_SWITCH_ATTR_REGISTER_READ/SAI_SWITCH_ATTR_REGISTER_WRITE.

PAI itself knows MIDIO access method. If it is SYSFS, it can implement its mdio read/write with SYSFS and ignore attr SAI_SWITCH_ATTR_REGISTER_READ/SAI_SWITCH_ATTR_REGISTER_WRITE.

Further more, we could create a common IPC channel between syncd and gbsyncd. MDIO access is a service on top of it. In the future new service could be implemented on it.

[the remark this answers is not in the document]
jiahua-wang: The MDIO SYSFS interface may depend on FPGA/CPLD based MDIO kernel driver implementation. There will be many different hardware implementation of the FPGA/CPLD hence the SYSFS interface could differ from vendor to vendor. The easiest way to add new MDIO access function is to add the function through library.
The current common IPC mechanism in SONiC is RedisDB. We have tried to use RedisDB, the performance is very bad. When any common IPC channel is used, it puts a limit on using a specific thread to process MDIO only, which affects the performance of firmward download.

doc/gearbox/gearbox_mdio-HLD.md
jimmyzhai: > assume a simple case where syncd context number is the same as the PHY id

Still do not understand the meaning of the assumption. There is only one context for gbsyncd.

doc/gearbox/gearbox_mdio-HLD.md
jimmyzhai: If `commandLineOptions->m_globalContext != 0`, will it only create one vendorSai? In the design, looks a vendorSai will dynamically load a PAI library for one PHY. How to load multiple PAI libs?

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
jimmyzhai: MDIO SYSFS access is one part of PAI lib, which is vendor specific.

For common IPC mechanism, I means your IPC client/server for MDIO access.

doc/gearbox/gearbox_mdio-HLD.md
jiahua-wang: The original gearbox design has one context for each PHY. That is why there is ongoing effort to aggregate multiple PHY in a single context. Our design is based on what is currently in the community sonic. There is different way to aggregate PHYs into a single context. It is outside the scope of this design document.

doc/gearbox/gearbox_mdio-HLD.md
jiahua-wang: This design does not cover the PHY aggregation issue. When there is no PHY aggregation, each PHY will have its own context and each PHY is controlled by one gbsyncd process. Each gbsyncd process will have a different context.

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
jiahua-wang: The IPC channel is dedicated to MDIO access. If the channel is shared with other IPC tasks, we could see performance issue.

doc/gearbox/gearbox_mdio-HLD.md
jimmyzhai: Currently Broncos/Broadcom PAI for PikeZ is working with PHY aggregation, i.e. one gbsyncd process with one context manages multiple PHYs. This design should cover it.

jiahua-wang: Please point us the PR of Broncos/Broadcom PAI for PikeZ PHY aggregation.

jimmyzhai: Please refer to PR https://github.com/sonic-net/sonic-buildimage/pull/11362, file device/arista/x86_64-arista_720dt_48s/Arista-720DT-48S/context_config.json.

jiahua-wang: I only see that context of a json file but I don't see the code processing the context. In order for us to adapt our design for the phy aggregation, we need to know how the context_config.json is processed and used.
We only can design based on what we know. Our original design has listed context limitation in the HLD. We can have a follow up HLD to overcome the limitation in this HLD when we see the design of the PHY aggregation.

kcudnik: im not sure about idea passing globalContext to VendorSai since global context is concept of syncd and sairedis not VendorSai

doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,298 @@
...
+        _In_ uint32_t attr_count,
+        _In_ const sai_attribute_t *attr_list)
+	{
+	...
+		if (objectType == SAI_OBJECT_TYPE_SWITCH)
+		{
+		    std::vector<sai_attribute_t> attr_copy;
+		    sai_attribute_t attr;
+		    
+		    for (auto i = 0; i < attr_count; i++)
```
kcudnik, line 184: im not sure it this should be handled here, or in syncd itself, since vendor sai should be just thin layer on SAI apis, and not interfering with attributes

doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,298 @@
...
+		    {
+		        if (attr_list[i].id == SAI_SWITCH_ATTR_REGISTER_READ && m_context !=0)
+		        {
+		        attr.id = SAI_SWITCH_ATTR_REGISTER_READ;
+		        attr.value.ptr =  (void *) m_mdio_read;
+		        attr_copy.push_back(attr);
+		        }
+		        else if (attr_list[i].id == SAI_SWITCH_ATTR_REGISTER_WRITE)
```
kcudnik, line 192: should here be also && m_context !=0 ?

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
jiahua-wang: We will change to pass above values to VendorSai constructor.

doc/gearbox/gearbox_mdio-HLD.md
jiahua-wang: We will change to handle in syncd itself.

doc/gearbox/gearbox_mdio-HLD.md
jiahua-wang: Actually SAI_SWITCH_ATTR_REGISTER_WRITE and SAI_SWITCH_ATTR_REGISTER_READ attributes are only passed by orchagent for PAI but not for SAI. As my answer to another comment, we are going to move this piece of code from VendorSai to syncd itself.

doc/gearbox/gearbox_mdio-HLD.md
jiahua-wang: We may add another command line options to specify PHY id. I will double check if global context is needed in VendorSai. Our existing code use VendorPaiDll class which inherits from VendorSai class. We found that we need to override most member functions but with a small change in each member function. We come to the conclusion that change VendorSai itself makes more sense.

doc/gearbox/gearbox_mdio-HLD.md
kcudnik: maybe there should be ParSay inheriting VendorSai to keep this separate from Syncd.cpp class

jiahua-wang: We will create VendorPaiDll inheriting VendorSai. We still need a small change inside VendorSai so that VendorPaiDll does not have to override most the member function.

kcudnik: yes

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
kcudnik: maybe it should not be called Dll, and just by simple VendorPai

doc/gearbox/gearbox_mdio-HLD.md
```diff
@@ -0,0 +1,337 @@
...
+The VendorPaiDll class will inherit most member functions from the VendorSai class. The VendorSai class needs some changes to accomdate the inheritance.
+
+When a syncd instance runs inside the syncd docker or inside the gbsyncd docker, the syncd global context number is unique for each instance. We can assume the syncd global context number is 0 for the syncd docker instance and the syncd global context number is larger than 0 for any gbsyncd docker instance.
+
+The syncd global context number can used as an index into the array of the "phys" section in the configuration file "gearbox\_config.json". There is information for PHY id, PAI library name, MDIO library name and MDIO mode stored as values for the JSON key of "phy\_id", "lib\_name", "phy\_access\_lib\_name" and "mdio\_cl22\_only" from the configuration file "gearbox\_config.json".
+
+	int syncd_main(int argc, char **argv)
+	{
+		...
+		if (commandLineOptions->m_globalContext != 0) {
```
jimmyzhai, line 124: For the condition `commandLineOptions->m_globalContext != 0`, suggest explicitly define a new command line option such as `m_useVendorPaiDll`. If it is **true**, it will create VendorPaiDll object.

This would be a backward compatible behavior. It does not impact the current case of one gbsyncd process managing multiple same-type PHYs, where it is also a truth of  `m_globalContext != 0`.

kcudnik: good point, this was one of the proposal on the meeting, and i think also it is good approach

kcudnik: also that json configuration should be passed to that object in the constructor

doc/gearbox/gearbox_mdio-HLD.md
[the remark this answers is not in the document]
jiahua-wang: Changed class name to VendorPai

doc/gearbox/gearbox_mdio-HLD.md
jiahua-wang: Will add commandLineOptions->m_paiInstance. It severs 2 purposes,  to differentiate VendorPai from VendorSai and to index into the "phys" section array of "gearbox_config.json". The json configuration name is already in commandLineOptions->m_contextConfig which will be passed the VendorPai constructor. The constructor can get all information by parsing the json configuration.

jimmyzhai (approved):

zhangyanzhao:
We still have another code PR of VendorPai class which is part of the [https://github.com/sonic-net/SONiC/pull/1045] HLD. We will try to make it as part of 202311 release.

Thanks,

Jiahua

PR #1047, opened 2022-08-04
Update psud according to the current implementation

stephenxs:
Update psud according to the current implementation


liat-grozovik (approved):

liat-grozovik:
well done of having hld follow up so it can be trusted document.

PR #1048, opened 2022-08-07
User Management HLD

Mohammedz93:
This PR provides high level design for the username management and configuration in Sonic.
It will define the default role (capabilities), default users and describe the supported configurations for users.


doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+```
+
+###  3.3 <a name='CLIYANGmodel'></a>CLI/YANG model
+
+##### YANG model
+```yang
+
+module sonic-user-mgmt {
+    yang-version 1.1;
+    namespace "http://github.com/Azure/sonic-user-mgmt";
```
zhangyanzhao, line 188: namespace should be https://github.com/sonic-net/sonic-user-mgmt ?  SONiC does not belong to Azure github org.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+} /* end of module sonic-user-mgmt */
+
+
+```
+
+```yang
+//filename:  sonic-role-mgmt.yang
+module sonic-role-mgmt {
+    yang-version 1.1;
+    namespace "http://github.com/Azure/sonic-role-mgmt";
```
zhangyanzhao, line 264: namespace should be https://github.com/sonic-net/sonic-role-mgmt ?  SONiC does not belong to Azure github org.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+
+==============================================================================
+root@host:~$ config username password --help
+Usage: config user password [OPTIONS] <username> <password>
+
+  Specify a user password
+
+Options:
+  -?, -h, --help  Show this message and exit.
+```
```
prvattem, line 372: Please consider providing configuration cli for new role additions as well.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+* 2. [Design](#Design)
+	* 2.1. [Architecture Design](#ArchitectureDesign)
+	* 2.2. [High-Level Design](#High-LevelDesign)
+	* 2.2. [Usercfd Role](#UsercfdRole)
+* 3. [Configuration and management](#Configurationandmanagement)
+		* 3.1. [ConfigDB Tables](#ConfigDBTables)
+		* 3.2. [ConfigDB schemas](#ConfigDBschemas)
+		* 3.3. [CLI/YANG model](#CLIYANGmodel)
+* 4. [Build](#Build)
+		* 4.1. [Compilation](#Compilation)
```
lguohan, line 23: suggest to have runtime configuration only, not build time. enablement can be controlled by config db, it would be easier for user to consume this feature.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+                }       
+
+                leaf primary_group {
+                    mandatory true;
+                    type string;
+                  	description "Primary linux group.";                    
+                }
+                
+                leaf secondary_groups {
+                    mandatory true;
```
venkatmahalingam, line 300: Is this field mandatory?

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+                    type string {
+                        pattern "admin|monitor";
+                    }
+                }                
+                
+                leaf full-name {
+                    type string;
+                }
+                
+                leaf hashed-password {
```
venkatmahalingam, line 235: Use underscore in the field name.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+        } /* end of container ROLE_TABLE */
+
+    } /* end of container sonic-role-mgmt */
+
+} /* end of module sonic-role-mgmt */
+
+
+```
+
+##### Config CLI
```
venkatmahalingam, line 317: Hope we'll use auto Click command generator utility based on YANG model for these commands, please mention the commands that cant be auto-generated here.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+	prefix user-mgmt;
+
+    description "User Management YANG Module for SONiC OS";
+
+	revision 2022-12-01 {
+        description "First Revision";
+    }
+   container sonic-user-mgmt {
+
+        container USER_TABLE {
```
venkatmahalingam, line 198: Can we have the container name as "USER" instead of USER_TABLE? generally..we use _TABLE suffix in the APP DB tables.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+                        pattern '[a-zA-Z_]([a-zA-Z0-9_-]{0,31}|[a-zA-Z0-9_-]{0,30}$)';
+                        length 1..32 {
+                              error-message "Invalid length for the username.";
+                              error-app-tag username-invalid-length;
+                        }				  
+				          }
+                  description "User name.";
+                }                
+
+                leaf state {
```
venkatmahalingam, line 217: Generally state name is used for the read-only field, can we use "status" or some other appropriate name?

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+	**secondary groups**: sudo,docker,redis,adm
+
+* **monitor**: a role for monitoring users with read permissions only. <br/>
+	Each monitor user will be part of the following groups:<br/>
+	**primary group**: adm<br/>
+	**secondary groups**: N/A
+
+In addition, we will define two default users:
+* **admin**: User role is admin.<br/>
+	The only change from the current admin in Sonic is the additional group: adm
```
jeffhtgt, line 115: Should this be "users"?  Having adm and admin could lead to confusion.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+Usercfgd is the daemon that get events from USER_TABLE in CONF_DB and apply the new configuration on the system.<br/>
+The new daemon should start running after "config-setup.service" (i.e. after database is ready) but before systemd enables "system-logind.service" (i.e. before user can login to system)<br/>
+
+
+![services_diagram](services_diagram.png))<br/>
+
+
+##  3 <a name='Configurationandmanagement'></a>Configuration and management
+
+###  3.1 <a name='ConfigDBTables'></a>ConfigDB Tables
```
jeffhtgt, line 135: Using CONFIG_DB seems like a security issue.  CONFIG_DB is exported into "show techsupport".   Superuser/su is not required for "show techsupport".  This will allow non-superuser to export hashed passwords.  Wouldn't it be better to utilize a new keyspace and add support for exporting/importing like config save/import?

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+
+```
+USER_TABLE:{
+  "<username>":{
+    "state": {{enabled/disabled}}
+    "full-name": {{string}}
+    "hashed-password": {{string}}
+    "hashed-password_history": {{string}}
+    "role": {{admin/monitor}}
+    "passwd_date_of_last_change": {{uint32}}
```
jeffhtgt, line 145: Do we need a parameter for "expires" (e.g. user expires in 30 days)?

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+###  3.2 <a name='ConfigDBschemas'></a>ConfigDB schemas
+
+```
+; Defines schema for User configuration attributes in USER_TABLE:
+key                                   = "<username>"                ;user configuration
+; field                               = value
+STATE                                 = "enabled" / "disabled"      ; user enable/disable
+FULL-NAME                             = STRING                      ; Full-name/Description of user
+HASHED-PASSWORD                       = STRING                      ; Hashed password
+HASHED-PASSWORD_HISTORY               = STRING                      ; List of old passwords (0-99)
```
jeffhtgt, line 167: Do we need a "expires" parameter?

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+    } /* end of container sonic-role-mgmt */
+
+} /* end of module sonic-role-mgmt */
+
+
+```
+
+##### Config CLI
+
+```
```
jeffhtgt, line 319: All of these command should require superuser/su.  If this is the case, please explicitly state this in the HLD so appropriate test cases are run.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+
+
+
+###  6 <a name='SAIAPI'></a>SAI API
+No change.
+
+
+
+###  7 <a name='WarmbootandFastbootDesignImpact'></a>Warmboot and Fastboot Design Impact
+Not relevant.
```
jeffhtgt, line 567: If the migration logic is always run, wouldn't this impact timings?

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+  - Verify default values
+  - Add/Delete user
+  - Enable/Disable user
+  - Configure role of user
+  - Configure password of user
+  - Check password last change date after configure password
+  - Configure full-name of user
+
+###### Configuration - Negative flow
+  - Change default users role
```
jeffhtgt, line 591: I think the following tests would also be appropriate:
- Add user and attributes but do not perform config save.  Perform reboot and verify expected behavior (this would be applicable for attribute changes and user deletion)
- Add new user with new attributes, config save, then downgrade to older SONiC version and verify expected results

doc/user_mgmt/user_mgmt_hld.md
jeffhtgt, line 372: Should "show" commands also be added (and only available to su)?  e.g. "show users" could display the list of users/attributes but not the hashed passwords

doc/user_mgmt/user_mgmt_hld.md
jeff-yin: +1 -- can you also share the specific reasons why this is starting off as disabled by default? Are we concerned that this design may not be accepted by all users -- if so, can we update the design in such a way that it is?

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+The goal is to standardize and define the User management configurations and functionality.<br/>
+It will define the default roles (capabilities), default users and describe the supported configurations
+for users.<br/>
+
+Note:
+The compilition of the feature code will be controlled by a build flag and it will be disabled by default.
+
+###  1.5. <a name='Requirements'></a>Requirements
+####  1.5.1. <a name='Definedefaults'></a>Define defaults
+* Default roles: admin and monitor<br/>
```
jeff-yin, line 69: Propose updating the roles to `admin` and `operator` (instead of `monitor`). I *think* the `operator` term is more widely used for the least-privileged / read-only role in other RBAC implementations, but I might be biased by my own experience at Dell.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+
+
+###  1.1. Revision
+|  Rev  |  Date   |      Author      | Change Description |
+| :---: | :-----: | :--------------: | ------------------ |
+|  0.1  | 12/2022 | Mohammed Zayadna | Phase 1 Design     |
+
+###  1.2. <a name='Scope'></a>Scope
+
+This HLD document described the requirements, architecture and configuration details of User Management feature in switches Sonic OS based.
```
jeff-yin, line 50: Does the scope of this document cover local user accounts only?

I feel that we should try to make sure that this design makes space for, or at least avoids conflicting with, remote authentication servers and the management of remotely authenticated users. 

There was a "AAA Improvements" HLD proposed and merged a few years ago (https://github.com/sonic-net/SONiC/blob/master/doc/aaa/AAA%20Improvements/AAA%20Improvements.md) but the code PRs didn't make it in (https://github.com/sonic-net/sonic-buildimage/pull/5553). 
I'm not saying this needs to be revived necessarily, but it is indicative of the complexity that may be involved.

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+##  3 <a name='Configurationandmanagement'></a>Configuration and management
+
+###  3.1 <a name='ConfigDBTables'></a>ConfigDB Tables
+
+```
+USER_TABLE:{
+  "<username>":{
+    "state": {{enabled/disabled}}
+    "full-name": {{string}}
+    "hashed-password": {{string}}
```
jeff-yin, line 142: +1 agree with @jeffhtgt 's comment above. The hashed user pw is already stored in `/etc/shadow`, which is permission-restricted in the Linux file system. Putting the hashed pw in CONFIG_DB will circumvent those permission restrictions.

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: We have received a request from several vendors to add a compilation flag as they are using remote authentication only.
If we change it to be controlled in runtime, it will necessitate usercfd to run always, which will undoubtedly have an impact on the initialization time even when the feature is disabled.

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: I have reviewed the implementation of remote authentication servers and I think that there is no conflict.
I still need to do some manual tests then I will add a new section for AAA integration and elaborate about it.
Regarding "AAA Improvements" HLD, I wasn't aware of it.
Is there still an intention to merge it?
it is obviously conflicts with this HLD

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: I agree. 
I am working with the team on a new design for storing users passwords.
Can you please elaborate about the keyspace ? where I can learn more about it ?
I would appreciate if you could provide an example

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: I agree. 
I am working with the team on a new design for storing users passwords.

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: "passwd_date_of_last_change" is the actually the third field in shadow password 
(Please see: [understanding-etcshadow-file](https://www.cyberciti.biz/faq/understanding-etcshadow-file/))
Linux force the password ageing by comparing the current epoch with this field.
We need this field in DB to preserve the expiration date after upgrade.
it also will allow us to apply the California-SB237 law feature easily.
[California-SB237 feature](https://github.com/sonic-net/SONiC/tree/master/doc/California-SB237)

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: See the comment above: [https://github.com/sonic-net/SONiC/pull/1048/files#r1073802100](https://github.com/sonic-net/SONiC/pull/1048/files#r1073802100)

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: Sure I will fix it

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+                    default "enabled";
+                    type string {
+                        pattern "enabled|disabled";
+                    }
+                }
+                
+                leaf role {
+                    default "admin";
+                    type string {
+                        pattern "admin|monitor";
```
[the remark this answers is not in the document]
Mohammedz93: Sure

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: I have seen examples in CONFIG DB for tables with "_TABLE" Suffix: ACL_TABLE, FLEX_COUNTER_TABLE, PBH_TABLE...
Is there a guideline document ?

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: it is already used in many other yang models
for example: [sonic-feature.yang](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-feature.yang)

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: I have seen both formats in other yang models
I couldn't find any conventions document. If you are aware of one, can you please share it?

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+                    mandatory true;
+                    type string;
+                }
+                
+                leaf password_history {
+                    type string;
+                  	description "List of hashed passwords seperated by comma."; 
+                }
+
+                leaf passwd_date_of_last_change {
```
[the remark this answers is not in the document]
Mohammedz93: I don't think so
This field is actually the Unix epoch time of the last password change
It will be used to preserve ageing after upgrade, and it won't be exposed to user.
so it is better to keep it in the original format.

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: Sure

doc/user_mgmt/user_mgmt_hld.md
```diff
@@ -0,0 +1,595 @@
...
+	prefix role-mgmt;
+
+    description "Role Management YANG Module for SONiC OS";
+
+	revision 2022-12-01 {
+        description "First Revision";
+    }
+   container sonic-role-mgmt {
+
+        container ROLE_TABLE {
```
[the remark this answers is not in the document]
Mohammedz93: This table will be added to init_cfgd with admin and monitor roles only.
so we always expect to find these roles there.
If for some reason usercfgd fails to fetch this info, then it will report an error.

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: In a second thought, no.
I will delete it.

doc/user_mgmt/user_mgmt_hld.md
venkatmahalingam: I understand few exceptions today, IMO we should not keep that has a practice for new tables.

venkatmahalingam: https://github.com/sonic-net/SONiC/blob/master/doc/mgmt/SONiC_YANG_Model_Guidelines.md

doc/user_mgmt/user_mgmt_hld.md
venkatmahalingam: It's a suggestion to change the field name but I leave it upto you.

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: operators usually have limited write permissions. our monitor users will have read-only permissions
Please see this RBAC document: [RBAC](https://docs.jboss.org/author/display/WFLY9/RBAC.html)
Monitor – a read-only role. Cannot modify any resource.
Operator – Monitor permissions, plus can modify runtime state, but cannot modify anything that ends up in the persistent configuration. Could, for example, restart a server.


I found a similar description of operator in Dell website:
[Dell users](https://infohub.delltechnologies.com/l/bare-metal-orchestrator-1-4-command-line-interface-reference-guide/users-and-roles-overview-4)

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: Sure

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: Sure I will add it

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: I will add the role commands to the HLD
but I think that we won't implement it them at phase 1 of this feature

Mohammedz93: If you are referring to "show" command for users, then it is already in the HLD, "show username"
We can't use "show users" because it already in CLI and it shows that current connected users.
I don't think it should be available to su only since in Linux any user can see the list of the current users on the system (e.g. "lslogins")

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: I will test the warm-boot and fast-boot again and update the results in the HLD.

doc/user_mgmt/user_mgmt_hld.md
Mohammedz93: I will add these tests to the list.

doc/user_mgmt/user_mgmt_hld.md
jeffhtgt: See /var/run/redis/sonic-db/database_config.json on any running switch (keyspace = DB).  Maybe an unused keyspace/DB could be used and not exported during a "show techsupport".  Here is a .json from a switch running 202111 SONiC:
{
    "INSTANCES": {
        "redis": {
            "hostname": "127.0.0.1",
            "port": 6379,
            "unix_socket_path": "/var/run/redis/redis.sock",
            "persistence_for_warm_boot": "yes"
        }
    },
    "DATABASES": {
        "APPL_DB": {
            "id": 0,
            "separator": ":",
            "instance": "redis"
        },
        "ASIC_DB": {
            "id": 1,
            "separator": ":",
            "instance": "redis"
        },
        "COUNTERS_DB": {
            "id": 2,
            "separator": ":",
            "instance": "redis"
        },
        "LOGLEVEL_DB": {
            "id": 3,
            "separator": ":",
            "instance": "redis"
        },
        "CONFIG_DB": {
            "id": 4,
            "separator": "|",
            "instance": "redis"
        },
        "PFC_WD_DB": {
            "id": 5,
            "separator": ":",
            "instance": "redis"
        },
        "FLEX_COUNTER_DB": {
            "id": 5,
            "separator": ":",
            "instance": "redis"
        },
        "STATE_DB": {
            "id": 6,
            "separator": "|",
            "instance": "redis"
        },
        "SNMP_OVERLAY_DB": {
            "id": 7,
            "separator": "|",
            "instance": "redis"
        },
        "RESTAPI_DB": {
            "id": 8,
            "separator": "|",
            "instance": "redis"
        },
        "GB_ASIC_DB": {
            "id": 9,
            "separator": ":",
            "instance": "redis"
        },
        "GB_COUNTERS_DB": {
            "id": 10,
            "separator": ":",
            "instance": "redis"
        },
        "GB_FLEX_COUNTER_DB": {
            "id": 11,
            "separator": ":",
            "instance": "redis"
        },
        "APPL_STATE_DB": {
            "id": 14,
            "separator": ":",
            "instance": "redis"
        }
    },
    "VERSION": "1.0"
}

doc/user_mgmt/user_mgmt_hld.md
jeff-yin: I am seeking to revive the effort to contribute the AAA improvements code to SONiC, as the original contributor is no longer assigned to the SONiC project. It would be ideal if we modified both proposals to be compatible with each other so that we could allow for the eventual contribution of AAA enhancements. 
What are some of the conflicts you see? Is it just that the proposed DB schemas are different?

liat-grozovik:
@Mohammedz93 can you please share a plan when the HLD should be updated following the comments provided?

PR #1049, opened 2022-08-08
Syslog rate limit design

Junchao-Mellanox:
**Overview:**

Logging in SONiC is organized with rsyslogd. Each container has its own rsyslogd instance plus a daemon running on a host. The rsyslogd instance which is running on the host is used to collect the messages from within containers and store them at /var/log/syslog path. Rsyslog config file are generated from templates:

- Container scope:
  - Multi ASIC: https://github.com/Azure/sonic-buildimage/blob/master/files/image_config/rsyslog/rsyslog-container.conf.j2
  - Single ASIC: https://github.com/Azure/sonic-buildimage/blob/master/dockers/docker-base/etc/rsyslog.conf
- Host scope: https://github.com/Azure/sonic-buildimage/blob/master/files/image_config/rsyslog/rsyslog.conf.j2

Currently, each container has hardcoded message rate limiting to avoid receiving flooded log messages:

```
$SystemLogRateLimitInterval 300
$SystemLogRateLimitBurst 20000
```
There is no rate limiting configured on host side for now.

The SystemLogRateLimitInterval determines the amount of time that is being measured for rate limiting. The SystemLogRateLimitBurst defines the amount of messages, that have to occur in the time limit of SystemLogRateLimitInterval, to trigger rate limiting. For example, SystemLogRateLimitInterval=300, SystemLogRateLimitBurst=20000, it means that if one daemon generate more than 20000 messages in 300 seconds, rsyslogd will start to drop messages after that(FIFO).

This feature allows user to configure SystemLogRateLimitInterval and SystemLogRateLimitBurst for host, containers.

**Related PR:**
| [[sonic-host-services] Support host side rate limit configuration ](https://github.com/sonic-net/sonic-host-services/pull/22) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-utilities] Support host side rate limit configuration ](https://github.com/sonic-net/sonic-utilities/pull/2454) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[YANG] Support host side rate limit configuration ](https://github.com/sonic-net/sonic-buildimage/pull/12488) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[infra] Support syslog rate limit configuration ](https://github.com/sonic-net/sonic-buildimage/pull/12490) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[containercfgd] Add containercfgd and syslog rate limit configuration support ](https://github.com/sonic-net/sonic-buildimage/pull/12489) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-mgmt] Add test case for syslog rate limit feature ](https://github.com/sonic-net/sonic-mgmt/pull/6627) | GitHub issue/pull request detail | GitHub pull request check contexts |


doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,367 @@
...
+#### Unit Test cases
+1. Verify command "config syslog rate-limit-host"
+2. Verify command "config syslog rate-limit-container"
+3. Verify command "show syslog rate-limit-host"
+4. Verify command "config syslog rate-limit-container"
+
+#### sonic-mgmt Test cases
+
+A new test case shall be added, test flow like below:
+1.	Loop each container
```
wen587, line 361: I think this design also covers host side rate limit setting. Do we need host side rate limit test?

Junchao-Mellanox:
Hi @wen587 , comment fixed. Could you please review again?

wen587:
LGTM

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+{% if rate_limit_burst is defined %}
+$SystemLogRateLimitBurst {{ rate_limit_burst }}
+{% endif %}
+```
+
+#### rsyslog-container.conf.j2
+
+```
+{% set rate_limit_interval = '300' %}
+{% set rate_limit_burst = '20000' %}
```
nazariig, line 125: @Junchao-Mellanox why do we have defaults in different places?

Junchao-Mellanox: Do you want me to remove this and keep the default value in init_cfg.json.j2?

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+            "customized-rate-limit": "false"
+        }
+        ...
+    }
+    ...
+}
+```
+
+support-rate-limit: indicates if this APP extension supports configuring rate limit. This field affected CLI behavior: if true, CLI shall put rate limit configuration to CONFIG DB; otherwise, CLI shall reject the configuration.
+customized-rate-limit: indicates if this APP extension is using "SONiC official way" to configure rate limit. Effective only if support-rate-limit is true. This field affected CLI behavior: if false, CLI shall generated rsyslogd configuration file by rendering rsyslog-container.conf.j2, coping it to relevant container and restart rsyslogd by supervisorctl command; otherwise, app extension shall handle CONFIG DB change and set log rate limit accordingly. "SONiC official way" means:
```
nazariig, line 172: @Junchao-Mellanox  IMHO, this approach exposes too much container internal implementation. Also rendering and copying config files from CLI to container contradicts SONiC APP extension philosophy

doc/syslog/syslog-rate-limit-design.md
nazariig: @Junchao-Mellanox if i am not mistaken this was a review comment during HLD presentation

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+
+rsyslog.conf.j2 and rsyslog-container.conf.j2 shall be extended to accept template variable for SystemLogRateLimitInterval and SystemLogRateLimitBurst. rsyslog.conf for single ASIC platform will be removed and replaced by rsyslog-container.conf.j2.
+
+Host side:
+
+![host_config_on_fly](/doc/syslog/images/host_rate_limit_config_flow.svg).
+
+Change syslog message rate limit while container is running:
+
+![container_config_on_fly](/doc/syslog/images/container_rate_limit_config_flow.svg).
```
nazariig, line 60: @Junchao-Mellanox the container flow diagram shows only the modification flow. Who will put container initial rsyslogd configuration in that case?

doc/syslog/syslog-rate-limit-design.md
nazariig: @Junchao-Mellanox during HLD review the original agreement was to have it in `init_cfg.j2` only

Chenjunchao: Will fix

doc/syslog/syslog-rate-limit-design.md
Chenjunchao: There is no change for container startup flow.

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+
+For app extension, CLI shall only put the configuration to CONFIG DB and let app extension itself to handle it. How app extension handle CONFIG DB change is out of the feature scope.
+
+Load syslog message rate limit configuration during startup is already covered by existing flow, no change of the flow is required:
+
+- Host: rsyslog-config service automatically loads rate limit configuration from CONFIG DB and apply it to rsyslog
```
lguohan, line 66: is rsyslog-config a new service? why not reuse hostcfgd?

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+```
+root@sonic:/home/admin# cat /etc/sonic/config_db.json
+{
+    "FEATURE": {
+        ...
+        "what-just-happened": {
+            ...
+            "support_syslog_rate_limit": "True",
+            "has_customized_syslog_rate_limit": "False"
+            ...
```
TencentSwitchGroup, line 191: Can we make it configurable for rsyslog to suppress duplicate messages?

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+
+```
+root@sonic:/home/admin# cat /etc/sonic/config_db.json
+{
+    "FEATURE": {
+        ...
+        "what-just-happened": {
+            ...
+            "support_syslog_rate_limit": "True",
+            "has_customized_syslog_rate_limit": "False"
```
tzack000, line 190: Can we make it configurable for rsyslog to suppress duplicate messages?

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+
+> Note: Code present in this design document is only for demonstrating the design idea, it is not production code.
+
+#### init_cfg.json.j2
+New tables shall be added to CONFIG DB to store the rate limit configuration. init_cfg.json.j2 shall be extended to define the default value for each built-in containers.
+
+```
+...
+    "SYSLOG_CONFIG": {
+        "GLOBAL": {
```
venkatmahalingam, line 83: GLOBAL seems to indicate that the settings are applicable for Host and Containers, suggest to have to better key name.

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+
+
+container FEATURE {
+
+    description "feature table in config_db.json";
+
+    list FEATURE_LIST {
+        ...
+
+        leaf support_syslog_rate_limit {
```
venkatmahalingam, line 323: Not sure why we need this when we have SYSLOG_CONFIG_FEATURE table for container specific syslog config.

doc/syslog/syslog-rate-limit-design.md
venkatmahalingam: Please keep App ext related handling for user containers separate from SONiC core container config handling.

doc/syslog/syslog-rate-limit-design.md
venkatmahalingam: I believe, we can use hostcfgd or new daemon (in host-services) to handle SYSLOG configs specific to containers and restart the corresponding container by executing "docker exec -i <container-name> supervisorctl restart restart-rsyslogd".

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+        ...
+    }
+}
+```
+
+### Warmboot and Fastboot Design Impact
+
+N/A
```
venkatmahalingam, line 342: There might be some impact for fastboot, please add TBD

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+
+### Restrictions/Limitations
+
+1. Cannot support container not registered to FEATURE table
+2. Persist syslog configuration for database container will not be loaded on next reboot/reload.
```
venkatmahalingam, line 347: Any specific reason?

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: Will fix

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: There are 6 solutions provided by community review meeting, none of them are perfect, but we need choose one to move forward. The discussion is still in progress. Will update here if conclusion available.

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: There is no change for container initial flow. Existing container init flow will render rsyslog-container.conf.j2 and "docker cp" it to container. So, I did not draw the chart here.

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: I prefer to have a new service "syslogcfgd" to handle it. But it is still under discussion. Will update once done.

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: It is out of the feature scope. We may need a new feature on top of this feature to support it.

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: GLOBAL is a typical way to indicate that the entry is for host side. Please refer to https://github.com/sonic-net/sonic-buildimage/blob/53a281008a96edcc22d0536ca4357a18f071c04e/src/sonic-yang-models/yang-models/sonic-auto_techsupport.yang#L31

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: This is a typical way to handle container based configuration. Please refer to https://github.com/sonic-net/sonic-buildimage/blob/53a281008a96edcc22d0536ca4357a18f071c04e/src/sonic-yang-models/yang-models/sonic-auto_techsupport.yang#L90

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: Sure

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: Some template variables to render rsyslog-container.conf.j2 are getting from redis. So, there is nothing can be done if database container is not even up.

venkatmahalingam: Why cant we start the rsyslog after the redis is fully up in the database container?

doc/syslog/syslog-rate-limit-design.md
venkatmahalingam: Ok, please a comment/description in the YANG for clarity, thanks!

doc/syslog/syslog-rate-limit-design.md
venkatmahalingam: Ok, please a comment/description in the YANG for clarity, thanks!

doc/syslog/syslog-rate-limit-design.md
venkatmahalingam: Please share the design points here for wider community visibility.

doc/syslog/syslog-rate-limit-design.md
venkatmahalingam: Sure, please add any mail thread that's happening in the community, not sure non app-ext users would want to combine the code, however, would like to see the community view.

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+            "support-rate-limit": "true",
+            "customized-rate-limit": "false"
+        }
+        ...
+    }
+    ...
+}
+```
+
+support-rate-limit: indicates if this APP extension supports configuring rate limit. This field affected CLI behavior: if true, CLI shall put rate limit configuration to CONFIG DB; otherwise, CLI shall reject the configuration.
```
madhupalu, line 171: - Shouldn't the design support configurable for rsyslog to suppress duplicate messages? 
- Shouldn't the design handle lose of the syslog message when the rsyslog daemon restarts?

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,376 @@
...
+customized-rate-limit: indicates if this APP extension is using "SONiC official way" to configure rate limit. Effective only if support-rate-limit is true. This field affected CLI behavior: if false, CLI shall generated rsyslogd configuration file by rendering rsyslog-container.conf.j2, coping it to relevant container and restart rsyslogd by supervisorctl command; otherwise, app extension shall handle CONFIG DB change and set log rate limit accordingly. "SONiC official way" means:
+
+- Using rsyslogd
+- Configuration of rsyslogd is generated by rsyslog-container.conf.j2
+- rsyslogd can be managed by supervisorctl
```
madhupalu, line 176: - should there be an option to specify immediate restart ? or next boot ? for users to select the option

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: Chart provided

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: Fixed this.

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: Thanks for the comment. 
1. Suppress duplicate messages is another small feature which can be implemented on top of this one. It is not in this scope. 
2. Yes, there could be lose of the syslog message during restarting rsyslog daemon. This is the nature of rsyslogd, there is no much we can do from SONIC scope.

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: CLI is designed to configure things immediately. If you want to delay the change to next boot, you can manually change config_db.json and do reboot/reload.

wen587 (approved):
LGTM. Please ask for more approvals to merge.

Junchao-Mellanox:
@venkatmahalingam , @lguohan , @madhupalu , @tzack000 , could you please review and sign-off?

madhupalu (approved):

qiluo-msft (approved):

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,383 @@
...
+
+![container_rate_limit_load_flow](/doc/syslog/images/container_rate_limit_load_flow.svg).
+
+> Note: docker support copying file to a stopped container
+
+##### Runtime configuration
+
+![container_config_on_fly](/doc/syslog/images/container_rate_limit_config_flow.svg).
+
+> Note: according to test, syslog rate limit configuration on host side would not affect container side.
```
nazariig, line 89: @Junchao-Mellanox IMHO, it's bit confusing comparing to how it's done in AUTO TECHSUPPORT, where global configuration overrides feature rate limit configuration. Here, using 'GLOBAL', does not reflect the actual intentions.

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,383 @@
...
+3. A new daemon containercfgd shall be added to each container to handle container side rsyslog configuration
+4. rsyslog.conf.j2 and rsyslog-container.conf.j2 shall be extended to accept template variable for SystemLogRateLimitInterval and SystemLogRateLimitBurst.
+5. rsyslog.conf for single ASIC platform will be removed and replaced by rsyslog-container.conf.j2
+6. App extension shall handle rsyslog configuration itself
+
+#### CLI flow
+
+![cli_rate_limit_config_flow](/doc/syslog/images/cli_rate_limit_config_flow.svg)
+
+#### Host side flow
```
nazariig, line 63: @Junchao-Mellanox this approach introduces double host side configuration: first time with `rsyslog-config` service and second time with `hostcfgd`

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,383 @@
...
+
+There is an existing service rsyslog-config.service. On switch start up, this service renders rsyslog.conf.j2 and restarts rsyslog service. No extra changes for this feature.
+
+![host_rate_limit_load_flow](/doc/syslog/images/host_rate_limit_load_flow.svg)
+
+##### Runtime configuration
+
+![host_config_on_fly](/doc/syslog/images/host_rate_limit_config_flow.svg).
+
+#### Container side flow
```
nazariig, line 75: @Junchao-Mellanox this approach introduces double container side configuration: first time with container `preStartAction` and second time with `containercfgd`.

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,383 @@
...
+
+### SAI API
+
+N/A
+
+### Configuration and management
+
+#### CLI change
+
+Config rate limit:
```
nazariig, line 226: @Junchao-Mellanox why not to have like this:
```
config
|--- syslog
     |--- rate-limit
          |--- host --interval <interval> --burst <burst>
          |--- container <service_name> --interval <interval> --burst <burst>
```

doc/syslog/syslog-rate-limit-design.md
```diff
@@ -0,0 +1,383 @@
...
+Example:
+config syslog rate-limit-host --interval 300 --burst 20000
+config syslog rate-limit-host --interval 300
+config syslog rate-limit-host --burst 20000
+config syslog rate-limit-container bgp --interval 300 --burst 20000
+```
+
+> Note: set interval or burst to 0 will disable rate limit.
+
+Show rate limit:
```
nazariig, line 241: @Junchao-Mellanox why not to have like this:
```
show
|--- syslog
     |--- rate-limit
          |--- host
          |--- container <service_name>
```

nazariig (changes requested):

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: How about HOST?

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: no, if configuration does not change, hostcfgd will not restart rsyslog

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: no, if configuration does not change, hostcfgd will not restart rsyslog

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: Good suggestion. But I don't see an obviously benefit comparing to the current design. As the code has been implemented for a long time, I would like to keep it as `rate-limit-host` and `rate-limit-container`

liat-grozovik:
Last call on HLD review, @tzack000 and @venkatmahalingam  if you have further comments this is the time. Otherwise I will cont with merging it.
PR should be ready very soon and will be tracked with this PR description.

doc/syslog/syslog-rate-limit-design.md
nazariig: @Junchao-Mellanox unfortunately i don't have neither a strong opinion here, nor a good suggestion

doc/syslog/syslog-rate-limit-design.md
nazariig: @Junchao-Mellanox how do you plan to achieve that? The only reliable way which i see is to parse the `config_db.json` (taking into consideration also `init_cfg.json`) and fill daemon's internal cache on start

doc/syslog/syslog-rate-limit-design.md
nazariig: @Junchao-Mellanox how do you plan to achieve that? The only reliable way which i see is to parse the `config_db.json` (taking into consideration also `init_cfg.json`) and fill daemon's internal cache on start

doc/syslog/syslog-rate-limit-design.md
nazariig: @Junchao-Mellanox ok. Got it.

doc/syslog/syslog-rate-limit-design.md
nazariig: @Junchao-Mellanox ok. Got it.

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: We will parse /etc/rsyslog.conf for the first time, and comparing its configuration with CONFIG DB value.

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: We will parse /etc/rsyslog.conf for the first time, and comparing its configuration with CONFIG DB value.

nazariig: @Junchao-Mellanox what is the guarantee that `rsyslog.conf` will be up-to-date before the `hostcfgd` is started?

doc/syslog/syslog-rate-limit-design.md
nazariig: @Junchao-Mellanox makes sense

doc/syslog/syslog-rate-limit-design.md
Junchao-Mellanox: Please check hostcfgd logic: https://github.com/sonic-net/sonic-host-services/blob/bc8698d1d760fefedaeb4742ad19b25ef2b3c17b/scripts/hostcfgd#L1661. Hostcfgd will delay itself by waiting for other system services. Thus, rsyslog-config.service and rsyslog.service should be started before hostcfgd,

zhangyanzhao:
@qiluo-msft  can you please merge this PR? Thanks.

kannankvs:
@Junchao-Mellanox  @liat-grozovik  The document seems to have special characters and hence the page deployment workflow is failing. Can you check if the failure on page deployment workflow is not due to this merge ?

Junchao-Mellanox:
> @Junchao-Mellanox @liat-grozovik The document seems to have special characters and hence the page deployment workflow is failing. Can you check if the failure on page deployment workflow is not due to this merge ?

Hi, what do you mean by "page deployment workflow"? where can I get the log?

zhangyanzhao:
@StormLiangMS  can you please back port this to 202211? Thanks.

PR #1054, opened 2022-08-11
Adding P4RT app configuration to hld.

ntoorchi:
P4RT app configuration is added to HLD.


ntoorchi:
@rhalstea @prsunny please review.

rhalstea:
Thanks @ntoorchi 

@prsunny do you mind reviewing? This change is just updating the existing HLD to include an example config based on sonic-net/sonic-buildimage#10725

prsunny (approved):

PR #1084, opened 2022-09-19
Port Profile Init HLD

noaOrMlnx:
Add Port Profile Init feature HLD

The purpose of this HLD is to use BULK approach for port creation and reduce uptime after reboot.


doc/port-profile-init/port-profile-init-design.md
```diff
@@ -0,0 +1,179 @@
...
+#### Initialization Flow
+
+As part of PortsOrch initialization we query for all pre-configured ports in SAI and put that in a list for the later comparison logic.
+
+Now that SAI will no longer initialize all ports from SAI profile file, this query is not relevant.
+It will no longer be part of PortsOrch initialization and will only be done in case bulk option is not supported.
+
+2 checks will be performed in initialization flow.
+1. Is bulk option supported? It will be checked using the following flow:
+
```
venkatmahalingam, line 99: Can we have the explicit switch level capability check for port bulk support instead implicitly assuming with create_ports API? May be we can refer SAI_SWITCH_ATTR_SUPPORTED_FAILOVER_MODE

https://github.com/opencomputeproject/SAI/blob/master/inc/saiswitch.h#L2557

doc/port-profile-init/port-profile-init-design.md
```diff
@@ -0,0 +1,179 @@
...
+### SAI API
+
+https://github.com/opencomputeproject/SAI/pull/1460/files SAI PR introduces new SAI APIs to support bulk creation.
+
+sai_port module will now have the following APIs:
+
+```
+sai_bulk_object_create_fn              create_ports;
+sai_bulk_object_remove_fn              remove_ports;
+sai_bulk_object_set_attribute_fn       set_ports_attribute;
```
venkatmahalingam, line 132: Please see if we can use bulk set_ports_attribute instead of individual set attribute API.

doc/port-profile-init/port-profile-init-design.md
```diff
@@ -0,0 +1,179 @@
...
+
+The above logic requires multiple calls to SAI APIs and can take significant amount of time, which can challenge fast-boot time restriction.
+
+SAI introduced new capability which allows using bulk operation for creating any type of object. Using these capability will allow us saving API calls and decrease ports creation time.
+
+### Requirements
+
+* SONiC will support both bulk approach and legacy approach.
+* Before creating the ports, SONiC will check if bulk operation is supported by SAI. If not, it will fallback to legacy approach.
+* Ports creation bulk approach in SONiC expectes SAI to not create the port in create_switch() phase.
```
venkatmahalingam, line 53: Can we share the performance data for ports create as part of create_switch (SAI profile file) vs bulk ports create from portsorch?

doc/port-profile-init/port-profile-init-design.md
```diff
@@ -0,0 +1,179 @@
...
+
+In order to support feature in SONiC side, Port configuration flow will be changed in sonic-swss PortsOrch module.
+
+<p align=left>
+<img src="img/old_flow.svg" alt="Figure 1. Old Flow">
+</p>
+
+
+<p align=left>
+<img src="img/new_flow.svg" alt="Figure 2. New Flow">
```
venkatmahalingam, line 77: In the new flow, instead of keeping the port information in the queue until we receive PORT_CONFIG_DONE, can we keep the port information in the local cache and process them upon getting PORT_CONFIG_DONE in portsorch? this helps to avoid processing the pending messages every second from queue.

doc/port-profile-init/port-profile-init-design.md
madhupalu, line 53: This requirement must be removed from the HLD and we should support both legacy static mode (via sai.profile)and bulk api.

doc/port-profile-init/port-profile-init-design.md
```diff
@@ -0,0 +1,179 @@
...
+SAI introduced new capability which allows using bulk operation for creating any type of object. Using these capability will allow us saving API calls and decrease ports creation time.
+
+### Requirements
+
+* SONiC will support both bulk approach and legacy approach.
+* Before creating the ports, SONiC will check if bulk operation is supported by SAI. If not, it will fallback to legacy approach.
+* Ports creation bulk approach in SONiC expectes SAI to not create the port in create_switch() phase.
+* Before creating the ports, even if bulk is suported, SONiC will check if SAI created ports as before. If so, it will fallback to legacy approach.
+* SONiC will send the bulk request only when all ports are configured in APP DB, PORT table. Meaning, only when "PortConfigDone" flag is raised.
+* In case create_ports() API is called with an empty list of ports, SONiC expects SAI to return a status different from SAI_STATUS_NOT_IMPLEMENTED or SAI_STATUS_NOT_SUPPORTED. This call will be made in order to check if bulk approach is supported and implemented in vendor's SAI.
```
madhupalu, line 56: How does the bulk API handles or propagate errors back to caller/ application in case of errors while do bulk ports?

doc/port-profile-init/port-profile-init-design.md
```diff
@@ -0,0 +1,179 @@
...
+* Execute regression test for fast boot and perform time analyze, in order to make sure fast-boot time decreased as expected. In addition, manual time testing for fast-reboot up time.
+* A new sonic-mgmt test will need to be added. the test will check fast-reboot with dynamic port breakout.
+* A full regression testing to make sure evertything is working as before.
+
+
+#### Unit Test cases
+
+A new test will be added to DVS tests.
+
+The test will push ports configuration to Config DB and expect ASIC_STATE:SAI_OBJECT_TYPE_PORT* keys count to be as the number of the ports in Config DB.
```
madhupalu, line 172: please add error cases for bulk api create or delete failures?

doc/port-profile-init/port-profile-init-design.md
```diff
@@ -0,0 +1,179 @@
...
+<img src="img/new_flow.svg" alt="Figure 2. New Flow">
+</p>
+
+
+The above sequence diagram shows that a lot of calls to SAI and SDK APIs are being saved.
+In terms of Fast Reboot, when system is going up, it can save a lot of time and not challenge the 30 seconds restriction.
+
+Notes:
+
+* Multi-Asic is covered for these changes. In Multi-Asic environment, there are multiple instances of swss and Redis DB and each DB has it's own PORT table. Hence, port configuration is per Asic.
```
madhupalu, line 86: Can you list what are the use cases need bulk API support? what is wrong with existing SAI create ports? Is that DPB needs bulk API support ?

doc/port-profile-init/port-profile-init-design.md
noaOrMlnx: Done

doc/port-profile-init/port-profile-init-design.md
noaOrMlnx: Done

doc/port-profile-init/port-profile-init-design.md
noaOrMlnx: Done

doc/port-profile-init/port-profile-init-design.md
noaOrMlnx: Bulk API is needed in order to save time calling the regular APIs one by one. 
It doesn't mean there is something wrong with the current approach, just that there is a faster way which will help the initialization to be faster and the plan is to save time for fast-boot so the uptime will not cross the 30 seconds limitation.

doc/port-profile-init/port-profile-init-design.md
noaOrMlnx: will do once we will have it.
But please note that performance is depend on vendor's implementation so can change between vendors.

doc/port-profile-init/port-profile-init-design.md
noaOrMlnx: Please check the design after changes. 
We allow vendor to implement create_ports() API while not implementing other APIs like remove_ports and set_ports_attribute. 
for this purpose, we will have to check for each API if it is implemented. 

Furthermore, if we will use the suggested approach, it means we will need to wait for PortConfigDone before configuring any port, even if bulk is not supported. and if it is not supported, syncd will translate the bulk command to the regular one. this can delay the process we have today and will not gain the purpose we wanted.

zhangyanzhao:
@noaOrMlnx  can you please add the code PRs into this HLD by referring to #806

liat-grozovik:
@zhangyanzhao the doc is updated based on comments for more than a month with no feedback.
I suggest to merge it.

liat-grozovik (approved):

PR #1089, opened 2022-09-28
Update CLA part with up-to-date info

mathieulaunay:
- Update CLA part with up-to-date info
- fix markdown syntax


lihuay (approved):
Thanks for fixing this!

PR #1092, opened 2022-10-02
Added doc/sonic-front-pannel-port-prefix-regex.md

itamar-talmon:
In order to support distinguishing between different types of front panel ports in a maintainable fashion.
Specifically, we are planning to bring up a system with 'service' ports (in addition to the regular ethernet data ports) - these
are lower speed ports that used for connection to accelerators, internal loopbacks and more.


itamar-talmon:
@lihuay - Hi I created new PR with the correct user (last one was not signed the CLA). Thanks.

Blueve:
Should this be a part of doc: https://github.com/sonic-net/SONiC/blob/master/doc/sonic-port-name.md ?

yxieca:
> Should this be a part of doc: https://github.com/sonic-net/SONiC/blob/master/doc/sonic-port-name.md ?

I agree with this suggestion. The design sounds benight but the actual change touches quite some code. Can we call out these areas needs to be changed?

itamar-talmon:
I will update  https://github.com/sonic-net/SONiC/blob/master/doc/sonic-port-name.md and the scope of this change. Thanks

itamar-talmon:
@yxieca - I have updated  https://github.com/sonic-net/SONiC/blob/master/doc/sonic-port-name.md and added details about the scope of this change. It is simple but touches many places.

lguohan:
@itamar-talmon , we should have a sync with @rlhui on what has already been added to chassis, i remember we have some internal links, loopback defined for chassis system. we need to check what exist today and see if some of the new things can be accomodated by chassis proposal already?

itamar-talmon:
Hi @rlhui - could you please review this proposal?

itamar-talmon:
I close this since I opened new one from a new branch that is not my master - https://github.com/sonic-net/SONiC/pull/1777

PR #1095, opened 2022-10-08
teamd warm-restart in fast mode hld

tianshangfei:
We expect that the restart of teamd docker should not cause link flapping or any traffic loss. All lags at data plane should remain the same. But it's hard to implement in some scenarios.

During teamd warm-restart, the control plane remains up for a maximum of 90 seconds in LACP slow mode. However, in LACP fast mode, the control plane can only remain up for 3 seconds. This is because LACPDUs are sent every second. LACP protocol considers a LAG to be down if three LACPDUs are not received.

Teamd containers are not restarted that fast, so teamd warm-restart in LACP fast mode always results in lag down and the kernel LAG state in mess. However, in a data center, it is necessary to set LACP to fast mode to ensure faster link convergence and less traffic loss.

Therefore, supporting teamd warm-restart in LACP fast mode is very important. With this feature, we can support teamd bug hotfix and smooth upgrades.

This design supports teamd warm-restart in fast mode by switching between active and standby teamd container.
Signed-off-by: tianshangfei <tian_shangfei@163.com>


zhangyanzhao:
HLD was reviewed on Oct 11, 2022. @tianshangfei  can you please help to add the code PRs to this HLD? Thanks.

tianshangfei:
> HLD was reviewed on Oct 11, 2022. @tianshangfei can you please help to add the code PRs to this HLD? Thanks.

The code PR is not ready yet, and the implementation has been delayed because some urgent work needs to be done.

PR #1106, opened 2022-10-26
[system-health] Led color shall be controlled by configuration when system is booting

Junchao-Mellanox:
Currently, system status LED is always red during booting stage. User cannot tell if it is a real issue or system is under initialization. This HLD change introduces a booting stage for the system health daemon and allow user to configure the system status led color for booting state.

PRs:
| [[system-health] Led color shall be controlled by configuration when system is booting ](https://github.com/sonic-net/sonic-buildimage/pull/12487) | GitHub issue/pull request detail | GitHub pull request check contexts |


keboliu (approved):

doc/system_health_monitoring/system-health-HLD.md
```diff
@@ -125,16 +125,16 @@ The configuration shall be:
...
+### 1.5 System status LED color definition
 
-default system status LED color definition is like
+System status LED is set based on the status of system. The system status is defined as following
+
+ | Status           | Led Color            | Description                                                               |
+ |:----------------:|:--------------------:|:-------------------------------------------------------------------------:|
+ | booting          | ${led_color.booting} | System up time is less that 5 mins and not all services/devices are ready |
+ | normal           | ${led_color.normal}  | All services/devices are in good state                                    |
+ | fault            | ${led_color.fault}   | Not all services/devices are in good state                                |
```
liat-grozovik, line 136: Do get into fault state also when boot time is more than 5 min and not all is ok, right?

Junchao-Mellanox: yes

liat-grozovik:
@zhangyanzhao I suggest to go a head with the merge. Any objection?
we would like this change to be applied to 202205 and master/202211 as it is considered as 'bug fix'

doc/system_health_monitoring/system-health-HLD.md
```diff
@@ -148,8 +148,15 @@ Considering that different vendors platform may have different LED color capabil
...
 ```
 
-The field "booting" is deprecated because there is no booting stage anymore. For backward compatible, user can still configure this field but it won't take effect.
+If not configuration is provided, default configuration is used:
 
+```python
+DEFAULT_LED_CONFIG = {
+	'fault': 'red',
+	'normal': 'green',
+	'booting': 'red'
```
prgeor, line 157: can we have some other color other than 'red' which is already used for indicating 'fault'?

Junchao-Mellanox: Before this PR, there is no booting stage. So, only two cases:
1. System is all good, LED is green
2. System is not all good, LED is red

Booting stage is introduced by this PR. So, setting "red" as the default value for booting stage is for backward compatible. User can override this value by using a system health configuration file.

PR #1107, opened 2022-10-26
Preserve CoPP table HLD

arfeigin:
This is HLD for preserving CoPP table during reboot in order to improve fast-reboot.

This will enable faster LAG creation which will shorten dataplane downtime.


dgsudharsan (approved):

vaibhavhd:
Can this be presented in a community meeting for assessing if any other platform can benefit from this change?

prsunny (approved):

zhangyanzhao:
HLD is reviewed by community on 1/3/2023. No YANG model change, no CLI change.

doc/copp/Preserve_COPP_tables_to_improve_fast-reboot.md
```diff
@@ -0,0 +1,46 @@
...
+
+With the new proposal, the CoPP tables contents will be preserved during reboot, i.e. they won't be cleared by DB migrator. Then, initializing CoPP tables in startup phase for any key-value entry it will be checked if such entry exists, in case it does, the entry will be ignored. In case there is an entry with the same key but with different value preserved from prior reboot the existing entry will be deleted and a new entry will be added to the CoPP tables with the key and the new value.
+In addition, for backwards compatibility, in case CoPP tables preserve an entry with a key that is not supported (i.e. such key is not present in the json default initialization file) it will be deleted from the CoPP tables during merge.
+The solution of deleting old entry and creating a new one instead is proposed since there is no SAI implementation to check for overwrites and this might lead to trying to re-create create-only entries which will cause orchagent crash.
+
+# Flows
+
+## DB migrator copp tables handling logic
+
+The following flow captures the DB migrator proposed functionality.
```
rck-innovium, line 32: It is not clear if there is any differences in the calls made to syncd/SAI. Can you explain why this check needs to be platform specific? Please also explaine the difference in terms of SAI API calls?

doc/copp/Preserve_COPP_tables_to_improve_fast-reboot.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Design Proposal
+
+## Current behavior
+
+In the current implementation DB migrator clears CoPP tables contents and it is being initialized with default values at startup. This process takes some time and leads to that LACP are being missed shortly after reboot since LACP trap is not set yet, thus delaying LAG creation and extending dataplane downtime during fast-reboot.
+
+## Proposed behavior
+
+With the new proposal, the CoPP tables contents will be preserved during reboot, i.e. they won't be cleared by DB migrator. Then, initializing CoPP tables in startup phase for any key-value entry it will be checked if such entry exists, in case it does, the entry will be ignored. In case there is an entry with the same key but with different value preserved from prior reboot the existing entry will be deleted and a new entry will be added to the CoPP tables with the key and the new value.
```
svshah-intel, line 24: can you please clarify if current startup coppmgr logic would still be performed or only new merge logic in case when app-db is not cleared in migrator

arfeigin: This is just an enhancement of the current startup coppmgr logic. It uses the existing merge logic and adds a new check for duplicates or overwrites.

doc/copp/Preserve_COPP_tables_to_improve_fast-reboot.md
```diff
@@ -0,0 +1,46 @@
...
+In addition, for backwards compatibility, in case CoPP tables preserve an entry with a key that is not supported (i.e. such key is not present in the json default initialization file) it will be deleted from the CoPP tables during merge.
```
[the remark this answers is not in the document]
arfeigin: The file referred to is /etc/sonic/copp_cfg.json which is being parsed in coppmgr startup as the default configuration. It is being merged with user configuration but as far as I understand user can not add other keys not supported in the default init file.
The implementation implies that any key present either in the json init file or by user configuration and passed the existing merge config logic is considered as supported - other preserved keys will be deleted as treated as invalid keys.
image
...
image

doc/copp/Preserve_COPP_tables_to_improve_fast-reboot.md
arfeigin: No difference in the calls made to syncd/SAI - the only difference is that it will avoiding trying to set duplicates. In case of overwrites, it will delete existing entry (which currently is done by db migrator) and set the new entry (which is done as well in the current flow of coppmgr startup).

doc/copp/Preserve_COPP_tables_to_improve_fast-reboot.md
svshah-intel: ok, looking at the code it is clear

doc/copp/Preserve_COPP_tables_to_improve_fast-reboot.md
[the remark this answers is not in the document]
svshah-intel: what code is doing makes sense. the description here in document does not match. my understanding is that user configured entries can allow to create traps not present in default copp_cfg.json file. And thus you allow entries from preserved ones to carry forward if it exists in default or current user-config (it is not only ones from the default copp_cfg.json file)

PR #1111, opened 2022-10-31
Address configure clis

qinchuanares:
remove unnecessary -- in config CLI commands


mihirpat1 (commented):
It will be great if you can update the spec with the below for PM since I couldn't find it.

1. post-fec errored frames ratio for min, avg and max
2. evm min, avg and max
3. cd min, avg and max
4. Laser temperature

qinchuanares:
> It will be great if you can update the spec with the below for PM since I couldn't find it.
> 
> 1. post-fec errored frames ratio for min, avg and max
> 2. evm min, avg and max
> 3. cd min, avg and max
> 4. Laser temperature

I included the min/max specs for important optical and DSP specs.

mihirpat1 (approved):

prgeor (approved):

PR #1124, opened 2022-11-21
Use empty target and origin to support new GNMI config interface.

ganglyu:
sonic-gnmi needs to support sonic management framework and new GNMI config interface at the same time.
sonic-gnmi does not use origin field for translib, so we propose to use empty target and specified origin to support new GNMI config interface.


ganglyu:
@adyeung @anand-kumar-subramanian @venkatmahalingam 
https://github.com/openconfig/gnmi/blob/master/proto/gnmi/gnmi.proto#L144
We will use origin field to support new GNMI config interface, but I'm not sure if sonic management has any plan to use this field. Is there any conflict with this origin field?
Would you please confirm?

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -456,25 +456,14 @@ The full configuration request will be overwritten by subsequent full configurat
...
-      STATE_DB        = 6;
-      // For none-DB data
-      OTHERS          = 100;
-      // For mixed schema
-      MIXED_SCHEMA    = 101;
-    }
+SONiC telemetry is using empty target to support sonic management framework.
+
+We will below solution to support new GNMI config interface:
+<img src="images/single_target.svg" alt="overwritten-config" width="1000px"/>
```
qiluo-msft, line 462: Suggest merge this diagram into original https://github.com/sonic-net/SONiC/blob/master/doc/mgmt/gnmi/images/data_flow.svg
No need add extra diagram.

qiluo-msft (dismissed):

PR #1128, opened 2022-11-24
Update Auto_ts doc to include orchagent abort case

vivekrnv:
Update the Auto-TS HLD to include a special handling when orchagent aborts due to SAI programming failure

| Repo  | PR title |
| ---------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------- |
| sonic-buildimage | [Update syncd stop script to collect saisdkdump during orch abort](https://github.com/sonic-net/sonic-buildimage/pull/12936)  |
| sonic-swss | [[orchagent] Set ABRT signal in STATE_DB during a SAI failure](https://github.com/sonic-net/sonic-swss/pull/2556)  |
| sonic-utilities | [Updated TS and auto-TS to collect orch abrt saisdkdump files](https://github.com/sonic-net/sonic-utilities/pull/2540)  |


dgsudharsan (approved):

doc/auto_techsupport/auto_techsupport_and_coredump_mgmt.md
```diff
@@ -458,12 +460,45 @@ Although if the admin feels otherwise, these values are configurable.
...
 ### 7.8 Techsupport Locking
 
-Recently, an enhancement was made to techsupport script to only run one instance at a time by using a locking mechanism. When other script instance of techsupport tries to run, it'll exit with a relevent code. This would apply nevertheless of how a techsupport was invoked i.e. manual or through auto-techsupport. 
+Recently, an enhancement was made to techsupport script to only run one instance at a time by using a locking mechanism. When other instance of techsupport tries to run, it'll exit with a relevent code. This would apply nevertheless of how a techsupport was invoked i.e. manual or through auto-techsupport. 
 
 With this change, rate-limit-interval of zero would not make any difference. The locking mechanism would implicitly impose a minimum rate-limit-interval of techsupport execution time.  And since, the techsupport execution time can't be found out and varies based on underlying machine and system state, the range of values configurable for the rate-limit-interval is left unchanged
 
 A relevant message will be logged to syslog when the invocation fails because of LOCKFAIL exit code. 
 
+### 7.9 Orchagent abort consideration
```
liat-grozovik, line 469: @vivekrnv  it is ok to have the implementation for Nvidia/Mellanox syncd only, the question if the flow can be invoked on any ASIC vendor if they will add the support for that. if so, I think it should be considered as generic based on code availability yet as any other features in SAI. what do you think?

doc/auto_techsupport/auto_techsupport_and_coredump_mgmt.md
```diff
@@ -458,12 +460,45 @@ Although if the admin feels otherwise, these values are configurable.
...
+```
+ORCH_ABRT_STATUS
+<1|empty>
+
+Eg:
+root@sonic:/home/admin# sonic-db-cli STATE_DB GET ORCH_ABRT_STATUS
+1
+```
+
+During sai programming failure, orchagent will set the status to ORCH_ABRT_STATUS flag in STATE_DB. syncd.sh script checks if the ORCH_ABRT_STATUS flag is set in STATE_DB before stopping the syncd container and if yes proceeds with collecting saisdkdump to `/var/log/orch_abrt_saisdkdump/` on the host and also creates a file under /tmp named 'saidump_collection_notify_flag'. This is used to synchronize b/w auto-techsupport and syncd.
```
liat-grozovik, line 497: assuming it will be generic and some ASIC vendors will nor refer to the new state db adds, what will be the system behaviour?

doc/auto_techsupport/auto_techsupport_and_coredump_mgmt.md
vivekrnv: code availability will be present for all members. Every SAI vendor is expected to implement sai_dbg_generate_dump call which is used in saisdkdump. But it's not possible to determine if the dump is important for a particular vendor. As we already know only Nvidia is using saisdkdump according to techsupport. 

So, i think we should keep it specific to Nvdia for now. if and when other vendors decide if it's important, they can add enable this for their platform.

doc/auto_techsupport/auto_techsupport_and_coredump_mgmt.md
vivekrnv: Orchagent will write to STATE_DB irrespective of the vendor. In the syncd.sh scrip
    if [[ x"$(${SONIC_DB_CLI} STATE_DB GET ORCH_ABRT_STATUS)" == x"1" ]]; then
        # Collecting saisdkdump before restarting syncd
        # Runs when orchagent is aborted because of SAI failure.
        # Only enabled for mellanox platform
        if [[ x$sonic_asic_platform == x"mellanox" ]]; then
            collect_saisdkdump
        fi
        # This is used to notify auto-techsupport process
        touch /tmp/saidump_collection_notify_flag
    fi

liat-grozovik (approved):

zhangyanzhao:
@liat-grozovik  is this closed intentionally?

dgsudharsan:
> @liat-grozovik is this closed intentionally?

@zhangyanzhao We have a new design to cover this https://github.com/sonic-net/SONiC/pull/1212

PR #1203, opened 2023-01-03
Config Reload Enhancement

dgsudharsan:
Design to address the issue https://github.com/sonic-net/sonic-buildimage/issues/12071
Enhancing config reload to sequence the services and faster system initialization.

- Immediately restart the critical services during config reload.
- The non critical should be started only after all the ports are initialized.
- Services can be configured to be started immediately or delayed. This can be using a field in FEATURE table.
- The existing timers should be removed by this event driven approach.
- This flow is applicable in case of all reboots (warm/fast/cold) as well as config reload.


skg-net:
@dgsudharsan can you please add how this works along with ZTP? ZTP might use some of the GNMI services during reload cases.

dgsudharsan:
> @dgsudharsan can you please add how this works along with ZTP? ZTP might use some of the GNMI services during reload cases.

@skg-net I don't think it is going to affect ZTP. In the ZTP too, there will be swss initialization, followed by PortInitDone message posted to APP_DB. Once it is received, all the delayed services include telemetry and mgmt-framework should start. Even today they start in a delayed fashion (3m 30 seconds) but with this change, they should be starting soon after port initialization.

doc/config_reload/config_reload_enhancement.md
```diff
@@ -0,0 +1,122 @@
...
+
+#### DB Migrator
+
+The 'has_timer' field in FEATURE table will be changed to 'delayed'. Hence db_migrator is required to modify the configurations to reflect this change.
+
+### YANG model changes
+
+Yang model needs to be updated for FEATURE_TABLE. The 'has_timer' field will be removed and replaced with 'delayed'
+
+```
```
prvattem, line 96: When and how is this flag set per feature?

doc/config_reload/config_reload_enhancement.md
```diff
@@ -0,0 +1,122 @@
...
+### Warmboot and Fastboot Considerations
+
+In case of Warmboot and fastboot, Hostcfgd itself is currently delayed today by a timer. This timer will also be removed and replaced by hostcfgd waiting for warmboot/fastboot completion using waitAdvancedBootDone.
+Additionally there will be checks added to Warmboot command to ensure all services including delayed services are started before proceeding.
+
+### Testing Design
+
+#### Unit tests
+Hostcfgd tests would be enhanced to cover the new flow.
+- Ensure the service start are done for critical services immediately without delay.
```
prvattem, line 115: What is the criteria to decide whether a service is critical or not?

doc/config_reload/config_reload_enhancement.md
```diff
@@ -0,0 +1,122 @@
...
+During the hostcfgd initialization it will cache these delayed services based on the configuration in the feature table. The hostcfgd will also subscribe to PORT_TABLE in the APPL_DB. Once the switch is initialized and all the ports are created in ASIC and Kernel the PortSyncd will publish PortInitDone key in the APPL_DB. On receiving this key the hostcfgd will go through the delayed services list and enables them. There will also be a timeout defined for the services to start if PortInitDone is not defined within the specific timeout. This is to ensure the management related services start even if there is some failure for the switch to initialize.
+
+
+The below diagram explains the sequence when config reload is executed. 
+![](/images/config_reload/Enhance_config_reload.JPG)
+The below diagram explains the new sequence inside hostcfgd.
+![](/images/config_reload/Hostcfgd_flow_for_delayed_services.JPG)
```
prvattem, line 74: Please tag*yes/no) the flows towards "Enable Service" & "Disable Service"

doc/config_reload/config_reload_enhancement.md
```diff
@@ -0,0 +1,122 @@
...
+### SAI API Requirements
+None
+
+### Configuration and management
+
+#### Config command
+
+No new commands are introduced as part of this design.
```
prvattem, line 82: Apparently modifying critical/non-critical is not straight forward.
Providing a config knob would be good(might internally include changing the feature table and the service file).
Also providing the current snapshot of tagged services is also a good idea.

doc/config_reload/config_reload_enhancement.md
```diff
@@ -0,0 +1,122 @@
...
+Primary requirements for sequencing the config reload are
+- Immediately restart the critical services during config reload.
+- The non critical should be started only after all the ports are initialized.
+- Services can be configured to be started immediately or delayed. This can be using a field in FEATURE table.
+- The existing timers should be removed by this event driven approach.
+- This flow is applicable in case of all reboots (warm/fast/cold) as well as config reload.
+
+
+### High-Level Design
+Currently hostcfgd controls the services based on the feature table. The feature table has a specific field 'has_timer' for the non essential services which needs to be delayed during the reboot flow. This field will be now replaced by new field called "delayed". These services will controlled by hostcfgd.
```
prsunny, line 67: what if just hostcfgd is restarted, since it's now subscribing for portinitdone?

doc/config_reload/config_reload_enhancement.md
dgsudharsan: Currently we have feature table defined in init_cfg.j2 and these have has_timer field. This field will be replaced with  'delayed'. Additionally the user can configure this field in config_db.json

doc/config_reload/config_reload_enhancement.md
dgsudharsan: It is determined as per the existing approach where we already delay a list of services. In my opinion all the services that are required for faster initialization and traffic convergence are critical. All monitoring services like snmp, lldp, telemetry can be delayed

doc/config_reload/config_reload_enhancement.md
dgsudharsan: Hostcfgd will restart the delayed services on seeing the portinitdone. If the services are already started this will be a no op. Today hostcfgd does the same as we already start many services before hostcfgd starts and we encounter service twice. The 2nd time through hostcfgd is a no-op.

doc/config_reload/config_reload_enhancement.md
dgsudharsan: Yes. Modifying the feature table in config_db.json is in line with the approach where the user can decide if a service needs to be delayed or not.

doc/config_reload/config_reload_enhancement.md
dgsudharsan: Done

doc/config_reload/config_reload_enhancement.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Testing Design
+
+#### Unit tests
+Hostcfgd tests would be enhanced to cover the new flow.
+- Ensure the service start are done for critical services immediately without delay.
+- Ensure the service start for timers are programmed immediately for delayed services.
+- Ensure the delayed services are started as soon as PortInitDone is seen in APPL_DB table
+
+#### System tests
```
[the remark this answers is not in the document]
dgsudharsan: Done

liat-grozovik:
@prvattem @prsunny  would you please help to review/approve the HLD? code PRs should be available soon and better to close HLD before

stepanblyschak (approved):

liat-grozovik:
@prsunny @prvattem could you please review following feedback provided/answered?

liat-grozovik:
@zhangyanzhao this HLD is ready for weeks now and already passed community review. what is next?

venkatmahalingam (approved):

prvattem (approved):

liat-grozovik (approved):

skg-net:
@dgsudharsan Can you please update the Quality Metric (Alpha/Beta/GA) for the feature either in this PR comments or in HLD itself based on https://github.com/sonic-net/SONiC/blob/master/doc/SONiC%20feature%20quality%20definition.md
Thanks

PR #1211, opened 2023-01-07
Add firmware update control HLD

fastiuk:
Add an ability to control ASIC FW update
* Ability to enable/disable FW update
* Ability to choose FW source: image/user-defined

Signed-off-by: Yevhen Fastiuk <yfastiuk@nvidia.com>


liat-grozovik:
closing this HLD to re think of the right approach for it

PR #1231, opened 2023-01-23
Reset factory hld

Mohammedz93:
This PR provides high level design for Reset factory.

| [[sonic-buildimage] Support Reset factory ](https://github.com/sonic-net/sonic-buildimage/pull/14105) | GitHub issue/pull request detail | GitHub pull request check contexts
| [[sonic-mgmt] Reset factory tests ](https://github.com/sonic-net/sonic-mgmt/pull/7652) | GitHub issue/pull request detail | GitHub pull request check contexts


skg-net (commented):
Thanks for the HLD. 

1) Can you elaborate on the keep-basic?  I see you have mentioned few tables, how the dependencies between the tables will be addressed. Is there a guideline which tables are allowed etc?
2) Can you add details how ssh keys will be handled during factory reset? 
3) I assume the users will also be deleted during factory reset, in that case I'm not sure why to retain the home directory.
4) Is there a clear logs command in SONiC? if yes, how is keep-all-config different from clear logs command?

doc/reset_factory/ResetFactoryHLD.md
```diff
@@ -0,0 +1,210 @@
...
+It will allow user to track operations such as "reset-factory".
+
+6. Reboot
+
+
+## 3 <a name='Functionality'></a>Functionality
+###  3.1 <a name='Reset-Factory'></a>Reset-Factory
+
+##### Default
+It will generate the default config_db.json file by running command:<br/>
```
madhupalu, line 111: What is really factory reset means? How is different from brand new device vs the factory reset device? 
Ex: Use case - if RMA device or send device back to vendor wanted to delete all the data for security reasons?
1. Remove /etc/sonic files 
2. users & passwd
3. dockers overlay filesystem

doc/reset_factory/ResetFactoryHLD.md
```diff
@@ -0,0 +1,210 @@
...
+```
+
+
+## 4 <a name='CLI'></a>CLI
+
+#### reset-factory
+```
+==============================================================================
+root@host:~$ sudo reset-factory --help
+ Usage:  reset-factory < keep-all-config | only-config | keep-basic >
```
venkatmahalingam, line 165: Add 'none' or 'default' option as well.

doc/reset_factory/ResetFactoryHLD.md
```diff
@@ -0,0 +1,210 @@
...
+#### config-setup factory
+```
+==============================================================================
+root@host:~$ sudo config-setup factory --help
+ Usage:  config-setup factory < keep-basic >
+
+         Create factory default configuration and save it to
+         to ${CONFIG_DB_JSON}.
+
+         keep-basic  - Preserves basic configurations only.
```
venkatmahalingam, line 186: keep-basic name doesnot look right, is it possible to have a default 'filter-list' with tables (e.g MGMT_PORT,MGMT_INTERFACE,PASSW_HARDENING), hope user can dynamically provide the tables to filter in a file as an option during factory-reset command execution..

doc/reset_factory/ResetFactoryHLD.md
Mohammedz93: There are a few use cases:
1. System was tested in the vendor's lab prior to shipment to customer. before shipment any configuration should be deleted. but software installation should remain as is.
2. Customer got to configuration conflict and wishes to start from scratch.
3. Wiping information for security reason, before moving the system (including the case of shipping back to the vendor).

I would say that we want the system to **function** as a newly manufactured system.
This does not mean that the file system must be in the same state.
We want to achieve that by:
-  Resetting configurations to defaults (config_db.json)
-  Clear files that may potentially affect the functionality or user experience :
    This is very challenging as sudo user can change almost any file/directory in the system.
    We will try to cover the files that might be changed by the system.
    If the user choose to change the file system directly rather than using the CLI/official APIs, then the responsibility falls on him.
    
 1. Remove /etc/sonic files:
 I believe we can clear the /etc/sonic directory under the upperdir of overlayfs (e.g. /host/image*/rw/etc/sonic)
 but I need to test it in different scenarios first.
 
 2. users & passwd:
This will be handled by [User MGMT](https://github.com/sonic-net/SONiC/pull/1048) feature.
but since Reset Factory might be merged before User MGMT, we will handle this also.
We will delete non-default users and configure the default password for default users (e.g. admin )
 
 3. dockers overlay filesystem:
 As discussed in the HLD review meeting, this will be covered by removing all docker containers on system.
 
 I will update this HLD with the latest changes very soon,

doc/reset_factory/ResetFactoryHLD.md
```diff
@@ -0,0 +1,210 @@
...
+SSH_SERVER,USER_TABLE,ROLE_TABLE
+```
+
+
+## 4 <a name='CLI'></a>CLI
+
+#### reset-factory
+```
+==============================================================================
+root@host:~$ sudo reset-factory --help
```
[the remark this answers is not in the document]
Mohammedz93: Why we need yang model ? isn't only for DB tables ?
We add any new DB table.

doc/reset_factory/ResetFactoryHLD.md
```diff
@@ -0,0 +1,210 @@
...
+
+We will extend this list with the following tables after the merge of features:
+```
+SSH_SERVER,USER_TABLE,ROLE_TABLE
+```
+
+
+## 4 <a name='CLI'></a>CLI
+
+#### reset-factory
```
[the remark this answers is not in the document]
Mohammedz93: "reset" is already utilized in Linux for terminal initialization. please see [reset](https://linux.die.net/man/1/reset)
The idea was to implement it as bash script ( same as config-setup) for better performances, as we are going only to clear files and run Linux shell commands.

doc/reset_factory/ResetFactoryHLD.md
Mohammedz93: I think it would be confusing to print something like:
`Usage:  reset-factory < none | keep-all-config | only-config | keep-basic >`
I think it will be better to explain about the default flow before the parameters (line 185)

doc/reset_factory/ResetFactoryHLD.md
```diff
@@ -0,0 +1,210 @@
...
+Command: config-setup factory keep-basic<br/>
+It will clear system logs and files.<br/>
+
+##### Keep-all-config
+It will keep the same configuration but will clear system logs and files.
+
+##### Only-config
+It will generate the default config_db.json file by running command:<br/>
+config-setup factory only-config<br/>
+It won't clear system logs and files.
```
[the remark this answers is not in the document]
Mohammedz93: Please see my [comment](https://github.com/sonic-net/SONiC/pull/1231#discussion_r1088065516)
I believe this will cover it.
But I would like to know that exact flow and path of folder, Can you please share the relevant HLD/File ?

doc/reset_factory/ResetFactoryHLD.md
```diff
@@ -0,0 +1,210 @@
...
+    - it will restore "config_db.json".<br/>
+    - it will reboot the system if "sonic.target" already stopped.
+
+###  6 <a name='TestPlan'></a>Test Plan
+####  6.1 <a name='Cases'></a>Cases
+###### Good flow
+  - Run reset-factory without parameters
+  - Run reset-factory with keep-all-config/only-config/keep-basic
+  - Run config-setup factory without parameters
+  - Run config-setup factory with keep-basic
```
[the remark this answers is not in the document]
Mohammedz93: Are you referring to files under "/host/warmboot" ?
If so, I believe we can just clear this directory as we are going also to clear the reboot-cause folder "/host/reboot-cause"

doc/reset_factory/ResetFactoryHLD.md
```diff
@@ -0,0 +1,210 @@
...
+
+## 3 <a name='Functionality'></a>Functionality
+###  3.1 <a name='Reset-Factory'></a>Reset-Factory
+
+##### Default
+It will generate the default config_db.json file by running command:<br/>
+config-setup factory<br/>
+It will clear system logs and files.
+
+##### Keep-basic
```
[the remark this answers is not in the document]
Mohammedz93: We want to allow the user to preserve a basic set of configurations that he will likely always want to to preserve and settings necessary for the box to be reachable from the outside. 
So user won't need to reconfigure his MGMT interface, ssh servers and users of the system ( in future).

In this design, we are defining the default basic tables that we will preserve in the new "config_db.json" file. but vendors might choose to overwrite/extend this list by changing "config-setup..conf" file.

doc/reset_factory/ResetFactoryHLD.md
Mohammedz93: Please see my [comment](https://github.com/sonic-net/SONiC/pull/1231#discussion_r1088131084)
We need to take into account the user use case.
We want to allow the user to reset the system to factory settings without losing basic configurations.
I feel that "keep-basic" suits the use case and it explains the functionality of the command better than "filter-list".
Vendor can still extend/replace this list by changing "config-setup.conf" as explained in the design.

liat-grozovik:
@madhupalu  @venkatmahalingam please review HLD fixes following your comments.

liat-grozovik:
@madhupalu please review comments reply. if you have no further concern please approve the PR.
@venkatmahalingam and further comments from your side?

Note: if no other comments provided by end of this week, the HLD PR will be merged.

madhupalu (approved):

liat-grozovik:
@venkatmahalingam any further comment? if, not please approve so we can merge the HLD and start with the code PR review.

venkatmahalingam (approved):

skg-net (approved):

liat-grozovik (approved):

qiluo-msft:
@rajendra-dendukuri Could you help review (even it is already merged)?

doc/reset_factory/ResetFactoryHLD.md
```diff
@@ -0,0 +1,236 @@
...
+default user accounts names and default hashed passwords.
+The file will have same permissions as /etc/shadow "-rw-------" (600).
+```
+{
+    "admin": {
+        "password": "$y$j9T$qGjjyhXSHtuHwowV3DE44.$wrDK6/vecdYIE8ujXyVPkGHvlDHP.ySCsBTLvOdg.31"
+    }
+}
+```
+We will remove all non-default user accounts (e.g. 1000 < UID <= 6000),
```
qiluo-msft, line 129: You intention is to restore the /etc/shadow to the original one at build time. I propose to mount/host/image-CURRENT/fs.squashfs, and copy the original file to the current file system.

Mohammedz93: Thanks for the suggestion
I think it might be better to do the following ( similar to /etc/sonic clear step):
1. rm -f /host/image-CURRENT/rw/etc/shadow
2. mount -o remount /

I will test it and update the code accordingly

Mohammedz93: After investigating it more deeply, I found out that modifying the shadow file directly isn't recommended at all
The recommended way to update these file is through the official API commands ( useradd, usermod...)
In addition, it is not enough to restore shadow file only, we also should restore /etc/passwd, /etc/group
(probably some other files also)
I believe the current approach is much safer.

PR #1268, opened 2023-02-20
Egress Sflow Enhancement.

rajkumar38:
Updated existing sFlow HLD for egress Sflow support.

| Repo |  PR Title | State |
| ------------- | ------------- |------------- |
| sonic-utilities  | [[sflow] Added egress Sflow support](https://github.com/sonic-net/sonic-utilities/pull/2790)  | Open |
| sonic-swss  | [Egress Sflow Support ](https://github.com/sonic-net/sonic-swss/pull/2731) |  Merged  |
| sonic-buildimage  | [[sflow] Add egress sflow support](https://github.com/sonic-net/sonic-buildimage/pull/14630)  |  Merged  |


jeff-yin (approved):

doc/sflow/sflow_hld.md
```diff
@@ -380,9 +438,11 @@ AGENT_ID            = ifname        ; Interface name
 
 A new SFLOW_SESSION table would be added.
 ```
-key SFLOW_SESSION:interface_name
+key SFLOW_SESSION:interface_name|direction
 ADMIN_STATE     = "up" / "down"
-SAMPLE_RATE     = 1*7DIGIT      ; average number of packets skipped before the sample is taken
+SAMPLE_RATE     = 1*7DIGIT      ; average number of packets 
+skipped before the sample is taken
+direction can have value of  rx | tx
```
venkatmahalingam, line 445: do we need both direction as well here?

doc/sflow/sflow_hld.md
```diff
@@ -68,9 +68,8 @@ sFlow will be implemented in multiple phases:
 2. sFlow extended router support.
 
 ### Not planned to be supported:
-1. Egress sampling support.
-2. sFlow backoff mechanism (Drop the packets beyond configured CPU Queue rate limit).
-3. sFlow over vlan interfaces.
+1. sFlow backoff mechanism (Drop the packets beyond configured CPU Queue rate limit).
+2. sFlow over vlan interfaces.
```
venkatmahalingam, line 72: Do we need to add sFlow over port-channel interfaces?

doc/sflow/sflow_hld.md
```diff
@@ -68,9 +68,8 @@ sFlow will be implemented in multiple phases:
 2. sFlow extended router support.
```
venkatmahalingam, line 69: @rajkumar38 Do you know whether all 3 above phases are completed? if yes, mentioning the current status would be helpful.

dgsudharsan:
@vadymhlushko-mlnx FYI

zhangyanzhao:
This is reviewed in sonic community and recording is https://zoom.us/rec/share/ftw0da4bBQdfjGpKWj0KpRUS-Jn0FZiiUCD2Ia3oxByheb02XoJfvsKTgYIbDrRP.MaLAuB2VyRHtcHap

sflow:
Hello all,

The PSAMPLE netlink channel defines the integer attribute PSAMPLE_ATTR_SAMPLE_GROUP.  The sFlow daemon on SONiC, hsflowd,  expects ingress samples to appear with group=1,  and egress samples to appear with group=2.

Starting in version 2.0.45-1, hsflowd for SONiC will automatically be ready to accept egress samples if they appear on the netlink channel.  So no need to change anything else in the sflow container if you upgrade hsflowd to that release.

Regards,
Neil McKee, InMon Corp.

pphaal:
Hi rajkumar38,

While sFlow supports bidirectional packet sampling, it requires that the sampling  rate be the same in each direction, i.e. ingress_sampling_rate == egress_sampling_rate for bidirectionall sampled interfaces. 

I would suggest an option 3 that would simplify the proposal and minimize issues with backward compatibility:

```
"SFLOW_SESSION": {
    "Ethernet0|rx": {
       "admin_state": "down",
       "sample_rate": "40000",
       "sample_direction":"both"           
    },
```
The semantics of the proposed CLI settings should change:
`sflow interface sample-rate {interface-name} {value} [direction {rx|tx|both}]`
The sampling direction should be treated as part of the value, so changing an interface setting from "1000 direction ingress" to "1000 direction egress" would remove ingress sampling on the interface. 

The "show sflow" command should also minimize changes, i.e. first three columns should remain the same, a fourth column could be added to indicate direction:
```
Interface     Admin Status   Sampling rate  Sampling direction
---------     ------------   -------------      --------------
Ethernet0     down           40000                   both
```
Similarly the DB schema can be extended with a SAMPLE_DIRECTION attribute added to SFLOW_SESSION_TABLE. When migrating from an older SONiC version, the default sampling direction should be set to "ingress"

Regards,
Peter Phaal, InMon Corp.

doc/sflow/sflow_hld.md
```diff
@@ -68,9 +68,8 @@ sFlow will be implemented in multiple phases:
 2. sFlow extended router support.
 
 ### Not planned to be supported:
-1. Egress sampling support.
-2. sFlow backoff mechanism (Drop the packets beyond configured CPU Queue rate limit).
-3. sFlow over vlan interfaces.
+1. sFlow backoff mechanism (Drop the packets beyond configured CPU Queue rate limit).
```
shishao7sxm, line 71: Hi @rajkumar38, Enabling both direction sample may double the number of packets trapped to CPU, meanwhile any drop of sample packet cause inaccuracy. Agree that backoff mechanism is too heavy, have you considered adding a monitor in sFlow container to alert sample packet loss due to Copp? That will help users to use it more easily.

pphaal: The automated sampling rate backoff mechanism described in the sFlow spec tends to cause more problems than it solves. Instead, the sFlow agent should maintain the configured sampling rates and report internal drops using the drops counter in sFlow packet sample records (defined as, number of times that the sFlow agent detected that a packet marked to be sampled was dropped due to lack of resources). If excessive drops are detected then the network operator may decide to configure a higher sampling rate to improve accuracy during peak loads.

The psample PSAMPLE_ATTR_GROUP_SEQ allows the ASIC driver to indicate to the sFlow agent that samples have been dropped/rate limited. In addition to incrementing the counter each time a psample message is sent, it should also increment the counter by the number of packet samples lost to the policer. The hsflowd agent uses discontinuitites in the PSAMPLE_ATTR_GROUP_SEQ to increment the sFlow drops counters.

shishao7sxm: Thanks, sounds good.

rajkumar38:
> "sample_direction":"both"

SAI spec allows both ingress and egress sample-rate to be different. 
Also your proposal deviates from open-config model (openconfig allows ingress-sample-rate and egress-sample-rate).
Hence we want to give same flexibility in SONiC.

doc/sflow/sflow_hld.md
rajkumar38: > @rajkumar38 Do you know whether all 3 above phases are completed? if yes, mentioning the current status would be helpful.

To our understanding, there is no progress on phase II and III. 
It is not clear if anyone else is already started working on this.

pphaal:
> > "sample_direction":"both"
> 
> SAI spec allows both ingress and egress sample-rate to be different. Also your proposal deviates from open-config model (openconfig allows ingress-sample-rate and egress-sample-rate). Hence we want to give same flexibility in SONiC.

The sFlow HLD is an implementation the [sFlow specification](https://sflow.org/developers/specifications.php), which requires symmetric sampling rates when performing bi-directional sampling. 

The CLI and ConfigDB schema must faithfully implement the sFlow specification. Permitting asymmetric sampling to be configured puts the agent out of complience and will result in erroneous traffic data being reported by sFlow analyzers. The proposed, "sampling_direction":"both", ensures compliance with the sFlow specification, is simpler to implement, requires fewer HLD changes, and minimizes backward compatibility issues with implementations of the current HLD.

The sFlow agent abstracts the low level features of the ASIC exposed by SAI in order to implement the sFlow data model. The sFlow agent will consume SAI's ability to set ingress-sample-rate and egress-sample-rate in order to implement symmetric sampling, but sFlow configuration model protects the user from misconfiguring SAI.

The open-config model for sFlow configuration should not be used as a reference since it is flawed (for the reasons described above). Instead, the open-config model should be changed to comply with the sFlow specification.

vadymhlushko-mlnx:
> Hi rajkumar38,
> 
> While sFlow supports bidirectional packet sampling, it requires that the sampling rate be the same in each direction, i.e. ingress_sampling_rate == egress_sampling_rate for bidirectionall sampled interfaces.
> 
> I would suggest an option 3 that would simplify the proposal and minimize issues with backward compatibility:
> 
> ```
> "SFLOW_SESSION": {
>     "Ethernet0": {
>        "admin_state": "down",
>        "sample_rate": "40000",
>        "sample_direction":"both"           
>     },
> ```
> 
> The sflow sample-rate setting should remain unchanged. Instead, a new CLI settings should be added to configure direction, i.e.: `sflow interface direction {rx|tx|both}` Decoupling sampling rate and sampling direction in the CLI is simpler and leaving the existing sampling-rate command unchanged simplifies forward migration. If unspecified, the default direction should be "rx" for backward compatibility (and because this is the most common setting).
> 
> The "show sflow" command should also minimize changes, i.e. first three columns should remain the same, a fourth column could be added to indicate direction:
> 
> ```
> Interface     Admin Status   Sampling rate  Sampling direction
> ---------     ------------   -------------      --------------
> Ethernet0     down           40000                   both
> ```
> 
> Similarly the DB schema can be extended with a SAMPLE_DIRECTION attribute added to SFLOW_SESSION_TABLE. When migrating from an older SONiC version, the default sampling direction should be set to "rx"
> 
> Regards, Peter Phaal, InMon Corp.

Agree with Peter Phaal. This approach is cleaner & simple than in the proposal.

rajkumar38:
@pphaal @vadymhlushko-mlnx 
Thanks for the inputs. We agree to this option3. I will update the HLD accordingly. In addition, we would like to introduce new "direction" command at global level. If not set default is "rx" for backward compatibility. On upgrade, db_migrator will set this to "rx" in CFG_SFLOW_TABLE and APP_SFLOW_TABLE_NAME . This allows operator to retain previous behavior like before upgrade.

_config sflow direction { rx | tx | both }_

rajkumar38:
> Depending on the capabilities of the ASIC/driver, egress samples may be able to expose additional PSAMPLE attributes. I think it would be helpful to point this out by adding a paragraph after the paragraph discussing PSAMPLE_ATTR_SAMPLE_GROUP in section 6.9 - not to propose a change, but to raise awareness. Something along the lines:
> 
> Depending on hardware capabilities, driver may report additional attributes defined in https://github.com/torvalds/linux/blob/master/include/uapi/linux/psample.h. For example, PSAMPLE_ATTR_OUT_TC (egress queue), PSAMPLE_ATTR_OUT_TC_OCC (egress queue depth), and PSAMPLE_ATTR_LATENCY (transit delay) populate the sFlow Transit Delay Structures (https://sflow.org/sflow_transit.txt).

Updated the HLD. Pls check

pphaal:
> > Depending on the capabilities of the ASIC/driver, egress samples may be able to expose additional PSAMPLE attributes. I think it would be helpful to point this out by adding a paragraph after the paragraph discussing PSAMPLE_ATTR_SAMPLE_GROUP in section 6.9 - not to propose a change, but to raise awareness. Something along the lines:
> > Depending on hardware capabilities, driver may report additional attributes defined in https://github.com/torvalds/linux/blob/master/include/uapi/linux/psample.h. For example, PSAMPLE_ATTR_OUT_TC (egress queue), PSAMPLE_ATTR_OUT_TC_OCC (egress queue depth), and PSAMPLE_ATTR_LATENCY (transit delay) populate the sFlow Transit Delay Structures (https://sflow.org/sflow_transit.txt).
> 
> Updated the HLD. Pls check

Looks good. Thanks.

vadymhlushko-mlnx (approved):

jeff-yin (approved):

zhangyanzhao:
@shishao7sxm can you please help to review this PR? Alibaba registered this PR to review. Thanks.

venkatmahalingam (approved):

rajkumar38:
@shishao7sxm @zhangyanzhao 
If no further comments, can you help to merge this PR.

skg-net:
@rck-innovium Can you please update the Quality Metric (Alpha/Beta/GA) for the feature either in this PR comments or in HLD itself based on https://github.com/sonic-net/SONiC/blob/master/doc/SONiC%20feature%20quality%20definition.md
Thanks

rck-innovium:
> @rck-innovium Can you please update the Quality Metric (Alpha/Beta/GA) for the feature either in this PR comments or in HLD itself based on https://github.com/sonic-net/SONiC/blob/master/doc/SONiC%20feature%20quality%20definition.md Thanks

Quality metric: Beta

Test plan has been reviewed and partially implemented: https://github.com/sonic-net/sonic-mgmt/pull/8275/files

PR #1285, opened 2023-03-15
sonic-gnmi: Master Arbitration

sallylsy:
Adding doc for Master Arbitration HLD

Signed-off-by: sallylyu@google.com

| Repo  | PR Title   | State          | 
|------|------------|--------------------|
| sonic-gnmi | [Master Arbitration ](https://github.com/sonic-net/sonic-gnmi/pull/96) |  |

#### Why I did it
Demonstrate the high level design of Master Arbitration feature.

#### How I did it
Provide the markdown document for the high level design.

#### How to verify it
We designed a set of unit test and system test cases to verify the Master Arbitration is working as expected. 

#### Which release branch to backport (provide reason below if selected)

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111

N/A. Master Arbitration high level design is not a fix. No backporting is needed.

#### Description for the changelog

No change in changelog

#### Link to config_db schema for YANG module changes

No change in config_db schema.

#### A picture of a cute animal (not mandatory but encouraged)
piggy


tomek-US (approved):
LGTM

doc/mgmt/gnmi/master_arbitration.md
```diff
@@ -0,0 +1,331 @@
...
+
+Reference: https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-master-arbitration.md
+
+### Overview
+
+For high availability, a system may run multiple replicas of a gNMI client (controller). Among the replicas, only one client should be elected as master and do gNMI operations that mutate any state on the target. However, in the event of a network partition, there can be two or more replicas thinking themselves as master. For read-only RPCs, this is OK. But if they both call the `Set` RPC, the target may be incorrectly configured by the stale master. Therefore, "Master Arbitration" is needed when multiple clients exist.
+
+**Definition**
+
+When Master Arbitration feature is turned on, the gNMI server receives a proto with the election ID (EID) specified in it. The election ID is a 128-bit integer MUST be guaranteed to be monotonically increasing -- a new master's election ID MUST always be larger than its predecessor. The process through which election IDs are allocated to client replicas is out-of-scope of this specification.
```
mm0930, line 47: "a new master's election ID MUST always be larger than its predecessor"
In theory, on the roll over of the 128-bit integer, wouldn't this assumption be breached ?

doc/mgmt/gnmi/master_arbitration.md
```diff
@@ -0,0 +1,331 @@
...
+
+The client carries the election ID in all of its `Set` requests. When the target receives any `Set` request, it looks at the election ID carried with the requests and stores the largest ID it sees. An election ID that is equal to or greater than the currently stored one should be accepted as the master. On the other hand, Set requests with no election ID or with a smaller election ID are rejected and return an error.
```
mm0930, line 49: " An election ID that is equal to or greater than the currently stored one should be accepted as the master."
How does the arbitration work if there are two replicas that send the same (largest) election ID ?

doc/mgmt/gnmi/master_arbitration.md
```diff
@@ -0,0 +1,331 @@
...
+basis. There can be one master for each role, i.e. multiple master clients are
+allowed only if they belong to different roles.
+
+The role is identified by its `id`, which is assigned offline in agreement
+across the entire control plane. The target will use the role ID to arbitrate
+such that each role has one and only one master controller instance.
+
+To simplify for use-cases where multiple roles are not needed, the client can
+leave the role message unset in MasterArbitration. This implies default role.
+All `Set` RPCs that uses a default role are arbitrated in the same group.
```
mm0930, line 65: Is there a possibility of a split brain on the Client side where more than one client is sending the same role 'id'. If yes, what is the expectation of the GNMI server? Is there any "deny-list" of clients maintained to deny both the clients if they send the same role ?

doc/mgmt/gnmi/master_arbitration.md
```diff
@@ -0,0 +1,331 @@
...
+}
+
+// ...
+```
+
+### High-Level Design
+
+When Master Arbitration is enabled, `SetRequest` with `MasterArbitration` message sent from clients (controllers) will get checked by gNMI. The client with the highest `election_id` will be treated as master with the condition of `ReqFromMaster = True`. The rest clients are non-masters. After the initial `SetRequest`, only request from master will be handled by gNMI server, the requests from non-master clients will be dropped.
+
+The implementation of Master Arbitration, a new built-in feature in SONIC, will involve modifications in the telemetry module and the sonic-gnmi repository. However, it is important to note that this change will not affect other modules or sub-modules. It is worth mentioning that Master Arbitration does not require any changes to be made to SWSS or Syncd. A new field will be created in `CONFIG_DB` to record the status of Master Arbitration.
```
mm0930, line 131: Is this new field role-scoped?

doc/mgmt/gnmi/master_arbitration.md
```diff
@@ -0,0 +1,331 @@
...
+
+* Multiple Master Arbitration extension protobufs when Master Arbitration is enabled
+  * The last one should be used.
+
+* Change of the master taking place
+  * Accepted call from EID = 1
+  * Accepted call from EID = 2
+  * Rejected call from EID = 1
+
+* Restart of the switch resets the expected master EID to 0
```
mm0930, line 300: If two controllers send conflicting requests with the same largest EID=2, then one config will overwrite the other. That appears to be concerning. 
Also, I assume all controller instances will exchange the same certificate for authentication. Please confirm.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: The implementation of this feature is as designed and described in the spec [link](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-master-arbitration.md). Assignment of ID is out of switch control. The controller designers have to handle this problem by, say, restarting the switch before the ID rolls over, and make sure the ID is unique.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: If two controllers are sending SET requests with the same EID, the switch will treat them as one controller and execute SET from both.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: This new field is recording the status of Master Arbitration, say {"with-master-arbitration": true} or {"with-master-arbitration": false}. No plan to make it role-scoped.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: > If two controllers send conflicting requests with the same largest EID=2, then one config will overwrite the other. That appears to be concerning.
Yes. This concerning is EXACTLY the one when Master Arbitration is not available at all. The discussion of implementation of EID is out of the scope of switch implementation though. 

> Also, I assume all controller instances will exchange the same certificate for authentication. Please confirm.
Each controller (no matter it is selected as Master or not) has its own authentication. The exchange is not necessary.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: If more than one clients share the same role "id", the one with the largest "EID" among them will be considered as Master. The SET gRPC requests from other clients will be dropped. And, the GET gRPC requests from non-Master clients is not impacted. Therefore, there is not need for any "deny-list" of clients.

anand-kumar-subramanian (dismissed):

zhangyanzhao:
@sallylsy can you please add the code PRs if you have now? Thanks.

sallylsy:
@zhangyanzhao sure. Please find it here https://github.com/sonic-net/sonic-gnmi/pull/96

zhangyanzhao:
community review recording https://zoom.us/rec/share/fhMIOkkDYx3INSShyKdXw9dXl8HqhBqQLxzG1mEUyvJrxdYbdGhofRS9Xs4MCh2g.aBeFosVoWBv3VBXi

doc/mgmt/gnmi/master_arbitration.md
prvattem, line 49: Where is the last larget ID seen is stored for reference?

doc/mgmt/gnmi/master_arbitration.md
```diff
@@ -0,0 +1,331 @@
...
+
+### Restrictions/Limitations
+
+Master Arbitration is designed to enable one master client for one role. However, only default role is supported for now.  In other words, the content of the role field is ignored.
+
+### Testing Requirements/Design
+The main focus of the test is on the coverage of the new feature. The user cases are listed as following:
+
+* No Master Arbitration extension protobuf when Master Arbitration is disabled
+  * Nothing should happen
```
rathnasabapathyv, line 284: **Regarding the issue, we discussed in community-meeting to avoid issues with EID out-of-sync between new-controller-master & GNMI-server till new SET from user through new controller:**  ==> I was also suggesting to learn new EID from get-request from new-master rather than waiting for set initiated by user through new-master

* **While checking further, this GNMI spec** (https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-master-arbitration.md#32-setrequest-handling), there is a special mention about this case to sync latest EID immediately through empty-set rather than waiting for set-from-user, after new master change. 

**Exact snippet from spec:**
image

Assuming controller on master change will immediately does empty-set with new EID. So from SONIC GNMI-server perspective, hope we will handle empty-GNMI-set and update latest EID through that, and ignore the subsequent sets from old-master. 

Can you please confirm this EID-sync handling in GNMI-server for empty-sets?

doc/mgmt/gnmi/master_arbitration.md
venkatmahalingam: Not sure restarting the switch before the ID rolls over is the good idea, do we have any mechanism to validate whether the switches are using the right EID, also the switches are using the right master controller.

doc/mgmt/gnmi/master_arbitration.md
venkatmahalingam: Can there be more than 2 controllers?

doc/mgmt/gnmi/master_arbitration.md
```diff
@@ -0,0 +1,331 @@
...
+across the entire control plane. The target will use the role ID to arbitrate
+such that each role has one and only one master controller instance.
+
+To simplify for use-cases where multiple roles are not needed, the client can
+leave the role message unset in MasterArbitration. This implies default role.
+All `Set` RPCs that uses a default role are arbitrated in the same group.
+
+### Requirements
+
+Master Arbitration is a HA (high-availability) SDN-focused feature therefore it cannot be enabled by default. On the other hand, when it is required, it has to be enabled from the very moment the gNMI server starts. As an optional feature, Master Arbitration could be turned off when gNMI server restarted. And it should not impact the overall perfermance.
```
venkatmahalingam, line 69: I think, we can have this feature enable/disable via config DB field. We just need to handle config DB event for any on-the-fly enable/disable.

doc/mgmt/gnmi/master_arbitration.md
venkatmahalingam: Can you explain how are we storing the last received EID in the config-DB and how it gets restored upon gNMI server or switch restart?

doc/mgmt/gnmi/master_arbitration.md
```diff
@@ -0,0 +1,331 @@
...
+
+```
+message MasterArbitration {
+  Role role = 1;
+  Uint128 election_id = 2;
+}
+
+message Uint128 {
+  uint64 high = 1;
+  uint64 low = 2;
```
venkatmahalingam, line 83: What is the behaviour in gNMI server when there are changes in the low & high values of the EID in the set Request?

doc/mgmt/gnmi/master_arbitration.md
venkatmahalingam: Also, how do we ensure conflicting requests are not received from multiple controllers with same EID? is there any use-case like this?

doc/mgmt/gnmi/master_arbitration.md
venkatmahalingam: Is there any possibility of receiving set requests from clients with different roles and EIDs to the same switch? if yes, do we need to store all these info. into the config_DB.

doc/mgmt/gnmi/master_arbitration.md
venkatmahalingam: >> Restart of the switch resets the expected master EID to 0
If we reset EID to 0 upon switch reset, wont we accept the first set request with lower EID > 0 from the client?

doc/mgmt/gnmi/master_arbitration.md
```diff
@@ -0,0 +1,331 @@
...
+
+![MA0](images/MA0.png)
+
+If only one controller is interacting with the gNMI server and Master Arbitration is enabled, the behavior of the controller-switch is identical to the scenario where there is no Master Arbitration.
+
+![MA1](images/MA1.png)
```
venkatmahalingam, line 160: As discussed in the meeting, please include controller-mgr in the diagram.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: > Can there be more than 2 controllers? if yes, please capture it accordingly in the diagrams.
Yes. There could be more. I can update the graph.

> Also, how do we ensure conflicting requests are not received from multiple controllers with same EID? is there any use-case like this?
If controllers send the same EID, basically this is the same senario as MA disabled case. However, controller manager should make sure that no EID should be unique.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: We do not store EID in config-DB. Once gNMI server, or switch restarts, the stored EID is reset to 0. In such case, controller manager should still keep the highest EID for the previous master controller. In such case, as long as MA is enabled for gNMI server, the same master controller should still be picked by switch. Worst senario will be EID rollover and at the meantime, gNMI server restarts. In such case, we still recommend controller manager to handle the EID of the previous master controller. If gNMI stores some large EID value after restart, no controller may reach switch after EID rollover with MA enabled, since in-coming EID probably will be smaller. So, we don't think keep EID on switch side is a good idea.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: It is possible. As long as the controller manager be able to assign the highest EID to the previous master controller, the previous master controller will take control eventually. Comparing the drawback of possible unreachable switch after EID rollover vs loose of master controller for a short time period, we think the later one is more acceptable.

doc/mgmt/gnmi/master_arbitration.md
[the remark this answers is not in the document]
sallylsy: `"TELEMETRY|gnmi"` table will carry the flag.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: The last largest ID seen is stored in the gNMI client ([code](https://github.com/sonic-net/sonic-gnmi/pull/96/files#diff-27de8bb6147404b50ed13655e76620985b572253e2e8105ebad6d2d6bc4e8c0fR45)).

doc/mgmt/gnmi/master_arbitration.md
sallylsy: The uint128 type comparison is implemented [here](https://github.com/sonic-net/sonic-gnmi/pull/96/files#diff-27de8bb6147404b50ed13655e76620985b572253e2e8105ebad6d2d6bc4e8c0fR536). high values are compared first. If there is a tie in high values, low values are compared.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: We would like to add this function in next release.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: Currently we only have one role, which is the default one. For the detail about multiple-role case, we would like to include in the future release.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: Added. Please refer the EID manager in the new graph.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: Yes. EID check is done for every SET request (for both empty-set and non-empty-set). As long as new master controller sending empty-set, GNMI-server will sync the master EID.

doc/mgmt/gnmi/master_arbitration.md
sallylsy: I don't think there is a `RIGHT` EID in the case. I understand your concern. But this can only be addressed outside of switch (EID manager should do the job). Switch side has no knowledge to make sure the `RIGHT` controller is sending request.

anand-kumar-subramanian (approved):

doc/mgmt/gnmi/master_arbitration.md
venkatmahalingam: >> as long as MA is enabled for gNMI server, the same master controller should still be picked by switch
Since the gNMI server restart sets the EID to 0, whichever controller first sends the set request will be accepted by the switch irrespective of the higher EID master controller  previously chosen before the gNMI restart. How do we make sure that the previous master controller is chosen after the gNMI restart?

venkatmahalingam: > **as long as MA is enabled for gNMI server, the same master controller should still be picked by switch**
> 
> -> Since the gNMI server restart sets the EID to 0, whichever controller first sends the set request will be accepted by the switch irrespective of the higher EID master controller previously chosen before the gNMI restart. How do we make sure that the previous master controller is chosen after the gNMI restart?

Looks like you answered this question already, thanks!

doc/mgmt/gnmi/master_arbitration.md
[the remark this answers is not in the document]
venkatmahalingam: **A new field will be created in `CONFIG_DB` to record the status of Master Arbitration.**
-> Is this the user configuration? or when the GNMI server is started with "--with-master-arbitration" in the command line, you'll update the CONFIG_DB master_arbitration_enabled field?

Wondering if we need to make the configuration for MA in the CONFIG_DB and read the config in the gNMI server to enable/disable the MA feature.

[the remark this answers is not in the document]
sallylsy: This is for record purpose. So client knows that MA is enabled/disabled. 
>-> Is this the user configuration? or when the GNMI server is started with "--with-master-arbitration" in the command line, you'll update the CONFIG_DB master_arbitration_enabled field?

We will update the CONFIG_DB master_arbitration_enabled field

>Wondering if we need to make the configuration for MA in the CONFIG_DB and read the config in the gNMI server to enable/disable the MA feature.

Currently, we think MA is better set with the start of gNMI server.

tomek-US (approved):

sallylsy:
Hi @zhangyanzhao, the implementation of this HLD is merged. Can we get this HLD merged as well? We still need one more approval from community.

venkatmahalingam (approved):

prvattem (approved):

doc/mgmt/gnmi/master_arbitration.md
rathnasabapathyv: ok thanks.

rathnasabapathyv (approved):

adyeung (approved):

skg-net:
@sallylsy Can you please update the Quality Metric (Alpha/Beta/GA) for the feature either in this PR comments or in HLD itself based on https://github.com/sonic-net/SONiC/blob/master/doc/SONiC%20feature%20quality%20definition.md
Thanks

PR #1289, opened 2023-03-16
Sonic Chassis http-server HLD

anamehra:
Signed-off-by: anamehra <anamehra@cisco.com>

Sonic Chassis network http-server HLD

Repo | PR title | State
-- | -- | --
sonic-buildimage | [database-chassis python http-server implementation PR13345](https://github.com/sonic-net/sonic-buildimage/pull/13345) |


anamehra:
Hi @rlhui , @abdosi , Please review. Thanks

anamehra:
Hi @abdosi , @qiluo-msft , please review.

doc/chassis-http-server/chassis_http_server.md
```diff
@@ -0,0 +1,112 @@
...
+ ```
+
+ **Example of config in chassisdb.conf:**
+ ```
+  start_http_server=1
+  chassis_db_address=127.0.0.10
+  http_server_dir=/var/www/tftp
+ ```
+
+ The service will be added as a new process in docker database-chassis. dockers/docker-database/supervisord.conf.j2 will be modified to support the new service. This service should run only on database-chassis server, which also hosts redis_chassis.server. Other database dockers should not run the http-server.
```
lguohan, line 72: not sure why re-use database docker not starting its own docker for this http_server?

doc/chassis-http-server/chassis_http_server.md
```diff
@@ -0,0 +1,112 @@
...
+### Requirements 
+
+ The http-server shall run on the internal midplane network and use the internal Supervisor midplane IP, provided by the platform, to bind.
+ The http-server shall use port 8000 to bind on the internal network.
+
+ The use of this http-server shall be limited to internal midplane and not designed to service requests on the mgmt/external interface.
+
+### Architecture Design 
+
+ The new http-server shall use the current Sonic architecture and shall not require any architecture design change. The server shall run only on the supervisor and inside the existing database-chassis docker, which runs only on supervisor.
```
qiluo-msft, line 43: Why put this service in database container? I do not see any dependency.

doc/chassis-http-server/chassis_http_server.md
anamehra: Hi @lguohan , this feature was discussed in Chassis community group last year and adding http-server docker was one of the options. But as this is a light weight python http-server, community was more in favor of adding a host side http-server feature. A PR was raised for that but that was an issue enabling the http-server on host side as a feature as it required the service to run a docker.

Later in the community discussion it was decided as its needed in 202205 and only for chassis, to add it as a light weight service in database-chassis which only runs for chassis.
This service is only required for Chassis, binds only to midplane/internal network and uses redis_chassis.server IP to bind to host the server.

@abdosi , please add if I missed anything.

Let me know if we need to have a quick meeting to close on this. 
Thanks

doc/chassis-http-server/chassis_http_server.md
anamehra: It's in database-chassis container which runs only on Supervisor. Please see above comment for more details. Thanks

PR #1302, opened 2023-03-22
Delete config_reload duplicate image to fix `git clone`

mdanish-kh:
### What I did
Delete duplicate image to fix git clone on NTFS-formatted hard-drives.

### Why I did it
Git Clone of this repo will cause errors on a case-insensitive path filesystem (which is NTFS if you're on a Windows machine) because two files with same names but different casing exist on this repo.

'images/config_reload/Hostcfgd_flow_for_delayed_services.`JPG`'
'images/config_reload/Hostcfgd_flow_for_delayed_services.`jpg`'

Screenshot of a git clone of SONiC repo on a Windows machine:

The result is that, as Windows does not allow the two files to exist simultaneously, your git repo will never be in sync with the upstream SONiC repo as their will always be 1 local change (file deletion) that'll cause problems in your git workflow.

Even on macOS, you will have to reformat volume as case-insensitive/sensitive or make a new volume altogether.

### Solution

Delete the duplicate image **"images/config_reload/Hostcfgd_flow_for_delayed_services.`jpg`**", as it is not being used in the [HLD](https://github.com/sonic-net/SONiC/blob/master/doc/config_reload/config_reload_enhancement.md)
https://github.com/sonic-net/SONiC/blob/668987fb25a9eacae3ae509d5a81a51ec04e9726/doc/config_reload/config_reload_enhancement.md?plain=1#L74


mdanish-kh:
@dgsudharsan Please review as this image is part of the HLD authored by you.

dgsudharsan (approved):

mdanish-kh:
@liat-grozovik @lihuay Can someone please merge this PR?

liat-grozovik (approved):

PR #1311, opened 2023-04-01
The script for counting commits based on organizations

eddieruan-alibaba:
The script for counting commits based on organizations. It also help each organization to check if company name is set in associated githubids.

This is not the final script for calculating SII. We could grow that calculating tools from this script.


tsc/scripts/git_count.py
```diff
@@ -0,0 +1,495 @@
...
+        return True, "NVIDIA"
+    if "nokia" in key:
+        return True, "NOKIA"
+    if "barefootnetworks" in key:
+        return True, "INTEL"
+    if "innovium" in key:
+        return True, "INNOVIUM"
+    if "cavium" in key:
+        return True, "CAVIUM"
+
```
mdanish-kh, line 204:
```suggestion
    if "cavium" in key:
        return True, "CAVIUM"
    if "xflow" in key:
        return True, "XFLOW"

```

Please include xFlow Research as it's both a General Member and one of the contributing organizations. Please see https://sonic-net.github.io/SONiC/index.html for reference

zhangyanzhao:
@eddieruan-alibaba can you please resubmit this to contributor map repo? Thanks.

tsc/scripts/git_count.py
```diff
@@ -0,0 +1,495 @@
...
+#
+# List of repros
+#
+repros = [
+    "SONiC",          "sonic-host-services",  "sonic-mgmt-framework",    "sonic-quagga",     "sonic-swss",         "sonic-ztp",
+    "sonic-dbsyncd",  "sonic-linux-kernel",   "sonic-platform-common",   "sonic-restapi",    "sonic-swss-common",  "sonic-buildimage",
+    "sonic-fips",     "sonic-mgmt",           "sonic-platform-daemons",  "sonic-sairedis",   "sonic-telemetry",
+    "sonic-frr",      "sonic-mgmt-common",    "sonic-py-swsssdk",        "sonic-snmpagent",  "sonic-utilities"
+]
```
adyeung, line 19: Per TSC Election doc, the contribution count should include OCP SAI repo

tsc/scripts/git_count.py
```diff
@@ -0,0 +1,495 @@
...
+    # loop through all repros
+    #
+    for repro in repros:
+        num_repro = num_repro +1
+        companys={}
+
+        #
+        # Collect commit info from each repro
+        #
+        count_repro_commits_via_short_log(repro, years)
```
adyeung, line 450: Besides code PR, there are many other contribution to be accounted for SII like reviews, issue triage etc, I assume they will be updated in the next patchset

tsc/scripts/git_count.py
adyeung: sonic-mgmt (test), SONiC (HLD) carry different score than other code PRs, you may need to process them separately

tsc/scripts/git_count.py
```diff
@@ -0,0 +1,495 @@
...
+        comp = fixup_companyname(comp)
+    else:
+        ret, comp = use_email_to_find_company(email, src)
+
+    return comp
+
+#
+# Company name mapping
+#
+def company_name_check(key):
```
adyeung, line 160: Pls use contributor map posted by Guohan

tsc/scripts/git_count.py
```diff
@@ -0,0 +1,495 @@
...
+
+    # Per company per repor
+
+    if comp in companys:
+        companys[comp] = companys[comp] +count
+    else:
+        companys[comp] = count
+
+    # Total per repro        
+    if "Total" in companys:
```
dgsudharsan, line 50: The current logic resets companys array while looping through in an upper level API. This is very hard to read. I prefer indexing companys to have repo like companys[repo][comp] and then going over this dic to print at the end

dgsudharsan (changes requested):

dgsudharsan (changes requested):

tsc/scripts/git_count.py
```diff
@@ -0,0 +1,495 @@
...
+        #if ret and "NOTFOUND_" not in comp:
+            # Found cached info, no need to trigger lookup
+            update_company_commit_counts(comp, ccount)
+        else:
+            # Step 2: Check email first for easy cases
+            ret, comp = company_name_check(email)
+            if not ret:
+                # Step 3: use name to find githubid,then use github id to find profile.
+                # Try to get company information from profile
+                url = "https://api.github.com/search/users?q={}".format(name)
```
dgsudharsan, line 318: Search by name is wrong. There might be many GitHub users with the same name. The search should be based on a unique key (e.g. email)

dgsudharsan (changes requested):

tsc/scripts/git_count.py
```diff
@@ -0,0 +1,495 @@
+#!/usr/bin/env python3
```
dgsudharsan, line 1: What if a specific user has quit an organization? Their github ID will no more have details of the previous organization and the final org counts will not be correct. Anyway to overcome this problem?

dgsudharsan (changes requested):

PR #780, opened 2021-04-21
SONiC CLI Auto-generation tool HLD

vadymhlushko-mlnx:
| PR title | state | context |
| ---------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------ |
| [[sonic-cli-gen] first phase implementation of the SONiC CLI Auto-generation tool](https://github.com/Azure/sonic-utilities/pull/1644) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-package-manager] support sonic-cli-gen and packages with YANG model](https://github.com/Azure/sonic-utilities/pull/1650) | GitHub issue/pull request detail | GitHub pull request check contexts |

Signed-off-by: Vadym Hlushko <vadymh@nvidia.com>
The high-level design document for the SONiC CLI Auto-generation tool.

PR #1315, opened 2023-04-05
SONiC PAC Support HLD

samitabh:
SONiC PAC Support HLD


adyeung (commented):
Initial comments

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,1152 @@
...
+### 3.6.3 FDB
+PAC uses **SAI_FDB_ENTRY_ATTR_PACKET_ACTION** with **SAI_PACKET_ACTION_DROP** to put the static FDB entry in discard state.   
+**SAI_PACKET_ACTION_FORWARD** is used to put the static FDB entry into forwarding state post successful client authentication.   
+
+
+
+
+## 3.7 Manageability
+
+### 3.7.1 Yang Model
```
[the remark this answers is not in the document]
samitabh: Removed.

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,1152 @@
...
+The following scale is supported:
+
+| Configuration / Resource                 | Scale |
+| ---------------------------------------- | ----- |
+| Total number of authenticated clients on a port configured in Multiple Authentication host mode | 16    |
+| Total number of authenticated clients supported by the switch | 128   |
+
+
+
+# 5 Unit Test
```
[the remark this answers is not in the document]
samitabh: Removed.

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,804 @@
...
+
+mabd informs pacd about the result of the authentication. mabd also passes all the authorization parameters it receives from the RADIUS Server to the pacd. These are used for configuring the NAS to allow authenticated client traffic.
+
+
+## 3.5 SyncD
+
+No specific changes are needed in syncd for PAC.
+
+## 3.6 SAI
+
```
adyeung, line 631: Pls post a note here to clarify the followings are existing supported SAI attributes, no new SAI attribute requirement.

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,804 @@
...
+**SAI_PACKET_ACTION_FORWARD** is used to put the static FDB entry into forwarding state post successful client authentication.   
+
+
+
+
+## 3.7 Manageability
+
+### 3.7.1 Yang Model
+Yang Models are available for managing PAC, hostapd and MAB modules.
+
```
adyeung, line 655: Please clarify in the HLD that SONIC YANG will be added as part of this contribution

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,804 @@
...
+### 3.7.4 Clear Commands
+
+**sonic-clear authentication sessions <interface <all | interface | mac \>\>**
+
+This command clears information for all Auth Manager sessions. All the authenticated clients are re-initialized and forced to authenticate again.
+
+
+
+# 4 Scalability
+
```
adyeung, line 770: Unless this is defined by the SW data construct. Scalability performance being HW dependent, I would suggest to remove this claim from the HLD

doc/pac/Port Access Control.md
samitabh: Updated.

doc/pac/Port Access Control.md
samitabh: Updated.

doc/pac/Port Access Control.md
samitabh: Yes, these are SW defined constraints.

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,844 @@
...
+
+### 3.1.2 EAPoL receive flow
+
+![pac-eapol-rx-flow](images/PAC_EAPoL_Rx_Flow.JPG)
+
+
+**Figure 3: EAPoL receive flow**
+
+1. EAPoL packet is received by hardware on a front panel interface and trapped to the CPU by COPP rules for EAP. The packet gets delivered to the hostapd socket listening on EtherType 0x888E.
```
madhupalu, line 295: How does design handle malicious client attach with filling up with FDB space? I know there is a rate limit 128 discussed however how does it handle in case good clients and many malicious clients (> 129) there is a chance that good client never get a chance to join the system. 
Use case: Lets DoS attach occurred in campus network where all the kids try to connect network while DoS attach occurred,.

rck-innovium (changes requested):

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,844 @@
...
+
+Existing SAI attributes are used for this implementation and there is no new SAI attribute requirement.
+
+### 3.6.1 Host interface traps
+
+Leveraged **SAI_HOSTIF_TRAP_TYPE_EAPOL** to trap EAP packets (Ethertype - 0x888E) to the CPU.   
+
+
+### 3.6.2 Bridge port learning modes
+PAC uses the following bridge port learning modes to drop/trap all unknown source MAC packets.
```
dgsudharsan, line 677: Can we please consider using SAI_BRIDGE_PORT_FDB_LEARNING_MODE_FDB_NOTIFICATION? This mode was introduced for the purpose of dot1x

    /** Notify unknown source MAC using FDB callback. Do not learn in hardware. Do not forward.
      * When a packet from unknown source MAC comes this mode will trigger a new learn notification
      * via FDB callback for the MAC address. This mode will generate only one notification
      * per unknown source MAC to FDB callback.
      */
    SAI_PORT_FDB_LEARNING_MODE_FDB_NOTIFICATION,

dgsudharsan (changes requested):

thovikeerthi:
As per the received update during HLD review meeting, whenever RADIUS server connectivity with Switch (i.e. Authenticator) is down (due to any issues), then no new supplicant's shall be allowed to get authenticated i.e. until connectivity with RADIUS server is restored, which results in a down-time w.r.t authentication functionality for new supplicants. 
Please check, can we avoid this error case by supporting 'Local Authentication' mechanism as well along with Remote Authentication via RADIUS.
On supporting both mechanisms, we can avoid authentication down-time for new supplicants.
Note:
- As known to all, Hostapd has support for both local & remote authentication.

thovikeerthi:
IEE802.1X defines the encapsulation of EAP over Wired networks and Wireless networks. Please confirm, what is the support provided/extended in your design w.r.t this note ?

thovikeerthi:
As per the received update during HLD review meeting i.e. 'On session-timeout', the static FDB MAC entry aging happens and entry will be removed. Please confirm, who manages (i.e. start/stop/run etc) this timer ? what is the role of SWITCH here ?
Also, please explain these Timer's state-machine or transactions handling in the HLD and their impact on the switch.

thovikeerthi:
Please specify the details of your design handling for below use-cases:
1. When a supplicant is authenticated on a Port and then that port/interface becomes operationally down momentarily and gets restores state ?
2. Dynamic port break-out

kishorgovind:
The design does not consider or explains following use case scenario.
Step-1: MAC-A is authenticated on Port "X"
Step-2:  MAC-A client moves from Port "X" to Port "Y" in the same switch
Step-3: MAC-A gets authenticated
Then design should explain mechanism how duplicate entry fro same MAC-A client gets amanged

kishorgovind:
Since there is MAC limit of 128 entries in FDB, I see challenge with the design that the legitimate MAC client may not get access/authenticated as explained below.
If there are burst of unauthenticated MAC addresses are received say due to DOS attack. At the same time Radius server is not reachable or for some reason Radius server is not providing quick response and current design does not support local authentication. In this scenario unauthenticated clients may occupy 128 FDB entries (even though traffic is not forwarded). Now  since 128 entries are full and when next client (which is a legitimate client) tries to authenticate it will be denied the service since FDB entries reached the limit.

Pl. address this scenario in the design. According to me the solution to this problem is FDB entry should not be created for incoming client till this client is authenticated.

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,844 @@
...
+PAC supports only 1 RADIUS server. The highest priority server will be picked up for authentication. 
+
+
+## 3.2 DB Changes
+
+### 3.2.1 Config DB
+
+**PAC_PORT_CONFIG**   
+```   
+"PAC_PORT_CONFIG": {
```
lguohan, line 338: why do we need to create a new PAC_PORT_CONFIG, why cannot we add to existing port table?

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,844 @@
...
+
+### 2.2.6 Warmboot
+
+After a Warm Boot, the authenticated client sessions are torn down and they need to authenticate again.
+
+# 3 Design
+
+## 3.1 Overview
+
+[Figure 2](#configuration-flow) shows the high level design overview of PAC services in SONiC. The existing "macsec" docker is leveraged.
```
lguohan, line 261: why are we re-using macsec docker? why cannot use build a new pac docker?

thovikeerthi:
Please specify Code PRs if created already for the feature. Thanks.

doc/pac/Port Access Control.md
samitabh: DoS actually happens as PAC does not attempt authentication beyond 128 clients. It is not an FDB limit that causes the DoS.
PAC needs to have the logic to not even attempt authentication for MACs classified as "malicious". This will be taken up a future release.

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,844 @@
...
+
+PAC works with port learning modes and FDB entries to block or allow traffic for authenticated clients as needed. 
+
+- **Multiple Host mode**: A single client can be authenticated on the port. With no client authenticated on the port, the learning mode is set to DROP or CPU_TRAP (if MAB is enabled on the port). Once a client is authenticated on a port, the learning mode is set to HW. All clients connected to the port are allowed access and FDB entries are populated dynamically.
+
+- **Single Host and Multiple Authentication Modes**: All clients on the port need to authenticate. The learning mode of the port is always set to CPU_TRAP. Once a client starts the authentication process, the client is no longer unknown to PAC. PAC installs a static FDB entry to mark the client known so that the incoming traffic does not flood the CPU. The entry is installed with discard bits set to prevent client traffic from being forwarded. In effect, the packets are not flooded to the CPU nor forwarded to other ports during the authentication process. When the client is authenticated, the discard bits of the installed FDB entry are reset to allow client traffic.
```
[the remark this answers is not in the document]
samitabh: All packets trapped to CPU from the L2 table come via CoS queue 0. Configuring rate limit for a CoS queue is outside the purview of PAC.

doc/pac/Port Access Control.md
samitabh: In our current approach of CPU TRAP, we install the FDB entry with the discard bit once the authentication is in progress. We will never install more than max supported overall clients which is 128.

However in the FDB notification mechanism, the MACs are learned in HW with the pending bit set. So in case of a DoS attack, the HW L2 table gets filled up over and beyond what PAC can cater to (128). This can affect the overall system.
That makes TRAP to CPU a better mechanism to use.

doc/pac/Port Access Control.md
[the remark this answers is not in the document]
samitabh: When the client session is terminated, these static FDB entries are explicitly deleted by PAC.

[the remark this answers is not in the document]
samitabh: Yes it is Trap to CPU. Will correct / update the HLD with this information.

samitabh:
> As per the received update during HLD review meeting, whenever RADIUS server connectivity with Switch (i.e. Authenticator) is down (due to any issues), then no new supplicant's shall be allowed to get authenticated i.e. until connectivity with RADIUS server is restored, which results in a down-time w.r.t authentication functionality for new supplicants. Please check, can we avoid this error case by supporting 'Local Authentication' mechanism as well along with Remote Authentication via RADIUS. On supporting both mechanisms, we can avoid authentication down-time for new supplicants. Note:
> 
> * As known to all, Hostapd has support for both local & remote authentication.

Local authentication is not scalable and managable. All the Access Switches need to be in sync with the User Profile configurations in the RADIUS servers.

samitabh:
> IEE802.1X defines the encapsulation of EAP over Wired networks and Wireless networks. Please confirm, what is the support provided/extended in your design w.r.t this note ?

802.1x Authenticator functionality confirms to the 2004 standard wrt EAP encapsulation.

samitabh:
> As per the received update during HLD review meeting i.e. 'On session-timeout', the static FDB MAC entry aging happens and entry will be removed. Please confirm, who manages (i.e. start/stop/run etc) this timer ? what is the role of SWITCH here ? Also, please explain these Timer's state-machine or transactions handling in the HLD and their impact on the switch (i.e. under use-cases like reboot etc).

PACd handles all client session timeouts and it is the repository of 802.1x and MAB clients. Will update in the HLD.

samitabh:
> The design does not consider or explains following use case scenario. Step-1: MAC-A is authenticated on Port "X" Step-2: MAC-A client moves from Port "X" to Port "Y" in the same switch Step-3: MAC-A gets authenticated Then design should explain mechanism how duplicate entry fro same MAC-A client gets amanged

Yes will add in the HLD.

samitabh:
> Since there is MAC limit of 128 entries in FDB, I see challenge with the design that the legitimate MAC client may not get access/authenticated as explained below. If there are burst of unauthenticated MAC addresses are received say due to DOS attack. At the same time Radius server is not reachable or for some reason Radius server is not providing quick response and current design does not support local authentication. In this scenario unauthenticated clients may occupy 128 FDB entries (even though traffic is not forwarded). Now since 128 entries are full and when next client (which is a legitimate client) tries to authenticate it will be denied the service since FDB entries reached the limit.
> 
> Pl. address this scenario in the design. According to me the solution to this problem is FDB entry should not be created for incoming client till this client is authenticated.

DoS actually happens as PAC does not attempt authentication beyond 128 clients. It is not an FDB limit that causes the DoS.
Local Authentication is not scalable and managable.
PAC needs to have the logic to not even attempt authentication for MACs classified as "malicious". This will be taken up a future release.

doc/pac/Port Access Control.md
samitabh: This was the design choice as we did not want to overload the existing PORT Table with PAC configuration.

doc/pac/Port Access Control.md
samitabh: MACsec, 802.1x, PAC and MAB are all related features.
Today Switch-to-Switch MACsec is implemented in the macsec docker.
It may be logically extended to Host-to-Switch MACsec which uses 802.1x to authenticate clients and then derive keys needed for traffic encryption between the client and the switch.
PAC is the central client manager which also uses MAB to authenticate 802.1x unaware clients.
Hence it makes sense to club all these features in the existing MACsec docker.

samitabh:
> Please specify Code PRs if created already for the feature. Thanks.

There are none at the moment.

samitabh:
> 1. It would help to list the SAI APIs and attributes used to implement each mode.
Yes will update the HLD

> 2. I am assuming in multiple host mode, before any client gets authenticated, the port will have its learning mode as SAI_BRIDGE_PORT_FDB_LEARNING_MODE_DROP. 
[Amitabha] Yes

The expectation is that EAPOL packets sent by the client will be trapped to CPU via the EAPOL host-if trap.  Given this expectation, other Switch Traps like STP, LACP etc. will also get trapped to CPU and get processed.  However, these packets should be dropped and not be sent to the daemons.

[Amitabha] Will get back on this.

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,844 @@
...
+The following scale is supported:
+
+| Configuration / Resource                 | Scale |
+| ---------------------------------------- | ----- |
+| Total number of authenticated clients on a port configured in Multiple Authentication host mode | 16    |
+| Total number of authenticated clients supported by the switch | 128   |
+
+
+
+# 5 Appendix: Sample configuration
```
prvattem, line 820: Please add the step to create the VLAN as well in this config flow to make it a complete example.

zhangyanzhao:
2nd part of community review https://zoom.us/rec/share/ns8FrPvIRAdaFEDkMXrdrJxoSmgR73B2UIEyBj3UHYeEIo_iZc2oT9H5zbI_Kp8W.8HwWzx-3Lr9EeNV1

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,844 @@
...
+```
+
+
+
+# 6 Future Enhancements
+
+1. Add configurability support for 802.1x and MAB timers.
+2. Add support for fallback VLANs like Guest, Unauth etc. VLAN. These are used to authorize clients if they fail authentication under various circumstances.
+3. Add support for RADIUS Authorization attributes like ACLs.
+4. Add support for multiple RADIUS servers.
```
prvattem, line 840: Please include the thoughts on supporting on SONIC-VS image.

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,844 @@
...
+### 3.6.2 Bridge port learning modes
+PAC uses the following bridge port learning modes to drop/trap all unknown source MAC packets.   
+- SAI_BRIDGE_PORT_FDB_LEARNING_MODE_DROP
+- SAI_BRIDGE_PORT_FDB_LEARNING_MODE_HW
+- SAI_BRIDGE_PORT_FDB_LEARNING_MODE_CPU_TRAP
+
+
+
+### 3.6.3 FDB
+PAC uses **SAI_FDB_ENTRY_ATTR_PACKET_ACTION** with **SAI_PACKET_ACTION_DROP** to put the static FDB entry in discard state.
```
kselvaraj2, line 685: @rck-innovium Whether Both Innovium and Prestera (AC5x) line support src-discard per FDB entry ?

doc/pac/Port Access Control.md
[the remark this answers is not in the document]
rck-innovium: *   Can you please give the reference where (either SAI or SONiC) it is defined that "packets trapped to CPU via L2 learn mode CPU\_TRAP come via CoS queue 0". 
*   How would the end user know that packets trapped to CPU via MAB feature will be received on Queue 0?

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,844 @@
...
+### 3.7.1 Yang Model
+Yang Models are available for managing PAC, hostapd and MAB modules. SONiC YANG is added as part of this contribution.
+
+### 3.7.2 Configuration Commands
+
+The following commands are used to configure PAC.  
+
+| CLI Command                              | Description                              |
+| :--------------------------------------- | :--------------------------------------- |
+| config authentication port-control interface <auto \| force-authorized \| force-unauthorized \>  <interface\> | This command configures the authentication mode to use on the specified interface. Default is force-authorized. |
```
venkatmahalingam, line 702: Please use Click auto-generation tool to generate the CLIs. Refer below for more information.
https://github.com/sonic-net/SONiC/pull/780

doc/pac/Port Access Control.md
venkatmahalingam: This is important as there are early validations with GNS3 before actual deployments.

kishorgovind:
> > As per the received update during HLD review meeting, whenever RADIUS server connectivity with Switch (i.e. Authenticator) is down (due to any issues), then no new supplicant's shall be allowed to get authenticated i.e. until connectivity with RADIUS server is restored, which results in a down-time w.r.t authentication functionality for new supplicants. Please check, can we avoid this error case by supporting 'Local Authentication' mechanism as well along with Remote Authentication via RADIUS. On supporting both mechanisms, we can avoid authentication down-time for new supplicants. Note:
> > 
> > * As known to all, Hostapd has support for both local & remote authentication.
> 
> Local authentication is not scalable and managable. All the Access Switches need to be in sync with the User Profile configurations in the RADIUS servers.

For port Authentication option, we need to provide local authentication as an option

kishorgovind:
> > Since there is MAC limit of 128 entries in FDB, I see challenge with the design that the legitimate MAC client may not get access/authenticated as explained below. If there are burst of unauthenticated MAC addresses are received say due to DOS attack. At the same time Radius server is not reachable or for some reason Radius server is not providing quick response and current design does not support local authentication. In this scenario unauthenticated clients may occupy 128 FDB entries (even though traffic is not forwarded). Now since 128 entries are full and when next client (which is a legitimate client) tries to authenticate it will be denied the service since FDB entries reached the limit.
> > Pl. address this scenario in the design. According to me the solution to this problem is FDB entry should not be created for incoming client till this client is authenticated.
> 
> DoS actually happens as PAC does not attempt authentication beyond 128 clients. It is not an FDB limit that causes the DoS. Local Authentication is not scalable and managable. PAC needs to have the logic to not even attempt authentication for MACs classified as "malicious". This will be taken up in a future release.

Here when I say DOS attack, once client might be generating series of MAC address and fills FDB table. I am not saying MAC address as malicious or not. My point is that since you are creating FDB entry for MAC before client gets authenticated and hence FDB may get filled up. Instead logic should be you add MAC entry only after authentication.

doc/pac/Port Access Control.md
samitabh: Yes will do so.

doc/pac/Port Access Control.md
samitabh: OK. Thanks.

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,844 @@
...
+2. pacd sends a "Client Authenticate" message along with the received packet MAC to mabd.
+3. mabd interacts with RADIUS server to authenticate the given client based on the MAC.
+4. On successful authentication of a client, mabd sends an "Client Authenticated" message to pacd with all the authorization parameters like VLAN, Session-Timeout, etc.
+5. pacd proceeds to authorize the client. RADIUS authorization parameters like client VLAN membership, is communicated to relevant modules (VLAN, FDB) by writing on their tables on STATE_DB.  Authenticated clients are updated in PAC_AUTHENTICATED_CLIENT_OPER table in STATE_DB.
+6. VLAN, FDB further process these STATE_DB updates from PAC and write into their STATE_DB and APPL_DB tables.
+7. Orchagent in SWSS docker gets notified about changes in APPL_DB and responds by translating the APPL_DB changes to respective sairedis calls.
+8. Sairedis APIs write into ASIC_DB.
+9. Syncd gets notified of changes to ASIC_DB and in turn calls respective SAI calls. The SAI calls translate to respective SDK calls to program hardware.
+10. EAP success message (EAPoL PDU) is sent to the client.
```
prvattem, line 322: Does MAB client understand EAPoL PDUs?

samitabh: We may be doing MAB (if configured a higher priority method) for a client which initiated Dot1x. Hence EAPoL Success is needed for the client's Supplicant state machine to move to Authorized state. If it is a Dot1x unaware cilent, it will ignore the EAPoL-Success.

samitabh:
> The expectation is that EAPOL packets sent by the client will be trapped to CPU via the EAPOL host-if trap.  Given this expectation, other Switch Traps like STP, LACP etc. will also get trapped to CPU and get processed.  However, these packets should be dropped and not be sent to the daemons

[Amitabha] Generally these clients are laptops, printers etc. and will not be sending STP and LACP packets. We will however make a note and come back on this in a further release.

doc/pac/Port Access Control.md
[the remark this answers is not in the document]
samitabh: Yes there may be other factors like the packets priority tag that will define which CPU CoS queue. The admin will not know for sure which CPU queue the client's packet will queued on. That said, client packets that don't hit a COPP rule but get trapped via L2 learn mode CPU_TRAP will mostly come on CPU queue0 as laptops/printers etc will not tag their packets. The admin however still can configure rate limiting on other queues if need be.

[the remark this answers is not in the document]
AnandhiDhanabalan: "That said, client packets that don't hit a COPP rule but get trapped via L2 learn mode CPU_TRAP will mostly come on CPU queue0" Is this mentioned in COPP? Can you please give reference to it ?

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,852 @@
...
+
+No specific changes are needed in syncd for PAC.
+
+## 3.6 SAI
+
+Existing SAI attributes are used for this implementation and there is no new SAI attribute requirement.
+
+### 3.6.1 Host interface traps
+
+Leveraged **SAI_HOSTIF_TRAP_TYPE_EAPOL** to trap EAP packets (Ethertype - 0x888E) to the CPU.
```
AnandhiDhanabalan, line 680: What does Leveraged refer here ?

doc/pac/Port Access Control.md
AnandhiDhanabalan: Can we make thus scaling numbers configurable ?

samitabh: We want keep an upper limit which we can qualify. Hence these hard non-configurable limits.

doc/pac/Port Access Control.md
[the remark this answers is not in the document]
samitabh: As I mentioned, there may be factors that may lead to packets getting trapped to a certain queue. This proposal currently does not intend to force a queue to these packets. The admin however still can configure rate limiting on other queues if need be.

doc/pac/Port Access Control.md
samitabh: These learning modes are as per the current SAI spec. This proposal does not intend to cater to vendors that are not complaint to the SAI spec.

doc/pac/Port Access Control.md
samitabh: Used the trap type as defined in the SAI spec.

doc/pac/Port Access Control.md
samitabh: Yes this will be taken up in a future release.

samitabh:
@thovikeerthi 
Can you please review the updated HLD and my comments and approve the HLD.

doc/pac/Port Access Control.md
AnandhiDhanabalan: @samitabh Hard non-configurable limits might not be future scalable. Consider adding switch limits to this item and you can still qualify the switch limit currently supported.

samitabh: We do have configurable max users on a port from 1 to a MAX value. Why would we want to have MAX itself configurable ?

AnandhiDhanabalan: For future use case to support more than 128 clients per port, you need to modify the MAX value. This can be avoided if this is configurable.

samitabh: Sure. We will reevaluate this in a future release.

kishorgovind:
> > As per the received update during HLD review meeting, whenever RADIUS server connectivity with Switch (i.e. Authenticator) is down (due to any issues), then no new supplicant's shall be allowed to get authenticated i.e. until connectivity with RADIUS server is restored, which results in a down-time w.r.t authentication functionality for new supplicants. Please check, can we avoid this error case by supporting 'Local Authentication' mechanism as well along with Remote Authentication via RADIUS. On supporting both mechanisms, we can avoid authentication down-time for new supplicants. Note:
> > 
> > * As known to all, Hostapd has support for both local & remote authentication.
> 
> Local authentication is not scalable and managable. All the Access Switches need to be in sync with the User Profile configurations in the RADIUS servers.

As discussed during community discussion we are going to have single HLD. Hence we need to update HLD with local authentication. Pl let me know how CG can do this?

samitabh:
> > > As per the received update during HLD review meeting, whenever RADIUS server connectivity with Switch (i.e. Authenticator) is down (due to any issues), then no new supplicant's shall be allowed to get authenticated i.e. until connectivity with RADIUS server is restored, which results in a down-time w.r.t authentication functionality for new supplicants. Please check, can we avoid this error case by supporting 'Local Authentication' mechanism as well along with Remote Authentication via RADIUS. On supporting both mechanisms, we can avoid authentication down-time for new supplicants. Note:
> > > 
> > > * As known to all, Hostapd has support for both local & remote authentication.
> > 
> > 
> > Local authentication is not scalable and managable. All the Access Switches need to be in sync with the User Profile configurations in the RADIUS servers.
> 
> As discussed during community discussion we are going to have single HLD. Hence we need to update HLD with local authentication. Pl let me know how CG can do this?

Let's have a separate HLD for Local Authentication with this HLD as a base reference HLD.

kishorgovind (approved):
Approved

kishorgovind (approved):

prvattem (approved):

adyeung:
@msosyak feature contribution is planned for 202405, expect code PR posting in the next few weeks

adyeung (approved):

zhangyanzhao:
@adyeung  please help to update the HLD with the code PRs, so that we can track the feature progress. Thanks.

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,852 @@
...
+PAC as a PAE Authenticator for 802.1x is essentially a passthrough for client Authentication exchange messages. Hence different EAP authentication methods like EAP-MD5, EAP-PEAP, EAP-TLS, etc. are supported. These are essentially the 802.1x Supplicant and RADIUS server functionalities.
+
+PAC as a PAE Authenticator for MAB mimics the Supplicant role for MAB clients. Authentication methods EAP-MD5, EAP-PAP and EAP-CHAP are supported.
+
+***Authorization***
+
+Once a client is authenticated, authorization parameters from RADIUS can be sent for the client. The Authenticator switch processes these RADIUS attributes to apply to the client session. Following attributes are supported.
+
+- *VLAN Id*: This is the VLAN ID sent by a RADIUS server for the authenticated client. This VLAN should be a pre-created VLAN on the switch. 
+- *Session Timeout*: This is the timeout attribute of the authenticated client session.
```
praveenraja1, line 228: Do we have default session for un-authenticated clients as well? 

For instance lets say if a MAB  auth-request is rejected by Radius server and if we dont have a session timeout and immediately remove the client. Then the next packet from the client would trigger another Radius request, incase if the client is generating too many packets, the Radius connection might be overloaded. 

One way to maintain a session timer for auth-failed clients as well. Please let me know if this was considered during the design

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,852 @@
...
+
+- **Multiple Host mode**: A single client can be authenticated on the port. With no client authenticated on the port, the learning mode is set to DROP or CPU_TRAP (if MAB is enabled on the port). Once a client is authenticated on a port, the learning mode is set to HW. All clients connected to the port are allowed access and FDB entries are populated dynamically.
+
+- **Single Host and Multiple Authentication Modes**: All clients on the port need to authenticate. The learning mode of the port is always set to CPU_TRAP. Once a client starts the authentication process, the client is no longer unknown to PAC. PAC installs a static FDB entry to mark the client known so that the incoming traffic does not flood the CPU. The entry is installed with discard bits set to prevent client traffic from being forwarded. In effect, the packets are not flooded to the CPU nor forwarded to other ports during the authentication process. When the client is authenticated, the discard bits of the installed FDB entry are reset to allow client traffic. 
+
+  ​
+
+### 2.2.5 VLAN
+1. PAC associates authenticated clients to a VLAN on the port.
```
praveenraja1, line 244: Does this change involve writing to config_db? 

What if the user does 'config save' while the PAC has authenticated the client and added a port to Radius specified VLAN, would this make the configuration persistent? 

I'm assuming we only do 'Vlan member add' and not 'VLAN create' based on the radius response, what if the radius specified VLAN is not present in the system? Do we error out?

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,852 @@
...
+
+![pac-config-flow](images/PAC_Config_Flow.JPG)
+
+**Figure 2: PAC service daemon and configuration flow**
+
+1. Mgmt interfaces like CLI write the user provided configuration to CONFIG_DB.
+2. The pacd, mabd and hostapdmgrd gets notified about their respective configurations.
+3. hostapd being a standard Linux application gets its configuration from a hostapd.conf file. hostapdmgrd  generates the hostapd.conf file based on the relevant CONFIG_DB tables. hostapdmgrd informs hostapd about the list of ports it needs to run on. This port list is dynamic as it depends of port link/admin state, port configuration etc. hostapdmgrd  keeps hostapd updated about these changes. 
+4. These modules communicate amongst themselves via socket messages.
+5. hostapd listens to EAPoL PDUs on the provided interface list. When it receives a PDU, it consults pacd and proceeds to authenticate the client. pacd also listens to "unknown src MAC" and triggers MAB, if configured on the port, to authenticate the client.
```
praveenraja1, line 289: What is client sends a EAPOL pkt followed by regular ETH packet, im assuming there is  a possibility that both dot1x (triggered by hostapd) & MAB authentication can happen concurrently. 

How does pacd behave during this? Also how do we make we give higher priority to a specific auth-method lets say 'dot1x'?

doc/pac/Port Access Control.md
```diff
@@ -0,0 +1,852 @@
...
+    ],
+    "priority_list": [
+      "dot1x",
+      "mab"
+    ],
+    "port_pae_role": "authenticator",
+    "port_control_mode": "auto",
+    "host_control_mode": "multi_auth",
+    "reauth_period": 60,
+    "reauth_enable": "true",
```
praveenraja1, line 357: Just a suggestion, maybe consider removing this field and indicting disabled by setting reauth_period as 0. This would be inline with how its done in hostapd.conf 

# EAP reauthentication period in seconds (default: 3600 seconds; 0 = disable
# reauthentication).
#eap_reauth_period=3600

praveenraja1 (commented):
Overall looks good, left some minor comments/questions

zhangyanzhao:
@adyeung multiple code PRs listed in this HLD are not merged yet and what is the plan for those code PRs? Thanks.

adyeung:
@zhangyanzhao 15 PAC code PRs already merged to master, we are tracking a few more to complete the phase 1 for 202411, the remaining code PRs will be deferred to phase 2

PR #1342, opened 2023-05-19
Local Mitigation - HLD

renukamanavalan:


renukamanavalan:
> What is the source of the data used for detection, is it just from xxx-db？Can I use this function to do some missions that the host intrusion detection system do? (such as Suspicious connection detection, suspicious process detection)

First cut -- Source is Redis-DB only.
RFE: We could extend .. But all the time running within the context of container.

renukamanavalan:
> Firstly, since all the data is sourced from the database, won't there be significant limitations? Secondly, are there any experimental data on how much resources the LoM service will consume?

SONiC is going towards maintaining all its state & config in DB only. This is the beautiful model (_not just for this project_) but by general design principals. Many of the operations are possible, like service restart is possible via DB and more coming to DB. So there are *no* significant limitations.

As we mature, we could increase our reach by mounting more dirs as RO and we could potentially extend our capabilities via D-Bus. 
As for resources, it could be set at Docker level. Most of our code will go on subscribe  model and fallback to poll where subscribe can be expensive. Expect code to sleep most of the time in a healthy box and wake up only upon some signals.

PR #1416, opened 2023-07-06
[FEC] Design for auto-fec

dgsudharsan:
This HLD introduces design for auto fec. This document tries to bring in a deterministic approach when FEC and autoneg are configured together which is currently left to vendor implementation.

| Repo  | PR title | State |
| sonic-swss | [FEC]Auto FEC initial changes(https://github.com/sonic-net/sonic-swss/pull/2874) | GitHub issue/pull request detail |
| sonic-sairedis | [FEC]Auto FEC initial changes(https://github.com/sonic-net/sonic-sairedis/pull/1271) | GitHub issue/pull request detail |
| sonic-swss | [FEC]Auto FEC initial changes(https://github.com/sonic-net/sonic-swss/pull/2893) | GitHub issue/pull request detail |
| sonic-utilities | [portconfig]Auto FEC initial changes(https://github.com/sonic-net/sonic-utilities/pull/2965 ) | GitHub issue/pull request detail |
| sonic-buildimage   | [yang] Update sonic-port yang model to support auto FEC (https://github.com/sonic-net/sonic-buildimage/pull/16389) | GitHub issue/pull request detail |

### DIFF
--- a/doc/port_auto_neg/auto-fec.md
+++ b/doc/port_auto_neg/auto-fec.md
@@ -1,0 +1,181 @@
+# SONiC Port Auto FEC Design #
+
+## Table of Content
+
+- [Revision](#revision)
+- [Scope](#scope)
+- [Definitions/Abbreviations](#definitionsabbreviations)
+- [Overview](#overview)
+- [Requirements](#requirements)
+- [High-Level Design](#high-level-design)
+- [SAI API Requirements](#sai-api-requirements)
+- [SWSS Enhancements](#swss-enhancements)
+- [Configuration and management ](#configuration-and-management)
+    - [Config command](#config-command)
+    - [Show command](#show-command)
+- [YANG Model changes](#yang-model-changes)
+- [Warmboot and Fastboot Considerations](#warmboot-and-fastboot-considerations)
+- [Restrictions/Limitations](#restrictionslimitations)
+- [Testing Design](#testing-design)
+    - [VS Test cases](#vs-test-cases)
+
+
+### Revision
+
+ | Rev |     Date    |       Author        | Change Description                         |
+ |:---:|:-----------:|:-------------------:|--------------------------------------------|
+ | 0.1 |             |      Sudharsan      | Initial version                            |
+
+### Scope
+This document is the design document for port auto fec mode on SONiC. This includes the requirements, CLI change, yang model change and swss change.
+
+### Definitions/Abbreviations
+ FEC - Forward Error Correction
+
+### Overview
+SONiC supports the following FEC settings:
+
+**None**: FEC is disabled.
+
+**Reed Solomon (RS)**: IEEE 802.3 Clause 108 (CL108) on individual 25G channels and Clause 91 on 100G (4channels). This is the highest FEC algorithm, providing the best bit-error correction.
+
+**Fire Code (FC)**: IEEE 802.3 Clause 74 (CL74). Base-R provides less protection from bit errors than RS FEC but adds less latency.
+
+The behavior of FEC when autoneg is configured is currently not defined. By introducing a new FEC mode called 'auto' this document brings a deterministic approach in  behavior of FEC when autoneg is configured
+
+
+### Requirements
+
+Primary requirements for configuring FEC are
+- Honor the backward compatibility.
+- Allow user to enable FEC to be auto negotiated
+- Allow user to force set FEC
+
+
+### High-Level Design
+In order to facilitate the behavior of FEC when autoneg is enabled, the SAI attribute SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE was introduced. Before this attribute the behavior is not clearly defined and it is dependent on vendor SAI implementation. 
+
+A new FEC mode called 'auto' is introduced. 
+- When FEC mode is set to any of the legacy modes (none, rs, FC) SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be set to true. User FEC will take precedence
+- When FEC mode is seto 'auto' SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be set to false. Auto negotiated FEC will take precedence.
+- To maintain backward compatibility, the SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be queried and will be used only when it is supported.
+
+### SAI API Requirements
+
+The following is an existing attribute which was recently added to support the FEC override functionality.
+```
+    /**
+     * @brief FEC mode auto-negotiation override status
+     *
+     * If set to true, any auto-negotiated FEC mode will be
+     * overridden by the value configured in SAI_PORT_ATTR_FEC_MODE
+     *
+     * @type bool
+     * @flags CREATE_AND_SET
+     * @default false
+     */
+    SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE,
+```
+Any SAI vendors who want to make use of this deterministic behavior should implement the above attribute.
+
+In order to get operation FEC the following attributes will be introduced
+```
+    /**
+     * @brief Operational FEC mode
+     *
+     * If port is down, the returned value should be zero.
+     * If auto negotiation is on, the returned value should be the negotiated fec.
+     *
+     * @type sai_port_fec_mode_t
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_OPER_FEC,
+```
+
+### SWSS Enhancements
+
+As specified in the high level design, the SAI attribute SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be queried in portsorch. If it is not supported the existing behavior will continue to be in effect. If it is supported and FEC is configured, SAI_PORT_ATTR_FEC_MODE will be set to the configured FEC value followed by SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE set to true.
+If FEC is configured as 'auto' by the user SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be set to false. If AN is set to false, SAI_PORT_ATTR_FEC_MODE will be set to SAI_PORT_FEC_MODE_NONE.
+If AN is set to true and FEC is not set, no FEC related attributes will be programmed and existing behavior will continue.
+The below table covers different scenarios of what will be programmed in SAI when override is supported and not supported for various combinations of FEC and autoneg
+
+ | Idx | Override supported | Autoneg    |    FEC     | SAI Attributes set by orchagent                                                 |
+ |:---:|:------------------:|:----------:|:----------:|---------------------------------------------------------------------------------|
+ |  1  |       False        | True/False | none/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc                                               |
+ |  2  |       False        | True/False | auto       | No FEC attributes will be set                                                   |
+ |  3  |       False        | True/False | N/A        | No FEC attributes will be set                                                   |
+ |  4  |       True         | True       | none/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE=True |
+ |  5  |       True         | True       | auto       | SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE=False                                  |
+ |  6  |       True         | True       | N/A        | SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will not be set                        |
+ |  7  |       True         | False      | None/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc                                               |
+ |  8  |       True         | False      | auto       | No FEC attributes will be set                                                   |
+ |  9  |       True         | False      | N/A        | No FEC attributes will be set                                                   |
+ 
+ The portsorch will also be responsible to update  operational FEC. This operational mode is applicable only when the mode is 'auto'. When the FEC mode is auto and oper up is detected, SAI_PORT_ATTR_OPER_FEC will be queried and updated in the STATE_DB PORT_TABLE field 'fec'. 
+
+### Configuration and management
+
+#### Config command
+
+No new CLI commands are introduced to support auto FEC. The existing 'config interface fec' is extended to have additional option called 'auto'. If this option is set without autoneg enabled, it will default to 'none'
+
+```
+Format:
+  config interface fec <interface_name> <mode>
+
+Arguments:
+  interface_name: name of the interface to be configured. e.g: Ethernet0
+  mode: none/rs/fc/auto
+
+Example:
+  config interface fec Ethernet0 auto
+
+Return:
+  error message if interface_name or mode is invalid otherwise empty
+```
+
+#### Show command
+
+Currently the configured FEC mode is displayed in "show interfaces status" command. With the introduction of mode 'auto' it becomes necessary to display operational FEC. For modes other than auto, the config will match the operational FEC. In case of mode 'auto' the operational FEC will be queried from SAI during oper up sequence using SAI_PORT_ATTR_OPER_FEC. This will be updated to state_db PORT_TABLE field "fec". The value will contain the operational FEC and have the config auto in paranthesis as shown below.
+The show command will query the state DB and if there is FEC field, it would be displayed.
+If the field is not present, the config will be displayed, similar to the legacy behavior.
+
+````
+show interfaces  status
+   Interface            Lanes    Speed    MTU    FEC       Alias          Vlan    Oper    Admin             Type    Asym PFC
+------------  ---------------  -------  -----  ---------   -------  ------------  ------  -------  ---------------  ----------
+   Ethernet0          0,1,2,3     100G   9100    N/A       etp1        routed      up       up   QSFP+ or later         N/A
+   Ethernet4          4,5,6,7     100G   9100     rs       etp2  PortChannel1      up       up   QSFP+ or later         N/A
+   Ethernet8        8,9,10,11     100G   9100   rs(auto)   etp3        routed      up       up   QSFP+ or later         N/A
+````
+
+### YANG model changes
+
+The yang model will be modified to accept additional value 'auto' for FEC
+
+```
+leaf fec {
+	type string {
+		pattern "rs|fc|none|auto";
+	}
+}
+```
+
+### Warmboot and Fastboot Considerations
+
+No impact to warm or fastboot
+
+### Restrictions/Limitations
+
+Since previously the behavior of FEC with autoneg is undefined, if a vendor always gives precedence for auto negotiated FEC over user configured FEC, implementing the new attribute may lead to backward incompatibility issues when SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE is implemented. Since orchagent will always set SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE to true when FEC is explicitly configured, there will be difference in behavior before and after upgrade.
+
+In order to mitigate for those platforms, a db migrator plugin can be introduced which will be executed based on platform check. This plugin will change change 'fec' to 'auto' if the attribute exists in port table and 'autoneg' is enabled. This will make sure that previously when autoneg was given precedence over user configuration, it will continue to happen even after the upgrade.
+
+### Testing Design
+
+#### VS Test cases
+
+1. Add SWSS VS test to query SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE and if supported expect SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE to be set to true when FEC is configured.
+2. If SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE is not suppored it should not be set when configuring FEC
+3. Set FEC mode to auto and AN=true verify SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE is set to false
+4. Set FEC mode to auto and AN=false verify no FEC mode is programmed
--- a/doc/port_auto_neg/auto-fec.md
+++ b/doc/port_auto_neg/auto-fec.md
@@ -95,7 +95,7 @@ In order to get operation FEC the following attributes was recently introduced
 ### SWSS Enhancements
 
 As specified in the high level design, the SAI attribute SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be queried in portsorch. If it is not supported the existing behavior will continue to be in effect. If it is supported and FEC is configured, SAI_PORT_ATTR_FEC_MODE will be set to the configured FEC value followed by SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE set to true.
-If FEC is configured as 'auto' by the user SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be set to false. If AN is set to false, SAI_PORT_ATTR_FEC_MODE will be set to SAI_PORT_FEC_MODE_NONE. Only when autoneg is set to true, FEC mode auto will take effect.
+If FEC is configured as 'auto' by the user SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be set to false. With FEC as 'auto', if AN is set to false, SAI_PORT_ATTR_FEC_MODE will be set to SAI_PORT_FEC_MODE_NONE. Only when autoneg is set to true, FEC mode auto will take effect.
 If AN is set to true and FEC is not set, no FEC related attributes will be programmed and existing behavior will continue.
 The below table covers different scenarios of what will be programmed in SAI when override is supported and not supported for various combinations of FEC and autoneg. The override supported column is updated based on querying SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE from SAI , while Autoneg and FEC comes from user configuration.
 
--- a/doc/port_auto_neg/auto-fec.md
+++ b/doc/port_auto_neg/auto-fec.md
@@ -62,7 +62,7 @@ A new FEC mode called 'auto' is introduced.
 
 ### SAI API Requirements
 
-The following is an existing attribute which was recently added to support the FEC override functionality.
+The following is an existing attribute which was added to support the FEC override functionality.
 ```
     /**
      * @brief FEC mode auto-negotiation override status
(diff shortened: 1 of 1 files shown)


dgsudharsan:
@prgeor @ridahanif96  Please review this HLD

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,191 @@
...
+
+### SWSS Enhancements
+
+As specified in the high level design, the SAI attribute SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be queried in portsorch. If it is not supported the existing behavior will continue to be in effect. If it is supported and FEC is configured, SAI_PORT_ATTR_FEC_MODE will be set to the configured FEC value followed by SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE set to true.
+If FEC is configured as 'auto' by the user SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be set to false. If AN is set to false, SAI_PORT_ATTR_FEC_MODE will be set to SAI_PORT_FEC_MODE_NONE.
+If AN is set to true and FEC is not set, no FEC related attributes will be programmed and existing behavior will continue.
+The below table covers different scenarios of what will be programmed in SAI when override is supported and not supported for various combinations of FEC and autoneg. The override supported column is updated based on querying SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE from SAI , while Autoneg and FEC comes from user configuration.
+
+
+ | Idx | Override supported | Autoneg    |    FEC     | SAI Attributes set by orchagent                                                 |
```
venkatmahalingam, line 103: Can we have another column for current behavior? you can mention platform specific/undeterministic for some cases.

Can we mention Override supported (SAI capability) in the column name if possible?

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,191 @@
...
+The show command will query the state DB and if there is FEC field, it would be displayed.
+If the field is not present, the config will be displayed, similar to the legacy behavior.
+
+````
+show interfaces  status
+   Interface            Lanes    Speed    MTU    FEC       Alias          Vlan    Oper    Admin             Type    Asym PFC
+------------  ---------------  -------  -----  ---------   -------  ------------  ------  -------  ---------------  ----------
+   Ethernet0          0,1,2,3     100G   9100    N/A       etp1        routed      up       up   QSFP+ or later         N/A
+   Ethernet4          4,5,6,7     100G   9100     rs       etp2  PortChannel1      up       up   QSFP+ or later         N/A
+   Ethernet8        8,9,10,11     100G   9100   rs(auto)   etp3        routed      up       up   QSFP+ or later         N/A
```
venkatmahalingam, line 153: rs(auto) -> auto (negotiated: rs)?

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,191 @@
...
+}
+```
+
+### Warmboot and Fastboot Considerations
+
+No impact to warm or fastboot
+
+### Restrictions/Limitations
+
+Since previously the behavior of FEC with autoneg is undefined, if a vendor always gives precedence for auto negotiated FEC over user configured FEC, implementing the new attribute may lead to backward incompatibility issues when SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE is implemented. Since orchagent will always set SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE to true when FEC is explicitly configured, there will be difference in behavior before and after upgrade.
```
venkatmahalingam, line 174: Since orchagent will always set SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE to true when FEC is explicitly configured, there will be difference in behavior before and after upgrade.
-> Can orchagent configure SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE to true only when SAI is capable?

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,191 @@
...
+The below table covers different scenarios of what will be programmed in SAI when override is supported and not supported for various combinations of FEC and autoneg. The override supported column is updated based on querying SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE from SAI , while Autoneg and FEC comes from user configuration.
+
+
+ | Idx | Override supported | Autoneg    |    FEC     | SAI Attributes set by orchagent                                                 |
+ |:---:|:------------------:|:----------:|:----------:|---------------------------------------------------------------------------------|
+ |  1  |       False        | False      | none/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc                                               |
+ |  2  |       False        | False      | auto       | CLI will throw error. No FEC attributes will be set                             |
+ |  3  |       False        | False      | N/A        | No FEC attributes will be set                                                   |
+ |  4  |       False        | True       | none/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc. Behavior is undeterministic when AN is enabled and user has configured FEC |
```
ridahanif96, line 108: What will be priority precedence for rs/fc when auto mode is configured? Can you please add this as well?

doc/port_auto_neg/auto-fec.md
dgsudharsan: I prefer rs(auto) or auto(rs) to avoid making the show interface status output wider.

doc/port_auto_neg/auto-fec.md
dgsudharsan: This is queried at init and set only when supported. If the override is not supported, the attribute will not be programmed. The UT will cover both scenarios where query is supported and not supported

doc/port_auto_neg/auto-fec.md
dgsudharsan: There is no precedence. It depends on the auto-negotiation algorithm which is not in the scope of this HLD.

doc/port_auto_neg/auto-fec.md
dgsudharsan: Updated the column header.
The first 6 rows with override supported= false is the current behavior and hence I don't think a new column is required.

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Warmboot and Fastboot Considerations
+
+No impact to warm or fastboot
+
+### Restrictions/Limitations
+
+Since previously the behavior of FEC with autoneg is undefined, if a vendor always gives precedence for auto negotiated FEC over user configured FEC, implementing the new attribute may lead to backward incompatibility issues when SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE is implemented. Since orchagent will always set SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE to true when FEC is explicitly configured, there will be difference in behavior before and after upgrade.
+
+In order to mitigate for those platforms, a db migrator plugin can be introduced which will be executed based on platform check. This plugin will change change 'fec' to 'auto' if the attribute exists in port table and 'autoneg' is enabled. This will make sure that previously when autoneg was given precedence over user configuration, it will continue to happen even after the upgrade.
```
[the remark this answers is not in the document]
dgsudharsan: Hi Madhu,
In the HLD I have added section in config command describing the error that will be received when auto mode is configured on unsupported platform.
In SWSS enhancements section, I mentioned the error log that would be seen when orchagent receives auto configuration but vendor doesn't support it.

doc/port_auto_neg/auto-fec.md
ridahanif96: Ok, thanks

doc/port_auto_neg/auto-fec.md
ridahanif96: I would suggest to consider: auto(rs)

dgsudharsan:
@venkatmahalingam @ridahanif96 @madhupalu @prgeor Can you please review and sign off?

venkatmahalingam (approved):

ridahanif96 (approved):

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+    SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE,
+```
+Any SAI vendors who want to make use of this deterministic behavior should implement the above attribute.
+
+In order to get operation FEC the following attributes was recently introduced
+```
+    /**
+     * @brief Operational FEC mode
+     *
+     * If port is down, the returned value should be zero.
```
prgeor, line 86: @dgsudharsan return value zero or none?

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+     * If auto negotiation is on, the returned value should be the negotiated fec.
```
prgeor, line 87: @dgsudharsan can we just mention, return value is the operational FEC value as configured in the hw (irrespective of AN enabled)

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+     *
+     * @type sai_port_fec_mode_t
+     * @flags READ_ONLY
+     */
+    SAI_PORT_ATTR_OPER_FEC,
```
prgeor, line 92: @dgsudharsan can you share the SAI PR for this?

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+### SWSS Enhancements
+
+As specified in the high level design, the SAI attribute SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be queried in portsorch. If it is not supported the existing behavior will continue to be in effect. If it is supported and FEC is configured, SAI_PORT_ATTR_FEC_MODE will be set to the configured FEC value followed by SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE set to true.
```
prgeor, line 97: @dgsudharsan why not the sequence of SAI OVERRIDE FEC first followed by SAI_PORT_ATTR_FEC_MODE

prgeor:
@dgsudharsan can you attach the code PRs in the description. @yutongzhang-microsoft  FYI

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+If FEC is configured as 'auto' by the user SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be set to false. If AN is set to false, SAI_PORT_ATTR_FEC_MODE will be set to SAI_PORT_FEC_MODE_NONE. Only when autoneg is set to true, FEC mode auto will take effect.
```
prgeor, line 98: @dgsudharsan when AN is set to false, user could still set the FEC mode to 'rs' but will this change affect that because i read `If AN is set to false, SAI_PORT_ATTR_FEC_MODE will be set to SAI_PORT_FEC_MODE_NONE`

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+If AN is set to true and FEC is not set, no FEC related attributes will be programmed and existing behavior will continue.
+The below table covers different scenarios of what will be programmed in SAI when override is supported and not supported for various combinations of FEC and autoneg. The override supported column is updated based on querying SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE from SAI , while Autoneg and FEC comes from user configuration.
+
+
+ | Idx | Override supported (SAI capability)  | Autoneg    |    FEC     | SAI Attributes set by orchagent                                                 |
+ |:---:|:------------------:|:----------:|:----------:|---------------------------------------------------------------------------------|
+ |  1  |       False        | False      | none/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc                                               |
+ |  2  |       False        | False      | auto       | CLI will throw error. No FEC attributes will be set                             |
```
prgeor, line 106: @dgsudharsan does CLI has the ability to query SAI capabilities? If not, how will this CLI implementation look like?

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+ |  5  |       False        | True       | auto       | No FEC attributes will be set                                                   |
+ |  6  |       False        | True       | N/A        | No FEC attributes will be set                                                   |
+ |  7  |       True         | True       | none/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE=True |
+ |  8  |       True         | True       | auto       | SAI_PORT_ATTR_FEC_MODE=none, SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE=False     |
+ |  9  |       True         | True       | N/A        | SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will not be set                        |
+ | 10  |       True         | False      | None/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc                                               |
+ | 11  |       True         | False      | auto       | SAI_PORT_ATTR_FEC_MODE=none, SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE=False. However the auto will take effect only when AN is enabled and until then the mode will be none. |
```
prgeor, line 115: @dgsudharsan please throw the error to make it clear to the user that auto need AN to be enabled.

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+ | 12  |       True         | False      | N/A        | No FEC attributes will be set                                                   |
```
prgeor, line 116: @dgsudharsan what do you mean by "N/A"

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+
+
+ | Idx | Override supported (SAI capability)  | Autoneg    |    FEC     | SAI Attributes set by orchagent                                                 |
+ |:---:|:------------------:|:----------:|:----------:|---------------------------------------------------------------------------------|
+ |  1  |       False        | False      | none/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc                                               |
+ |  2  |       False        | False      | auto       | CLI will throw error. No FEC attributes will be set                             |
+ |  3  |       False        | False      | N/A        | No FEC attributes will be set                                                   |
+ |  4  |       False        | True       | none/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc. Behavior is undeterministic when AN is enabled and user has configured FEC |
+ |  5  |       False        | True       | auto       | No FEC attributes will be set                                                   |
```
prgeor, line 109: @dgsudharsan why not throw the error if the platform does not support auto?

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+ |  6  |       False        | True       | N/A        | No FEC attributes will be set                                                   |
+ |  7  |       True         | True       | none/rs/fc | SAI_PORT_ATTR_FEC_MODE=none/rs/fc SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE=True |
```
prgeor, line 111: @dgsudharsan how the case is handled when user configures the FEC (at which time AN is not enabled) first and later on user configures AN=enabled

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+ |  8  |       True         | True       | auto       | SAI_PORT_ATTR_FEC_MODE=none, SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE=False     |
+ |  9  |       True         | True       | N/A        | SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will not be set                        |
```
prgeor, line 113: @dgsudharsan why not set override = False. What is the issue?

doc/port_auto_neg/auto-fec.md
prgeor, line 115: @dgsudharsan how is 11 different from idex=8 case where also SAI_PORT_ATTR_FEC_MODE=None. Throwing the error will make the distinction between 8 and 11 clear

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+
+**Fire Code (FC)**: IEEE 802.3 Clause 74 (CL74). Base-R provides less protection from bit errors than RS FEC but adds less latency.
+
+The behavior of FEC when autoneg is configured is currently not defined. By introducing a new FEC mode called 'auto' this document brings a deterministic approach in  behavior of FEC when autoneg is configured
+
+
+### Requirements
+
+Primary requirements for configuring FEC are
+- Honor the backward compatibility.
```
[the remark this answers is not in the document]
dgsudharsan: Backward compatibility means if an image upgrade is done from an image without the feature to image with the feature, there should be no issues. E.g a port with 'rs' fec and AN enabled should work the same before and after the upgrade to image with the feature.

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+- Allow user to force set FEC
+
+
+### High-Level Design
+In order to facilitate the behavior of FEC when autoneg is enabled, the SAI attribute SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE was introduced. Before this attribute the behavior is not clearly defined and it is dependent on vendor SAI implementation. 
+
+A new FEC mode called 'auto' is introduced. 
+- When FEC mode is set to any of the legacy modes (none, rs, FC) SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be set to true. User FEC will take precedence
+- When FEC mode is seto 'auto' SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be set to false. Auto negotiated FEC will take precedence.
+- To maintain backward compatibility, the SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE will be queried and will be used only when it is supported.
```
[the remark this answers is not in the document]
dgsudharsan: It is captured in line 98 https://github.com/sonic-net/SONiC/pull/1416/files#diff-34d144e1c4293c17a7a99c6901da1b61bdce1c817bad4eae34c5e2a4ebb4495cR98

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+Primary requirements for configuring FEC are
+- Honor the backward compatibility.
+- Allow user to enable FEC to be auto negotiated
+- Allow user to force set FEC
+
+
+### High-Level Design
+In order to facilitate the behavior of FEC when autoneg is enabled, the SAI attribute SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE was introduced. Before this attribute the behavior is not clearly defined and it is dependent on vendor SAI implementation. 
+
+A new FEC mode called 'auto' is introduced.
```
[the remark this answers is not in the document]
dgsudharsan: It is captured in line 98 https://github.com/sonic-net/SONiC/pull/1416/files#diff-34d144e1c4293c17a7a99c6901da1b61bdce1c817bad4eae34c5e2a4ebb4495cR98

doc/port_auto_neg/auto-fec.md
dgsudharsan: Updated from latest SAI headers

doc/port_auto_neg/auto-fec.md
dgsudharsan: https://github.com/opencomputeproject/SAI/pull/1801

doc/port_auto_neg/auto-fec.md
dgsudharsan: Sequence doesn't matter here as the port will be admin down during setting of FEC. Logically I felt its better to have FEC attribute defined first and then we request SAI to override it using override attribute.

doc/port_auto_neg/auto-fec.md
dgsudharsan: This line is applicable only when when FEC is auto. I have rephrased for better clarity.

doc/port_auto_neg/auto-fec.md
dgsudharsan: CLI doesn't query SAI capabilities. Orchagent queries SAI capabilities and updates in state DB. CLI will query state DB and act based on it. This is the existing flow for regular FEC. We extend it with auto mode.

doc/port_auto_neg/auto-fec.md
dgsudharsan: I don't think this is an error scenario. When the configuration is done, AN can be configured first followed by FEC and vice-versa. Throwing error will mislead the user. Instead I can have a notice log.

doc/port_auto_neg/auto-fec.md
dgsudharsan: N/A should be tread as not available. This is what we use in show interfaces and other commands when a particular configuration is not available.

doc/port_auto_neg/auto-fec.md
dgsudharsan: Updated

doc/port_auto_neg/auto-fec.md
dgsudharsan: Both are independent configurations. However FEC will get auto negotiated only when AN is enabled. Without AN enabled we will set it to NONE which is SAI default. This is explained in line 98
https://github.com/sonic-net/SONiC/pull/1416/files#diff-34d144e1c4293c17a7a99c6901da1b61bdce1c817bad4eae34c5e2a4ebb4495cR98

doc/port_auto_neg/auto-fec.md
dgsudharsan: This is the case when there is no fec config. N/A means not available. FEC override will be set only when FEC is configured.

doc/port_auto_neg/auto-fec.md
dgsudharsan: As I said we should not throw error. User can configure in any order and even during config reload, the configurations can be processed in any order. The FEC will be programmed to override false which will take effect only when auto neg is enabled. Otherwise it will default to none.

doc/port_auto_neg/auto-fec.md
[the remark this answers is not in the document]
prgeor: @dgsudharsan thanks. can you capture this in the doc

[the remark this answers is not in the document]
dgsudharsan: Done

doc/port_auto_neg/auto-fec.md
prgeor: @dgsudharsan FEC override = False or FEC override not set aren't the same?

doc/port_auto_neg/auto-fec.md
prgeor: @dgsudharsan ok. please add notice log

doc/port_auto_neg/auto-fec.md
prgeor: @dgsudharsan I think we should have separate column (`oper-FEC`) for operation FEC vs user configured FEC in CONFIG_DB. This is not just for auto. Even if a vendor does not support FEC override, we should have a separte column to show operational FEC. This way its clear for the user just like we distinguish operational link status (UP/down) vs user configured admin status (UP/down)

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Warmboot and Fastboot Considerations
+
+No impact to warm or fastboot
+
+### Restrictions/Limitations
+
+Since previously the behavior of FEC with autoneg is undefined, if a vendor always gives precedence for auto negotiated FEC over user configured FEC, implementing the new attribute may lead to backward incompatibility issues when SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE is implemented. Since orchagent will always set SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE to true when FEC is explicitly configured and the attribute is supported by SAI, there will be difference in behavior before and after upgrade.
+
+In order to mitigate for those platforms, a db migrator plugin can be introduced which will be executed based on platform check. This plugin will change change 'fec' to 'auto' if the attribute exists in port table and 'autoneg' is enabled. This will make sure that previously when autoneg was given precedence over user configuration, it will continue to happen even after the upgrade.
```
prgeor, line 187:
```suggestion
In order to mitigate for those platforms, a db migrator plugin can be introduced which will be executed based on platform check. This plugin will change 'fec' to 'auto' if the attribute exists in port table and 'autoneg' is enabled. This will make sure that previously when autoneg was given precedence over user configuration, it will continue to happen even after the upgrade.
```

doc/port_auto_neg/auto-fec.md
```diff
@@ -0,0 +1,202 @@
...
+}
+```
+
+### Warmboot and Fastboot Considerations
+
+No impact to warm or fastboot
+
+### Restrictions/Limitations
+
+Since previously the behavior of FEC with autoneg is undefined, if a vendor always gives precedence for auto negotiated FEC over user configured FEC, implementing the new attribute may lead to backward incompatibility issues when SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE is implemented. Since orchagent will always set SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE to true when FEC is explicitly configured and the attribute is supported by SAI, there will be difference in behavior before and after upgrade.
```
prgeor, line 185: `SAI_PORT_ATTR_AUTO_NEG_FEC_MODE_OVERRIDE` support by the platform itself means that the platform is going to OVERRIDE its existing behavior. isn't it? If so, the difference in behavior is expected...i mean it will be a breaking change though...If the platform choose to keep the original behavior, then it means the platform is not supporting the override SAI attribute in which case why support override attribute?

dgsudharsan: If there is  difference in behavior, it should be also made backward compatible through db_migrator. It is upto the vendors to decide whether they want to support new attribute and introduce db_migrator if they need it for solving the backward incompatibility or avoid introducing the attribute altogether.

doc/port_auto_neg/auto-fec.md
dgsudharsan: FEC override attribute will be set only in the flow of setting FEC. When FEC itself is not set there will be no override attribute that needs to be programmed.

dgsudharsan: In addition by default the attribute is false, so if not set the default value expected from SAI is false. However the table is to capture what is programmed from orchagent perspective on specific combinations

prgeor (approved):

skg-net:
@dgsudharsan Can you please update the Quality Metric (Alpha/Beta/GA) for the feature either in this PR comments or in HLD itself based on https://github.com/sonic-net/SONiC/blob/master/doc/SONiC%20feature%20quality%20definition.md
Thanks

PR #1428, opened 2023-07-14
NETCONF Server HLD

ntt-omw:
NETCONF Server HLD


doc/mgmt_netconf/hld_NETCONF-NTT.md
```diff
@@ -0,0 +1,790 @@
...
+
+Example of get-config
+
+```
+> print(m.get_config(source="running", filter=("subtree", interfaces)))
+
+<rpc-reply xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="urn:uuid:4e02f166-fc2d-4701-a7a1-b15145cc01d6"><data><interfaces xmlns="http://openconfig.net/yang/interfaces"><interface><name>Ethernet0</name><config><name>Ethernet0</name></config><subinterfaces><subinterface><index>0</index><config><index>0</index></config><ipv4 xmlns="http://openconfig.net/yang/interfaces/ip"><addresses><address><ip>10.1.1.1</ip><config><ip>10.1.1.1</ip><prefix-length>24</prefix-length></config></address></addresses></ipv4></subinterface></subinterfaces></interface></interfaces></data></rpc-reply>
+```
+
+## Open/Action items - if any
```
kwangsuk, line 701: It would be good to mention the notification support (https://datatracker.ietf.org/doc/html/rfc8639) in case for scaling up capabilities in the future.

mbalachandar:
Will the NETCONF server be disabled by default ? since we already have REST, and Telemetry running.

mbalachandar:
NETCONF server will be running in separate docker container?

doc/mgmt_netconf/hld_NETCONF-NTT.md
```diff
@@ -0,0 +1,790 @@
...
+| unlock             | No        | [*4]   |
+| close-session      | Yes       |        |
+| kill-session       | Yes       |        |
+
+- [*1] only below options are supported:
+  - operation : merge,delete
+  - target : running (as commit feature is not supported, only Running config is applicable)
+  - test-option : test-and-set
+  - error-option : rollback-on-error
+- [*2] get-config will return config in netopeer2/sysrepo and not from CONFIG_DB
```
kwangsuk, line 185: It is understood that the sysrepo serves running datastore. But, I am not clear on how to init the sysrepo at the server init. Can you elaborate?

ebiken-ntt: As of this release, sysrepo will serve as running database only for the config modified by NETCONF. Thus, it will be initialized with no data on startup. In future, we are planning to introduce functionality to sync sysrepo with CONFIG_DB but that is not scope of this release.
Would this answer your question? Let me know if you were asking from different aspect. (I will update HLD doc otherwise)

zyboy2000:
How do ensure that the NetConf running database and Redis database are synchronized?   for example CLI ?  Is it possible for NetConf to fetch directly from the Redis database and deprecate its own running database

ridahanif96:
In the context of NETCONF capabilities, does this HLD will cover extended capabilities as well?

ebiken-ntt:
Noting as future reference. (Comment from Kwan Kim)

In today’s (14th Sep) meeting, one of questions from NTT is about NETCONF Get-config operation.

NETCONF Get-config can be translated to REST Get with the “content” query parameter, i.e. content=config.
https://datatracker.ietf.org/doc/html/rfc8040#section-4.8.1

We have opened PRs for query parameter support.
https://github.com/sonic-net/sonic-mgmt-common/pull/102
https://github.com/sonic-net/sonic-mgmt-framework/pull/119

zhangyanzhao:
not target 202311 per Kanji update. Keep a record.

burnCalories:
@ntt-omw @nakano-omw 
Hi ，
I have read the document [you](https://github.com/ntt-omw/SONiC/blob/ntt_netconf/doc/mgmt_netconf/hld_NETCONF-NTT.md) submitted and I fully agree with your design for integrating netconf into Sonic. I wanted to inquire if you have already completed the development work to support netconf in Sonic. Recently, I have been considering enabling netconf functionality in my Sonic switch. Although your documentation is very detailed, I am not sure where to start. Could you provide me with some advice or suggest some source code for me to study?

PR #1432, opened 2023-07-25
OA crash handling to reinitialize port through xcvrd

mihirpat1:
Proposal to reinitialize ports through xcvrd during OA crash.
The current proposal is created to resolve https://github.com/sonic-net/sonic-platform-daemons/issues/356

| [[sonic-swss] [orchagent]: admin-disable port before setPortSerdesAttribute()](https://github.com/sonic-net/sonic-swss/pull/2831) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [Changes for port reinitialization in case of syncd/swss/oa crash and NPU SI settings update for CMIS transceivers](https://github.com/sonic-net/sonic-platform-daemons/pull/403) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [APIs to help in finding NPU SI settings](https://github.com/sonic-net/sonic-platform-common/pull/410) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[orchagent]Modifying NPU_SI_SETTINGS_SYNC_STATUS based on NPU SI settings application status](https://github.com/sonic-net/sonic-swss/pull/2951) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [Backup STATE_DB PORT_TABLE during warm-reboot](https://github.com/sonic-net/sonic-utilities/pull/3111) | GitHub issue/pull request detail | GitHub pull request check contexts |

### DIFF
--- a/doc/sfp-cmis/Interface-Link-bring-up-sequence.md
+++ b/doc/sfp-cmis/Interface-Link-bring-up-sequence.md
@@ -195,7 +195,7 @@ Following infra will ensure port reinitialization by xcvrd in case of syncd/swss
 1. XCVRD main thread init
 	- XCVRD main thread creates the key CMIS_REINIT_REQUIRED in PORT_TABLE:\<port\> (APPL_DB) with value as true for ports which do NOT have this key present 
 	- XCVRD main thread creates the key MEDIA_SETTINGS_SYNC_STATUS in PORT_TABLE:\<port\> (APPL_DB) with value MEDIA_SETTINGS_DEFAULT for ports which do NOT have this key present.  
-      - For transceivers which do not support media settings, MEDIA_SETTINGS_SYNC_STATUS will stay with value MEDIA_SETTINGS_DEFAULT
+      - For transceivers which do not require media settings, MEDIA_SETTINGS_SYNC_STATUS will stay with value MEDIA_SETTINGS_DEFAULT
 
   Following table describes the various values for MEDIA_SETTINGS_SYNC_STATUS  
 
--- a/doc/sfp-cmis/Interface-Link-bring-up-sequence.md
+++ b/doc/sfp-cmis/Interface-Link-bring-up-sequence.md
@@ -199,6 +199,13 @@ In case of continuous restart of xcvrd, both the keys will still hold the same v
 
 Following infra will ensure port re-initialization by xcvrd in case of syncd/swss/orchagent crash
 
+Pre-requisites for the infra to work:
+  - APPL_DB should be cleared after syncd/swss/orchagent crash/restart, config reload and cold reboot
+  - APPL_DB should be retained after 
+    - pmon crash/restart
+    - xcvrd crash/restart (does not apply to xcvrd restart triggered due to a different docker container crash/restart)
+    - warm reboot
+
 1. XCVRD main thread init
 	- XCVRD main thread creates the key CMIS_REINIT_REQUIRED in PORT_TABLE:\<port\> (APPL_DB) with value as true for ports which do NOT have this key present 
 	- XCVRD main thread creates the key NPU_SI_SETTINGS_SYNC_STATUS in PORT_TABLE:\<port\> (APPL_DB) with value NPU_SI_SETTINGS_DEFAULT for ports which do NOT have this key present.  
@@ -442,13 +449,14 @@ sequenceDiagram
 | -------------- | ------------------------ | --------------- | ------------------------ | ------------------------------------------------------------------------------ | ---------------------- | --------- |
 | Xcvrd restart | N | Y | N | NPU_SI_SETTINGS_DONE | N | N |
 | Pmon restart | N | Y | N | NPU_SI_SETTINGS_DONE | N | N |
+| orchagent restart | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
 | Swss restart | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
 | Syncd restart | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
 | config reload | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
 | Cold reboot | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
 | Warm reboot | N | Y | N | NPU_SI_SETTINGS_DONE | N | N |  
-| config interface shut | N | N | N | NPU_SI_SETTINGS_DONE | N | N/A |
-| config interface no shut | N | N | N | NPU_SI_SETTINGS_DONE | N | N/A |
+| config interface shutdown | N | N | N | NPU_SI_SETTINGS_DONE | N | N/A |
+| config interface startup | N | N | N | NPU_SI_SETTINGS_DONE | N | N/A |
 
 **Transceiver OIR testplan**  
 | Event | APPL_DB_<asic_n> cleared | Xcvrd restarted | NPU SI settings notified | NPU_SI_SETTINGS_SYNC_STATUS value upon event completion | CMIS init triggered |
--- a/doc/sfp-cmis/Interface-Link-bring-up-sequence.md
+++ b/doc/sfp-cmis/Interface-Link-bring-up-sequence.md
@@ -233,7 +233,7 @@ There will be a timeout of 5s for every retry
 6. XCVRD will subscribe to PORT_TABLE in APPL_DB and trigger self-restart if the PORT_TABLE is deleted for the namespace.  
 All threads will be gracefully terminated and xcvrd deinit will be performed followed by issuing a SIGABRT to ensure XCVRD is restarted automatically by supervisord. After respawn, CMIS re-init and NPU_SI_SETTINGS notified is triggered for the ports belonging to the affected namespace
 
-7. syncd/swss/orchagent restart clears the entire APPL-DB, including “NPU_SI_SETTINGS_SYNC_STATUS” and "CMIS_REINIT_REQUIRED" in PORT_TABLE
+7. syncd/swss/orchagent restart (restart triggered due to docker container crash) clears the entire APPL-DB, including “NPU_SI_SETTINGS_SYNC_STATUS” and "CMIS_REINIT_REQUIRED" in PORT_TABLE
 
 8. In case of warm reboot, the APPL_DB is not cleared and hence, once xcvrd is spawned after the reboot, the ports are not initialized again.
 
@@ -436,18 +436,25 @@ sequenceDiagram
 ```
 
 ## Test plan and expectation
+
+**Process and device crash/restart and interface config command handling testplan**  
 | Event          | APPL_DB_<asic_n> cleared | Xcvrd restarted | NPU SI settings renotify | NPU_SI_SETTINGS_SYNC_STATUS value on xcvrd boot-up for initialized transceiver | CMIS re-init triggered | Link flap |
 | -------------- | ------------------------ | --------------- | ------------------------ | ------------------------------------------------------------------------------ | ---------------------- | --------- |
-| Xcvrd restart  | N                        | Y               | N                        | NPU_SI_SETTINGS_DONE                                                           | N                      | N         |
-| Pmon restart   | N                        | Y               | N                        | NPU_SI_SETTINGS_DONE                                                           | N                      | N         |
-| Swss restart   | Y                        | Y               | Y                        | NPU_SI_SETTINGS_DEFAULT                                                        | Y                      | N/A       |
-| Syncd restart  | Y                        | Y               | Y                        | NPU_SI_SETTINGS_DEFAULT                                                        | Y                      | N/A       |
-| config reload  | Y                        | Y               | Y                        | NPU_SI_SETTINGS_DEFAULT                                                        | Y                      | N/A       |
-| Cold reboot    | Y                        | Y               | Y                        | NPU_SI_SETTINGS_DEFAULT                                                        | Y                      | N/A       |
-| Config shut    | N                        | N               | N                        | NPU_SI_SETTINGS_DONE                                                           | N                      | N/A       |
-| Config no shut | N                        | N               | N                        | NPU_SI_SETTINGS_DONE                                                           | N                      | N/A       |
-| Warm reboot    | N                        | Y               | N                        | NPU_SI_SETTINGS_DONE                                                           | N                      | N         |
+| Xcvrd restart | N | Y | N | NPU_SI_SETTINGS_DONE | N | N |
+| Pmon restart | N | Y | N | NPU_SI_SETTINGS_DONE | N | N |
+| Swss restart | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
+| Syncd restart | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
+| config reload | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
+| Cold reboot | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
+| Warm reboot | N | Y | N | NPU_SI_SETTINGS_DONE | N | N |  
+| config interface shut | N | N | N | NPU_SI_SETTINGS_DONE | N | N/A |
+| config interface no shut | N | N | N | NPU_SI_SETTINGS_DONE | N | N/A |
 
+**Transceiver OIR testplan**  
+| Event | APPL_DB_<asic_n> cleared | Xcvrd restarted | NPU SI settings notified | NPU_SI_SETTINGS_SYNC_STATUS value upon event completion | CMIS init triggered |
+| -------------- | ------------------------ | --------------- | ------------------------ | ------------------------------------------------------------------------------ | ---------------------- |
+| Transceiver Removal | N | N | Y | NPU_SI_SETTINGS_DEFAULT | Y |
+| Transceiver Insertion | N | N | Y | NPU_SI_SETTINGS_DONE | Y |
 # Out of Scope 
 Following items are not in the scope of this document. They would be taken up separately
 1. CMIS API feature is not part of this design and the APIs will be used in this design. For CMIS HLD, Please refer to:
--- a/doc/sfp-cmis/Interface-Link-bring-up-sequence.md
+++ b/doc/sfp-cmis/Interface-Link-bring-up-sequence.md
@@ -191,48 +191,50 @@ if transceiver is not present:
 ## Overview
 
 When syncd/swss/orchagent crashes, all ports in the corresponding namespace will be reinitialized by xcvrd irrespective of the current state of the port.  
-If just xcvrd crashes and restarts, then forced reinitialization (CMIS reinit + media settings notify) of ports will not be performed.  
-CMIS_REINIT_REQUIRED and MEDIA_SETTINGS_SYNC_STATUS keys in APPL_DB PORT_TABLE:\<port\> are used to determine if port reinitialization is required or not.
+If just xcvrd crashes and restarts, then forced reinitialization (CMIS reinit + NPU SI settings notification) of ports will not be performed.  
+CMIS_REINIT_REQUIRED and NPU_SI_SETTINGS_SYNC_STATUS keys in PORT_TABLE:\<port\> (APPL_DB) are used to determine if port reinitialization is required or not.
   - CMIS_REINIT_REQUIRED key states if CMIS reinitialization is required for a port after xcvrd is spawned. CMIS_REINIT_REQUIRED helps in mainly driving CMIS reinitialization after syncd/swss/orchagent crash since it will allow reinitializing ports belonging to the relevant namespace of the crashing process. This key is not planned to drive CMIS initialization after transceiver insertion. 
-  - MEDIA_SETTINGS_SYNC_STATUS key is used as a means to communicate the status of applying media settings for a transceiver requiring media settings. This key is used to update the media settings application status between SfpStateUpdateTask, CmisManagerTask and Orchagent. In case of warm reboot or xcvrd restart, this key will prevent application of media settings on the port if media settings are already applied. In case of transceiver insertion, media settings will be applied irrespective of the media settings application status for the port.  
+  - NPU_SI_SETTINGS_SYNC_STATUS key is used as a means to communicate the status of applying NPU SI settings for a transceiver requiring NPU SI settings. This key is used to update the NPU SI settings application status between SfpStateUpdateTask, CmisManagerTask and Orchagent. In case of warm reboot or xcvrd restart, this key will prevent application of NPU SI settings on the port if the settings are already applied. In case of transceiver insertion, NPU SI settings will be applied irrespective of the NPU SI settings application status for the port.  
 
 Following infra will ensure port reinitialization by xcvrd in case of syncd/swss/orchagent crash
 
 1. XCVRD main thread init
 	- XCVRD main thread creates the key CMIS_REINIT_REQUIRED in PORT_TABLE:\<port\> (APPL_DB) with value as true for ports which do NOT have this key present 
-	- XCVRD main thread creates the key MEDIA_SETTINGS_SYNC_STATUS in PORT_TABLE:\<port\> (APPL_DB) with value MEDIA_SETTINGS_DEFAULT for ports which do NOT have this key present.  
-      - For transceivers which do not require media settings, MEDIA_SETTINGS_SYNC_STATUS will stay with value MEDIA_SETTINGS_DEFAULT
+	- XCVRD main thread creates the key NPU_SI_SETTINGS_SYNC_STATUS in PORT_TABLE:\<port\> (APPL_DB) with value NPU_SI_SETTINGS_DEFAULT for ports which do NOT have this key present.  
+      - For transceivers which do not require NPU SI settings, NPU_SI_SETTINGS_SYNC_STATUS will stay with value NPU_SI_SETTINGS_DEFAULT  
 
-  Following table describes the various values for MEDIA_SETTINGS_SYNC_STATUS  
+    Following table describes the various values for NPU_SI_SETTINGS_SYNC_STATUS  
 A transceiver is classified as CMIS SM driven transceiver if its module type is CMIS and it does not have flat memory  
 
-| Value                   | Modifier thread and event                                                                                                                                                     | Consumer thread and purpose                                                                    |
-| ----------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
-| MEDIA_SETTINGS_DEFAULT  | 1\. XCVRD main thread during cold start of XCVRD<br>2. SfpStateUpdateTask during transceiver removal                                                                   | XCVRD main thread during boot-up for deciding to notify media settings                         |
-| MEDIA_SETTINGS_NOTIFIED | 1\. SfpStateUpdateTask while updating and notifying the media settings for non-CMIS SM driven transceivers<br> 2. CmisManagerTask while updating and notifying the media settings for CMIS SM driven transceivers | Not being used currently                                                                       |
-| MEDIA_SETTINGS_DONE     | Orchagent after applying the SI settings                                                                                                                                      | CmisManagerTask for proceeding from CMIS_STATE_MEDIA_SETTINGS_WAIT to CMIS_STATE_AP_CONF state |
+| Value                    | Modifier thread and event                                                                                                                                                                                           | Consumer thread and purpose                                                                     |
+| ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------- |
+| NPU_SI_SETTINGS_DEFAULT  | 1\. XCVRD main thread during cold start of XCVRD<br>2. SfpStateUpdateTask during transceiver removal                                                                                                                | XCVRD main thread during boot-up for deciding to notify NPU SI settings                         |
+| NPU_SI_SETTINGS_NOTIFIED | 1\. SfpStateUpdateTask while updating and notifying the NPU SI settings for non-CMIS SM driven transceivers<br> 2. CmisManagerTask while updating and notifying the NPU SI settings for CMIS SM driven transceivers | Not being used currently                                                                        |
+| NPU_SI_SETTINGS_DONE     | Orchagent after applying the SI settings                                                                                                                                                                            | CmisManagerTask for proceeding from CMIS_STATE_NPU_SI_SETTINGS_WAIT to CMIS_STATE_DP_INIT state |
 
-2. Update and notify media settings to OA  
-      For non-CMIS SM driven transceivers, SfpStateUpdateTask thread will update and notify the media settings to OA based on the value of PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS  
-      If PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS != MEDIA_SETTINGS_DONE, notify media settings will be invoked and will be set to MEDIA_SETTINGS_NOTIFIED for a port requiring media settings.  
-      For CMIS SM driven transceivers, CmisManagerTask thread will update and notify the media settings to OA based on the value of PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS
 
-3. The OA upon receiving media settings will
+2. Update and notify NPU SI settings to OA  
+      For non-CMIS SM driven transceivers, SfpStateUpdateTask thread will update NPU SI settings in the PORT_TABLE (APPL_DB) and notify to OA based on the value of PORT_TABLE:\<port\>.NPU_SI_SETTINGS_SYNC_STATUS  
+      If PORT_TABLE:\<port\>.NPU_SI_SETTINGS_SYNC_STATUS != NPU_SI_SETTINGS_DONE, update and notify NPU SI settings will be invoked and will be set to NPU_SI_SETTINGS_NOTIFIED for a port requiring NPU SI settings.  
+      For CMIS SM driven transceivers, based on the value of PORT_TABLE:\<port\>.NPU_SI_SETTINGS_SYNC_STATUS, CmisManagerTask thread will update NPU SI settings in the PORT_TABLE (APPL_DB) and notify to OA for a port requiring NPU SI settings. The CMIS SM will then transition from CMIS_STATE_AP_CONF to CMIS_STATE_NPU_SI_SETTINGS_WAIT. If port doesn't require NPU SI settings, CMIS SM will transition to CMIS_STATE_DP_INIT state.  
+
+3. The OA upon receiving NPU SI settings will  
 	- Disable port admin status
-	- Request SAI-SDK to apply the media settings via syncd
-		- If SAI-SDK returns success, OA will update the PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS to MEDIA_SETTINGS_DONE
+	- Request SAI-SDK to apply the NPU SI settings via syncd
+		- If SAI-SDK returns success, OA will update the PORT_TABLE:\<port\>.NPU_SI_SETTINGS_SYNC_STATUS to NPU_SI_SETTINGS_DONE
 		- In case of failure, OA will log an error message and proceed to handling the next port
 
-4. For transceivers requiring media settings, media settings will be updated and notified to OA in the CMIS_STATE_AP_PRE_CONF state. Then, CMIS SM will transition to CMIS_STATE_MEDIA_SETTINGS_WAIT state.  
-    If port doesn't require media settings to be applied, CMIS SM will transition to CMIS_STATE_AP_CONF state.  
-
-5. CMIS_STATE_MEDIA_SETTINGS_WAIT state will wait for MEDIA_SETTINGS_DONE and upon reaching to MEDIA_SETTINGS_DONE, CMIS SM will transition to CMIS_STATE_AP_CONF state.  
+4. CMIS_STATE_NPU_SI_SETTINGS_WAIT state will wait for NPU_SI_SETTINGS_DONE and upon reaching to NPU_SI_SETTINGS_DONE, CMIS SM will transition to CMIS_STATE_DP_INIT state.  
 There will be a timeout of 5s for every retry
-6. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED" to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
-7. XCVRD will subscribe to PORT_TABLE in APPL_DB and trigger self-restart if the PORT_TABLE is deleted for the namespace.  
-All threads will be gracefully terminated and xcvrd deinit will be performed followed by issuing a SIGABRT to ensure XCVRD is restarted automatically by supervisord. After respawn, CMIS re-init and media_settings notified is triggered for the ports belonging to the affected namespace
-8. syncd/swss/orchagent restart clears the entire APPL-DB, including “MEDIA_SETTINGS_SYNC_STATUS” and "CMIS_REINIT_REQUIRED" in PORT_TABLE
-9. In case of warm reboot, the APPL_DB is not cleared and hence, once xcvrd is spawned after the reboot, the ports are not initialized again.
+
+5. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED" to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
+
+6. XCVRD will subscribe to PORT_TABLE in APPL_DB and trigger self-restart if the PORT_TABLE is deleted for the namespace.  
+All threads will be gracefully terminated and xcvrd deinit will be performed followed by issuing a SIGABRT to ensure XCVRD is restarted automatically by supervisord. After respawn, CMIS re-init and NPU_SI_SETTINGS notified is triggered for the ports belonging to the affected namespace
+
+7. syncd/swss/orchagent restart clears the entire APPL-DB, including “NPU_SI_SETTINGS_SYNC_STATUS” and "CMIS_REINIT_REQUIRED" in PORT_TABLE
+
+8. In case of warm reboot, the APPL_DB is not cleared and hence, once xcvrd is spawned after the reboot, the ports are not initialized again.
 
 ## XCVRD init sequence to support port reinitialization during syncd/swss/orchagent crash
 
@@ -244,21 +246,21 @@ sequenceDiagram
     participant SfpStateUpdateTask
     participant DomInfoUpdateTask
 
-    Note over XCVRDMT: Load new platform specific api class,<br> sfputil class and load namespace details<br> load_media_settings()
+    Note over XCVRDMT: Load new platform specific api class,<br> sfputil class and load namespace details<br> load_NPU_SI_SETTINGS()
     XCVRDMT ->> XCVRDMT: Wait for port config completion
     loop lport in logical_port_list
         alt if CMIS_REINIT_REQUIRED not in PORT_TABLE:<lport>
             XCVRDMT ->> APPL_DB: PORT_TABLE:<lport>.CMIS_REINIT_REQUIRED = true
         end
-        alt if MEDIA_SETTINGS_SYNC_STATUS not in PORT_TABLE:<lport>
-            XCVRDMT ->> APPL_DB: PORT_TABLE:<lport>.MEDIA_SETTINGS_SYNC_STATUS = MEDIA_SETTINGS_DEFAULT
+        alt if NPU_SI_SETTINGS_SYNC_STATUS not in PORT_TABLE:<lport>
+            XCVRDMT ->> APPL_DB: PORT_TABLE:<lport>.NPU_SI_SETTINGS_SYNC_STATUS = NPU_SI_SETTINGS_DEFAULT
         end
     end
-    Note over APPL_DB: PORT_TABLE:<lport><br>CMIS_REINIT_REQUIRED : true/false<br>MEDIA_NOTIFY_REQUIRED : true/false
+    Note over APPL_DB: PORT_TABLE:<lport><br>CMIS_REINIT_REQUIRED : true/false<br>NPU SI_NOTIFY_REQUIRED : true/false
     XCVRDMT ->> CmisManagerTask: Spawns
     XCVRDMT ->> DomInfoUpdateTask: Spawns
     XCVRDMT ->> SfpStateUpdateTask: Spawns
-    par XCVRDMT, CmisManagerTask, SfpStateUpdateTask, DomInfoUpdateTask
+    par XCVRD main thread, CmisManagerTask, SfpStateUpdateTask, DomInfoUpdateTask
         loop Wait for stop_event else poll every 60s
             DomInfoUpdateTask->>DomInfoUpdateTask: Update TRANSCEIVER_DOM_SENSOR,<br>TRANSCEIVER_STATUS (HW section)<br>TRANSCEIVER_PM tables
         end
@@ -276,7 +278,7 @@ sequenceDiagram
     end
 ```
 
-## SfpStateUpdateTask's role to notify media settings to OA during xcvrd boot-up
+## SfpStateUpdateTask's role to notify NPU SI settings to OA during xcvrd boot-up
 
 ```mermaid
 sequenceDiagram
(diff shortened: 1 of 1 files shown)


mihirpat1:
@prgeor @shyam77git @jaganbal-a @amnsinghal - It will be great if you can help in reviewing this HLD

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,7 +185,235 @@ if transceiver is not present:
...
+| Syncd restart    | Y               | Y               | Y              | MEDIA_SETTINGS_DEFAULT                                                        | Y                      | Y         |
+| config   reload  | Y               | Y               | Y              | MEDIA_SETTINGS_DEFAULT                                                        | Y                      | Y         |
+| Cold reboot      | Y               | Y               | Y              | MEDIA_SETTINGS_DEFAULT                                                        | Y                      | Y         |
+| Config shut      | N               | N               | N              | MEDIA_SETTINGS_DONE                                                           | N                      | Y         |
+| Config no   shut | N               | N               | N              | MEDIA_SETTINGS_DONE                                                           | N                      | Y         |
+| Warm   reboot    | N               | Y               | N              | MEDIA_SETTINGS_DONE                                                           | N                      | N         |
 # Out of Scope
```
shivuv, line 417: Please update "Out of scope" section as some of it's covered as part of updated HLD.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+| MEDIA_SETTINGS_DEFAULT  | XCVRD main thread during cold   start of xcvrd         | XCVRD main thread   during boot-up for deciding to notify media settings                     |
+|                         | SfpStateUpdateTask during   transceiver removal        |                                                                                              |
+| MEDIA_SETTINGS_NOTIFIED | SfpStateUpdateTask while   updating the media settings | Not being used currently                                                                     |
+| MEDIA_SETTINGS_DONE     | Orchagent after applying the SI   settings             | CmisManagerTask for proceeding   to CMIS_STATE_DP_DEINIT from CMIS_STATE_MEDIA_SETTINGS_WAIT |
+
+2. SfpStateUpdateTask thread will notify the media settings to OA based on the value of PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS  
+If PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS != MEDIA_SETTINGS_DONE, notify media settings will be invoked and will be set to MEDIA_SETTINGS_NOTIFIED for a port supporting media settings.
```
amnsinghal, line 210: Should we also check for MEDIA_SETTINGS_NOTIFIED also? Could there be a scenario where SfpStateUpdateTask tries to renotify media settings before MEDIA_SETTINGS_DONE?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+	- PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS = MEDIA_SETTINGS_DONE
+4. In the CMIS_STATE_INSERTED state, if 'admin_status' is up and 'host_tx_ready' is true, CmisManagerTask thread will check if
+	- the port supports media settings (will be checked using g_dict and finding valid SI values) and
+	- MEDIA_SETTINGS_SYNC_STATUS != MEDIA_SETTINGS_DONE  
+If all the above conditions are true, CMIS SM transitions to CMIS_STATE_MEDIA_SETTINGS_WAIT state.  
+If port doesn't require media settings to be applied, CMIS SM will proceed with normal code flow (transitions to CMIS_STATE_DP_DEINIT)  
+Overall, no functionality change related to CMIS SM transitions is intended for ports not supporting media settings
+5. CMIS_STATE_MEDIA_SETTINGS_WAIT state will wait for MEDIA_SETTINGS_DONE and upon reaching to MEDIA_SETTINGS_DONE, CMIS SM will transition to CMIS_STATE_DP_DEINIT.  
+There will be a timeout of 5s for every retry
+6. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED" to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
```
amnsinghal, line 223: Shouldn't CMIS_REINIT_REQUIRED be set to True for all states other than CMIS_STATE_READY, to allow CmisManagerTask to reinit the module if xcvrd crashes?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+    participant STATE_DB
+    participant OA
+    participant APPL_DB
+    participant CmisManagerTask
+    participant SfpStateUpdateTask
+
+    SfpStateUpdateTask ->> SfpStateUpdateTask : event = SFP_STATUS_REMOVED
+    SfpStateUpdateTask -x STATE_DB : Delete TRANSCEIVER_INFO table for the port
+    par         CmisManagerTask, SfpStateUpdateTask
+        CmisManagerTask ->> CmisManagerTask : Transition CMIS SM to CMIS_STATE_REMOVED
```
amnsinghal, line 347: Should we set CMIS_REINIT_REQUIRED to True upon module removal?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+## Overview
+
+When syncd/swss/orchagent crashes, all ports in the corresponding namespace will be reinitialized by xcvrd irrespective of the current state of the port.  
+If just xcvrd crashes and restarts, then forced reinitialization (CMIS reinit + media settings notify) of port will not be performed.  
+Following infra will ensure port reinitialization by xcvrd in case of syncd/swss/orchagent crash:
+
+1. XCVRD main thread init
+	- XCVRD main thread creates the key CMIS_REINIT_REQUIRED in PORT_TABLE:\<port\> (APPL_DB) with value as true for ports which do NOT have this key present 
+	- XCVRD main thread creates the key MEDIA_SETTINGS_SYNC_STATUS in PORT_TABLE:\<port\> (APPL_DB) with value MEDIA_SETTINGS_DEFAULT for ports which do NOT have this key present.  
+      - For transceivers which do not support media settings, MEDIA_SETTINGS_SYNC_STATUS will stay with value MEDIA_SETTINGS_DEFAULT
```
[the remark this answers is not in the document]
mihirpat1: I have modified this now.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: SfpStateUpdateTask could try to renotify the media settings before MEDIA_SETTINGS_DONE if xcvrd crashes just after notifying the media settings (in this case, MEDIA_SETTINGS_SYNC_STATUS would be MEDIA_SETTINGS_NOTIFIED).
In such case, OA would apply the media settings but CmisManagerTask will not perform CMIS initialization (along with handling port toggle message from OA) since CmisManagerTask is already dead.
Eventually, once xcvrd spawn again, SfpStateUpdateTask would renotify the media settings to OA followed by CmisManagerTask proceeding with CMIS initialization.
Hence, to allow port toggle message to be sent from OA to CmisManagerTask and allow CMIS initialization in the above xcvrd crash scenario, I am currently not checking for MEDIA_SETTINGS_NOTIFIED. Do you think there could be an issue if we apply the media settings twice?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: I don't think so since CMIS_REINIT_REQUIRED is created mainly to force CMIS re-initialization in case of xcvrd/OA crash. Hence, if a port reaches to any of the steady states above (apart from CMIS_STATE_READY), it means that the port has genuinely failed initialization (CMIS_STATE_FAILED) or it doesn't require forced re-initialization (CMIS_STATE_REMOVED).
Any subsequent triggers for CMIS initialization would still be addressed via https://github.com/sonic-net/sonic-platform-daemons/blob/76baca3d71542dbeeeeac6bea45b762abab7166f/sonic-xcvrd/xcvrd/xcvrd.py#L1557.
This would still ensure that CMIS initialization gets triggered.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: I am not currently changing the value of CMIS_REINIT_REQUIRED after setting it as False since the purpose of this flag is to drive CMIS re-initialization and not CMIS initialization.
The existing flow in the CMIS SM already ensures that CMIS initialization will be performed after module insertion.

Let me know your thoughts on this.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+	- XCVRD main thread creates the key MEDIA_SETTINGS_SYNC_STATUS in PORT_TABLE:\<port\> (APPL_DB) with value MEDIA_SETTINGS_DEFAULT for ports which do NOT have this key present.  
+      - For transceivers which do not require media settings, MEDIA_SETTINGS_SYNC_STATUS will stay with value MEDIA_SETTINGS_DEFAULT
+
+  Following table describes the various values for MEDIA_SETTINGS_SYNC_STATUS  
+
+|          Value          |                Modifier thread and event               |                                  Consumer thread and purpose                                 |
+|:-----------------------:|:------------------------------------------------------:|:--------------------------------------------------------------------------------------------:|
+| MEDIA_SETTINGS_DEFAULT  | XCVRD main thread during cold   start of xcvrd         | XCVRD main thread   during boot-up for deciding to notify media settings                     |
+|                         | SfpStateUpdateTask during   transceiver removal        |                                                                                              |
```
shyam77git, line 205: Please correct indentation (unnecessary gaps between words)

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+1. XCVRD main thread init
+	- XCVRD main thread creates the key CMIS_REINIT_REQUIRED in PORT_TABLE:\<port\> (APPL_DB) with value as true for ports which do NOT have this key present 
+	- XCVRD main thread creates the key MEDIA_SETTINGS_SYNC_STATUS in PORT_TABLE:\<port\> (APPL_DB) with value MEDIA_SETTINGS_DEFAULT for ports which do NOT have this key present.  
+      - For transceivers which do not require media settings, MEDIA_SETTINGS_SYNC_STATUS will stay with value MEDIA_SETTINGS_DEFAULT
+
+  Following table describes the various values for MEDIA_SETTINGS_SYNC_STATUS  
+
+|          Value          |                Modifier thread and event               |                                  Consumer thread and purpose                                 |
+|:-----------------------:|:------------------------------------------------------:|:--------------------------------------------------------------------------------------------:|
+| MEDIA_SETTINGS_DEFAULT  | XCVRD main thread during cold   start of xcvrd         | XCVRD main thread   during boot-up for deciding to notify media settings                     |
```
shyam77git, line 204: In this row, I believe consumer thread captures both uses cases - boot-up and transceiver insertion.
In that case, please add 'transceiver insertion' here too

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
shyam77git: In xcvrd restart scenario, re-applying or re-notifying media settings  might have an impact to the system i.e.
OA is just a conduit. Re-applying would lead to re-notifying media_settings from OA->syncd->SAI->SDK and in turn making port toggle.
Should avoid port toggling here

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+|                         | SfpStateUpdateTask during   transceiver removal        |                                                                                              |
+| MEDIA_SETTINGS_NOTIFIED | SfpStateUpdateTask while   updating the media settings | Not being used currently                                                                     |
+| MEDIA_SETTINGS_DONE     | Orchagent after applying the SI   settings             | CmisManagerTask for proceeding   to CMIS_STATE_DP_DEINIT from CMIS_STATE_MEDIA_SETTINGS_WAIT |
+
+2. SfpStateUpdateTask thread will notify the media settings to OA based on the value of PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS  
+If PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS != MEDIA_SETTINGS_DONE, notify media settings will be invoked and will be set to MEDIA_SETTINGS_NOTIFIED for a port supporting media settings.
+3. The OA upon receiving media settings will
+	- Disable port admin status
+	- Apply SI settings
+	- PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS = MEDIA_SETTINGS_DONE
```
shyam77git, line 214: [1] Please elaborate "Apply SI settings" to something on lines...Ask SAI-SDK to apply SI settings
[2] Add Next bullet - Based on the  update/response as SUCCESS from SAI call,  
set PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS = MEDIA_SETTINGS_DONE
Enable port admin status

In case of failure update/response from SAI call, what would be the system flow? how would the failure be handled?
I beleive host_tx_ready settings is part of this workflow. Please add it here to the workflow

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+4. In the CMIS_STATE_INSERTED state, if 'admin_status' is up and 'host_tx_ready' is true, CmisManagerTask thread will check if
+	- the port supports media settings (will be checked using g_dict and finding valid SI values) and
+	- MEDIA_SETTINGS_SYNC_STATUS != MEDIA_SETTINGS_DONE  
+If all the above conditions are true, CMIS SM transitions to CMIS_STATE_MEDIA_SETTINGS_WAIT state.  
+If port doesn't require media settings to be applied, CMIS SM will proceed with normal code flow (transitions to CMIS_STATE_DP_DEINIT)  
+Overall, no functionality change related to CMIS SM transitions is intended for ports not supporting media settings
```
shyam77git, line 220: "If port doesn't require media settings to be applied,..."
This should be checked as part of previous point/bullet (i.e. the port supports media settings...)
bail out from there itself (i.e. no further action) in case port doesn't require media settings

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+5. CMIS_STATE_MEDIA_SETTINGS_WAIT state will wait for MEDIA_SETTINGS_DONE and upon reaching to MEDIA_SETTINGS_DONE, CMIS SM will transition to CMIS_STATE_DP_DEINIT.  
+There will be a timeout of 5s for every retry
```
shyam77git, line 222: retry attempt would be logged?
How many retries planned? Once retries exhaust, how would the system flow/handling would be w.r.t CMIS FSM for that port?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
-
+# Port reinitialization during syncd/swss/orchagent crash
+## Overview
+
+When syncd/swss/orchagent crashes, all ports in the corresponding namespace will be reinitialized by xcvrd irrespective of the current state of the port.  
+If just xcvrd crashes and restarts, then forced reinitialization (CMIS reinit + media settings notify) of port will not be performed.  
+Following infra will ensure port reinitialization by xcvrd in case of syncd/swss/orchagent crash:
+
+1. XCVRD main thread init
+	- XCVRD main thread creates the key CMIS_REINIT_REQUIRED in PORT_TABLE:\<port\> (APPL_DB) with value as true for ports which do NOT have this key present
```
shyam77git, line 196: Isn't this the PORT_TABLE of STATE_DB?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+        XCVRDMT -x XCVRDMT : Kill XCVRD with SIGABRT
+
+    else
+        XCVRDMT -x XCVRDMT : Graceful exit
+    end
+    deactivate XCVRDMT
+```
+
+## Test plan and expectation
+|       Event      | APPL_DB cleared | Xcvrd restarted | Media renotify | MEDIA_SETTINGS_SYNC_STATUS value on   xcvrd boot-up for initialized transceiver | CMIS re-init triggered | Link flap |
```
shyam77git, line 412: Change Media renotify to Media settings renotify

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+|:----------------:|:---------------:|:---------------:|:--------------:|:-----------------------------------------------------------------------------:|:----------------------:|:---------:|
+| Xcvrd restart    | N               | Y               | N              | MEDIA_SETTINGS_DONE                                                           | N                      | N         |
+| Pmon restart     | N               | Y               | N              | MEDIA_SETTINGS_DONE                                                           | N                      | N         |
+| Swss restart     | Y               | Y               | Y              | MEDIA_SETTINGS_DEFAULT                                                        | Y                      | Y         |
```
shyam77git, line 416: this is not a case of link flap per se.
All critical SW components went down along with HW ones (NPU, data-path power zone etc.) so its equivalent to SW/NOS restart.
We should mark this case as N/A under this link flap column

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+| Syncd restart    | Y               | Y               | Y              | MEDIA_SETTINGS_DEFAULT                                                        | Y                      | Y         |
+| config   reload  | Y               | Y               | Y              | MEDIA_SETTINGS_DEFAULT                                                        | Y                      | Y         |
+| Cold reboot      | Y               | Y               | Y              | MEDIA_SETTINGS_DEFAULT                                                        | Y                      | Y         |
```
shyam77git, line 419: config reload and cold reboot triggers/use-cases doesn't have link flap (until link come operationally up and then go down post bring up).
Since all SW components went down along with HW ones (NPU, PHY, optics etc.), so should mark this case  as N/A under this link flap column

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: I think the confusion here is because PORT_TABLE exists in APPL_DB as well as STATE_DB (although, both are unique). I have now modified the below line to be more precise to APPL_DB
"Check for changes in PORT_TABLE and act upon receiving DEL event" -> "Check for changes in APPL_DB:PORT_TABLE and act upon receiving DEL event"

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+4. In the CMIS_STATE_INSERTED state, if 'admin_status' is up and 'host_tx_ready' is true, CmisManagerTask thread will check if
+	- the port supports media settings (will be checked using g_dict and finding valid SI values) and
+	- MEDIA_SETTINGS_SYNC_STATUS != MEDIA_SETTINGS_DONE  
+If all the above conditions are true, CMIS SM transitions to CMIS_STATE_MEDIA_SETTINGS_WAIT state.  
+If port doesn't require media settings to be applied, CMIS SM will proceed with normal code flow (transitions to CMIS_STATE_DP_DEINIT)  
+Overall, no functionality change related to CMIS SM transitions is intended for ports not supporting media settings
+5. CMIS_STATE_MEDIA_SETTINGS_WAIT state will wait for MEDIA_SETTINGS_DONE and upon reaching to MEDIA_SETTINGS_DONE, CMIS SM will transition to CMIS_STATE_DP_DEINIT.  
+There will be a timeout of 5s for every retry
+6. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED" to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
+7. XCVRD will subscribe to PORT_TABLE in APPL_DB and trigger self-restart if the PORT_TABLE is deleted for the namespace.
```
shyam77git, line 224: In a multi-ASIC system, DB belongs to a particular ASIC/NPU namespace.
Please update _DB references in these workflows to _DB@<ASIC_<n> or _DB@0 , _DB@1 etc

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+
+    CMIS_STATE_DP_DEINIT --> CMIS_STATE_AP_CONF
+    CMIS_STATE_AP_CONF --> CMIS_STATE_DP_INIT
+    CMIS_STATE_DP_INIT --> CMIS_STATE_DP_TXON
+    CMIS_STATE_DP_TXON --> CMIS_STATE_DP_ACTIVATE
+    CMIS_STATE_DP_ACTIVATE --> CMIS_STATE_READY
+```
+
+## Transceiver OIR handling
+
```
eddyk-nvidia, line 335: I guess OIR is On-the-fly Insertion / Removal but better to add it to the abbreviation list in the beginning

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+        After 5s timeout, force_cmis_reinit will be called
+    end note
+
+    CMIS_STATE_DP_DEINIT --> CMIS_STATE_AP_CONF
+    CMIS_STATE_AP_CONF --> CMIS_STATE_DP_INIT
+    CMIS_STATE_DP_INIT --> CMIS_STATE_DP_TXON
+    CMIS_STATE_DP_TXON --> CMIS_STATE_DP_ACTIVATE
+    CMIS_STATE_DP_ACTIVATE --> CMIS_STATE_READY
+```
+
```
eddyk-nvidia, line 333: Probably not related to this specific design but what is the reason for setting CMIS_STATE_READY when either "host_tx_ready" is not yet TRUE or admin_status is not UP ? Why do we skip to the final state of the CMIS state machine in this use-case ?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+        else
+            Note over SfpStateUpdateTask: retry_eeprom_set.add(lport)
+        end
+    end
+    Note over SfpStateUpdateTask: _init_port_sfp_status_tbl<br>Subscribe to CONFIG_DB
+    loop Wait for stop_event
+        SfpStateUpdateTask ->> SfpStateUpdateTask: Handle config change event<br>retry_eeprom_reading()<br>_wrapper_get_transceiver_change_event
+    end
+```
+
```
eddyk-nvidia, line 301: When we set the MEDIA_SETTING_NOTIFY I would also mention on the flow that actual SI params from media_setting.json are pushed to PORT TABLE in APP DB. Someone can think that only Media Setting notification is set in DB

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+When syncd/swss/orchagent crashes, all ports in the corresponding namespace will be reinitialized by xcvrd irrespective of the current state of the port.  
+If just xcvrd crashes and restarts, then forced reinitialization (CMIS reinit + media settings notify) of port will not be performed.  
+Following infra will ensure port reinitialization by xcvrd in case of syncd/swss/orchagent crash:
+
+1. XCVRD main thread init
+	- XCVRD main thread creates the key CMIS_REINIT_REQUIRED in PORT_TABLE:\<port\> (APPL_DB) with value as true for ports which do NOT have this key present 
+	- XCVRD main thread creates the key MEDIA_SETTINGS_SYNC_STATUS in PORT_TABLE:\<port\> (APPL_DB) with value MEDIA_SETTINGS_DEFAULT for ports which do NOT have this key present.  
+      - For transceivers which do not require media settings, MEDIA_SETTINGS_SYNC_STATUS will stay with value MEDIA_SETTINGS_DEFAULT
+
```
eddyk-nvidia, line 199: To know that  the XCVRD main thread might need to parse "media_settings.json" file. Ports which have no match in this file don't require media settings. Is it the intention to do ? I know that with recent changes parsing/setting of media settings moved from init code of the XCVRD main thread to init of SfpStateUpdate task. May be this setting after parsing of .json file should be done there ? Or MEDIA_SETTING_DEFAULT can be put for all ports regardless media settings for them in .json file ? Then if SfpStateUpdate taks finds data for a port in .json it will set MEDIA_SETTING_NOTIFIED ? What is the impact of having MEDIA_SETTING_DEFAULT for all ports ?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+| MEDIA_SETTINGS_DEFAULT  | XCVRD main thread during cold   start of xcvrd         | XCVRD main thread   during boot-up for deciding to notify media settings                     |
+|                         | SfpStateUpdateTask during   transceiver removal        |                                                                                              |
+| MEDIA_SETTINGS_NOTIFIED | SfpStateUpdateTask while   updating the media settings | Not being used currently                                                                     |
+| MEDIA_SETTINGS_DONE     | Orchagent after applying the SI   settings             | CmisManagerTask for proceeding   to CMIS_STATE_DP_DEINIT from CMIS_STATE_MEDIA_SETTINGS_WAIT |
+
+2. SfpStateUpdateTask thread will notify the media settings to OA based on the value of PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS  
+If PORT_TABLE:\<port\>.MEDIA_SETTINGS_SYNC_STATUS != MEDIA_SETTINGS_DONE, notify media settings will be invoked and will be set to MEDIA_SETTINGS_NOTIFIED for a port supporting media settings.
+3. The OA upon receiving media settings will
+	- Disable port admin status
```
eddyk-nvidia, line 212: Toggling done only if Port's Admin status is UP
What about the situation when Admin Status is DOWN when a module is plugged. The the SI data will be sent to SDK/FW. MEDIA_SETTINGS_DONE is set. But actual SI data will be applied to SERDES only after the port's Admin status is set to UP. Do you see any problem here ?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+Overall, no functionality change related to CMIS SM transitions is intended for ports not supporting media settings
+5. CMIS_STATE_MEDIA_SETTINGS_WAIT state will wait for MEDIA_SETTINGS_DONE and upon reaching to MEDIA_SETTINGS_DONE, CMIS SM will transition to CMIS_STATE_DP_DEINIT.  
+There will be a timeout of 5s for every retry
+6. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED" to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
+7. XCVRD will subscribe to PORT_TABLE in APPL_DB and trigger self-restart if the PORT_TABLE is deleted for the namespace.  
+All threads will be gracefully terminated and xcvrd deinit will be performed followed by issuing a SIGABRT to ensure XCVRD is restarted automatically by supervisord. After respawn, CMIS re-init and media_settings notified is triggered for the ports belonging to the affected namespace
+8. syncd/swss/orchagent restart clears the entire APPL-DB, including “MEDIA_SETTINGS_SYNC_STATUS” and "CMIS_REINIT_REQUIRED" in PORT_TABLE
```
Junchao-Mellanox, line 226: how about warm reboot/fast reboot?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +185,246 @@ if transceiver is not present:
...
+      - For transceivers which do not support media settings, MEDIA_SETTINGS_SYNC_STATUS will stay with value MEDIA_SETTINGS_DEFAULT
+
+  Following table describes the various values for MEDIA_SETTINGS_SYNC_STATUS  
+
+|          Value          |                Modifier thread and event               |                                  Consumer thread and purpose                                 |
+|:-----------------------:|:------------------------------------------------------:|:--------------------------------------------------------------------------------------------:|
+| MEDIA_SETTINGS_DEFAULT  | XCVRD main thread during cold   start of xcvrd         | XCVRD main thread   during boot-up for deciding to notify media settings                     |
+|                         | SfpStateUpdateTask during   transceiver removal        |                                                                                              |
+| MEDIA_SETTINGS_NOTIFIED | SfpStateUpdateTask while   updating the media settings | Not being used currently                                                                     |
```
Junchao-Mellanox, line 206: what is this state used for? no consumer?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: When this was initially implemented, the thought at that time was to move the CMIS SM to a fixed state rather than waiting for an event.
However, if you think we should create a new state to handle host_tx_ready and admin_status  rather than moving to CMIS_STATE_READY if either fields have the value false/down respectively, I can plan to implement it accordingly. Let me know your opinion on this.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: Currently, it seems that the SI data is applied to SERDES irrespective of the port's admin status. Can you please correct me if I am wrong.

eddyk-nvidia: it might be vendor-specific behavior. On some platforms SI is applied only when Admin status is set changed to UP. Nvidia is going to add soon the PR for port toggling on applying SI configuration for SERDES's

mihirpat1: Will check with @prgeor on this.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +187,271 @@ if transceiver is not present:
...
+  - NPU_SI_SETTINGS_SYNC_STATUS key is used as a means to communicate the status of applying NPU SI settings for a transceiver requiring NPU SI settings. This key is used to update the NPU SI settings application status between SfpStateUpdateTask, CmisManagerTask and Orchagent. In case of warm reboot or xcvrd restart, this key will prevent application of NPU SI settings on the port if the settings are already applied. In case of transceiver insertion, NPU SI settings will be applied irrespective of the NPU SI settings application status for the port.  
+
+Following infra will ensure port reinitialization by xcvrd in case of syncd/swss/orchagent crash
+
+1. XCVRD main thread init
+	- XCVRD main thread creates the key CMIS_REINIT_REQUIRED in PORT_TABLE:\<port\> (APPL_DB) with value as true for ports which do NOT have this key present 
+	- XCVRD main thread creates the key NPU_SI_SETTINGS_SYNC_STATUS in PORT_TABLE:\<port\> (APPL_DB) with value NPU_SI_SETTINGS_DEFAULT for ports which do NOT have this key present.
```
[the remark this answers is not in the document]
mihirpat1: I have currently mentioned the APPL-DB clearing part as part of the infra of OA crash handling. Do you want me to add a specific pre-requisite section for it in addition to the below?

"7. syncd/swss/orchagent restart clears the entire APPL-DB, including “NPU_SI_SETTINGS_SYNC_STATUS” and "CMIS_REINIT_REQUIRED" in PORT_TABLE"

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +187,271 @@ if transceiver is not present:
  - All the workflows mentioned above will reamin same ( or get exercised) till host_tx_ready field update
  - xcvrd will not perform any action on receiving host_tx_ready field update 
 
+# Port reinitialization during syncd/swss/orchagent crash
+## Overview
+
+When syncd/swss/orchagent crashes, all ports in the corresponding namespace will be reinitialized by xcvrd irrespective of the current state of the port.
```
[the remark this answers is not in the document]
mihirpat1: I have addressed this now.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +187,271 @@ if transceiver is not present:
  - All the workflows mentioned above will reamin same ( or get exercised) till host_tx_ready field update
  - xcvrd will not perform any action on receiving host_tx_ready field update 
 
+# Port reinitialization during syncd/swss/orchagent crash
+## Overview
+
+When syncd/swss/orchagent crashes, all ports in the corresponding namespace will be reinitialized by xcvrd irrespective of the current state of the port.  
+If just xcvrd crashes and restarts, then forced reinitialization (CMIS reinit + NPU SI settings notification) of ports will not be performed.  
+CMIS_REINIT_REQUIRED and NPU_SI_SETTINGS_SYNC_STATUS keys in PORT_TABLE:\<port\> (APPL_DB) are used to determine if port reinitialization is required or not.
+  - CMIS_REINIT_REQUIRED key states if CMIS reinitialization is required for a port after xcvrd is spawned. CMIS_REINIT_REQUIRED helps in mainly driving CMIS reinitialization after syncd/swss/orchagent crash since it will allow reinitializing ports belonging to the relevant namespace of the crashing process. This key is not planned to drive CMIS initialization after transceiver insertion.
```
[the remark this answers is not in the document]
mihirpat1: I have addressed this now.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +187,272 @@ if transceiver is not present:
...
+
+4. CMIS_STATE_NPU_SI_SETTINGS_WAIT state will wait for NPU_SI_SETTINGS_DONE and upon reaching to NPU_SI_SETTINGS_DONE, CMIS SM will transition to CMIS_STATE_DP_INIT state.  
+There will be a timeout of 5s for every retry
+
+5. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED" to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
+
+6. XCVRD will subscribe to PORT_TABLE in APPL_DB and trigger self-restart if the PORT_TABLE is deleted for the namespace.  
+All threads will be gracefully terminated and xcvrd deinit will be performed followed by issuing a SIGABRT to ensure XCVRD is restarted automatically by supervisord. After respawn, CMIS re-init and NPU_SI_SETTINGS notified is triggered for the ports belonging to the affected namespace
+
+7. syncd/swss/orchagent restart clears the entire APPL-DB, including “NPU_SI_SETTINGS_SYNC_STATUS” and "CMIS_REINIT_REQUIRED" in PORT_TABLE
```
shyam77git, line 236: This section is referring to "syncd/swss/orchagent crash" handling scenario.
Here "syncd/swss/orchagent restart" is mentioned. 
- Does 'restart' of these docker containers has the same workflow?
- Please check from present codebase/image working state as well and ensure same/different handling (if any)

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +187,272 @@ if transceiver is not present:
...
+| Syncd restart  | Y                        | Y               | Y                        | NPU_SI_SETTINGS_DEFAULT                                                        | Y                      | N/A       |
+| config reload  | Y                        | Y               | Y                        | NPU_SI_SETTINGS_DEFAULT                                                        | Y                      | N/A       |
+| Cold reboot    | Y                        | Y               | Y                        | NPU_SI_SETTINGS_DEFAULT                                                        | Y                      | N/A       |
+| Config shut    | N                        | N               | N                        | NPU_SI_SETTINGS_DONE                                                           | N                      | N/A       |
+| Config no shut | N                        | N               | N                        | NPU_SI_SETTINGS_DONE                                                           | N                      | N/A       |
```
shyam77git, line 448: Correct 'config shut' and 'config no shut' events to 'config interface shut' and 'config interface startup'

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +187,272 @@ if transceiver is not present:
...
+    else
+        XCVRDMT -x XCVRDMT : Graceful exit
+    end
+    deactivate XCVRDMT
+```
+
+## Test plan and expectation
+| Event          | APPL_DB_<asic_n> cleared | Xcvrd restarted | NPU SI settings renotify | NPU_SI_SETTINGS_SYNC_STATUS value on xcvrd boot-up for initialized transceiver | CMIS re-init triggered | Link flap |
+| -------------- | ------------------------ | --------------- | ------------------------ | ------------------------------------------------------------------------------ | ---------------------- | --------- |
+| Xcvrd restart  | N                        | Y               | N                        | NPU_SI_SETTINGS_DONE                                                           | N                      | N         |
```
shyam77git, line 441: To this table, add transceiver removal and transceiver insertion events depicting how all these columns would be handled

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: Yes, 'restart' of these docker containers has the same workflow.
I have now verified from that restarting swss indeed clears the APPL_DB. I have also added "restart swss" as part of my test plan to cover this scenario.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: I have addressed this now.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: I have addressed this now.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: Confirmed that this approach should not cause problem.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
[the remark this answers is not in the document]
mihirpat1: I have added the pre-requisite section now.

jaganbal-a (approved):

Junchao-Mellanox (changes requested):
This solution looks compliated. How about following:

1. If xcvrd crashes and restarts, it always notify media settings. And orchagent should keep a cache of media settings, only apply media settings when configuration changes
2. If a cable is plug out and plug in, xcvrd should remove media settings from APP DB and re-apply it.
3. if orchagent crashes, it should re-read from redis automatically, so, no issue.

Please share your thought. Thanks.

mihirpat1:
> This solution looks compliated. How about following:
> 
> 1. If xcvrd crashes and restarts, it always notify media settings. And orchagent should keep a cache of media settings, only apply media settings when configuration changes
> 2. If a cable is plug out and plug in, xcvrd should remove media settings from APP DB and re-apply it.
> 3. if orchagent crashes, it should re-read from redis automatically, so, no issue.
> 
> Please share your thought. Thanks.

Hi Junchao,

With the above suggestion, we will not be able to hold CMIS initialization (for CMIS based modules) while the NPU SI settings are being applied by the orchagent.
Currently, the orchagent disables the port before applying NPU SI settings followed by enabling the port again after application.
Hence, we can have a case wherein CMIS initialization has completed for a port and the link is up. After sometime, the orchagent applies NPU SI settings. The port now goes through the disable->enable sequence by orchagent which causes the link flap.

To solve this issue (link flap), the current HLD has a separate code path for CMIS modules to notify orchagent with NPU SI settings.

prgeor (commented):
@mihirpat1 can you add revision 1.0 to the Revision table at the top of this document?

prgeor:
> Proposal to reinitialize porta through xcvrd during OA crash. The current proposal is created to resolve [sonic-net/sonic-platform-daemons#356](https://github.com/sonic-net/sonic-platform-daemons/issues/356)
> 
> The PR details will be updated soon.

@mihirpat1 please update the PR details (including OA PR as well)

mihirpat1:
> > This solution looks compliated. How about following:
> > 
> > 1. If xcvrd crashes and restarts, it always notify media settings. And orchagent should keep a cache of media settings, only apply media settings when configuration changes
> > 2. If a cable is plug out and plug in, xcvrd should remove media settings from APP DB and re-apply it.
> > 3. if orchagent crashes, it should re-read from redis automatically, so, no issue.
> > 
> > Please share your thought. Thanks.
> 
> Hi Junchao,
> 
> With the above suggestion, we will not be able to hold CMIS initialization (for CMIS based modules) while the NPU SI settings are being applied by the orchagent. Currently, the orchagent disables the port before applying NPU SI settings followed by enabling the port again after application. Hence, we can have a case wherein CMIS initialization has completed for a port and the link is up. After sometime, the orchagent applies NPU SI settings. The port now goes through the disable->enable sequence by orchagent which causes the link flap.
> 
> To solve this issue (link flap), the current HLD has a separate code path for CMIS modules to notify orchagent with NPU SI settings.

In addition to the above, XCVRD can miss the port toggle message being sent from orchagent to xcvrd if the CmisTaskManager thread is busy handling other tasks.
The port toggle message is sent from orchagent to xcvrd to disable port before NPU settings are being applied and then enable port after the settings have been applied.
The current design of redis-db ensures only the final state of the DB being handled when the worker thread starts handling requests. So if CmisTaskManager thread is busy handling other events while port toggle message is sent out, both the port disable and enable events could be missed.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +187,290 @@ if transceiver is not present:
...
+CMIS_REINIT_REQUIRED and NPU_SI_SETTINGS_SYNC_STATUS keys in PORT_TABLE:\<port\> (APPL_DB) are used to determine if port re-initialization is required or not.
+  - CMIS_REINIT_REQUIRED key states if CMIS re-initialization is required for a port after xcvrd is spawned. CMIS_REINIT_REQUIRED helps in mainly driving CMIS re-initialization after syncd/swss/orchagent crash since it will allow reinitializing ports belonging to the relevant namespace of the crashing process. This key is not planned to drive CMIS initialization after transceiver insertion.  
+  - NPU_SI_SETTINGS_SYNC_STATUS key is used as a means to communicate the status of applying NPU SI settings for a transceiver requiring NPU SI settings. This key is used to update the NPU SI settings application status between SfpStateUpdateTask, CmisManagerTask and Orchagent. In case of warm reboot or xcvrd restart, this key will prevent application of NPU SI settings on the port if the settings are already applied. In case of transceiver insertion, NPU SI settings will be applied irrespective of the NPU SI settings application status for the port.
```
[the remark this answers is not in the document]
mihirpat1: I have added the description for it now.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +187,290 @@ if transceiver is not present:
...
+	- Request SAI-SDK to apply the NPU SI settings via syncd
+		- If SAI-SDK returns success, OA will update the PORT_TABLE:\<port\>.NPU_SI_SETTINGS_SYNC_STATUS to NPU_SI_SETTINGS_DONE
+		- In case of failure, OA will log an error message and proceed to handling the next port
+
+4. CMIS_STATE_NPU_SI_SETTINGS_WAIT state will wait for NPU_SI_SETTINGS_DONE and upon reaching to NPU_SI_SETTINGS_DONE, CMIS SM will transition to CMIS_STATE_DP_INIT state.  
+There will be a timeout of 5s for every retry
+
+5. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED" to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
+
+6. XCVRD will subscribe to PORT_TABLE in APPL_DB and trigger self-restart if the PORT_TABLE is deleted for the namespace.
```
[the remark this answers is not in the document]
mihirpat1: I have covered the same in "Test plan and expectation" section.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +187,290 @@ if transceiver is not present:
...
+| ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------- |
+| NPU_SI_SETTINGS_DEFAULT  | 1\. XCVRD main thread during cold start of XCVRD<br>2. SfpStateUpdateTask during transceiver removal                                                                                                                | XCVRD main thread during boot-up for deciding to notify NPU SI settings                         |
+| NPU_SI_SETTINGS_NOTIFIED | 1\. SfpStateUpdateTask while updating and notifying the NPU SI settings for non-CMIS SM driven transceivers<br> 2. CmisManagerTask while updating and notifying the NPU SI settings for CMIS SM driven transceivers | Not being used currently                                                                        |
```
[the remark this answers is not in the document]
mihirpat1: I have added the description for it now.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +187,290 @@ if transceiver is not present:
  - All the workflows mentioned above will reamin same ( or get exercised) till host_tx_ready field update
  - xcvrd will not perform any action on receiving host_tx_ready field update 
 
-
+# Port re-initialization during syncd/swss/orchagent crash
+## Overview
+
+When syncd/swss/orchagent crashes, all ports in the corresponding namespace will be reinitialized by xcvrd irrespective of the current state of the port. All the corresponding ports are expected to experience link down until the initialization is complete.  
+If just xcvrd crashes and restarts, then forced re-initialization (CMIS reinit + NPU SI settings notification) of ports will not be performed. Hence, the ports will not experience link downtime during scenario.
+CMIS_REINIT_REQUIRED and NPU_SI_SETTINGS_SYNC_STATUS keys in PORT_TABLE:\<port\> (APPL_DB) are used to determine if port re-initialization is required or not.
```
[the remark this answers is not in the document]
mihirpat1: I have added the description for it now.

mihirpat1:
> @mihirpat1 can you add revision 1.0 to the Revision table at the top of this document?

I have added it now.

mihirpat1:
> > Proposal to reinitialize porta through xcvrd during OA crash. The current proposal is created to resolve [sonic-net/sonic-platform-daemons#356](https://github.com/sonic-net/sonic-platform-daemons/issues/356)
> > The PR details will be updated soon.
> 
> @mihirpat1 please update the PR details (including OA PR as well)

Added OA PR for now. Will add other PRs once I upload.

mihirpat1:
@eddyk-nvidia - It will be great if you can help in reviewing this

aravindmani-1 (approved):

Junchao-Mellanox (approved):
LGTM

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +188,294 @@ if transceiver is not present:
...
+  Also, this key helps in preventing additional link flap which can happen if CMIS state machine initializes a port before NPU SI settings are applied (since application of NPU SI settings involves disabling the port followed by enabling it).  
+In case of continuous restart of xcvrd, both the keys will still hold the same value as before the restart. This would ensure that the port re-initialization is resumed from the last known state.  
+
+Rationale for choosing APPL_DB over STATE_DB for storing CMIS_REINIT_REQUIRED and NPU_SI_SETTINGS_SYNC_STATUS keys
+ - APPL_DB is the first DB which gets cleared as part of SWSS crash handling. This inturn helps XCVRD to detect DB removal and terminate child threads and itself earlier than other DBs.
```
prgeor, line 203: @mihirpat1 should we reconsider APPL_DB because, i do see oper status in APPL_DB as well
image

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -184,18 +188,294 @@ if transceiver is not present:
...
+| NPU_SI_SETTINGS_DEFAULT  | 1\. XCVRD main thread during cold start of XCVRD<br>2. SfpStateUpdateTask during transceiver removal                                                                                                                | XCVRD main thread during boot-up for deciding to notify NPU SI settings                         |
+| NPU_SI_SETTINGS_NOTIFIED | 1\. SfpStateUpdateTask while updating and notifying the NPU SI settings for non-CMIS SM driven transceivers (this approach was chosen to preserve the existing behavior for non-CMIS SM driven transceivers)<br> 2. CmisManagerTask while updating and notifying the NPU SI settings for CMIS SM driven transceivers | Not being used currently                                                                        |
+| NPU_SI_SETTINGS_DONE     | Orchagent after applying the SI settings                                                                                                                                                                            | CmisManagerTask for proceeding from CMIS_STATE_NPU_SI_SETTINGS_WAIT to CMIS_STATE_DP_INIT state |
```
prgeor, line 226: @mihirpat1 does platform that does not need SI settings, do those port's CMIS state machine transition to DP init?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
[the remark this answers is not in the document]
prgeor: @mihirpat1 can you also describe the rationale behind restoring Xcvrd in case swss/syncd/OA restart/crash?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: Yes - For platforms or ports not requiring NPU SI settings, the CMIS state machine will transition to DP init state (i.e. skip transition to CMIS_STATE_NPU_SI_SETTINGS_WAIT)

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: @prgeor - I have now updated the HLD with the usage of STATE_DB for storing the `NPU_SI_SETTINGS_SYNC_STATUS` and `CMIS_REINIT_REQUIRED` keys.

prgeor (approved):

PR #1445, opened 2023-08-16
Add the high level design doc for SmartSwitch high availability feature.

r12f:
High level design doc for SmartSwitch high availability feature.

It includes both doc and an overview slides for a quick glance of everything.

Recording: https://zoom.us/rec/play/m5JG2kuCziV9F0xuqzyiFWmRWGNvE_U4YAxCqQS_QbW8KWxup36aJmlbAn0dKagfOOuhXeJDwKqhwgYW.tm1pHd870CjPk97i?canPlayFromShare=true&from=share_recording_detail&continueMode=true&componentName=rec-play&originRequestUrl=https%3A%2F%2Fzoom.us%2Frec%2Fshare%2Fqldb-vtmXWV4fHsvtceyr_rvy5eFhDqNfPkUF9a9vDi_YPzmnPgzIlpk4kuUXJJn.RTitqeo260XpX5IS


doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2175 @@
...
+3. More than 2 DPUs participate in a single HA set.
+
+### 3.3. Assumptions
+
+When designing HA for SmartSwitch, we have a few assumptions:
+
+1. Stable decision on packet transformation: The same packet will always implement the same packet transformation as long as the DASH policies stay the same. Changes, such as random port selection, are not allowed in the current pipeline.
+2. No delay sync support: When creating a flow, we will replicate this flow to the standby DPU immediately with inline. The flow will not be truly created before receiving the ack from the standby DPU.
+3. No warm boot support: Warm boot is used to keep the data path continuing to work while upgrade happens. It is mostly used for upgrading switches that being considered as SPOF, as a gentler solution than a hard data plane down. However, it requires complicated coordination between many services on NPU and DPU. Gladly, since HA requires more than 1 switch to participate, we don’t need to worry about SPOF problem, hence we don’t need to consider warm boot support in our design.
```
chrispsommers, line 249: Suggest adding SPOF (Single Point of Failure) to terminology section

r12f: thanks Chris! it is updated now.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2176 @@
...
+### 3.1. Goals
+
+The goal of the SmartSwitch HA design is trying to achieve the following goals:
+
+1. 0 downtime on planned switchover.
+2. <2 sec downtime on failover to standalone setup for each ENI.
+3. Ability to resume connections in the event of both planned and unplanned failover.
+4. After both planned and unplanned failover and recovery, the flow on all DPUs will be aligned.
+5. Only focus on the case, in which only 2 DPUs are in the HA set, although conceptually, we can have more than 2, but it is not in our current scope. 
+6. No complicated hashing logic in switches to ensure packets land on right switch.
```
chrispsommers, line 230: This is a bit ambiguous to me. Do you mean, "No complex hash-based ECMP in T2 to steer packets to correct T1?" or "No complex logic in T1 NPU to steer packet to correct DPU?" or something else?

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2176 @@
...
+    * This can happen in cases, such as switchover or when DPUs lose connection to each other and start to act as standalone. If policies in these DPUs don’t match, flow conflict can happen when later the connection recovers.
+    * To resolve the flow conflict, HA will do best effort, but it won’t be able to solve the problem alone. In cases like planned switchovers, our upstream service must coordinate and make sure policies are aligned.
+3. More than 2 DPUs participate in a single HA set.
+
+### 3.3. Assumptions
+
+When designing HA for SmartSwitch, we have a few assumptions:
+
+1. Stable decision on packet transformation: The same packet will always implement the same packet transformation as long as the DASH policies stay the same. Changes, such as random port selection, are not allowed in the current pipeline.
+2. No delay sync support: When creating a flow, we will replicate this flow to the standby DPU immediately with inline. The flow will not be truly created before receiving the ack from the standby DPU.
```
chrispsommers, line 249: "Delay sync" sounds like an internal buzzword with lots of implications. Can you describe? Also, "inline" seems to have lots of connotations, can this be described? It seems to mean that a flow won't be added unless the HA standby unit has acknowledged the new flow to the active. What if the standby is dead? Can't new flows still be added by the active?

 If it's detailed further one, perhaps a link to the relevant section can be provided, e.g. "described [here](some-heading)"

Note, I'm deliberately flagging terms which appear to have loaded meanings amongst "insiders" but not necessarily to others. I hope that is OK.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2176 @@
...
+2. Ensures having enough redundant links for the 2 DPUs in the HA set to talk to each other.
+
+And this can be addressed by the ECMP setup in our network, as the "[Network topology for Flow HA](#network-topology-for-flow-ha)" section shows above.
+
+
+#### 4.2.1. Upstream-to-Switch communication channel
+
+In HA setup, our upstream service will talk to our northbound interface for:
+
+- Programming SDN dash policies.
```
chrispsommers, line 294: s/dash/DASH/

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2176 @@
...
+
+Now let’s put all tunnels together. This can happen when the packets land on other switches that help traffic forwarding in the middle of HA state transition.
+ 
+<p align="center"><img alt="Tunneling packet between DPUs" src="./images/npu-to-dpu-tunnel-2-stages.svg"></p>
+
+#### 4.2.3. DPU-to-DPU communication channel
+
+There are 3 types of DPU-to-DPU communication channel: 
+
+- **Control Plane Control Channel**: This channel is used for transferring lightweighted data, such as control commands like HA state machine. This channel can be implemented by gRPC.
```
chrispsommers, line 337: s/lightweighted/lightweight/

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2176 @@
...
+
+Having our own control channel is important. If we consider the switches as data plane and our upstream services as control plane, then data plane should be able to operate on its own, when control plane is disconnected. For example, when control plane is down and network is having trouble, we should still be able to react and adjust our services. Coupling data plane with control plane will introduce more failure modes also making the data plane unable to react properly.
+
+The DPU state machine management will run on NPU. Since each NPU manages multiple DPUs, all NPUs need to connect to all other NPUs, which forms a full mesh.
+
+The data path of this channel will look like below:
+<p align="center"><img alt="HA control plane control channel data path" src="./images/ha-control-plane-control-channel-data-path.svg"></p>
+
+Since DPU events are still coming from the DPU side via SAI/syncd, we will need to pass these events to the NPU side via ZMQ, which go through the PCIe bus.
```
chrispsommers, line 356: This is unclear to me. There are two levels of SONiC stacks, one on each DPU and one on NPU. So I don't know which SAI/syncd you're referring to here. I also don't know which "PCIe bus" you're describing, since presumably there could be PCIe between the CPU and NPU, and PCIe between CPU and each DPU.

Perhaps the spec needs a SW stack block diagram showing both of these in some appropriate level of detail. I'd like to understand the event flow you just described here including ZMQ role.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2176 @@
...
+
+##### 4.2.2.4. Put all tunnels together
+
+Now let’s put all tunnels together. This can happen when the packets land on other switches that help traffic forwarding in the middle of HA state transition.
+ 
+<p align="center"><img alt="Tunneling packet between DPUs" src="./images/npu-to-dpu-tunnel-2-stages.svg"></p>
+
+#### 4.2.3. DPU-to-DPU communication channel
+
+There are 3 types of DPU-to-DPU communication channel:
```
chrispsommers, line 335: it's my understanding that some of these channels will have standardized formats/protocols and some may have vendor-specific protocols (e.g. inter-DPU flow sync). Can you add commentary about this?

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2176 @@
...
+
+###### 4.2.3.1.3. HA control plane channel data path HA
+
+Although it might be counter-intuitive, we could consider that we can always get a working connection within a bounded time. This is achieved by leveraging our ECMP setup on T1s.
+
+The approach is simply brute force: as long as we have tried enough number of connections with different source port, we can get the channel recovered.
+
+This approach works because with our current setup, it is extremely unlikely that 2 switches cannot communicate with each other. Say, if the possibility of 1 link failing is P, then the possibility of 2 T2s failing talk to each other will be (2p – p^2)^20 to (2p – p^2)^40.
+
+To reduce the time of finding a good path and meet our SLA requirement, we could also try spray the network with a specific number and choose the first one that works.
```
chrispsommers, line 374: By "spray the network," do you mean the CPU uses some algorithm to systematically try a range of ports (i.e. change the next hop egress port) one at a time, until a working port is found? The term "spray" can also refer to one possible mode of a switching ASIC's ECMP logic, where the CPU assigns a port group and the ASIC randomly chooses an egress port for each packet.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: oops. it is updated now.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: typo removed after designing the control plane data channel. :D

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2176 @@
...
+
+#### 4.2.2. NPU-to-DPU traffic forward tunnel
+
+This channel is used for tunneling the packet from all T1s that has the traffic forwarding rules to the active side in the HA pair, no matter if it is locally or remotely.
+
+There is a limitation on the traffic forwarding rules, it can only forward the traffic to either local DPU or remote DPU, but never both as an ECMP group. This is caused by the limitation of routes. Therefore, the tunnel will always only have 1 next hop at maximum.
+
+It works as below in high level:
+
+1. The tunneled packets ***MUST*** have an VxLan encap as the outer packet, with a specified NPU tunnel VNI number as identifier.
```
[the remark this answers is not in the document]
r12f: From DASH perspective, I believe so. Limiting the tunnel to VxLan has 2 purposes - to simplify the system since we only need to handle 1 single encap, as well as achieving better data path availability because we can manipulate the source port in VxLan encap.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
chrispsommers: Sure, that's better, thanks,

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: this is a very good question. for the 3 channels, the detailed packet format of HA control channel and DPU data plane channel are shown in the graph, which contains the information. and I have updated the HA data channel.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: it is updated and pushed now. : D

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: yep, this is a great comment. I have it updated with a bit more explanation and added references too.

The origin of this "delay sync" approach is coming from the [public Sirius paper chapter 3.2](https://www.usenix.org/system/files/nsdi23-bansal.pdf), which could lead people to think it as a valid scenario too, so I added this in this section to make our goal clear.

image

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: Yep! I have updated the HA control plane overview section (see section 5.2 in the latest doc), which have this comment addressed. ZMQ is just a underlying communication channel we might or might not be using, which is not too important, so I have it removed.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: > The term "spray" can also refer to one possible mode of a switching ASIC's ECMP logic, where the CPU assigns a port group and the ASIC randomly chooses an egress port for each packet.

ouch! I mean the former one. do you have any recommended term to use here?

(btw, I have updated this section as part of the HA control plane data channel redesign, it will be section 5.2.2.2. I hope things will be more clear.)

chrispsommers: I don't know of a better unambiguous term, but I think using "rotation through available ports using a suitable algorithm in the controller" is clear enough.

r12f: I have updated it to this: Additionally, to reduce the time of finding a good path and meet our SLA requirement, we could also try to use multiple different source port to establish the new connection in parallel, and choose whichever one that works first.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2208 @@
...
+
+Here is an example from "[Flow replication data path overview – Data plane sync data path](#1221-data-plane-sync-channel-data-path)":
+
+<p align="center"><img alt="Flow replication data plane sync data path" src="./images/flow-replication-data-path.svg"></p>
+
+###### 4.2.3.0.2. Multi-path data plane channel
+
+Since data plane channel is inline, it is very important to make sure whenever gray network failure happens (traffic being blackholed partially), we should avoid causing the traffic for the entire ENI to be dropped.
+
+To provide HA for DPU-to-DPU communication, we can also leverage our ECMP network setup. When sending the tunneled packet, we must calculate the source port from the 5 tuples of the inner packet and fold it into a port range specified by SONiC. This helps us to spread the packets across all possible paths and avoid the blackhole problem.
```
chrispsommers, line 358: >we must calculate the source port from the 5 tuples of the inner packet and fold it into a port range specified by SONiC

Do you mean "The source port is one of the 5 members of the inner packet's 5-tuple and as such it can be chosen to ensure the egress ports are evenly used?" I'm also considering the next section's language:
>The algorithm for port selection doesn’t matter. As long as we can spread the packet across all ports, we should be good.

My concern is, if typical ECMP hash-based techniques are used (hash on a 5-tuple, modulo a list of egress ports), it's not straightforward to predict the resultant egress port for a given 5-tuple (it can vary across different ASIC implementations), and it is even less straightforward to ensure all ports are evenly chosen for a set of flows. (ECMP Polarization is one malignant side-effect of this.) So I'm not sure how one would "choose an algorithm" and derive a list of source ports to plug into these packets to effect balanced use of egress ports.

r12f: oops, the second one is missed when updating the requirements. will get it updated too.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2208 @@
...
+3. The flow replication packet will be sent out with the selected source port in outermost packet, as well as in the metadata. And on the other side, the ack packet can use the same source port too. This binds the packets for this flow to a single data path.
+4. If one of the links went down, all the packets with certain source ports will start to fail. Hence, if we could track the failure for each port in the pool, we can avoid using it.
+5. Alternatively, the DPU-to-DPU probe can use the same ports in the pool, in order to reflect the data path problem. This also helps us to see when the data path is recovered, and we can resume using this port again.
+
+#### 4.2.4. VM to DPU data plane
+
+Again, it is very important to avoid completely killing the VM’s connectivity if one link is blackholing traffic. Hence, when VM sends traffic to DPU, we should also make it leverage ECMP setup.
+
+When VM sends packet out, the packet will be encap’ed by VxLan (see below). Hence, similar to DPU-to-DPU data plane channel, we can also set the VxLan source port by calculating the hash of the inner packet. Then, different flow will go through different data path to reach the SmartSwitch.
```
chrispsommers, line 378: See above comment about source ports affecting HASH-based ECMP egress port selection.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2208 @@
...
+    1. Driving the HA state machine transitions.
+    2. Report every ENI HA state change and reasons, so upstream service knows what is happening and can make decisions for planned events.
+    3. Handle HA related requests from upstream service.
+    4. Monitor and handle unplanned events, and trigger defined mitigations, such as driving to standalone setup.
+
+> It is very important to enable SmartSwitch to drive the HA state machine transition on its own. Under unplanned events such as network failures, we could lose connectivity to upstream service. If we could not drive the state machines or make decisions, we would lose the ability to handle the failures.
+
+### 5.2. HA control plane overview
+
+Here is the overview of HA control plane, and we will walk through the components and communication channels in the following sections.
```
chrispsommers, line 406: Please define "FP Ports" (probably "front-panel ports) and BP Ports ("Back panel?" "Backplane?") My understanding is these are internally wired ports between DPU and NPU using direct electrical connections, not cabled or anything).

r12f: yep, they are.

chrispsommers, line 406: I still see the redis/zmq channel going over PCIe bus. My understanding was all channels are carried as packets over the traffic lanes.

r12f: will get the terminology updated.

r12f: the redis/zmq communication channels will indeed go through the PCIe bus instead of BP ports. the ethernet connection is provided by ethernet-over-pcie.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2208 @@
...
+
+<p align="center"><img alt="HA control plane control channel data path" src="./images/ha-control-plane-control-channel-data-path.svg"></p>
+
+Since DPU events are still coming from the DPU side SAI/syncd, we will need to pass these events to the NPU side via our internal service communication channel, which goes through the PCIe bus (see [HA control plane overview](#52-ha-control-plane-overview) for more details).
+
+##### 5.2.2.2. HA control plane data channel
+
+The data channel is used for transferring large trunks of data between DPUs, e.g. [flow bulk sync](#125-bulk-sync). Since control messages and data sync are going through different channels, this helps us avoid head-of-queue blocking for control messages, which ensures the HA control plane and state machine transition is always responsive.
+
+Because we expect large trunks of data being transferred, this channel is designed to:
```
chrispsommers, line 465: s/trunks/chunks/ ?

doc/smart-switch/high-availability/smart-switch-ha-hld.md
```diff
@@ -0,0 +1,2208 @@
+# SmartSwitch High Availability High Level Design
+
+| Rev | Date | Author | Change Description |
+| --- | ---- | ------ | ------------------ |
+| 0.1 | 08/09/2023 | Riff Jiang | Initial version |
+| 0.2 | 08/10/2023 | Riff Jiang | Simpify ENI-level traffic control, primary election algorithm |
+| 0.3 | 08/14/2023 | Riff Jiang | Adding DPU level standalone support |
+| 0.4 | 08/17/2023 | Riff Jiang | Redesign HA control plane data channel |
+
+1. [1. Background](#1-background)
```
chrispsommers, line 10: The ToC is a little odd, it has numbered values preceding the heading numbers. Have you tried using the "Markdown All-in-one" extension for VScode? It doesn't assign the heading numbers but it does format the ToC nicely.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: for the ECMP polarization issue, you are right. Hash-based ECMP will never be a perfect or even great solution for elephant flows or determine the data path of each flow, but it should be a good starting point for us. This is because the number of flows we are handling is large enough and we have enough number of links between each T1, so as long as they can be spread, some skews will be acceptable. And it is hard to determine the data path of each flows anyway (vendor implementation varies and link states changes overtime as well), so we rarely use the 5-tuple to query the next hop. :D

r12f: The doc is updated now.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: Have updated the doc now to make the algorithm align.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: yep, i am using VSCode with Markdown All-in-one. And the heading numbers are automatically added and updated by it.

image

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: lol, right! XD

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: done.

doc/smart-switch/high-availability/smart-switch-ha-hld.md
r12f: please let me know, if still have any concerns. :D

chrispsommers: >This is because the number of flows we are handling is large enough and we have enough number of links between each T1, so as long as they can be spread, some skews will be acceptable.

I'm still unclear how you'll achieve effective egress port spreading, with any confidence, based on some list of source ports. It sounds like "spray and pray" (hah hah I just thought of that).

r12f: XD, kinda yes and no at the same time. From DPU perspective, it is indeed "spray and pray". However:

- For network failures, the network itself is heavily monitored. So, when any issue happens, it can isolate itself and shift the traffic.
- For load spreading, the number links between T1s are much higher than the number of DPUs in the switch, so although plain hash-based ECMP might not sound beautiful, but the amount of skews are acceptable.

chrispsommers: OK, thanks for explaining!

r12f: thank-you for pointing it out!

zhangyanzhao:
The current design is tight coupled with the upstream service, any way to isolate SmartSwitch from the upstream service and make SmartSwitch autonomous, so that people who does not have upstream service can still use it?

zhangyanzhao:
The current design is tight coupled with the upstream service, any way to isolate SmartSwitch from the upstream service and make SmartSwitch autonomous

zhangyanzhao:
Please leave your github id if you want to be reviewer of this HLD. Thanks.

r12f:
> The current design is tight coupled with the upstream service, any way to isolate SmartSwitch from the upstream service and make SmartSwitch autonomous, so that people who does not have upstream service can still use it?

Discussed in today's session. 

The ENIs will be created by "some service" and will not come out of nowhere, so upstream service will be in the picture no matter how the design looks like. And to limit the impact that upstream service goes down, Current HA is designed to follow the 2 rules:

1. Upstream service only defines the desired state of each ENI on SmartSwitch as well as the actions that we should do when unplanned events happens, e.g., which side of DPU should become standalone.
2. SmartSwitch stores the desired state and actions, then handles the state machine transitions on its own.

With this design, upstream service is decoupled from the HA state machine transition and unplanned event handling, so it will not cause HA stop functioning.

More details can be found in "Working with upstream service" section.

lguohan (approved):

PR #1451, opened 2023-08-28
Create Secureboot_uefi_key_management.md

sacnaik:
Secure boot UEFI key management HLD

### DIFF
--- a/doc/secure_boot/Secureboot_uefi_key_management.md
+++ b/doc/secure_boot/Secureboot_uefi_key_management.md
@@ -74,31 +74,6 @@ There are multiple keys involved in the UEFI secure boot implementations. The pl
 ### 3.3 Authenticated variable
 These are a special type of UEFI variable that supports cryptographic authentication and integrity verification. The authenticates variable enables a secure way of storing and accessing UEFI variables within the UEFI environment and ensures that the UEFI variable remains tamper-proof and authentic.  Since authenticated variables are signed objects, the UEFI firmware can verify them before accessing or updating them.
 
-```
-Sample example of how authenticated variables support a secure way of adding, updating, removing, and revoking UEFI keys.
-
-uuidgen --random > GUID.txt
-
-**PK authenticated variable:**
-openssl req -newkey rsa:4096 -nodes -keyout PK.key -new -x509 -sha256 -days 3650 -subj "/CN=Platform key/" -out PK.crt
-openssl x509 -outform DER -in PK.crt -out PK.cer
-cert-to-efi-sig-list -g "$(< GUID.txt)" PK.crt PK.esl
-sign-efi-sig-list -g "$(< GUID.txt)" -k PK.key -c PK.crt PK PK.esl PK.auth
-
-**KEK authenticated variable:**
-openssl req -newkey rsa:4096 -nodes -keyout KEK.key -new -x509 -sha256 -days 3650 -subj "/CN=Key Exchange Key/" -out KEK.crt
-openssl x509 -outform DER -in KEK.crt -out KEK.cer
-cert-to-efi-sig-list -g "$(< GUID.txt)" KEK.crt KEK.esl
-sign-efi-sig-list -g "$(< GUID.txt)" -k PK.key -c PK.crt KEK KEK.esl KEK.auth
-
-**Db authenticated variable:**
-openssl req -newkey rsa:4096 -nodes -keyout db.key -new -x509 -sha256 -days 3650 -subj "/CN=Database key/" -out db.crt
-openssl x509 -outform DER -in db.crt -out db.cer
-cert-to-efi-sig-list -g "$(< GUID.txt)" db.crt db.esl
-sign-efi-sig-list -g "$(< GUID.txt)" -k KEK.key -c KEK.crt db db.esl db.auth
-```
-Using the above example one should be able to create an authenticated variable to add, remove, update, and revoke UEFI keys from the UEFI database.
-
 ## 4 UEFI key management
 
 The key management involves following the functional area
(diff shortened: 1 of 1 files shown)


zhangyanzhao:
community review recording https://zoom.us/rec/share/2iI8PrtCImHig7wTVU0JLRj2TmnqlzWugD7yJuCpN866CbS2CKLS6jAR94vMOMKs.YmbqVw1-zSzQI03F

doc/secure_boot/Secureboot_uefi_key_management.md
```diff
@@ -0,0 +1,217 @@
...
+This is to display the PK, KEK, DB, and DBx key lists.
+
+## 5 SONIC implementation
+The access of UEFI variables may vary from platform to platform depending on the specific implementation of UEFI firmware. Sonic will provide a common layer for accessing these variables and vendors can implement platform-specific parts. The SONIC will provide a plugin Python for vendors to implement these APIs and integrate them with the SONiC generic layer.
+
+### 5.1 Plugin class
+
+![image](https://github.com/sacnaik/SONiC/assets/25231205/a42cd039-0675-4a71-84e0-bce00eb2ca9d)
+
+### 5.2 OVMF based platform can use Linux implementation to access UEFI variable
```
lguohan, line 179: can you clarify the cisco code contribution here?

doc/secure_boot/Secureboot_uefi_key_management.md
```diff
@@ -0,0 +1,217 @@
...
+2. show platform security uefi variables pk
+3. show platform security uefi variables kek
+4. show platform security uefi variables db
+5. show platform security uefi variables dbx
+
+1. platform security uefi variables update pk  <file>
+2. platform security uefi variables update kek <file>   
+3. platform security uefi variables update db  <file>
+4. platform security uefi variables update dbx <file>  
+```
```
mheese, line 211: a couple of comments here:

- can we call the subcommand `secureboot` instead of `uefi`? This is more fitting as these commands are all specifically dealing with secureboot variables, and not other UEFI variables
- You outlined really well all the steps which are necessary in order to create an authenticated variable file. Honestly, this is the hard part, while running a command to update the variable will on most platforms simply translate to an `efi-updatevar` command or by direct access through `efivarfs`. It would be nice to see an easy UX here
- I'm missing access to the general `SecureBoot` variable which can get information if UEFI Secure Boot is enabled or not on the platform

mheese (commented):
Great to see some progress in SONiC around UEFI Secure Boot features!

Please see my inline comments, but I also have some general comments:

- enrolling and managing custom keys for Secure Boot is probably the pinnacle of security in this area, and the hardest to achieve for average users, yet these commands are really only helpful for that specific use-case
- using custom keys particularly in `db` can have dire consequences particularly if `SecureBoot=1`
- the system should also provide access to the `SecureBoot` variable
- if `SecureBoot=1`, there should be protection of the boot chain: if `db` or `dbx` get updated, the system should ensure that the current boot entries are actually still going to work. As you mentioned updating these databases requires a reboot, but the `shim` of a normal SONiC installation will need to be signed correctly as well for this to still work. So either there should be a check of at least the SONiC OS boot entry against the `db`/`dbx` before updating it, or maybe the system should refuse to reboot otherwise (?) --> this requires some more discussion IMHO
- the same goes for SONiC updates, they should fail if an update leads to installing a `shim` which is not going to be able to boot any longer
- I understand that the idea for specific SONiC CLI commands is to abstract potentially platform specific implementations away. However, would you mind providing examples where different implementations would be necessary? At least on x86 I would have yet to encounter hardware where the standard tooling would be insufficient.

doc/secure_boot/Secureboot_uefi_key_management.md
```diff
@@ -0,0 +1,217 @@
...
+   openssl x509 -inform der -in db-orig-3.der -out db-orig-3.pem
+4. Removal of db-orig-2.pem to be discarded
+5. Convert db-orig-1.pem and db-orig-3.pem to esl format
+   cert-to-efi-sig-list -g <GUID> db-orig-1.pem db-orig-1.esl
+   cert-to-efi-sig-list -g <GUID> db-orig-3.pem db-orig-3.esl
+6. Concatenate these two esl files 
+   cat db-orig-1.esl db-orig-3.esl > DB-new.esl
+7. Create authenticated variable and sign it
+   sign-efi-sig-list -g "<GUID>" -k KEK.key -c KEK.crt db DB-new.esl DB-new.auth
+8. Copy DB-new.auth to the system and use CLI to update UEFI database
```
Yarden-Z, line 151: What is the command used here?
Can you copy the DB-new.auth without providing the KEK.key to the system?

doc/secure_boot/Secureboot_uefi_key_management.md
```diff
@@ -0,0 +1,217 @@
...
+5. show platform security uefi variables dbx
+
+1. platform security uefi variables update pk  <file>
+2. platform security uefi variables update kek <file>   
+3. platform security uefi variables update db  <file>
+4. platform security uefi variables update dbx <file>  
+```
+
+## 7 Test considerations
+The sonic-mgmt can cover UEFI key management such as adding, updating, removing, and revoking UEFI keys.
```
Yarden-Z, line 214: Should add a note that this feature should enhance secure boot tests where we can install an image with a specific key, then move it to the dbx and therefor - the same image cannot be installed anymore (signed with the previous key).

doc/secure_boot/Secureboot_uefi_key_management.md
```diff
@@ -0,0 +1,217 @@
...
+   cert-to-efi-sig-list -g <GUID> db-orig-3.pem db-orig-3.esl
+6. Concatenate these two esl files 
+   cat db-orig-1.esl db-orig-3.esl > DB-new.esl
+7. Create authenticated variable and sign it
+   sign-efi-sig-list -g "<GUID>" -k KEK.key -c KEK.crt db DB-new.esl DB-new.auth
+8. Copy DB-new.auth to the system and use CLI to update UEFI database
+
+````
+
+### 4.4 Revoke keys
```
Yarden-Z, line 155: Worth mentioning that key revocation is a potentially destructive action.
If we are removing a DB entry that is currently being used - this will cause the system to enter fatal mode (recoverable only via UEFI).

doc/secure_boot/Secureboot_uefi_key_management.md
```diff
@@ -0,0 +1,217 @@
...
+
+/sys/firmware/efi/efivars/PK-<GUID>
+/sys/firmware/efi/efivars/KEK-<GUID>
+/sys/firmware/efi/efivars/DB-<GUID>
+/sys/firmware/efi/efivars/DBX-<GUID>
+```
+The platform plugin for this OVMF type UEFI can use either sysfs or efi-readvar and efi-updatevar to implement getVariable() and setVariable() methods.   
+   
+
+## 6 SONiC CLI commands
```
Yarden-Z, line 199: What are the potential return values here?
Updatevar might fail (it being immutable, specific platform constraints, etc').
What is the expected design for these type of commands? As I see that they only receive a file and perform an action.

doc/secure_boot/Secureboot_uefi_key_management.md
mheese: I believe the idea is to perform the signing outside of the system, and then in (8) only copy the new file to the system where you can then update it with the CLI

doc/secure_boot/Secureboot_uefi_key_management.md
mheese: totally agree. See also my comment above about protection of boot chain.

doc/secure_boot/Secureboot_uefi_key_management.md
sacnaik: The HLD is referred to as OVMF since it is Open firmware and available for someone to experiment with. Cisco BIOS has implemented a similar approach to OVMF.  When Cisco8000 implements SONiC UEFI key management for Cisco hardware similar code will be added under Cisco platform-specific code.  That means the common contribution here is the SONiC CLI and Python module defining the base class that is mentioned in the document.

doc/secure_boot/Secureboot_uefi_key_management.md
Yarden-Z: I understand, but there might be UEFI cases where you cannot upload a new entry to the UEFI without providing the private KEK key (worth checking if this is true, when I played with these items a few years ago this was the case).
Or if you are entering a new KEK, you'll need the PK.

abdosi:
@xumia for reference

zhangyanzhao:
@sacnaik can you please add the code PRs to this HLD by referring to #806? Thanks.

zhangyanzhao:
code PR is not ready, move to backlog for future release

zhangyanzhao:
no code PRs, move to backlog

PR #1453, opened 2023-08-29
CMIS Module Management Enhancement  HLD

noaOrMlnx:
Add design document to enhance host_tx_ready set process to State DB, to have full synchronization between asic and module configuration.

| PR title                                                                  | state                                                                                                                  | context                                                                                                                     |
| ------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
| [Add host_tx_ready enhancements](https://github.com/sonic-net/sonic-swss/pull/2930) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [Add STATE_TRANSCEIVER_INFO_TABLE_NAME to shcema.h](https://github.com/sonic-net/sonic-swss-common/pull/824)   | GitHub issue/pull request detail  | GitHub pull request check contexts  |
| [Add SAI Notification support for host_tx_ready](https://github.com/sonic-net/sonic-sairedis/pull/1307)    | GitHub issue/pull request detail    | GitHub pull request check contexts    |           |


doc/cmis-module-enhancement/cmis-module-enhancement.md
```diff
@@ -0,0 +1,224 @@
...
+<p align=left>
+<img src="img/The-Current-Flow.png" alt="Figure 1. Current Flow">
+</p>
+
+### 6.1. Problem Statement
+With move to SW-based management of CMIS modules - when the ASIC side is configured by vendor SDK, and module side by SONIC, some problems have been identified with the current approach.
+
+#### 6.1.2. Host TX Ready signal
+As mentioned earlier, currently SONIC assumes that as soon as the Admin Status is set to UP and the corresponding SAI call returns with OK/SUCCESS status, the "host_tx_ready" flag can be set in the STATE DB to trigger the CMIS State Machine for specific port. 
+But it is not always a truth. ASIC port initialization process takes some time and this time can increase with move to new transceiver technologies (e.g. BiDi).
```
madhupalu, line 54: SONiC taken (~ 10-20 sec )to bring up the ports and complete the FSM, do you see any performance impact with this design tx_ready_signal ?

doc/cmis-module-enhancement/cmis-module-enhancement.md
```diff
@@ -0,0 +1,224 @@
...
+Whenever one of them is set to false (e.g. TX signal is disabled due to removal event of Admin station is set to DOWN), the SDK/FW stops transmission of this signal.
+
+This approach allows to avoid the issue of unterminated transmission when ASIC starts the transmission of the high-speed signal even when no module is plugged-in causing cross-talks to adjacent ports, high EMI/BER and shortening the transceiver lifespan
+
+On port creation for platforms supporting proposed enhancements, the SAI_PORT_ATTR_HOST_TX_SIGNAL_ENABLE shall be explicitly set by Ports OA to FALSE since its default SAI value is defined as TRUE for backward-compatibility reasons.
+
+#### 6.3.3. Handling of HW-based “host_tx_ready” event
+Per description above, on the platform supporting new enhancements, the SDK/FW shall send an asynchronous notification of “host_tx_ready” to SWSS when the ASIC starts/stops sending high-speed signal to a module.
+
+The vendor SDK shall internally support this indication from ASIC (via the trap mechanism) and shall call the notification callback registered by Ports OA on initialization .
```
madhupalu, line 161: How does SONiC be notified the whethernew  tx_trap_signal trap supported or not? I believe it must be vendor CMIS implemetation must support this trap otherwise correct us!

eddyk-nvidia: SONIC (Ports OA) will perform the query capability for support of HW-based host_tx_ready notification. Then it will initializes the consumer for this notification only if Capability Query returned the indication of support on the platform. The vendor SAI will indicate its support of this notification ONLY if the trap from ASIC about start/stop of host_tx_ready signal is supported by ASIC/FW on the platform

doc/cmis-module-enhancement/cmis-module-enhancement.md
```diff
@@ -0,0 +1,224 @@
...
+Current Flow:
+<p align=left>
+<img src="img/The-Current-Flow.png" alt="Figure 1. Current Flow">
+</p>
+
+### 6.1. Problem Statement
+With move to SW-based management of CMIS modules - when the ASIC side is configured by vendor SDK, and module side by SONIC, some problems have been identified with the current approach.
+
+#### 6.1.2. Host TX Ready signal
+As mentioned earlier, currently SONIC assumes that as soon as the Admin Status is set to UP and the corresponding SAI call returns with OK/SUCCESS status, the "host_tx_ready" flag can be set in the STATE DB to trigger the CMIS State Machine for specific port.
```
jaganbal-a, line 53: There is no assumption on host_tx_ready flag being set,  the ASIC vendor SW have to make sure the valid transmission/tuning etc while processing/handling the attribute 'SAI_PORT_ATTR_ADMIN_STATE'. 

What is the advantage of returning from SAI_PORT_ATTR_ADMIN_STATE for Up case without enabling the proper transmission ? 

Can you please share the benchmark on completion of SAI_PORT_ATTR_ADMIN_STATE and reaching the transmission of high speed signal from your setup? IMO this will be couple of hundred ms.

doc/cmis-module-enhancement/cmis-module-enhancement.md
```diff
@@ -0,0 +1,224 @@
...
+
+#### 6.1.2. Host TX Ready signal
+As mentioned earlier, currently SONIC assumes that as soon as the Admin Status is set to UP and the corresponding SAI call returns with OK/SUCCESS status, the "host_tx_ready" flag can be set in the STATE DB to trigger the CMIS State Machine for specific port. 
+But it is not always a truth. ASIC port initialization process takes some time and this time can increase with move to new transceiver technologies (e.g. BiDi). 
+So, in some cases the module initialization can be triggered too early, before the high-speed signal started to be transmitted by ASIC to a module
+
+#### 6.1.3. Unterminated Transmission
+With move to the SW-based module management, the module presence is handled by SONIC, and FW might be unaware of the module presence status. 
+In this case, when the Admin status of a port is set to UP, FW can start transmitting the high-speed signal even without a plugged-in module. Such unterminated transmission can cause cross-talks to adjacent ports, high EMI, high BER and eventually shorten the transceiver lifetime so it is recommended that ASIC will not start sending the high-speed signal before a module is plugged.
```
jaganbal-a, line 59: I see a usecase where the some amount power is saved when the transmission is not enabled in ASIC.

Can you please elaborate on unterminated transmission ?   and how it cause cross-talks to adjacent ports?

prgeor:
@noaOrMlnx @liat-grozovik some open comments in this HLD to be taken care

doc/cmis-module-enhancement/cmis-module-enhancement.md
```diff
@@ -0,0 +1,224 @@
...
+To have the update status in STATE DB, one needs to refresh the status of “host_tx_ready” like it is done today for port’s operational status and speed.
+To do that a new logic related to host_tx_ready will be added in PortsOrch::initializePort().
+For platforms supporting proposed enhancements, SAI_PORT_ATTR_HOST_TX_READY_STATUS SAI attribute shall be queried and then the “host_tx_ready” flag in the STATE DB shall be updated accordingly.
+
+```
+bool PortsOrch::initializePort(Port &port)
+{
+    ...
+    ...
+    if (saiTxReadyNotifySupported && saiHwTxSignalEnabled)
```
prgeor, line 181: @noaOrMlnx what about platform that does not support this hw notification mechanism? Can we have parity in behavior for both?

doc/cmis-module-enhancement/cmis-module-enhancement.md
```diff
@@ -0,0 +1,224 @@
...
+The backward-compatibility is supported by a new design. Therefore, the handling shall be as follows:
+
+| Platform Type                            | Ports OA Behavior                                                                                                                                            |   |
+|------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------|---|
+| Platform supporting new enhancements     | On return from SAI call of setting port's Admin status, nothing is done, unless the notification event "host_tx_ready" is received and consumed by Ports OA. |   |
+| Platform not supporting new enhancements | Ports OA will set "host_tx_ready" in STATE DB right after Admin status SAI call returns.<br>(UP -> "true", DOWN -> "false")                                  |   |
+
+
+##### 6.3.3.1. Warm-boot handling
+After warm-boot, SONiC will not receive a new “host_tx_ready” notification from SAI (because nothing has changed in ASIC/module).
```
prgeor, line 171: @noaOrMlnx I am not sure what is the usecase to have host_tx_ready updated during warm reboot? Xcvrd is not going to act upon this host_tx_ready notification anyways during warm reboot

doc/cmis-module-enhancement/cmis-module-enhancement.md
noaOrMlnx: @jaganbal-a the advantage of returning from ADMIN state UP without enabling host_tx_ready, is to not start configuring the module before the high speed signal was transmitted. This situation is not desired, since module must be configured only after Asic is configured. and we cannot assume that Asic is configured when setting ADMIN to UP.

doc/cmis-module-enhancement/cmis-module-enhancement.md
noaOrMlnx: @jaganbal-a 
> I see a usecase where the some amount power is saved when the transmission is not enabled in ASIC.

The reason for this enhancement is not power. it is a side affect. 
The main reason here is to make sure the SerDes will not be damaged when high-speed signal will be sent without having a module plugged in to cage. 

> Can you please elaborate on unterminated transmission ? and how it cause cross-talks to adjacent ports?

Sending high-speed signal without having a plugged in module, can cause interruptions. for adjacent cables, it will cause more BER and again, can damage the SerDes.

doc/cmis-module-enhancement/cmis-module-enhancement.md
noaOrMlnx: @prgeor 
If a platform does not support notification mechanism, but does support the enable signal, then following situation can happen:
1. host tx ready will not set to true when admin status is set to up. 
2. "allow" signal will be sent to lower layers saying the module is plugged in. 
3. host_tx_ready will never be set to TRUE since a notification that high-speed signal has been transmitted will never arrive. 

That's the reason we need to make sure those 2 queries are supported.

doc/cmis-module-enhancement/cmis-module-enhancement.md
noaOrMlnx: @prgeor The only reason here is to have STATE DB updated with lower layers status. 
This was also discussed during HLD review, it is not mandatory but nice to have it all updated with the same status.

doc/cmis-module-enhancement/cmis-module-enhancement.md
```diff
@@ -0,0 +1,224 @@
...
+
+Once a module plug event occurs, Xcvrd in Pmon will update TRANSCEIVER_INFO table in STATE DB. 
+
+In order to know on module plug events, Ports OA will listen to changes in TRANSCEIVER_INFO table.
+According to information in TRANSCEIVER_INFO table, Ports OA will send SAI_PORT_ATTR_HOST_TX_SIGNAL_ENABLE enable/disable to SAI, indicating if sending high-speed signal to module is allowed or not. 
+
+#### 6.2.2. host_tx_ready
+When the ASIC starts transmitting the high-speed signal toward a plugged module the vendor SAI should notify the SONIC (SWSS) of that via a new notification - SAI_SWITCH_ATTR_PORT_HOST_TX_READY_NOTIFY.
+
+The SWSS shall use this notification to set "host_tx_ready" flag in STATE DB which will trigger the CMIS initialization of the module.
```
prgeor, line 106:
```suggestion
The OA shall use this notification to set "host_tx_ready" flag in STATE DB which will trigger the CMIS initialization of the module.
```

doc/cmis-module-enhancement/cmis-module-enhancement.md
```diff
@@ -0,0 +1,224 @@
...
+Ports OA shall use the status of support of these capabilities for its flow.
+For example, the notification consumer for a new SAI_SWITCH_ATTR_PORT_HOST_TX_READY_NOTIFY notification will be initialized only if the platform supports the asynchronous notification from SAI/SDK to SWSS.
+
+#### 6.3.2. Enabling/disabling “host_tx_signal”
+As described earlier, for platforms supporting this capability, high-speed signal shall be explicitly enabled/disabled by Ports OA on getting the Insertion/Removal event for a specific module.
+
+When a module is plugged in/out the PMON adds/deletes a per-port entry in the TRANSCIEVER_TABLE in STATE DB.
+This event shall be used by Ports OA as a trigger for enabling or disabling the host_tx_signal (using setting true/false to SAI_PORT_ATTR_HOST_TX_SIGNAL_ENABLE on the Port Object).
+
+To detect this insertion/removal event the Ports Orch Agent shall register for changes in the TRANSCIEVER_TABLE in STATE DB and will do the following:
```
prgeor, line 144: @noaOrMlnx AFAIK, STATE_DB TRANSCEIVER_TABLE is not preserved across warm-reboot. If so, my concern is during warm-reboot if the TRANSCEIVER_TABLE gets deleted, then SAI_PORT_ATTR_HOST_TX_SIGNAL_ENABLE could become FALSE after warm-reboot. Could you please check this.?

doc/cmis-module-enhancement/cmis-module-enhancement.md
prgeor: @noaOrMlnx can we preserve the STATE_DB's TRANSCEIVER_INFO table and PORT_TABLE (host_tx_ready) across warm-reboot so that it does not cause link flap.

doc/cmis-module-enhancement/cmis-module-enhancement.md
prgeor: @noaOrMlnx I was referring to host_tx_ready which gets deleted from STATE_DB table after a warm-reboot. With your proposal, for platform that support notification mechanis, the host_tx_ready will be restored after warm-reboot. But for other platforms, host_tx_ready will remain deleted from STATE_DB table causing Xcvrd to bring the link down. See my previous comment above: We should not delete PORT_TABLE from STATE_DB table during warm-reboot that way we can keep the behavior same for all platforms.

doc/cmis-module-enhancement/cmis-module-enhancement.md
noaOrMlnx: @prgeor 
Hi, 
When we execute warm-reboot, the whole table is deleted. Since we are performing a reboot, it will happen after OA will stop, so no notification is received. 
When OA is up again after reboot, the table is clean, so notification will be received only upon new info inserted to the table. 
So for me it looks like we should not worry about it.

doc/cmis-module-enhancement/cmis-module-enhancement.md
noaOrMlnx: Hi @prgeor 
Can you explain why do you expect link-flap when host_tx_ready is deleted with the deletion of PORT_TABLE? It is the exact same behavior as before. 

IMO reserving PORT_TABLE will not be the right solution.  
In PORT_TABLE there are some other keys as 'oper_status'. We do not want to have this key reserved between boots, but we need to know the real status - if the port link is up or not.
That's why we get the real oper status after boot is done via query to SAI.

doc/cmis-module-enhancement/cmis-module-enhancement.md
noaOrMlnx: @prgeor Hi Prince, As far as I checked, TRANSCEIVER_INFO table does not deleted when doing warm-reboot. 
only PORT_TABLE.
For PORT_TABLE, please check my above comment, which explains why I think it will not be the solution.

liat-grozovik:
@prgeor  kindly reminder to the HLD PR review. do we still have open issue?

doc/cmis-module-enhancement/cmis-module-enhancement.md
prgeor: @noaOrMlnx it seems except few TABLES mentioned [here](https://github.com/sonic-net/sonic-utilities/blob/master/scripts/fast-reboot#L247) other tables get deleted before warm-reboot

doc/cmis-module-enhancement/cmis-module-enhancement.md
prgeor: @noaOrMlnx oper_status i agree. I was talking about `host_tx_ready`. Have you tested if PORT_TABLE's `host_tx_ready` is not deleted ? The impact of host_tx_ready deleted during warm-reboot is Xcvrd re-initializing the CMIS optics causing link flap

noaOrMlnx: It is not deleted, but host_tx_ready value turns to false

doc/cmis-module-enhancement/cmis-module-enhancement.md
prgeor: @noaOrMlnx is your confirmation verified on a real switch? i.e if on warm-reboot `host_tx_ready` does not get deleted from STATE_DB and hence does not need to be preserved and restored before and after warm-reboot?

prgeor:
> @prgeor kindly reminder to the HLD PR review. do we still have open issue?

@liat-grozovik i have some concern, on warm-reboot, rest all looks good. have put in the HLD comment and PR comment as well. next will sync up with @noaOrMlnx offline

doc/cmis-module-enhancement/cmis-module-enhancement.md
noaOrMlnx: my bad, both tables are deleted during warm reboot

prgeor (approved):

eddyk-nvidia (approved):

skg-net:
@noaOrMlnx Can you please update the Quality Metric (Alpha/Beta/GA) for the feature either in this PR comments or in HLD itself based on https://github.com/sonic-net/SONiC/blob/master/doc/SONiC%20feature%20quality%20definition.md
Thanks

PR #1458, opened 2023-09-04
Enhanced Rule#18 to Disallow YANG LISTs with Over-lapping Keys

faraazbrcm:
It is observed that guideline does not describe the rules for YANG list keys. Added rules to make YANG LIST unique.

### DIFF
--- a/doc/mgmt/SONiC_YANG_Model_Guidelines.md
+++ b/doc/mgmt/SONiC_YANG_Model_Guidelines.md
@@ -503,12 +503,8 @@ container PORTCHANNEL_INTERFACE {
 #### Strategies for Ensuring Unique and Unambiguous Keys
 
 1. **Composite Keys with Different Number of Elements**: Utilize composite keys that have a different number of key elements to distinguish lists.
-
-2. **Different Data Types with Constraints**: Use constraints like length, pattern, or range to make keys of different data types non-overlapping.
   
-3. **Pattern Constraints**: Implement regular expressions to define specific patterns that keys must match.
-
-4. **Length and Range Constraints**: Limit the length of strings or the range of integers to make them unique.
+2. **Pattern Constraints**: Implement regular expressions to define specific patterns that keys must match (Only applicable to string datatype).
 
 #### ABNF
 ```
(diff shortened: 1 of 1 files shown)


adyeung:
@dutta-partha @praveen-li pls help review

doc/mgmt/SONiC_YANG_Model_Guidelines.md
```diff
@@ -550,12 +562,166 @@ container sonic-interface {
 ......
 ```
 
+#### Example 2: Composite Keys with Same Number of Elements(NOT Allowed case)
```
praveen-li, line 565: This one looks right to me.

doc/mgmt/SONiC_YANG_Model_Guidelines.md
```diff
@@ -550,12 +562,166 @@ container sonic-interface {
...
+			// ...
+		}
+	} 
+}
+......
+```
+
+#### Example 3: Pattern Constraints (Allowed case)
+
+In this example, `INTERFACE_LIST` uses a key `ifname` that must start with "Eth", while `INTERFACE_IPADDR_LIST` uses a key `ifname` that must start with "Vlan".
```
praveen-li, line 593: Does the translation support it. Also do we have use case for it? Kindly paste if we have.

doc/mgmt/SONiC_YANG_Model_Guidelines.md
```diff
@@ -550,12 +562,166 @@ container sonic-interface {
...
+			// ...
+		}
+	} 
+}
+......
+```
+
+#### Example 6: Length Constraints (NOT Allowed case)
+
+Here, both `INTERFACE_LIST` and `INTERFACE_ANOTHER_LIST` use a key ifname with overlapping length constraints, making them potentially ambiguous.
```
praveen-li, line 689: I have similar commands for examples, 4 to 6. Do we have valid use case right now, where these example or constraints can be used. Thanks.

doc/mgmt/SONiC_YANG_Model_Guidelines.md
faraazbrcm: you mean Rule holds good.. And example given makes sense. correct?

wen587: Hi @praveen-li , here is the issue created by @faraazbrcm : https://github.com/sonic-net/sonic-buildimage/issues/16356
The same key count cannot be parsed in GCU side. because two list share the same key count and same key type. When GCU is verifying the YANG model: [telemetry_client.yang](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-telemetry_client.yang#L68). GCU failed to parse TELEMETRY_CLIENT_SUB_LIST. It verify the list matches TELEMETRY_CLIENT_DS_LIST then return failure.

I would suggest to use different key count to protect GCU parse.

wen587: I can think of two fix on this: 
Combine two list into one with key count=2. Example: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-bgp-allowed-prefix.yang#L70-L71
Use two different key count to represent these lists. Example: https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-interface.yang#L45-L95

doc/mgmt/SONiC_YANG_Model_Guidelines.md
praveen-li: - [validator is on place.](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-mgmt/tests/libyang-python-tests/test_sonic_yang.py#L284)
-[test for translation and reverse translation](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-mgmt/tests/libyang-python-tests/test_sonic_yang.py#L324)
- [Config db used for above test](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/tests/files/sample_config_db.json)

Yes, seems for telemetry client, no config may be added in sample_config_db.json. That's why no build failures. This can be in Guideline to add config.

doc/mgmt/SONiC_YANG_Model_Guidelines.md
praveen-li: Composite Keys with Same Number of Elements. 

[PC]: Same number of element is allowed by translation. 'Same number of element plus Same type' is not allowed. But if we can support patterns, we can allow same type in future. I feel it should be allowed and should be added in translation.

doc/mgmt/SONiC_YANG_Model_Guidelines.md
praveen-li: Looking at telemetry YANG, seems there is a genuine case. If yes then.

-- Update Guideline. [please provide real example of today's config].
-- above should be done after translation PR is raised, which supports pattern if primary key has 'same number of element and same type.

doc/mgmt/SONiC_YANG_Model_Guidelines.md
faraazbrcm: @praveen-li There exists a config for telemetry in [sample_config_db.json](https://github.com/sonic-net/sonic-buildimage/blob/cc6a9b6832368a107ca29edb6c828c84849f3dd2/src/sonic-yang-models/tests/files/sample_config_db.json#L1207). Any idea why this has NOT caused the build failure?

doc/mgmt/SONiC_YANG_Model_Guidelines.md
faraazbrcm: @praveen-li IMO guideline PR has to go first then the translations. The translations are based on guideline. correct me.

doc/mgmt/SONiC_YANG_Model_Guidelines.md
faraazbrcm: @praveen-li Elements with distinct types but the same number can also lead to overlaps, especially since Redis always treats keys as strings. For instance, the number 123 can also be interpreted as the string '123'. To prevent this, we should apply specific constraints, like patterns for strings and ranges for integer types. This requirement should be clearly stated in the guideline.

wen587: @praveen-li , in the telemetry client sample, the list defination(`TELEMETRY_CLIENT_SUB_LIST` and `TELEMETRY_CLIENT_DS_LIST`) belongs to the "same number of element plus Same type". It should not be allowed.

adyeung:
@praveen-li @qiluo-msft @wen587 @faraazbrcm A webex meeting invite is sent to discuss this PR on 9/26 Tues 8:30pm PST, pls plan to join to get a closure.

@wen587 I don't have your email contact, I routed the invite thru Qi, pls reachout to Qi for webex info

praveen-li:
@faraazbrcm See this Page https://github.com/sonic-net/SONiC/pull/1458

Below, Instead of sonic-utilities, build sonic-yang-mgmt PKG. Then you can run tests. Also Note, sonic-yang-mgmt PKG have dependency on sonic-yang-models PKG, so if you change yang models then build sonic-yang-models PKG too.

```
Build the sonic-utilities Python wheel package inside the Bullseye slave container, and tell the build system to keep the container alive when finished

make NOSTRETCH=1 NOBUSTER=1 KEEP_SLAVE_ON=yes target/python-wheels/bullseye/sonic_utilities-1.2-py3-none-any.whl

    When the build finishes, your prompt will change to indicate you are inside the slave container. Change into the src/sonic-utilities/ directory

    user@911799f161a0:/sonic$ cd src/sonic-utilities/

    You can now make changes to the sonic-utilities source and build the package or run unit tests with the commands below. When finished, you can exit the container by calling exit.

[To build](https://github.com/sonic-net/sonic-utilities/tree/master#to-build)

python3 setup.py bdist_wheel

[To run unit tests](https://github.com/sonic-net/sonic-utilities/tree/master#to-run-unit-tests)

python3 setup.py test
```

praveen-li:
@faraazbrcm  For {yang list with [same number of keys + string type] }

In current case, a yang list with [same number of keys + string type] can be allowed. A restriction to include 'pattern of the key' will be added to differentiate. Without pattern keyword for key, translation will not be supported from config_db.json to sonic_yang.json.

wen587:
Hi @praveen-li , I want to remind that GCU only supports different keys count for now. That's why the TELEMETRY_CLIENT Yang fails in GCU process.

wen587:
Create a PR for **Path 2**: Do not permit lists to have the same number of keys. 
https://github.com/sonic-net/sonic-buildimage/pull/16861

wen587:
> Create a PR for **Path 2**: Do not permit lists to have the same number of keys. [sonic-net/sonic-buildimage#16861](https://github.com/sonic-net/sonic-buildimage/pull/16861)

Hi, @faraazbrcm and @praveen-li , could you help review this PR about **Path 2**

faraazbrcm:
@praveen-li, I've updated this PR in line with Path1, as discussed in subgroup email correspondence with the Alibaba team. I believe we agreed to proceed with Path1, which involves adding support for the pattern statement. Could you please review this PR? If everything looks satisfactory, kindly approve it. If any modifications are needed, please let me know.

wen587:
@qiluo-msft  add Qi for vis

kwangsuk:
As you see in the typical YANG modeling practice like ipv4/ip6, the union type of key node is commonly used in the YANG community like openconfig or ietf. So, in the similar line, the [path2](https://github.com/sonic-net/sonic-buildimage/pull/16861) approach would be better practice.

faraazbrcm:
@praveen-li @adyeung @yaqiangz  we should decide whether to go with path 1 or path 2 ASAP. This PR is aligned with path 1.

kwangsuk:
> @praveen-li @adyeung @yaqiangz we should decide whether to go with path 1 or path 2 ASAP. This PR is aligned with path 1.

Lets not compromise the effective modeling practices just because of internal constraints in consuming YANG. As we move on, we should be able to encourage SONiC contributors to abstract services in well versed and structured YANG models.

faraazbrcm:
closing because alternate path https://github.com/sonic-net/sonic-buildimage/pull/16861 is accepted

PR #1460, opened 2023-09-06
Update Arista platforms in supported list

Staphylo:
- Update existing Arista product to use the full SKU name
 - Add missing products from the list


Staphylo:
@lguohan @yxieca

Issue #1462, opened 2023-09-07
TACACS long wait when configuring non active server with high priority


ycoheNvidia:
We have encountered an issue while configuring and authenticating Tacacs servers.
** What is happening **
The scenario happens when a non active/non existent server (server A) is being configured with higher priority than another active server (server B), with aaa fallback and failthrough enabled.
If we try to authenticate with a remote user defined in server B using ssh for example, we get significant wait times for connection to be established, such as 30-50 seconds each time.
** Additional research **
For user names that were created locally (like admin or others using useradd command) or authenticated and established before with a radius server - we did not encounter these delays.
After examining the debug logs we suspect that the source of the issue is somewhere between linux pam and tacplus_pam, where while user connection pam calls tacacs server authentication multiple times, as it is checking user permissions - waiting full timeout for each check,
In addition, when using a valid server as first priority to authenticate - we still see these multiple authentication requests logged in tacplus_pam and libnss tacplus libraries, but since there is no significant delay for each request- the session is established in a reasonable time (mostly less than 2 seconds). 

We would like to know if this is a know limitation for TACACS in SONiC, since documentation in  pam_tacplus library used by SONiC specifically states that only one active server is being used after first authentication (from https://github.com/kravietz/pam_tacplus/blob/main/README.md): 
"Having more that one TACACS+ server defined for given management group has following effects on authentication:

if the first server on the list is unreachable or failing pam_tacplus will try to authenticate the user against the other servers until it succeeds

the first_hit option has been deprecated

when the authentication function gets a positive reply from a server, it saves its address for future use by account management function (see below)

The account management (authorization) function asks only one TACACS+ server and it ignores the whole server list passed from command line. It uses server saved by authentication function after successful authenticating user on that server. We assume that the server is authoritative for queries about that user."

** Reproduction steps: **
1. configure active server with low priority: sudo config tacacs add ACTIVE_SERVER_IP -k ACTIVE_SERVER_KEY -p 1
2. configure aaa authentication: sudo config aaa authentication login tacacs+ local
3. configure aaa failthrough: sudo config aaa authentication failthrough enable
try to authenticate with remote user (sshpass -p remoteuserpassword ssh remoteuser@switch)  -should be done almost immediately
4. configure non active/dummy server with higher priority: sudo config tacacs add 1.1.1.1 -p 2
Try to authenticate to remote user (sshpass -p remoteuserpassword ssh remoteuser@switch)  -should take significate time with long delay.

[...]

PR #1467, opened 2023-09-11
Add TACACS server monitor design document.

liuh-80:
Add TACACS server monitor design document.
The TACACS server monitor can change TACACS server priority based on syslog and can resolve following issue:
https://github.com/sonic-net/SONiC/issues/1462

### DIFF
--- a/doc/aaa/TACACS+ Server Monitor.md
+++ b/doc/aaa/TACACS+ Server Monitor.md
@@ -1,0 +1,80 @@
+# TACACS+ server monitor design
+
+## Overview
+
+SONiC device usually configured with multiple TACACS+ server, when a server is unreachable, SONiC device will try to connect with next TACACS+ server.
+
+SONiC device will communicate with TACACS+ server in following scenarios:
+1. Remote user login to SONiC device.
+2. Remote user run commands on SONiC device.
+
+There is a timeout for each server, the default value is 5 seconds, this means if the first server not reachable, SONiC device will stuck there when user login or running commands.
+
+To improve this issue, SONiC will add a TACACS+ server monitor to change server priority, a server unreachable or slow response will be downgrade.
+
+### Functional Requirement
+- Monit TACACS+ server unreachable event from syslog.
+- Monit TACACS+ server slow response event from syslog.
+- Change server priority based unreachable event and slow response event.
+- Not change any other server attribute.
+- Not change any other TACACS+ config.
+
+### Syslog format
+- TACACS+ server unreachable event format:
+```
+    failed to connect TACACS+ server [ip address]
+```
+
+- TACACS+ server slow response event format:
+```
+    connect TACACS+ server [ip address] take [time span] ms
+```
+
+# 3 Limitation
+
+- Service priority change will have 1 minutes delay, this is because monit service will run profile every 1 minutes.
+
+# 4 Design
+
+```
+       +------------+ 
+       |    Monit   | 
+       +-----+------+  
+             |          
++------------v--------------+       +---------------------+
+|                           |       |                     |
+|                           |       |                     |
+|     TACACS+ Monitor       |------>|       syslog        |
+|                           |       |                     |
+|                           |       |                     |
++------------+--------------+       +---------------------
+             |                                
+   +---------v---------+               +-------+--------+
+   |                   |               |                |
+   | TACACS config cli +--------------->    ConifgDB    |
+   |                   |               |                |
+   +-------------------+               +-------+--------+
+
+```
+- TACACS+ monitor is a Monit profile.
+- TACACS+ monitor will monit syslog for TACACS+ server unreachable or slow response event.
+- Monit time window will be 5 minutes, this will hardcode in profile script.
+- When a TACACS+ server event happen, monitor will change server priority according following rules:
+    - If this is the only TACACS+ server, keeps no change.
+    - If server with issue already has priority 1:
+        - If any other server also has priority 1, change that server to higher priority, value will be: index*10
+        - If all other server already has higher priority, keep no change. 
+    - If server with issue not has priority 1:
+        - change server priority to 1.
+        - If any other server also has priority 1, change that server to higher priority, value will be: index*10
+        - If all other server already has higher priority, keep no change. 
+- sonic-utilities need change to support 'sudo config tacacs update' operation. 
+    - Today there is only 'add' and 'delete'
+    - Using 'add' and 'delete' in script have risk to lost all server.
+
+# 5 References
+
+## TACACS+ Authentication
+https://github.com/sonic-net/SONiC/blob/master/doc/aaa/TACACS%2B%20Authentication.md
+## SONiC TACACS+ improvement
+https://github.com/sonic-net/SONiC/blob/master/doc/aaa/TACACS%2B%20Design.md


ycoheNvidia:
Thanks for addressing the issue.
Even though this might work - in my opinion this is not a good solution for this issue and here is few reasons:
- user who configures authentication methods doesn't expect configuration to change without any notice or active action from his side.  
- logs can be instable, we have experienced inconsistency of Debian logs in the past, I suspect it might cause bugs for this solution
- The main issue here is based on an issue with pam/tacplus_pam. In my opinion - it should be resolved there. Adding a monitor workaround solution might be hard to maintain and not be optimal for this issue.

a-barboza:
cc: @shdasari 

The idea of adjusting the priority based on the responsiveness of the server is good. As mentioned, it should alleviate issue https://github.com/sonic-net/SONiC/issues/1462. 

However, I feel adjusting the configuration based on a temporary network event might not be the best approach. How about adding some state information to the STATE_DB (COUNTER_DB, or appropriate DB)? Adjust the tacplus pam configuration to account for temporarily unreachable server. Once the network event is detected to be resolved, then the temporary adjustment could be backed out, with suitable logs to advise the admin of the actions taken.

liuh-80:
> cc: @shdasari
> 
> The idea of adjusting the priority based on the responsiveness of the server is good. As mentioned, it should alleviate issue #1462.
> 
> However, I feel adjusting the configuration based on a temporary network event might not be the best approach. How about adding some state information to the STATE_DB (COUNTER_DB, or appropriate DB)? Adjust the tacplus pam configuration to account for temporarily unreachable server. Once the network event is detected to be resolved, then the temporary adjustment could be backed out, with suitable logs to advise the admin of the actions taken.

@a-barboza , currently if SONiC TACACS can connect to first server, it will not try connecting other low priority servers, which means can't detect low priority server network status. and if we want get connection status of low priority server, we need create a new daemon running in background to periodically check server.

Also in this design, the monitor change server priority based on a time window and event count threshold, for example more than 50% connection failed within 5 minutes. this is a simple solution can handle almost every scenario.

So, how about I update the design doc to:
1. First stage, implement monitor to change priority based on sys event, this will be high priority.
2. Second stage, update TACACS code to update COUNTER_DB data. create daemon to check TACACS server status. then we can update monitor to change priority based on COUNTER_DB and server reachability?

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+- TACACS+ server slow response event format:
+```
+    connect TACACS+ server [ip address] take [time span] ms
+```
+
+# 3 Limitation
+
+- Service priority change will have 1 minutes delay, this is because monit service will run profile every 1 minutes.
```
qiluo-msft, line 35: >1 minutes [](http://example.com/codeflow?start=36&length=9)

How do you choose the delay? Is it configurable? I am not seeing it needed to be in ConfigDB for runtime configurable, a conffile in image would be good enough.

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,80 @@
...
+   +---------v---------+               +-------+--------+
+   |                   |               |                |
+   | TACACS config cli +--------------->    ConifgDB    |
+   |                   |               |                |
+   +-------------------+               +-------+--------+
+
+```
+- TACACS+ monitor is a Monit profile.
+- TACACS+ monitor will monit syslog for TACACS+ server unreachable or slow response event.
+- Monit time window will be 5 minutes, this will hardcode in profile script.
```
qiluo-msft, line 61: >5 minutes [](http://example.com/codeflow?start=28&length=9)

How do you choose the time window? Is it configurable?

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,80 @@
...
+|                           |       |                     |
+|                           |       |                     |
+|     TACACS+ Monitor       |------>|       syslog        |
+|                           |       |                     |
+|                           |       |                     |
++------------+--------------+       +---------------------
+             |                                
+   +---------v---------+               +-------+--------+
+   |                   |               |                |
+   | TACACS config cli +--------------->    ConifgDB    |
```
qiluo-msft, line 54: >config cli [](http://example.com/codeflow?start=12&length=10)

ConfigDB is by design user config, and we do not expect daemons inside sonic box change them (there are some exception in current implementation, but we do not want to add more). I agree with @a-barboza "adding some state information to the STATE_DB (COUNTER_DB, or appropriate DB)?"

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,80 @@
...
+SONiC device will communicate with TACACS+ server in following scenarios:
+1. Remote user login to SONiC device.
+2. Remote user run commands on SONiC device.
+
+There is a timeout for each server, the default value is 5 seconds, this means if the first server not reachable, SONiC device will stuck there when user login or running commands.
+
+To improve this issue, SONiC will add a TACACS+ server monitor to change server priority, a server unreachable or slow response will be downgrade.
+
+### Functional Requirement
+- Monit TACACS+ server unreachable event from syslog.
```
qiluo-msft, line 16: >unreachable event from syslog [](http://example.com/codeflow?start=23&length=29)

Monitor local syslog file is not a robust method. How about improve pam component with more counters (possibly inside CountersDB)?

liuh-80: Fixed, will test TACACS server connection periodically and update latency data to COUNTER_DB

doc/aaa/TACACS+ Server Monitor.md
liuh-80: The delay is not configurable, it's default setting for monit service.

doc/aaa/TACACS+ Server Monitor.md
liuh-80: Fixed, new design will generate TACACS config by both CONFIG_DB and COUNTER_DB, and keep CONFIG_DB no change.

doc/aaa/TACACS+ Server Monitor.md
liuh-80: Fixed, add new CONFIG_DB schema to this document.

qiluo-msft (dismissed):

qiluo-msft:
@ycoheNvidia @a-barboza Would you like to review again? Do you have more comments?

liuh-80:
@a-barboza , I update design doc according to your suggestion, please give your comments.

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,86 @@
...
+#### TACPLUS_SERVER_LATENCY Table schema
+```
+; Key
+server_key           = IPAddress  ;  TACACS+ server’s address
+; Attributes
+latency              = 1*10DIGIT  ; server network latency in MS, -1 for connect to server timeout
+```
+
+### Config DB schema
+#### TACPLUS_SERVER_MONITOR Table schema
```
a-barboza, line 32: Would it make sense to rename this to TACPLUS_MONITOR (there is only one row in this table 'config') ?

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,86 @@
...
+```
+- TACACS+ monitor is a Monit profile.
+- TACACS+ monitor will perdically check TACACS server latency and update latency to COUNTER_DB.
+    - The latency in COUNTER_DB TACPLUS_SERVER_LATENCY table is average latency in recent time window.
+    - The time window side defined in CONFIG_DB TACPLUS_SERVER_MONITOR table.
+- TACACS+ monitor also will write warning message to syslog and re-generate TACACS server config file when following event happen in COUNTER_DB:
+    - Any server latency is -1, which means the server is unreachable.
+    - Any server latency is bigger than high_latency_threshold.
+- TACACS+ monitor will not change TACACS server config in CONFIG_DB, it only re-generate TACACS config file based on CONFIG_DB and COUNTER_DB.
+- The TACACS config file generate code will move to a new script file, both hostcfgd and TACACS+ monitor will use this file to re-generate TACACS config file.
```
a-barboza, line 75: If there are 2 independent process scripts modifying the same set of TACACS+ pam configuration files, there is a chance that the files may get corrupted (i.e. out-of-sync). Would it make sense to have hostcfgd itself monitor all tables (TACPLUS_SERVER, TACPLUS_SERVER_LATENCY, and TACPLUS_MONITOR) and write out the configuration? You could still keep the code to generate TACPLUS+ config file in a separate script, but have it only be executed from hostcfgd.

doc/aaa/TACACS+ Server Monitor.md
liuh-80: Fixed.
Because every table must have at least 1 row, so it's common in CONFIG_DB have 1 row tables for config.

doc/aaa/TACACS+ Server Monitor.md
liuh-80: Fixed, improve hostcfgd to also monitor TACPLUS_SERVER_LATENCY table change.

a-barboza (approved):

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,88 @@
...
+#### TACPLUS_SERVER_LATENCY Table schema
+```
+; Key
+server_key           = IPAddress  ;  TACACS+ server’s address
+; Attributes
+latency              = 1*10DIGIT  ; server network latency in MS, -1 for connect to server timeout
+```
+
+### Config DB schema
+#### TACPLUS_MONITOR Table schema
```
a-barboza, line 32: How can TACPLUS_MONITOR be disabled? For example, if the table is not defined, does it mean that TACPLUS Monitor is disabled, i.e. no monitoring, and effective priorities are same as configured priorities?

Yarden-Z:
How does a user get notified that the order has been changed?
Should there be some warning here?
Also - what is the overhead for this item? If we have 8 tacacs servers, defined with an 8 second timeout, and monit is running every minute - we might get to starvation. (might also occur if we have less, but timeout is bugger).
This seems like an odd solution towards solving another issue where the current timeout for tacacs is not being applied well.

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,88 @@
...
+    - The time window side defined in CONFIG_DB TACPLUS_MONITOR table.
+- TACACS+ monitor also will write warning message to syslog when following event happen:
+    - Any server latency is -1, which means the server is unreachable.
+    - Any server latency is bigger than high_latency_threshold.
+- Hostcfgd will monitor TACPLUS_SERVER_LATENCY table, and will re-generate TACACS config file when following event happen:
+    - Any server latency is -1, which means the server is unreachable.
+    - Any server latency is bigger than high_latency_threshold.
+- When hostcfgd generate TACACS config file, server priority calculated according to following rules:
+    - Get server priority info from CONFIG_DB TACPLUS_SERVER table.
+    - Change high latency server and un-reachable server priority to 1, this is because 1 is the smallest priority, and SONiC device will use high priority server first.
```
qiluo-msft, line 79: >un-reachable server [](http://example.com/codeflow?start=37&length=19)

Is it possible to treat "un-reachable server" differently from "high latency server"? Since they are unreachable, no need to include it in the generated TACACS config file.

liuh-80: Fixed, un-reachable server will not include in TACACS config file.

liuh-80:
> How does a user get notified that the order has been changed? Should there be some warning here? Also - what is the overhead for this item? If we have 8 tacacs servers, defined with an 8 second timeout, and monit is running every minute - we might get to starvation. (might also occur if we have less, but timeout is bugger). This seems like an odd solution towards solving another issue where the current timeout for tacacs is not being applied well.

When Monitor found latency or unreachable issue, warning message will write to syslog. in prod environment there need some other service check SONiC device health by syslog and send alert to user, which is not part of SONiC.

For timeout issue, if all 8 servers not reachable, Monit will handle it by send alert event for the timeout.

qiluo-msft (dismissed):

liuh-80:
@lguohan , could you review and signoff this PR?

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,89 @@
...
+- TACACS+ monitor also will write warning message to syslog when following event happen:
+    - Any server latency is -1, which means the server is unreachable.
+    - Any server latency is bigger than high_latency_threshold.
+- Hostcfgd will monitor TACPLUS_SERVER_LATENCY table, and will re-generate TACACS config file when following event happen:
+    - Any server latency is -1, which means the server is unreachable.
+    - Any server latency is bigger than high_latency_threshold.
+- When hostcfgd generate TACACS config file, server priority calculated according to following rules:
+    - Get server priority info from CONFIG_DB TACPLUS_SERVER table.
+    - Change high latency server to 1, this is because 1 is the smallest priority, and SONiC device will use high priority server first.
+    - Un-reachable server will not include in TACACS config file.
```
lguohan, line 80: if un-reachable server is excluded, later if the server becomes reachable, how can we include it back?

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+### Config DB schema
+#### TACPLUS_MONITOR Table schema
+```
+; Key
+config_key           = 'config'  ;  The configuration key
+; Attributes
+time_window              = 1*5DIGIT  ; Monitor time window in minute, default is 5
+high_latency_threshold   = 1*5DIGIT  ; High latency threshold in ms, default is 20
```
lguohan, line 38: missing yang mode design.

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,89 @@
...
+1. Remote user login to SONiC device.
+2. Remote user run commands on SONiC device.
+
+There is a timeout for each server, the default value is 5 seconds, this means if the first server not reachable, SONiC device will stuck there when user login or running commands.
+
+To improve this issue, SONiC will add a TACACS+ server monitor to change server priority, a server unreachable or slow response will be downgrade.
+
+### Functional Requirement
+- Monit TACACS+ server unreachable event from COUNTER_DB.
+- Monit TACACS+ server slow response event from COUNTER_DB.
```
lguohan, line 17: which component write to the counter_db, it is not clear from the design doc

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,89 @@
...
+- TACACS+ monitor will perdically check TACACS server latency and update latency to COUNTER_DB.
+    - The latency in COUNTER_DB TACPLUS_SERVER_LATENCY table is average latency in recent time window.
+    - The time window side defined in CONFIG_DB TACPLUS_MONITOR table.
+- TACACS+ monitor also will write warning message to syslog when following event happen:
+    - Any server latency is -1, which means the server is unreachable.
+    - Any server latency is bigger than high_latency_threshold.
+- Hostcfgd will monitor TACPLUS_SERVER_LATENCY table, and will re-generate TACACS config file when following event happen:
+    - Any server latency is -1, which means the server is unreachable.
+    - Any server latency is bigger than high_latency_threshold.
+- When hostcfgd generate TACACS config file, server priority calculated according to following rules:
```
lguohan, line 77: we need an option to maintain backward compatibility

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,89 @@
...
+
+```
+- TACACS+ monitor is a Monit profile.
+- TACACS+ monitor will perdically check TACACS server latency and update latency to COUNTER_DB.
+    - The latency in COUNTER_DB TACPLUS_SERVER_LATENCY table is average latency in recent time window.
+    - The time window side defined in CONFIG_DB TACPLUS_MONITOR table.
+- TACACS+ monitor also will write warning message to syslog when following event happen:
+    - Any server latency is -1, which means the server is unreachable.
+    - Any server latency is bigger than high_latency_threshold.
+- Hostcfgd will monitor TACPLUS_SERVER_LATENCY table, and will re-generate TACACS config file when following event happen:
```
lguohan, line 74: what is the threshold to determine high latency v.s. not. how do we choose the threshold.

lguohan (changes requested):
commented.

doc/aaa/TACACS+ Server Monitor.md
liuh-80: Added 'enable' flag for disable this feature. when feature disabled, configured priorities will same as configured priorities.

doc/aaa/TACACS+ Server Monitor.md
liuh-80: Fixed, yang model added.

doc/aaa/TACACS+ Server Monitor.md
liuh-80: Fixed, add 'enable' flag, this feature can be disable by this flag.

doc/aaa/TACACS+ Server Monitor.md
liuh-80: The threshold is 20ms. which is based on experience when handle TACACS server latency/unreachable issue, I can share more detail in review meeting, but may not necessarily to write that in public doc.

doc/aaa/TACACS+ Server Monitor.md
liuh-80: Fixed, hostcfgd will add server back when found an unreachable server become reachable.

doc/aaa/TACACS+ Server Monitor.md
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+
+### Config DB schema
+#### TACPLUS_MONITOR Table schema
+```
+; Key
+config_key           = 'config'  ;  The configuration key
+; Attributes
+enable                   = BOOLEAN   ; Enable Monitor feature
```
qiluo-msft, line 74: >enable [](http://example.com/codeflow?start=0&length=6)

There is a `FEATURE` table to control many optional features. And suggest we separate the TACACS monitor feature and TACACS auto downgrade feature, and control them with finer granularity.

wait2035:
Hello, every one. Is there any follow-up plans for this design?

PR #1470, opened 2023-09-12
HLD: DHCPv4 - Specify dhcp relay's Gateway explicitly with Primary address.

pushpraj:
This document describes the High Level Design of 'Secondary' interfaces of vlan. These secondary interfaces are also excluded in use in dhcpv4 relay.

A vlan can have more than one subnet assigned to it. Such additional interfaces should be tagged as 'secondary'. Sonic OS should support dhcpv4 over non-secondary interfaces. 

Use case: We want to increase the IP addresses range supported by a live switch. These additional IPs are to be used for new scenarios on existing physical server by additional IP addresses. 

Challenge: DHCPv4 is affected when sonic switch is configured with multiple interfaces in a vlan. When a dhcpv4 packet is forwarded via dhcprelayagent, it embeds its own gateway address as 'giaddr' for return communication.
In case of multiple interfaces, dhcprelayagent randomizes giaddr among all interfaces.
 
This behavior is fixed by marking additional subnets as 'secondary' and explicitly specifying the primary interface as gateway for dhcpv4.

| Repo  | PR title | State |
| ----------------- | ----------------- | ----------------- |
| sonic-buildimage | [Adding new secondary field to VLAN_INTERFACE table table](https://github.com/sonic-net/sonic-buildimage/pull/16827)| GitHub issue/pull request detail |
| sonic-utilities | [Add new cli to add a interface ip as secondary address](https://github.com/sonic-net/sonic-utilities/pull/3016)| GitHub issue/pull request detail |
| sonic-buildimage | [Modify dhcp relay to pick primary address](https://github.com/sonic-net/sonic-buildimage/pull/17012)| GitHub issue/pull request detail |
| sonic-buildimage | [Added new test case to check docker-dhcp-relay file ](https://github.com/sonic-net/sonic-buildimage/pull/17263)| GitHub issue/pull request detail |
| sonic-mgmt | [Modify dhcp_relay test case](https://github.com/sonic-net/sonic-mgmt/pull/10828)| GitHub issue/pull request detail |

### DIFF
--- a/doc/DHCPv4_Gateway/DHCPv4_gateway.md
+++ b/doc/DHCPv4_Gateway/DHCPv4_gateway.md
@@ -1,0 +1,127 @@
+# DHCPv4 - Specify Gateway explicitly #
+
+## Table of Content 
+1. [Scope](#Scope)
+2. [Definitions](#Definitions/Abbreviations)
+3. [Overview](#Overview)
+4. [Requirements](#Requirements)
+5. [High Level Design](#High-Level-Design)
+6. [Test](#Testing-Considerations)
+
+### Revision  
+
+### Scope  
+
+This document describes the High Level Design of 'Subnets' of vlan and use of same in dhcpv4.
+This HLD document represent the changes in PR: https://github.com/sonic-net/sonic-buildimage/pull/15969.
+
+### Definitions/Abbreviations 
+
+
+
+### Overview 
+
+'Subnets' specify the primary subnet of a vlan (IPv4;IPv6). When a dhcpv4 packet is forwarded via dhcprelayagent, it embeds its own gateway address as 'giaddr' for return communication.
+In case of multiple interfaces, dhcprelayagent misconfigures this giaddr to any of the interfaces. 
+This change fixes this behaviour by explicitly specifying the primary interface/gateway.
+This primary subnet is stored in db as 'subnets' of VLAN.
+
+### Requirements
+
+1. Support a new member 'Subnets' of VLAN in config_db.
+2. Support parsing and assignment of subnets from minigraph to config_db. 
+3. Support specifying subnet's first address as gateway address to command line arguments to /usr/sbin/dhcrelay as -g.
+4. isc-dhcp/dhcrelay to support '-g gateway' argument, by porting an existing patch.
+
+### Architecture Design 
+
+This HLD doesn't propose any sonic architectural changes. It follows the existing architecture of sonic - open source asc-dhcp is used for its dhcpv4 requirements.
+This HLD prposes to enhance the dhcpv4 capability by avoiding ambiguity in dhcpv4 packets wrt giaddr field.
+
+### High-Level Design 
+
+This section covers the high level design of the feature/enhancement. This section covers the following points in detail.
+		
+	- Is it a built-in SONiC feature or a SONiC Application Extension? - Built in feature.
+	- What are the modules and sub-modules that are modified for this design? - isc-dhcp.
+	- What are the repositories that would be changed? - sonic-subnet/sonic-buildimage.
+	- Module/sub-module interfaces and dependencies. 
+	- SWSS and Syncd changes in detail - N/A.
+	- DB and Schema changes (APP_DB, ASIC_DB, COUNTERS_DB, LOGLEVEL_DB, CONFIG_DB, STATE_DB) - sonic-vlan/sonic-vlan/VLAN/VLAN_LIST/Subnets.
+	- Sequence diagram if required. - N/A.
+	- Linux dependencies and interface - N/A.
+	- Warm reboot requirements/dependencies - N/A.
+	- Fastboot requirements/dependencies - N/A.
+	- Scalability and performance requirements/impact - N/A.
+	- Memory requirements - N/A.
+	- Docker dependency - N/A,
+	- Build dependency if any - N/A.
+	- Management interfaces - SNMP, CLI, RestAPI, etc.,
+	- Serviceability and Debug (logging, counters, trace etc) related design
+	- Is this change specific to any platform? Are there dependencies for platforms to implement anything to make this feature work? If yes, explain in detail and inform community in advance. - N/A.
+	- SAI API requirements, CLI requirements, ConfigDB requirements. Design is covered in following sections. - ConfigDB requirement is captured below.
+
+### SAI API 
+
+No change in SAI API.
+
+### Configuration and management 
+(This section should have sub-sections for all types of configuration and management related design. Example sub-sections for "CLI" and "Config DB" are given below. Sub-sections related to data models (YANG, REST, gNMI, etc.,) should be added as required.)
+
+#### Minigraph 
+
+<<Redis DB Structure | Json Style>>
+{
+"VLAN": {
+	"Vlan1000": {
+		"dhcp_servers": [
+			"192.0.0.1"
+		],
+		"members": [
+			"Ethernet0"
+		],
+		"vlanid": "1000",
+		“subnets”: “192.168.0.1/27”
+	}
+  }
+}
+<<End Json Style>>
+
+#### Manifest (if the feature is an Application Extension)
+
+N/A
+
+
+#### CLI/YANG model Enhancements 
+
+There is no change in CLI.
+
+#### Config DB Enhancements  
+
+Support a new member 'Subnets' of VLAN in config_db.
+
+
+### Warmboot and Fastboot Design Impact  
+This feature doesnt impact warmboot/fastboot. This feature is limited to dhcpv4 relay agent behaviour, which doesnt affect warmboot/fastboot. 
+
+### Memory Consumption
+No memory consumption is expected when the feature is disabled via compilation and no growing memory consumption while feature is disabled by configuration. 
+
+### Restrictions/Limitations  
+
+### Testing Requirements/Design  
+Explain what kind of unit testing, system testing, regression testing, warmboot/fastboot testing, etc.,
+Ensure that the existing warmboot/fastboot requirements are met. For example, if the current warmboot feature expects maximum of 1 second or zero second data disruption, the same should be met even after the new feature/enhancement is implemented. Explain the same here.
+Example sub-sections for unit test cases and system test cases are given below. 
+
+#### Unit Test cases  
+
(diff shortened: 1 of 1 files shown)


doc/DHCPv4_Gateway/DHCPv4_gateway.md
```diff
@@ -0,0 +1,127 @@
...
+"VLAN": {
+	"Vlan1000": {
+		"dhcp_servers": [
+			"192.0.0.1"
+		],
+		"members": [
+			"Ethernet0"
+		],
+		"vlanid": "1000",
+		“subnets”: “192.168.0.1/27”
```
venkatmahalingam, line 84: -> This change fixes this behaviour by explicitly specifying the primary interface/gateway
If your intention is to specify an IP address from the interface for the gateway address, why cant we just specify that particular IP address in the new field?
I dont understand the intention of subnets with just an IP and prefix len.

venkatmahalingam (changes requested):
Not sure about the solution to the problem, the dhcp relay code expects just the interface and IP address but the config from SONiC conflicts.

doc/DHCPv4_Gateway/DHCPv4_gateway.md
pushpraj: 'Subnets' of a vlan is a defined member in minigraph (eg [here](https://github.com/sonic-net/sonic-mgmt/blob/43b12cb120e00bf8beb41fad039b37de55e23e90/ansible/minigraph/switch-t0.xml#L205)). instead of adding a new field, we thought on using the same. This field can also be used in future for other purposes.

doc/DHCPv4_Gateway/DHCPv4_gateway.md
```diff
@@ -0,0 +1,127 @@
...
+### Overview 
+
+'Subnets' specify the primary subnet of a vlan (IPv4;IPv6). When a dhcpv4 packet is forwarded via dhcprelayagent, it embeds its own gateway address as 'giaddr' for return communication.
+In case of multiple interfaces, dhcprelayagent misconfigures this giaddr to any of the interfaces. 
+This change fixes this behaviour by explicitly specifying the primary interface/gateway.
+This primary subnet is stored in db as 'subnets' of VLAN.
+
+### Requirements
+
+1. Support a new member 'Subnets' of VLAN in config_db.
```
prsunny, line 31: Propose to modify INTERFACE_TABLE for specifying secondary IPs as this could extend the implementation for future use.

santoshdoke:
>> In case of multiple interfaces, dhcprelayagent misconfigures this giaddr to any of the interfaces.

Can you please explain more about this problem? DHCP relay is enabled on a given VLAN/Interface and the 'giaddr' is populated using the IP address assigned to that VLAN/interface. So where are the "multiple interfaces" coming from? Are you referring to a scenario whether a given VLAN/interface can have multiple IP addresses (one primary and several secondary)?

doc/DHCPv4_Gateway/DHCPv4_gateway.md
pushpraj: updated the HLD and PRs according to this suggestion.

pushpraj:
> > > In case of multiple interfaces, dhcprelayagent misconfigures this giaddr to any of the interfaces.
> 
> Can you please explain more about this problem? DHCP relay is enabled on a given VLAN/Interface and the 'giaddr' is populated using the IP address assigned to that VLAN/interface. So where are the "multiple interfaces" coming from? Are you referring to a scenario whether a given VLAN/interface can have multiple IP addresses (one primary and several secondary)?

Yes. this is the scenario where one VLAN can have multiple IP addresses. We have updated the design to mark such additional addresses as 'secondary' now.


> > > In case of multiple interfaces, dhcprelayagent misconfigures this giaddr to any of the interfaces.
> 
> Can you please explain more about this problem? DHCP relay is enabled on a given VLAN/Interface and the 'giaddr' is populated using the IP address assigned to that VLAN/interface. So where are the "multiple interfaces" coming from? Are you referring to a scenario whether a given VLAN/interface can have multiple IP addresses (one primary and several secondary)?

Yes.

doc/DHCPv4_Gateway/DHCPv4_gateway.md
```diff
@@ -0,0 +1,133 @@
...
+
+Extend CLI:
+
+	$ sudo config interface ip add eth0 20.11.12.13/27 20.11.12.254 --secondary
+	Usage: config interface ip add <interface_name> <ip_addr> <default gateway IP address> <secondary>
+
+Extend yang (sonic-vlan.yang):
+	list VLAN_INTERFACE_IPPREFIX_LIST {
+
+				leaf secondary	{
```
venkatmahalingam, line 100: Can we extend this support for all other supported L3 interfaces in DHCP relay?

doc/DHCPv4_Gateway/DHCPv4_gateway.md
```diff
@@ -0,0 +1,133 @@
...
+In case of multiple interfaces (one primary, multiple secondary), dhcprelayagent misconfigures this giaddr to any of the interfaces. 
+This change fixes this behaviour by explicitly specifying the primary interface/gateway.
+These secondary interfaces are attributed with 'secondary':'true' in config_db/yang/cli for VLAN_INTERFACE.
+
+### Requirements
+
+1. Support a new member 'secondary' of VLAN_INTERFACE in config_db.
+2. Support parsing and assignment of subnets from minigraph/json to config_db. 
+3. Support specifying non-secondary interfaces' gateway address to command line arguments to /usr/sbin/dhcrelay as -g.
+4. isc-dhcp/dhcrelay to support '-g gateway' argument, by porting an existing patch.
```
venkatmahalingam, line 36: Hope we handle any on-the-fly primary/secondary flag changes in CONFIG_DB for VLAN interface? i.e restarting the DHCP relay to pickup right primary IP address.

doc/DHCPv4_Gateway/DHCPv4_gateway.md
```diff
@@ -0,0 +1,133 @@
...
+### Configuration and management 
+(This section should have sub-sections for all types of configuration and management related design. Example sub-sections for "CLI" and "Config DB" are given below. Sub-sections related to data models (YANG, REST, gNMI, etc.,) should be added as required.)
+
+#### Minigraph 
+
+<<Redis DB Structure | Json Style>>
+{
+"VLAN_INTERFACE": {
+	"Vlan1000": {},
+	"Vlan1000|100.1.1.1/27": {"secondary": "true"},
```
venkatmahalingam, line 79: How does the upgrade work when VLAN_INTERFACES has more than one IP? we need to document the guideline for a user.

doc/DHCPv4_Gateway/DHCPv4_gateway.md
```diff
@@ -0,0 +1,133 @@
...
+#### Manifest (if the feature is an Application Extension)
+
+N/A
+
+
+#### CLI/YANG model Enhancements 
+
+Extend CLI:
+
+	$ sudo config interface ip add eth0 20.11.12.13/27 20.11.12.254 --secondary
```
venkatmahalingam, line 94: Looks like 'eth0' is used here instead of VLAN interface.

zhangyanzhao:
Community review recording https://zoom.us/rec/share/LiTYydJg8GJCntz_HL1DZk-sFQxMdcAA89Mj35l3zBAUYtLgxNR4XZyq4p9e_xqC.eAJyKMK62ORcaaFZ

doc/DHCPv4_Gateway/DHCPv4_gateway.md
pushpraj: This is not in scope for current design. It can be extended in future per the usecase.

doc/DHCPv4_Gateway/DHCPv4_gateway.md
pushpraj: config_db upgrade with this feature will not affect existing VLAN_INTERFACES, if any. This ensures that the feature is not enabled until user takes an action.

doc/DHCPv4_Gateway/DHCPv4_gateway.md
pushpraj: fixed.

doc/DHCPv4_Gateway/DHCPv4_gateway.md
pushpraj: Add relay address by cli would cause dhcp_relay container to restart https://github.com/sonic-net/sonic-buildimage/blob/c71fb3a30f3d384438f2e4861c80c28aea398fef/dockers/docker-dhcp-relay/cli/config/plugins/dhcp_relay.py#L77

pushpraj:
@venkatmahalingam can you please review this hld, and respective PRs too? thank you.

nmoray:
@pushpraj I tried back-porting this feature on 202111. I see that it is working.  But I observed one weird behaviour.  

Topology:  dhcp client <----> dhcp-relay  <-----> DHCP server
                                                   vlan 2:                    172.19.5.x
                                                    172.19.5.1/24
                                                     1.1.1.1/24 (secondary)

Issue: During the normal scenario it works fine as expected. But after reboot of the device, dhcp-relay honours the primary address and forwards the dhcp request via the correct subnet / gateway (172.19.5.x). But when it receives the dhcp reply from the server,  dhcp-relay forwards the IP back to the client via 1.1.1.x gateway (secondary) in place of 172.19.5.x. 

Are you also seeing similar behaviour?

Note: It is a default behaviour in SONIC, IPs under a VLAN gets sorted out and the one which is having the lowest numerical comes first.

zhangyanzhao:
@prsunny  @venkatmahalingam  can you please help to approve this PR if you are ok? Thanks.

gord1306:
I am wondering if it's possible to specify an IP address as the primary address for the DHCP relay service. In the current design, the 'secondary' parameter's meaning is applied under the interface, encompassing the entire interface view. However, this feature is intended only for the DHCP relay, not for the entire interface

zhangyanzhao:
@prsunny  @venkatmahalingam  are you ok to merge this PR? Thanks.

prsunny:
yes @zhangyanzhao , we can merge this PR

venkatmahalingam (approved):

prsunny (approved):

PR #1471, opened 2023-09-13
TACACSPLUS_PASSKEY_ENCRYPTION.md

madhupalu:
Adding support for TACACS+ passkey encryption for SONiC deployments.

Related PRs:
+++++++++
1. Security Cipher module - https://github.com/sonic-net/sonic-buildimage/pull/17201
2. Encryption using security cipher module - https://github.com/sonic-net/sonic-utilities/pull/3027
3. Decryption using security cipher module - https://github.com/sonic-net/sonic-host-services/pull/81

### DIFF
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -105,7 +105,12 @@ show tacacs
 
 TACACS passkey encryption adds an extra layer of security to safeguard the passkey on each device throughout the network. Moreover, the utilization of MAC address-based encryption ensures that each network device possesses its distinct encrypted passkey. This strategy effectively mitigates the risk of a major network breach in case one of the devices is compromised.
 
+### Limitation
+We choose MAC address is unique and better for management, however if the MAC address of the device is known, we could decrypt the key. If we want to use any other ciper we should manage / stored it in the DB, it is again defeating the purpose. So for now we are going ahead with mac as network admins can orchestartion system to use mac address for passkey encryption. However we are open to new ideas and thoughts to hardened the security and we should be able to accomodate with later releases.
 
 ### Testing Requirements
 
 Need to add new / update the existing TACACS testcases to incorporate this new feature
+Test cases to unit test encrypt and decrypt fucntions 
+Test cases to add test the TACACS+ functionality with passkey encryption
+Test cases to cover DB migration 
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -1,0 +1,111 @@
+# TACACS+ Passkey Encryption #
+
+
+## Table of Contents
+
+- [Revision](#revision)
+- [Scope](#scope)
+- [Abbreviations](#abbreviations)
+- [Overview](#overview)
+- [Requirements](#requirements)
+- [High-Level Design](#high-level-design)
+- [Implementation Details](#implementation-details)
+        - [Show CLI changes](@show-cli-changes)
+- [Benifits](#benifits)
+- [Testing Requirements](#testing-requirements)
+
+
+
+### Revision
+
+ | Rev |     Date    |       Author         | Change Description                |
+ |:---:|:-----------:|:--------------------:|-----------------------------------|
+ | 0.1 |             | Nikhil Moray (nmoray)| Initial version                   |
+ | 0.1 |             | Madhu Paluru (madhupalu)| Updated                        |
+
+
+ ### Scope
+
+This document describes the High Level Design of "TACACS+ Passkey Encryption" feature support in SONiC. It encompasses design and implementation considerations for enhancing the security of TACACS+ (Terminal Access Controller Access-Control System) authentication by introducing an encrypted passkey.
+
+
+### Abbreviations
+
+ | Term    |     Meaning                                                        |
+ |:-------:|:-------------------------------------------------------|
+ | TACACS  | Terminal Access Controller Access Control System Plus) |
+
+### Overview
+
+This addition constitutes a substantial improvement in bolstering the security of the TACACS+ authentication protocol. TACACS+ has a well-established reputation as a reliable means of managing access to network devices. However, the previous practice of utilising a sensitive passkey in plaintext during the authentication process posed security concerns. With this enhancement, the vulnerability is effectively mitigated by introducing robust passkey encryption mechanisms (on the client side), ensuring the safeguarding of authentication credentials and an overall strengthening of network security."
+
+
+### Requirements
+
+The primary objective of this feature is to safeguard the TACACS passkey, which is stored in its plaintext format within CONFIG_DB.
+
+
+### High-Level Design
+
+In line with the TACACS+ architecture, when a user initiates a SSH connection to a device with TACACS+ authentication enabled, they must utilize the TACACS+ login password. Conversely, the corresponding device must have the passkey provided by the TACACS+ server properly configured within its configDB. Should either of these elements be missing or incorrect, the user will be unable to access the device. Thus to meet the given requirement, passkey will be encrypted is the configuration phase itself.
+
+The current data flow among the various TACACS modules operates in the following manner.
+
+1. When a user configures the TACACS passkey using the SONIC CLI, it is initially stored in the CONFIG_DB.
+2. Subsequently, the same key is retrieved by the HostCfg Enforcer module to update the PAM configuration file(s). This configuration file is inherently included in the authentication processes for login or SSH within the Linux operating system. 
+3. When TACACS+ Authentication is activated on the device, a new PAM configuration file (common-auth-sonic) is generated and substituted in the login and SSH daemons. Importantly, the pre-existing configuration file remains unchanged.
+
+The revised data handling procedure among the modules is outlined as follows:
+
+1. When a user configures the TACACS passkey using the SONIC CLI, it will now be securely stored in encrypted format instead of plaintext.
+2. Subsequently, the HostCfg Enforcer module retrieves this encrypted key. However, before writing it into the PAM configuration file(s), the hostCfgd module decrypts it. 
+```
+       +-------+  +---------+
+       |  SSH  |  | Console |
+       +---+---+  +----+----+
+           |           |   
++----------v-----------v---------+                                      +---------------------+
+| AUTHENTICATION                 |                                      |                     |
+|   +-------------------------+  |  Decrypted passkey                   |  +------------+     |
+|   | PAM Configuration Files <------------------------------------------+ |  AAA Config |    |
+|   +-------------------------+  |                                      |  +------------+     |
+|                                |                                      |                     |
+|         +-------------+        |                                      |  HostCfg Enforcer   |
+|         |     PAM     |        |                                      +----------^----------+
+|         |  Libraries  |        |                                                 |
+|         +-------------+        |                                                 | Encrypted passkey
++---------------+----------------+                                                 |
+                |                                                                  |
+           +----v----+                                                     +-------+--------+
+           |         |             Encrypted passkey                       |                |
+           |   CLI   +---------------------------------------------------->|    ConifgDB    |
+           |         |                                                     |                |
+           +---------+                                                     +----------------+
+```
+This decryption step is crucial because the login or SSH daemon references the PAM config file to verify the TACACS secret / passkey. If it remains encrypted, the SSH daemon will be unable to recognize the passkey, leading to login failures. The depicted block diagram clearly showcase the enhanced capbalities of the existing submodules.
+
+
+### Implementation details
+
+The implementation stands on three key pillars.
+1. OPENSSL toolkit is used for encryption / decryption
+2. aes-128-cbc is the encoding format used for encryption / decryption
+3. Device MAC address is used as a Password for encryption / decryption
+
+
+#### Show CLI changes
+
+Furthermore, aside from encrypting the passkey stored within CONFIG_DB, this infrastructure ensures that the passkey itself remains concealed from any of the displayed CLI outputs. Consequently, the passkey field has been eliminated from the "show tacacs" output, and it will now solely indicate the status whether the passkey is configured or not. For instance.
+
+show tacacs
+["TACPLUS global passkey configured Yes / No"]
+
+
+### Benifits
+
+TACACS passkey encryption adds an extra layer of security to safeguard the passkey on each device throughout the network. Moreover, the utilization of MAC address-based encryption ensures that each network device possesses its distinct encrypted passkey. This strategy effectively mitigates the risk of a major network breach in case one of the devices is compromised.
+
+
+### Testing Requirements
+
+Need to add new / update the existing TACACS testcases to incorporate this new feature
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -90,7 +90,7 @@ This decryption step is crucial because the login or SSH daemon references the P
 The implementation stands on three key pillars.
 1. OPENSSL toolkit is used for encryption / decryption
 2. aes-128-cbc is the encoding format used for encryption / decryption
-3. Device MAC address is used as a Password for encryption / decryption
+3. A unique Device MAC address used as a salt/password to encrypt/decrypt the configured pass key 
 
 
 #### Show CLI changes
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -100,6 +100,8 @@ Furthermore, aside from encrypting the passkey stored within CONFIG_DB, this inf
 show tacacs
 ["TACPLUS global passkey configured Yes / No"]
 
+### DB migration 
+A DB migration script will be added for users to migrate existing config_db to convert tacacs passkey plaintext to encypted. 
 
 ### Benifits
 
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -90,6 +90,7 @@ show tacacs
 
 ### Yang Changes 
 Increase existing passkey leaf length to 256.
+Create a new leaf for newly introduced run time flag 'key_encrypt'.
 
 ### Config DB changes 
 A new run time flag to enable/disable the tacacs passkey encryption feature - "key_encrypt".
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -54,7 +54,7 @@ The revised data handling procedure among the modules is outlined as follows:
                 |                     |    +-------------+    |o |                 |
            +----v----+                |           |           |t |         +-------+--------+
            |         |                |           |            --          |                |
-           |   CLI   +----------------+           +------------------------>    ConifgDB    |
+           |   CLI   +----------------+           +------------------------>    ConfigDB    |
            |         |                                  Encrypted passkey  |                |
            +---------+                                                     +----------------+
 ```
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -74,14 +74,14 @@ The implementation as follows
 4. sonic_utilities extended to passkey encyption using the master key/passwd manager.
 5. User has to enter master key/passwd at the time of configuring the passkey, this is mandatory requirement only if "key_encrypt" run time flag is enabled.
 6. The encrypted passkey stored in config_db 
-7. The master key/paswd used for encryption/decryption and will be stored in the same device with root access previleges.
+7. The master key/paswd used for encryption/decryption and will be stored in the same device with root access previleges (/etc/cipher_pass).
 8. HostCfg will use the master key/passwd to decrypt the encrypted passkey and further store it in PAM configuration files. 
 
 #### CLI Changes 
-config tacacs passkey TEST1
-
+config tacacs passkey TEST1 --encrypt
 Password: 
-Note: It will ask for a master key/password only when the 'key_encrypt' flag is set under respective feature table (ex: TACPLUS) in config_db.
+
+Note: It will ask for a master key/password only when '--encrypt' flag is set.
 
 #### Show CLI changes
 Furthermore, aside from encrypting the passkey stored within CONFIG_DB, this infrastructure ensures that the passkey itself remains concealed from any of the displayed CLI outputs. Consequently, the passkey field has been eliminated from the "show tacacs" output, and it will now solely indicate the status whether the passkey is configured or not. For instance,
@@ -95,6 +95,16 @@ Create a new leaf for newly introduced run time flag 'key_encrypt'.
 ### Config DB changes 
 A new run time flag to enable/disable the tacacs passkey encryption feature - "key_encrypt".
 
+### Schema changes 
+```
+"TACPLUS": {
+        "global": {
+            "auth_type": "login",
+            "key_encrypt": "true",
+            "passkey": "<Entrypted_Passkey>"
+        }
+    }
+```
 ### Benefits
 TACACS passkey encryption adds an extra layer of security to safeguard the passkey on each device throughout the network. Furthermore, the implementation of master key/password manager encryption ensures that encrypted passkeys can be reused across network nodes without any complications. Consequently, there are no obstacles when it comes to utilizing the config_db.json file from one device on another. Additionally, the use of a root protected config file effectively reduces the risk of exposing the encryption/decryption master key/passwd since it is only accessible to root users and remains inaccessible to external entities.
 
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -58,6 +58,17 @@ The revised data handling procedure among the modules is outlined as follows:
            +---------+                                                     +----------------+
 ```
 This decryption step is crucial because the login or SSH daemon references the PAM config file to verify the TACACS secret / passkey. If it remains encrypted, the SSH daemon will be unable to recognize the passkey, leading to login failures. The depicted block diagram clearly showcases the enhanced capabilities of the existing submodules.
+
+### Approach 1:
+1. A runtime flag (via config_db) for Enable / disable feature: "key_encrypt"
+2. Use admin password from shadow file to encrypt the TACACS passkey
+3. Use the same password to decrypt the TACACS paskey
+
+### Approach 2:
+1. A runtime flag (via config_db) for Enable / disable feature: "key_encrypt"
+2. CLI will ask for a encryption password while configuring the TACACS passkey and at the backend it will be stored at /etc/encrypt_pass file
+3. Same file will be read while decrypting the passkey at hostcfgd
+   
 ### Implementation details
 The implementation stands on four key pillars.
 1. OPENSSL toolkit is used for encryption / decryption.
@@ -68,11 +79,13 @@ Following is the snippet captured from a sample shadow file.
 <snip_from_shadow_file>
 admin:$6$YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1:19446:0:99999:7:::
 <snip_from_shadow_file>
-Here, "YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1" is the encrypted version of the admin level password. 
+Here, "YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1" is the encrypted version of the admin level password.
+
 #### Show CLI changes
 Furthermore, aside from encrypting the passkey stored within CONFIG_DB, this infrastructure ensures that the passkey itself remains concealed from any of the displayed CLI outputs. Consequently, the passkey field has been eliminated from the "show tacacs" output, and it will now solely indicate the status whether the passkey is configured or not. For instance,
 show tacacs
 ["TACPLUS global passkey configured Yes / No"]
+
 ### DB migration
 A DB migration script will be added for users to migrate existing config_db to convert tacacs passkey plaintext to encrypted.
 ### Benefits
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -46,13 +46,13 @@ The revised data handling procedure among the modules is outlined as follows:
 |   | PAM Configuration Files    <------------+       +-----------------+  | AAA Config |     |
 |   +-------------------------+  |            |       |                 |  +------------+     |
 |                                |            |       |                 |                     |
-|         +-------------+        |         +-------------+              |  HostCfg Enforcer   |
-|         |     PAM     |        |         |             |   key-store  +----------^----------+
-|         |  Libraries  |        |    +---->  Master key |     __                  |
-|         +-------------+        |    |    |   Manager   |----|r |                 | Encrypted passkey
-+---------------+----------------+    |    |             |    |o |                 |
-                |                     |    +-------------+    |o |                 |
-           +----v----+                |           |           |t |         +-------+--------+
+|         +-------------+        |         +--------------+             |  HostCfg Enforcer   |
+|         |     PAM     |        |         |              |   key-store +----------^----------+
+|         |  Libraries  |        |    +---->  Master key  |    __                 |
+|         +-------------+        |    |    |   Manager    |---|r |                | Encrypted passkey
++---------------+----------------+    |    |securitycipher|   |o |                |
+                |                     |    +-------------+    |o |                |
+           +----v----+                |           |           |t |         +------+--------+
            |         |                |           |            --          |                |
            |   CLI   +----------------+           +------------------------>    ConfigDB    |
            |         |                                  Encrypted passkey  |                |
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -1,8 +1,5 @@
 # TACACS+ Passkey Encryption #
-
-
 ## Table of Contents
-
 - [Revision](#revision)
 - [Scope](#scope)
 - [Abbreviations](#abbreviations)
@@ -11,54 +8,32 @@
 - [High-Level Design](#high-level-design)
 - [Implementation Details](#implementation-details)
         - [Show CLI changes](@show-cli-changes)
-- [Benifits](#benifits)
+- [Benefits](#benefits)
 - [Testing Requirements](#testing-requirements)
-
-
-
 ### Revision
-
- | Rev |     Date    |       Author         | Change Description                |
- |:---:|:-----------:|:--------------------:|-----------------------------------|
- | 0.1 |             | Nikhil Moray (nmoray)| Initial version                   |
- | 0.1 |             | Madhu Paluru (madhupalu)| Updated                        |
-
-
+ | Rev |     Date    |       Author            | Change Description                |
+ |:---:|:-----------:|:-----------------------:|:----------------------------------|
+ | 0.1 |             | Nikhil Moray (nmoray)   | Initial version                   |
+ | 0.1 |             | Madhu Paluru (madhupalu)| Updated                           |
  ### Scope
-
 This document describes the High Level Design of "TACACS+ Passkey Encryption" feature support in SONiC. It encompasses design and implementation considerations for enhancing the security of TACACS+ (Terminal Access Controller Access-Control System) authentication by introducing an encrypted passkey.
-
-
 ### Abbreviations
-
- | Term    |     Meaning                                                        |
+ | Term    |     Meaning                                            |
  |:-------:|:-------------------------------------------------------|
  | TACACS  | Terminal Access Controller Access Control System Plus) |
-
 ### Overview
-
-This addition constitutes a substantial improvement in bolstering the security of the TACACS+ authentication protocol. TACACS+ has a well-established reputation as a reliable means of managing access to network devices. However, the previous practice of utilising a sensitive passkey in plaintext during the authentication process posed security concerns. With this enhancement, the vulnerability is effectively mitigated by introducing robust passkey encryption mechanisms (on the client side), ensuring the safeguarding of authentication credentials and an overall strengthening of network security."
-
-
+This addition constitutes a substantial improvement in bolstering the security of the TACACS+ authentication protocol. TACACS+ has a well-established reputation as a reliable means of managing access to network devices. However, the previous practice of utilising a sensitive passkey in plaintext during the authentication process posed security concerns. With this enhancement, the vulnerability is effectively mitigated by introducing robust passkey encryption mechanisms (on the client side), ensuring the safeguarding of authentication credentials and an overall strengthening of network security.
 ### Requirements
-
 The primary objective of this feature is to safeguard the TACACS passkey, which is stored in its plaintext format within CONFIG_DB.
-
-
 ### High-Level Design
-
-In line with the TACACS+ architecture, when a user initiates a SSH connection to a device with TACACS+ authentication enabled, they must utilize the TACACS+ login password. Conversely, the corresponding device must have the passkey provided by the TACACS+ server properly configured within its configDB. Should either of these elements be missing or incorrect, the user will be unable to access the device. Thus to meet the given requirement, passkey will be encrypted is the configuration phase itself.
-
+In line with the TACACS+ architecture, when a user initiates a SSH connection to a device with TACACS+ authentication enabled, they must utilize the TACACS+ login password. Conversely, the corresponding device must have the passkey provided by the TACACS+ server properly configured within its configDB. Should either of these elements be missing or incorrect, the user will be unable to access the device. Thus to meet the given requirement, the passkey will be encrypted in the configuration phase itself.
 The current data flow among the various TACACS modules operates in the following manner.
-
 1. When a user configures the TACACS passkey using the SONIC CLI, it is initially stored in the CONFIG_DB.
-2. Subsequently, the same key is retrieved by the HostCfg Enforcer module to update the PAM configuration file(s). This configuration file is inherently included in the authentication processes for login or SSH within the Linux operating system. 
+2. Subsequently, the same key is retrieved by the HostCfg Enforcer module to update the PAM configuration file(s). This configuration file is inherently included in the authentication processes for login or SSH within the Linux operating system.
 3. When TACACS+ Authentication is activated on the device, a new PAM configuration file (common-auth-sonic) is generated and substituted in the login and SSH daemons. Importantly, the pre-existing configuration file remains unchanged.
-
 The revised data handling procedure among the modules is outlined as follows:
-
 1. When a user configures the TACACS passkey using the SONIC CLI, it will now be securely stored in encrypted format instead of plaintext.
-2. Subsequently, the HostCfg Enforcer module retrieves this encrypted key. However, before writing it into the PAM configuration file(s), the hostCfgd module decrypts it. 
+2. Subsequently, the HostCfg Enforcer module retrieves this encrypted key. However, before writing it into the PAM configuration file(s), the hostCfgd module decrypts it.
 ```
        +-------+  +---------+
        |  SSH  |  | Console |
@@ -67,7 +42,7 @@ The revised data handling procedure among the modules is outlined as follows:
 +----------v-----------v---------+                                      +---------------------+
 | AUTHENTICATION                 |                                      |                     |
 |   +-------------------------+  |  Decrypted passkey                   |  +------------+     |
-|   | PAM Configuration Files <------------------------------------------+ |  AAA Config |    |
+|   | PAM Configuration Files <-----------------------------------------+  | AAA Config |     |
 |   +-------------------------+  |                                      |  +------------+     |
 |                                |                                      |                     |
 |         +-------------+        |                                      |  HostCfg Enforcer   |
@@ -78,41 +53,35 @@ The revised data handling procedure among the modules is outlined as follows:
                 |                                                                  |
            +----v----+                                                     +-------+--------+
            |         |             Encrypted passkey                       |                |
-           |   CLI   +---------------------------------------------------->|    ConifgDB    |
+           |   CLI   +------------------------------------------------------>    ConifgDB   |
            |         |                                                     |                |
            +---------+                                                     +----------------+
 ```
-This decryption step is crucial because the login or SSH daemon references the PAM config file to verify the TACACS secret / passkey. If it remains encrypted, the SSH daemon will be unable to recognize the passkey, leading to login failures. The depicted block diagram clearly showcase the enhanced capbalities of the existing submodules.
-
-
+This decryption step is crucial because the login or SSH daemon references the PAM config file to verify the TACACS secret / passkey. If it remains encrypted, the SSH daemon will be unable to recognize the passkey, leading to login failures. The depicted block diagram clearly showcases the enhanced capabilities of the existing submodules.
 ### Implementation details
-
-The implementation stands on three key pillars.
-1. OPENSSL toolkit is used for encryption / decryption
-2. aes-128-cbc is the encoding format used for encryption / decryption
-3. A unique Device MAC address used as a salt/password to encrypt/decrypt the configured pass key 
-
-
+The implementation stands on four key pillars.
+1. OPENSSL toolkit is used for encryption / decryption.
+2. aes-128-cbc is the encoding format used for encryption / decryption.
+3. A unique device admin password (encrypted hash from linux shadow file) will be used as a salt/password to encrypt/decrypt the configured pass key in ConfigDB.
+4. In absence of configured admin password (default behaviour), a fixed / hardcoded hash will be used as a salt/password for encryption/decryption and it will be saved locally not in any of the databases.
+Following is the snippet captured from a sample shadow file.
+<snip_from_shadow_file>
+admin:$6$YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1:19446:0:99999:7:::
+<snip_from_shadow_file>
+Here, "YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1" is the encrypted version of the admin level password. 
 #### Show CLI changes
-
-Furthermore, aside from encrypting the passkey stored within CONFIG_DB, this infrastructure ensures that the passkey itself remains concealed from any of the displayed CLI outputs. Consequently, the passkey field has been eliminated from the "show tacacs" output, and it will now solely indicate the status whether the passkey is configured or not. For instance.
-
+Furthermore, aside from encrypting the passkey stored within CONFIG_DB, this infrastructure ensures that the passkey itself remains concealed from any of the displayed CLI outputs. Consequently, the passkey field has been eliminated from the "show tacacs" output, and it will now solely indicate the status whether the passkey is configured or not. For instance,
 show tacacs
 ["TACPLUS global passkey configured Yes / No"]
-
-### DB migration 
-A DB migration script will be added for users to migrate existing config_db to convert tacacs passkey plaintext to encypted. 
-
-### Benifits
-
-TACACS passkey encryption adds an extra layer of security to safeguard the passkey on each device throughout the network. Moreover, the utilization of MAC address-based encryption ensures that each network device possesses its distinct encrypted passkey. This strategy effectively mitigates the risk of a major network breach in case one of the devices is compromised.
-
+### DB migration
+A DB migration script will be added for users to migrate existing config_db to convert tacacs passkey plaintext to encrypted.
+### Benefits
+TACACS passkey encryption adds an extra layer of security to safeguard the passkey on each device throughout the network. Furthermore, the implementation of shadow file-based encryption ensures that encrypted passkeys can be reused across network nodes without any complications. Consequently, there are no obstacles when it comes to utilizing the config_db.json file from one device on another. Additionally, the use of a shadow file effectively reduces the risk of exposing the encryption/decryption salt since it is only accessible to root users and remains inaccessible to external entities.
 ### Limitation
-We choose MAC address is unique and better for management, however if the MAC address of the device is known, we could decrypt the key. If we want to use any other ciper we should manage / stored it in the DB, it is again defeating the purpose. So for now we are going ahead with mac as network admins can orchestartion system to use mac address for passkey encryption. However we are open to new ideas and thoughts to hardened the security and we should be able to accomodate with later releases.
-
+The chosen way to encrypt the passkey is using an already encrypted admin password from the shadow file. Thus, the network admin needs to regenerate the encrypted passkey in ConfigDB only if there is a change in the admin password of the network.
+For now we are planning to go ahead with this approach. However we are open to new ideas and thoughts to harden the security and we should be able to accommodate that with later releases.
 ### Testing Requirements
-
-Need to add new / update the existing TACACS testcases to incorporate this new feature
-Test cases to unit test encrypt and decrypt fucntions 
+Need to add new / update the existing TACACS test cases to incorporate this new feature
+Test cases to unit test encrypt and decrypt functions
 Test cases to add test the TACACS+ functionality with passkey encryption
-Test cases to cover DB migration 
+Test cases to cover DB migration
--- a/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
+++ b/doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
@@ -15,6 +15,7 @@
  |:---:|:-----------:|:-----------------------:|:----------------------------------|
  | 0.1 |             | Nikhil Moray (nmoray)   | Initial version                   |
  | 0.1 |             | Madhu Paluru (madhupalu)| Updated                           |
+ | 0.1 |   11/09/2023| Madhu Paluru (madhupalu)| Addressed review comments         |
  ### Scope
 This document describes the High Level Design of "TACACS+ Passkey Encryption" feature support in SONiC. It encompasses design and implementation considerations for enhancing the security of TACACS+ (Terminal Access Controller Access-Control System) authentication by introducing an encrypted passkey.
 ### Abbreviations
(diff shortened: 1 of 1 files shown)


doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,111 @@
...
+```
+This decryption step is crucial because the login or SSH daemon references the PAM config file to verify the TACACS secret / passkey. If it remains encrypted, the SSH daemon will be unable to recognize the passkey, leading to login failures. The depicted block diagram clearly showcase the enhanced capbalities of the existing submodules.
+
+
+### Implementation details
+
+The implementation stands on three key pillars.
+1. OPENSSL toolkit is used for encryption / decryption
+2. aes-128-cbc is the encoding format used for encryption / decryption
+3. Device MAC address is used as a Password for encryption / decryption
```
venkatmahalingam, line 93: We can easily decrypt and know the actual password if we know the device MAC address, right?

Also, if I want to use the config_db (encrypted passkey) generated on one system to another (may be I forced to replace the failed node with same config), hope passkey cant be used on another device, what procedure should be followed for this use-case?

nmoray: Yes @venkatmahalingam  you are right. We can decrypt the passkey only if we knew the system MAC address. And if  we allow the user to use any custom password then it will be bit difficult in terms of maintainability. 

For instance, either we need to use some fixed cipher for encryption / decryption which will be more susceptible to the overall network breach. Alternatively,  we need to save the password somewhere (for instance in redis). This is the reason I chosen system MAC as a password which will make every encrypted passkey unique within the network.

An encrypted passkey on one node will not be applicable or useable on another as hostcfgd will not be able to decrypt that passkey. To handle this sort of scenario, the Orchestrator / network admin has to create the config_db.json or a encrypted passkey for every node (considering the orchestrator have the list of system MAC of all the nodes).

madhupalu:
Adding @Yarden-Z as reviewer who is interested in this feature. @venkatmahalingam @Yarden-Z thanks for reviewing and please let us know we will work on the code PRs.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,111 @@
...
+ | TACACS  | Terminal Access Controller Access Control System Plus) |
+
+### Overview
+
+This addition constitutes a substantial improvement in bolstering the security of the TACACS+ authentication protocol. TACACS+ has a well-established reputation as a reliable means of managing access to network devices. However, the previous practice of utilising a sensitive passkey in plaintext during the authentication process posed security concerns. With this enhancement, the vulnerability is effectively mitigated by introducing robust passkey encryption mechanisms (on the client side), ensuring the safeguarding of authentication credentials and an overall strengthening of network security."
+
+
+### Requirements
+
+The primary objective of this feature is to safeguard the TACACS passkey, which is stored in its plaintext format within CONFIG_DB.
```
venkatmahalingam, line 45: Can you add the section for DB migrator?

venkatmahalingam: Can you add the YANG change for the new TACACS passkey?

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,111 @@
...
+3. Device MAC address is used as a Password for encryption / decryption
+
+
+#### Show CLI changes
+
+Furthermore, aside from encrypting the passkey stored within CONFIG_DB, this infrastructure ensures that the passkey itself remains concealed from any of the displayed CLI outputs. Consequently, the passkey field has been eliminated from the "show tacacs" output, and it will now solely indicate the status whether the passkey is configured or not. For instance.
+
+show tacacs
+["TACPLUS global passkey configured Yes / No"]
+
```
venkatmahalingam, line 102: Add a section for limitations/restrictions, mention that if the MAC address of the device is known, we could decrypt the key and add some thoughts for the future work if you are not planning to fix in the initial release.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
madhupalu: @venkatmahalingam passkey is already part of yang model, no new schema introduced part of HLD - https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-system-tacacs.yang
Only the backend will use system mac as unique salt to encrypt/decrypt the passkey.

madhupalu: We don't need DB migrator as no change to the schema.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
madhupalu: Addressed. Added a section to the HLD describing the limitations and release plan.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
venkatmahalingam: How about existing plaintext passkey to encrypted key migration in the config-db during image upgrade?

venkatmahalingam: We should update the YANG to reflect the plaintext to encrypt key change in the passkey description.

madhupalu: @venkatmahalingam ok, will take care of yang model change and added db migration section in the HLD.

madhupalu: @venkatmahalingam The current yang model **description already reflect the change you requested** - '   description "Shared secret used for encrypting the communication";'  Is the sufficient or do we think still need an update? 

 leaf passkey {
                    type string {
                        length "1..65";
                        pattern "[^ #,]*" {
                            error-message 'TACACS shared secret (Valid chars are ASCII printable except SPACE, "#", and ",")';
                        }
                    }
                    description "Shared secret used for encrypting the communication";
                }

a-barboza:
cc: @shdasari @adyeung

The idea of encrypting the TACACS+ shared secret before it is written to the configuration (CONFIG_DB) is good.

However,  the drawbacks are: (one of more of these might already be mentioned in the limitations)

- The key encryption key is not secret because it is the MAC address of the device.
- It cannot be changed, therefore the admin has no control over its secrecy.
- The admin cannot use CONFIG_DB snippets unmodified on other devices, because they are keyed by MAC address.
- The mechanism of encryption is not described. There could be some concern with using the key encryption key (MAC address) directly if that is indeed the intent. Is that the intent?

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: @venkatmahalingam is [sonic-system-tacacs.yang](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-system-tacacs.yang) a pull model or push? If we need to configure the passkey via this model, we don't need any change as it will still be in a plaintext format only.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: @venkatmahalingam the reason behind choosing system MAC as the encryption salt is, it will be known only to the network admin. Additionally, it will be readily accessible within the Redis so, can be pulled in different modules without hardcoding anything.

Moreover, we could consider other alternate salt too but again it will be an overhead of creating that uniquely for every node and also maintaining the same some where so that it can be consumed during the orchestration process.

nmoray:
> * The mechanism of encryption is not described.



> cc: @shdasari @adyeung
> 
> The idea of encrypting the TACACS+ shared secret before it is written to the configuration (CONFIG_DB) is good.
> 
> However, the drawbacks are: (one of more of these might already be mentioned in the limitations)
> 
> * The key encryption key is not secret because it is the MAC address of the device.
[nmoray]--> I think I confused you. Let me clarify. The passkey used here is not the system MAC address. The MAC is used an a salt to encrypt the unique passkey (which is in plaintext).
> * It cannot be changed, therefore the admin has no control over its secrecy.
[nmoray]--> The encrypted passkey key will be unique for every node so this design will be less susceptible to the overall network breach. 
> * The admin cannot use CONFIG_DB snippets unmodified on other devices, because they are keyed by MAC address.
[nmoray]--> Yes, that is correct. But the easy way out is, network admin will simply execute "config tacacs passkey <actual_passkey_in_plaintext>" command after copying the CONFIG_DB snippets on a given device so that a new unique encrypted passkey will be generated
> * The mechanism of encryption is not described. There could be some concern with using the key encryption key (MAC address) directly if that is indeed the intent. Is that the intent?
[nmoray]--> Please refer to the "Implementation details" section where we have explained the encryption details. Can you please elaborate bit more about those concerns? 
The implementation is straight forward. An encryption API which will take two inputs, passkey in plaintext and the system MAC. It will use openssl lib to encrypt the provided passkey and use system MAC as a salt.


@a-barboza the answers are inline.

a-barboza:
> > * The mechanism of encryption is not described.
> 
> > cc: @shdasari @adyeung
> > The idea of encrypting the TACACS+ shared secret before it is written to the configuration (CONFIG_DB) is good.
> > However, the drawbacks are: (one of more of these might already be mentioned in the limitations)
> > 
> > * The key encryption key is not secret because it is the MAC address of the device.
> >   [nmoray]--> I think I confused you. Let me clarify. The passkey used here is not the system MAC address. The MAC is used an a salt to encrypt the unique passkey (which is in plaintext).
> > * It cannot be changed, therefore the admin has no control over its secrecy.
> >   [nmoray]--> The encrypted passkey will be unique for every node so this design will be less susceptible to the overall network breach.
> > * The admin cannot use CONFIG_DB snippets unmodified on other devices, because they are keyed by MAC address.
> >   [nmoray]--> Yes, that is correct. But the easy way out is, network admin will simply execute "config tacacs passkey <actual_passkey_in_plaintext>" command after copying the CONFIG_DB snippets on a given device so that a new unique encrypted passkey will be generated
> > * The mechanism of encryption is not described. There could be some concern with using the key encryption key (MAC address) directly if that is indeed the intent. Is that the intent?
> >   [nmoray]--> Please refer to the "Implementation details" section where we have explained the encryption details. Can you please elaborate bit more about those concerns?
> >   The implementation is straight forward. An encryption API which will take two inputs, passkey in plaintext and the system MAC. It will use openssl lib to encrypt the provided passkey and use system MAC as a salt.
> 
> @a-barboza the replies are inline.

[a-barboza] So, if the system MAC address is used as the salt, what is the key used for encryption ?  IMHO, the salt has to be unique or pseudo-random (even if it is not secret).

[a-barboza] Even if the encrypted passkey is unique for every node, how can it be changed by the admin? Most admins will not like to have their passwords assigned to them. They would rather choose their passwords. 

[a-barboza] ok, so the workaround is to use a CLI in a script to set the TACACS+ encrypted passkeys. What's the workaround if the admin would like to generate the encrypted passkeys remotely and just give a pre-cooked CONFIG_DB json for configuration?

[a-barboza] The Implementation details mentions "A unique Device MAC address used as a salt/password to encrypt/decrypt the configured pass key". So, it is not clear if the device MAC address is used as the salt (initialization vector), key, or both. So, if there are multiple TACACS+ passkeys, will the same salt be used for all of them?

nmoray:
> > > * The mechanism of encryption is not described.
> > 
> > 
> > > cc: @shdasari @adyeung
> > > The idea of encrypting the TACACS+ shared secret before it is written to the configuration (CONFIG_DB) is good.
> > > However, the drawbacks are: (one of more of these might already be mentioned in the limitations)
> > > 
> > > * The key encryption key is not secret because it is the MAC address of the device.
> > >   [nmoray]--> I think I confused you. Let me clarify. The passkey used here is not the system MAC address. The MAC is used an a salt to encrypt the unique passkey (which is in plaintext).
> > > * It cannot be changed, therefore the admin has no control over its secrecy.
> > >   [nmoray]--> The encrypted passkey will be unique for every node so this design will be less susceptible to the overall network breach.
> > > * The admin cannot use CONFIG_DB snippets unmodified on other devices, because they are keyed by MAC address.
> > >   [nmoray]--> Yes, that is correct. But the easy way out is, network admin will simply execute "config tacacs passkey <actual_passkey_in_plaintext>" command after copying the CONFIG_DB snippets on a given device so that a new unique encrypted passkey will be generated
> > > * The mechanism of encryption is not described. There could be some concern with using the key encryption key (MAC address) directly if that is indeed the intent. Is that the intent?
> > >   [nmoray]--> Please refer to the "Implementation details" section where we have explained the encryption details. Can you please elaborate bit more about those concerns?
> > >   The implementation is straight forward. An encryption API which will take two inputs, passkey in plaintext and the system MAC. It will use openssl lib to encrypt the provided passkey and use system MAC as a salt.
> > 
> > 
> > @a-barboza the replies are inline.
> 
> [a-barboza] So, if the system MAC address is used as the salt, what is the key used for encryption ? IMHO, the salt has to be unique or pseudo-random (even if it is not secret).
> 
> [a-barboza] Even if the encrypted passkey is unique for every node, how can it be changed by the admin? Most admins will not like to have their passwords assigned to them. They would rather choose their passwords.
> 
> [a-barboza] ok, so the workaround is to use a CLI in a script to set the TACACS+ encrypted passkeys. What's the workaround if the admin would like to generate the encrypted passkeys remotely and just give a pre-cooked CONFIG_DB json for configuration?
> 
> [a-barboza] The Implementation details mentions "A unique Device MAC address used as a salt/password to encrypt/decrypt the configured pass key". So, it is not clear if the device MAC address is used as the salt (initialisation vector), key, or both. So, if there are multiple TACACS+ passkeys, will the same salt be used for all of them?

[nmoray]  Yes, the salt will be system MAC address and passkey will be any user defined secret string. This combination always produces an unique encrypted passkey.

[nmoray] As I mentioned previously, network admin will be able produce different encrypted unique passkey. They will have two options or I can say two combinations.
1. Generate encrypted passkey with same plaintext passkey + system MAC
2.  Generate encrypted passkey with different plaintext passkey + system MAC
Note: OpenSSL generates a distinct hash each time if we input the same secret / plaintext passkey along with the same salt.
In this approach, the network admin has a luxury to change the plaintext passkey but not the salt or password.

[nmoray] The network admin certainly be able to generate the encrypted passkey remotely as the admin knows both, plaintext passkey + system MAC of that device. Thus, the encrypted passkey generation would be pretty simple. Alternatively, the admin can use the previously generated encrypted key for that device and pre-cook it with CONFIG_DB json. Both the options will allow the network admin to login successfully to that respective device.

[nmoray] First of all, multiple TACACS+ passkey support in currently not there in SONIC. Furthermore, in practice, the encryption mechanism will appear as follows.
```
1. passkey: User entered passkey in plaintext
2. pswd: system MAC

def encryption(passkey, pswd):
    cmd = [ 'openssl', 'enc', '-aes-128-cbc', -A',  '-a', '-salt', '-pbkdf2', '-pass', 'pass:' + pswd ]
    p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
    encryptedkey, err = p.communicate(input=passkey)
    return encryptedkey, err
```

a-barboza:
> > > > * The mechanism of encryption is not described.
> > > 
> > > 
> > > > cc: @shdasari @adyeung
> > > > The idea of encrypting the TACACS+ shared secret before it is written to the configuration (CONFIG_DB) is good.
> > > > However, the drawbacks are: (one of more of these might already be mentioned in the limitations)
> > > > 
> > > > * The key encryption key is not secret because it is the MAC address of the device.
> > > >   [nmoray]--> I think I confused you. Let me clarify. The passkey used here is not the system MAC address. The MAC is used an a salt to encrypt the unique passkey (which is in plaintext).
> > > > * It cannot be changed, therefore the admin has no control over its secrecy.
> > > >   [nmoray]--> The encrypted passkey will be unique for every node so this design will be less susceptible to the overall network breach.
> > > > * The admin cannot use CONFIG_DB snippets unmodified on other devices, because they are keyed by MAC address.
> > > >   [nmoray]--> Yes, that is correct. But the easy way out is, network admin will simply execute "config tacacs passkey <actual_passkey_in_plaintext>" command after copying the CONFIG_DB snippets on a given device so that a new unique encrypted passkey will be generated
> > > > * The mechanism of encryption is not described. There could be some concern with using the key encryption key (MAC address) directly if that is indeed the intent. Is that the intent?
> > > >   [nmoray]--> Please refer to the "Implementation details" section where we have explained the encryption details. Can you please elaborate bit more about those concerns?
> > > >   The implementation is straight forward. An encryption API which will take two inputs, passkey in plaintext and the system MAC. It will use openssl lib to encrypt the provided passkey and use system MAC as a salt.
> > > 
> > > 
> > > @a-barboza the replies are inline.
> > 
> > 
> > [a-barboza] So, if the system MAC address is used as the salt, what is the key used for encryption ? IMHO, the salt has to be unique or pseudo-random (even if it is not secret).
> > [a-barboza] Even if the encrypted passkey is unique for every node, how can it be changed by the admin? Most admins will not like to have their passwords assigned to them. They would rather choose their passwords.
> > [a-barboza] ok, so the workaround is to use a CLI in a script to set the TACACS+ encrypted passkeys. What's the workaround if the admin would like to generate the encrypted passkeys remotely and just give a pre-cooked CONFIG_DB json for configuration?
> > [a-barboza] The Implementation details mentions "A unique Device MAC address used as a salt/password to encrypt/decrypt the configured pass key". So, it is not clear if the device MAC address is used as the salt (initialisation vector), key, or both. So, if there are multiple TACACS+ passkeys, will the same salt be used for all of them?
> 
> [nmoray] Yes, the salt will be system MAC address and passkey will be any user defined secret string. This combination always produces an unique encrypted passkey.
> 
> [nmoray] As I mentioned previously, network admin will be able to produce different encrypted unique passkey. They will have two options or I can say two combinations.
> 
> 1. Generate encrypted passkey with same plaintext passkey + system MAC
> 2. Generate encrypted passkey with different plaintext passkey + system MAC
>    Note: OpenSSL generates a distinct hash each time if we input the same secret / plaintext passkey along with the same password.
>    In this approach, the network admin has a luxury to change the plaintext passkey but not the password.
> 
> [nmoray] The network admin certainly be able to generate the encrypted passkey remotely as the admin knows both, plaintext passkey + system MAC of that device. Thus, the encrypted passkey generation would be pretty simple. Alternatively, the admin can use the previously generated encrypted key for that device and pre-cook it with CONFIG_DB json. Both the options will allow the network admin to login successfully to that respective device.
> 
> [nmoray] First of all, multiple TACACS+ passkey support in currently not there in SONIC. Furthermore, in practice, the encryption mechanism will appear as follows.
> 
> ```
> 1. passkey: User entered passkey in plaintext
> 2. pswd: system MAC
> 
> def encryption(passkey, pswd):
>     cmd = [ 'openssl', 'enc', '-aes-128-cbc', -A',  '-a', '-salt', '-pbkdf2', '-pass', 'pass:' + pswd ]
>     p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
>     encryptedkey, err = p.communicate(input=passkey)
>     return encryptedkey, err
> ```

[a-barboza] the python snippet helps answer the mechanism of encryption.
[a-barboza] the concern about the admin not being able to change the -pass argument remains. In the description of the mechanism of encryption, the pswd is always the system MAC. The admin cannot change that, unless they change the system MAC, which may not be possible (or not desirable to do) always.
[a-barboza] the solution is currently catered to TACACS+ shared secrets only? If each protocol has it's own mechanism, this may become unmanageable quite quickly. Just my 2 cents. Thanks for your replies.

nmoray:
> which may not be possible (or not desirable to do) always.



> > > > > * The mechanism of encryption is not described.
> > > > 
> > > > 
> > > > > cc: @shdasari @adyeung
> > > > > The idea of encrypting the TACACS+ shared secret before it is written to the configuration (CONFIG_DB) is good.
> > > > > However, the drawbacks are: (one of more of these might already be mentioned in the limitations)
> > > > > 
> > > > > * The key encryption key is not secret because it is the MAC address of the device.
> > > > >   [nmoray]--> I think I confused you. Let me clarify. The passkey used here is not the system MAC address. The MAC is used an a salt to encrypt the unique passkey (which is in plaintext).
> > > > > * It cannot be changed, therefore the admin has no control over its secrecy.
> > > > >   [nmoray]--> The encrypted passkey will be unique for every node so this design will be less susceptible to the overall network breach.
> > > > > * The admin cannot use CONFIG_DB snippets unmodified on other devices, because they are keyed by MAC address.
> > > > >   [nmoray]--> Yes, that is correct. But the easy way out is, network admin will simply execute "config tacacs passkey <actual_passkey_in_plaintext>" command after copying the CONFIG_DB snippets on a given device so that a new unique encrypted passkey will be generated
> > > > > * The mechanism of encryption is not described. There could be some concern with using the key encryption key (MAC address) directly if that is indeed the intent. Is that the intent?
> > > > >   [nmoray]--> Please refer to the "Implementation details" section where we have explained the encryption details. Can you please elaborate bit more about those concerns?
> > > > >   The implementation is straight forward. An encryption API which will take two inputs, passkey in plaintext and the system MAC. It will use openssl lib to encrypt the provided passkey and use system MAC as a salt.
> > > > 
> > > > 
> > > > @a-barboza the replies are inline.
> > > 
> > > 
> > > [a-barboza] So, if the system MAC address is used as the salt, what is the key used for encryption ? IMHO, the salt has to be unique or pseudo-random (even if it is not secret).
> > > [a-barboza] Even if the encrypted passkey is unique for every node, how can it be changed by the admin? Most admins will not like to have their passwords assigned to them. They would rather choose their passwords.
> > > [a-barboza] ok, so the workaround is to use a CLI in a script to set the TACACS+ encrypted passkeys. What's the workaround if the admin would like to generate the encrypted passkeys remotely and just give a pre-cooked CONFIG_DB json for configuration?
> > > [a-barboza] The Implementation details mentions "A unique Device MAC address used as a salt/password to encrypt/decrypt the configured pass key". So, it is not clear if the device MAC address is used as the salt (initialisation vector), key, or both. So, if there are multiple TACACS+ passkeys, will the same salt be used for all of them?
> > 
> > 
> > [nmoray] Yes, the salt will be system MAC address and passkey will be any user defined secret string. This combination always produces an unique encrypted passkey.
> > [nmoray] As I mentioned previously, network admin will be able to produce different encrypted unique passkey. They will have two options or I can say two combinations.
> > 
> > 1. Generate encrypted passkey with same plaintext passkey + system MAC
> > 2. Generate encrypted passkey with different plaintext passkey + system MAC
> >    Note: OpenSSL generates a distinct hash each time if we input the same secret / plaintext passkey along with the same password.
> >    In this approach, the network admin has a luxury to change the plaintext passkey but not the password.
> > 
> > [nmoray] The network admin certainly be able to generate the encrypted passkey remotely as the admin knows both, plaintext passkey + system MAC of that device. Thus, the encrypted passkey generation would be pretty simple. Alternatively, the admin can use the previously generated encrypted key for that device and pre-cook it with CONFIG_DB json. Both the options will allow the network admin to login successfully to that respective device.
> > [nmoray] First of all, multiple TACACS+ passkey support in currently not there in SONIC. Furthermore, in practice, the encryption mechanism will appear as follows.
> > ```
> > 1. passkey: User entered passkey in plaintext
> > 2. pswd: system MAC
> > 
> > def encryption(passkey, pswd):
> >     cmd = [ 'openssl', 'enc', '-aes-128-cbc', -A',  '-a', '-salt', '-pbkdf2', '-pass', 'pass:' + pswd ]
> >     p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
> >     encryptedkey, err = p.communicate(input=passkey)
> >     return encryptedkey, err
> > ```
> 
> [a-barboza] the python snippet helps answer the mechanism of encryption. [a-barboza] the concern about the admin not being able to change the -pass argument remains. In the description of the mechanism of encryption, the pswd is always the system MAC. The admin cannot change that, unless they change the system MAC, which may not be possible (or not desirable to do) always. [a-barboza] the solution is currently catered to TACACS+ shared secrets only? If each protocol has it's own mechanism, this may become unmanageable quite quickly. Just my 2 cents. Thanks for your replies.

@a-barboza Okay, so are you suggesting is to keep the password configurable so that same password will be used to generate the encrypted passkey across the fleet? If this is indeed the situation, decoding a single node could lead to the compromise of all other nodes.

BTW, this mechanism may not be limited to TACACS and could be applied to other protocols as well, such as Radius. :-)

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,118 @@
...
+```
+This decryption step is crucial because the login or SSH daemon references the PAM config file to verify the TACACS secret / passkey. If it remains encrypted, the SSH daemon will be unable to recognize the passkey, leading to login failures. The depicted block diagram clearly showcase the enhanced capbalities of the existing submodules.
+
+
+### Implementation details
+
+The implementation stands on three key pillars.
+1. OPENSSL toolkit is used for encryption / decryption
+2. aes-128-cbc is the encoding format used for encryption / decryption
+3. A unique Device MAC address used as a salt/password to encrypt/decrypt the configured pass key
```
Yarden-Z, line 93: Does this mean that the key will only be the MAC address of the switch? If so - this is not a strong encryption (as it is not a random password or something else).
What if we do not want it to be the MAC address, is there an option to change it?

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,118 @@
...
+
+The implementation stands on three key pillars.
+1. OPENSSL toolkit is used for encryption / decryption
+2. aes-128-cbc is the encoding format used for encryption / decryption
+3. A unique Device MAC address used as a salt/password to encrypt/decrypt the configured pass key 
+
+
+#### Show CLI changes
+
+Furthermore, aside from encrypting the passkey stored within CONFIG_DB, this infrastructure ensures that the passkey itself remains concealed from any of the displayed CLI outputs. Consequently, the passkey field has been eliminated from the "show tacacs" output, and it will now solely indicate the status whether the passkey is configured or not. For instance.
```
Yarden-Z, line 98: Do we also remove this password from all log files in the system?
I assume it must remain in the tacplus_nss_conf and also in PAM, but we must make sure that it is not exposed in other files.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,118 @@
...
+
+show tacacs
+["TACPLUS global passkey configured Yes / No"]
+
+### DB migration 
+A DB migration script will be added for users to migrate existing config_db to convert tacacs passkey plaintext to encypted.
```
Yarden-Z, line 104: What about DB sync and DB export?
If we have device A with it's configuration and we would like to copy the configuration to device B, how would the redis export perform in the case we have encrypted the Tacacs password?

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,118 @@
...
+
+show tacacs
+["TACPLUS global passkey configured Yes / No"]
+
+### DB migration 
+A DB migration script will be added for users to migrate existing config_db to convert tacacs passkey plaintext to encypted. 
+
+### Benifits
+
+TACACS passkey encryption adds an extra layer of security to safeguard the passkey on each device throughout the network. Moreover, the utilization of MAC address-based encryption ensures that each network device possesses its distinct encrypted passkey. This strategy effectively mitigates the risk of a major network breach in case one of the devices is compromised.
```
Yarden-Z, line 108: This feature enforces password encryption across all tacacs password settings (both global and per-server passwords), correct?

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: We have considered system MAC address as a password for encrypting the user entered plaintext passkey. Network admin can change the passkey but not the password (which is system MAC). With this combination, OPENSSL is able to generate the random encrypted hash / passkey.  

If we provide a configurable password, we need to store that somewhere (for instance redis which will again defeat the purpose of safe guarding the secret). Additionally, we need to generate unique password for every node within the network which will be a cumbersome job.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: Yes @Yarden-Z. The passkey in plaintext format will only be visible in PAM config file. Else where it will be masked.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: If we have to copy configuration from device A to B, there are two ways.
1. Copy configuration from device A to B and later generate a new encrypted key in device B so that the hostcfgd on device B must be able to decode the encrypted key. This is required as the system MAC address is used as an encryption password.  

2. Generate the configuration for device B (including a new encrypted passkey for device B) and later apply the same.


Code Reference:
```
1. passkey: User entered passkey in plaintext
2. pswd: system MAC

def encryption(passkey, pswd):
    cmd = [ 'openssl', 'enc', '-aes-128-cbc', -A',  '-a', '-salt', '-pbkdf2', '-pass', 'pass:' + pswd ]
    p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
    encryptedkey, err = p.communicate(input=passkey)
    return encryptedkey, err
```

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Limitation
+We choose MAC address is unique and better for management, however if the MAC address of the device is known, we could decrypt the key. If we want to use any other ciper we should manage / stored it in the DB, it is again defeating the purpose. So for now we are going ahead with mac as network admins can orchestartion system to use mac address for passkey encryption. However we are open to new ideas and thoughts to hardened the security and we should be able to accomodate with later releases.
+
+### Testing Requirements
+
+Need to add new / update the existing TACACS testcases to incorporate this new feature
+Test cases to unit test encrypt and decrypt fucntions 
+Test cases to add test the TACACS+ functionality with passkey encryption
+Test cases to cover DB migration
```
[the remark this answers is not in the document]
nmoray: Can you please elaborate what are per server vs global test cases?

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: Yes, right.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
Yarden-Z: So there is no seamless transition between devices.
If we export a config file from device A it will be applicable only for device A and not for any other device in the network.

And, therefor - applying this configuration on another device requires some adjustments to the configuration itself (currently, it's only the Tacacs passkey, but this is the current state as this feature might be applied to other items as well in the system such as radius passkey, snmp community string, ntp key, etc')

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
Yarden-Z: I understand, can we make sure to test that the key is not present in anywhere else in the system once it is configured?
Also, might be worth checking that the files where this key is present is not globally readable.
If it is and an un-privileged user can read it, then encrypting it in the DB will serve minimal purpose.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
Yarden-Z: So the key is not something that is saved on top of the system, but rather a manipulation of the device MAC.
This means the the encryption key consists only of the system MAC (this means that anyone with network proximity to the device can figure out this key) and therefor we don't need to save anything additionally from this feature on top of the system (as device MAC already has a stored location).

Question - isn't using the system MAC (which is exposed via network proximity) not a risk?

nmoray: I would say, the encrypted key is essentially the manipulation of a passkey done using the system MAC address. And yes, there is an obvious risk if an intruder is having physical network proximity to the device. :-)

Alternatively, we can combine the system MAC with the plaintext user entered passkey and will use that as a password for the encryption.

Lastly, we can certainly consider a user configurable password similar to the TACACS passkey. But then where are we going to store it? and how are we going to create a unique password for all the devices within the network?

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: Yeah. To make the config export seamless between the devices, we have to generate an encrypted passkey which can work on all the devices. And for that we need to use same password for encryption across all the devices as passkey can not be changed per device.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
Yarden-Z: These are exactly my questions here :)
If we are going to encrypt this passphrase - what are we trying to protect ourselves against? If it's a user without privileges reading the passkey, then this won't suffice (as this user can review the MAC address).
If it's against someone getting the redis dump and figuring out the passkey - I think that is also not sufficient as he can get the MAC address as well from it (or maybe in this case - it is more difficult).

If we are creating some DB encryption key (let's call it like that for now), we can use a shadow file or something similar to protect it.
Question is - what are we trying to achieve here and from who are we trying to protect the passphrase for the tacacs server?

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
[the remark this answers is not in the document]
ludal35: Hello,

Could you please explain how TACACS key encryption is handled during configuration backups & restores ? As far as I understand, in case of RMA, the system MAC will change, therefore the encrypted key stored in the config file won't be usable on the new box :(

Could you please check and advise ?

Thanks.

Note : config backup/restore is explained here : https://github.com/sonic-net/SONiC/blob/master/doc/ztp/SONiC-config-setup.md

a-barboza:
> > which may not be possible (or not desirable to do) always.
> 
> > > > > > * The mechanism of encryption is not described.
> > > > > 
> > > > > 
> > > > > > cc: @shdasari @adyeung
> > > > > > The idea of encrypting the TACACS+ shared secret before it is written to the configuration (CONFIG_DB) is good.
> > > > > > However, the drawbacks are: (one of more of these might already be mentioned in the limitations)
> > > > > > 
> > > > > > * The key encryption key is not secret because it is the MAC address of the device.
> > > > > >   [nmoray]--> I think I confused you. Let me clarify. The passkey used here is not the system MAC address. The MAC is used an a salt to encrypt the unique passkey (which is in plaintext).
> > > > > > * It cannot be changed, therefore the admin has no control over its secrecy.
> > > > > >   [nmoray]--> The encrypted passkey will be unique for every node so this design will be less susceptible to the overall network breach.
> > > > > > * The admin cannot use CONFIG_DB snippets unmodified on other devices, because they are keyed by MAC address.
> > > > > >   [nmoray]--> Yes, that is correct. But the easy way out is, network admin will simply execute "config tacacs passkey <actual_passkey_in_plaintext>" command after copying the CONFIG_DB snippets on a given device so that a new unique encrypted passkey will be generated
> > > > > > * The mechanism of encryption is not described. There could be some concern with using the key encryption key (MAC address) directly if that is indeed the intent. Is that the intent?
> > > > > >   [nmoray]--> Please refer to the "Implementation details" section where we have explained the encryption details. Can you please elaborate bit more about those concerns?
> > > > > >   The implementation is straight forward. An encryption API which will take two inputs, passkey in plaintext and the system MAC. It will use openssl lib to encrypt the provided passkey and use system MAC as a salt.
> > > > > 
> > > > > 
> > > > > @a-barboza the replies are inline.
> > > > 
> > > > 
> > > > [a-barboza] So, if the system MAC address is used as the salt, what is the key used for encryption ? IMHO, the salt has to be unique or pseudo-random (even if it is not secret).
> > > > [a-barboza] Even if the encrypted passkey is unique for every node, how can it be changed by the admin? Most admins will not like to have their passwords assigned to them. They would rather choose their passwords.
> > > > [a-barboza] ok, so the workaround is to use a CLI in a script to set the TACACS+ encrypted passkeys. What's the workaround if the admin would like to generate the encrypted passkeys remotely and just give a pre-cooked CONFIG_DB json for configuration?
> > > > [a-barboza] The Implementation details mentions "A unique Device MAC address used as a salt/password to encrypt/decrypt the configured pass key". So, it is not clear if the device MAC address is used as the salt (initialisation vector), key, or both. So, if there are multiple TACACS+ passkeys, will the same salt be used for all of them?
> > > 
> > > 
> > > [nmoray] Yes, the salt will be system MAC address and passkey will be any user defined secret string. This combination always produces an unique encrypted passkey.
> > > [nmoray] As I mentioned previously, network admin will be able to produce different encrypted unique passkey. They will have two options or I can say two combinations.
> > > 
> > > 1. Generate encrypted passkey with same plaintext passkey + system MAC
> > > 2. Generate encrypted passkey with different plaintext passkey + system MAC
> > >    Note: OpenSSL generates a distinct hash each time if we input the same secret / plaintext passkey along with the same password.
> > >    In this approach, the network admin has a luxury to change the plaintext passkey but not the password.
> > > 
> > > [nmoray] The network admin certainly be able to generate the encrypted passkey remotely as the admin knows both, plaintext passkey + system MAC of that device. Thus, the encrypted passkey generation would be pretty simple. Alternatively, the admin can use the previously generated encrypted key for that device and pre-cook it with CONFIG_DB json. Both the options will allow the network admin to login successfully to that respective device.
> > > [nmoray] First of all, multiple TACACS+ passkey support in currently not there in SONIC. Furthermore, in practice, the encryption mechanism will appear as follows.
> > > ```
> > > 1. passkey: User entered passkey in plaintext
> > > 2. pswd: system MAC
> > > 
> > > def encryption(passkey, pswd):
> > >     cmd = [ 'openssl', 'enc', '-aes-128-cbc', -A',  '-a', '-salt', '-pbkdf2', '-pass', 'pass:' + pswd ]
> > >     p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
> > >     encryptedkey, err = p.communicate(input=passkey)
> > >     return encryptedkey, err
> > > ```
> > 
> > 
> > [a-barboza] the python snippet helps answer the mechanism of encryption. [a-barboza] the concern about the admin not being able to change the -pass argument remains. In the description of the mechanism of encryption, the pswd is always the system MAC. The admin cannot change that, unless they change the system MAC, which may not be possible (or not desirable to do) always. [a-barboza] the solution is currently catered to TACACS+ shared secrets only? If each protocol has it's own mechanism, this may become unmanageable quite quickly. Just my 2 cents. Thanks for your replies.
> 
> @a-barboza Okay, Are you suggesting that we should maintain a configurable password, ensuring that the same password is utilised for generating the encrypted passkey across the entire fleet? If this is indeed the situation, decoding a single node could lead to the compromise of all other nodes.

[a-barboza] There are multiple clauses or parts in that question. **No**, that is not the suggestion. My concern is that an admin is being forced to encrypt the TACACS+ protocol passkey with a key that is assigned, and cannot be changed. Would you use a combination lock with the condition that it be only locked with the combination which happens to be the serial number of that box which also happens to be engraved on it? Probably not, right?

> 
> BTW, this mechanism may not be limited to TACACS and could be applied to other protocols as well, such as Radius. :-)

[a-barboza] It makes sense to have a common framework/infrastructure that all protocols could use. So, a thorough analysis might be good thing!

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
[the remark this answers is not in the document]
madhupalu: @ludal35 for RMA case, Backup the config and while restore the config encrypt the passkey with new system mac, update the config and load it. Typically, customers integrate with ZTP process or orchestrion used.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: First of all, a secure passphrase is necessary to verify the identity of users or devices seeking access to network resources. Without that unauthorised users could gain access to sensitive systems and data. Thus, encryption of the plaintext passphrase is needed.

But yeah, I understand that the only MAC based encryption won't be enough. Using shadow file is also a good idea.

Yarden-Z: I agree that creating a secure passphrase is critical for protecting any sensitive data that we have on top of the system.
My question is whether this will suffice, as I have presented different vectors where a user can get this data therefor bypassing this protection that was proposed here.
I'm still trying to understand the original security mitigation proposed here and from what type of attacks/users it protects.
In the case we are going with a shadow file for the DB password/key/secret - we can explore that, but we need to elaborate here what we would like to do with said key, how should we store it, manage it and export it if required.

nmoray: We can make use of shadow file for replacing MAC address (as a salt) with an encrypted user specific password. For 
instance, 
<snip_from_shadowfile>
admin:$6$YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1:19446:0:99999:7:::
<snip_from_shadowfile>

"YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1" is the encrypted password for user "admin". We can use the same hash for encrypting the TACACS passkey. What are your thoughts @Yarden-Z ?

nmoray: This way, salt will only be visible to the root user and it will be same across all the devices. Additionally, same config_db.json can also be loaded on to another device without any intermediate modifications.

madhupalu:
After careful consideration of the feedback, we are proposing the **a fixed / hardcoded hash will be used as a salt/password for encryption/decryption and it will be saved locally not in any of the databases**.

The **unique salt** will be the device admin password (encrypted hash from linux shadow file) and it will be saved local to the device and never need to store it any of the the databases. 
For How to - please refer the implementation section. We are working on the code changes will raise PR's soon. 
Thanks

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,87 @@
...
+           |   CLI   +------------------------------------------------------>    ConifgDB   |
+           |         |                                                     |                |
+           +---------+                                                     +----------------+
+```
+This decryption step is crucial because the login or SSH daemon references the PAM config file to verify the TACACS secret / passkey. If it remains encrypted, the SSH daemon will be unable to recognize the passkey, leading to login failures. The depicted block diagram clearly showcases the enhanced capabilities of the existing submodules.
+### Implementation details
+The implementation stands on four key pillars.
+1. OPENSSL toolkit is used for encryption / decryption.
+2. aes-128-cbc is the encoding format used for encryption / decryption.
+3. A unique device admin password (encrypted hash from linux shadow file) will be used as a salt/password to encrypt/decrypt the configured pass key in ConfigDB.
```
liuh-80, line 65: How to handle password change?

There was a password hardening feature will make the password expired: https://github.com/sonic-net/SONiC/blob/master/doc/passw_hardening/hld_password_hardening.md?plain=1

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,87 @@
...
+4. In absence of configured admin password (default behaviour), a fixed / hardcoded hash will be used as a salt/password for encryption/decryption and it will be saved locally not in any of the databases.
+Following is the snippet captured from a sample shadow file.
+<snip_from_shadow_file>
+admin:$6$YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1:19446:0:99999:7:::
+<snip_from_shadow_file>
+Here, "YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1" is the encrypted version of the admin level password. 
+#### Show CLI changes
+Furthermore, aside from encrypting the passkey stored within CONFIG_DB, this infrastructure ensures that the passkey itself remains concealed from any of the displayed CLI outputs. Consequently, the passkey field has been eliminated from the "show tacacs" output, and it will now solely indicate the status whether the passkey is configured or not. For instance,
```
liuh-80, line 73: If the passkey field eliminated or eliminated, how user get the original passkey, will there a new CLI to show the decrypted the passkey?

This will make admin difficult to handle passkey related incident.

I think 'show tacacs' command need decrypt the passkey and keeps the output no change.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,87 @@
...
+           |         |             Encrypted passkey                       |                |
+           |   CLI   +------------------------------------------------------>    ConifgDB   |
+           |         |                                                     |                |
+           +---------+                                                     +----------------+
+```
+This decryption step is crucial because the login or SSH daemon references the PAM config file to verify the TACACS secret / passkey. If it remains encrypted, the SSH daemon will be unable to recognize the passkey, leading to login failures. The depicted block diagram clearly showcases the enhanced capabilities of the existing submodules.
+### Implementation details
+The implementation stands on four key pillars.
+1. OPENSSL toolkit is used for encryption / decryption.
+2. aes-128-cbc is the encoding format used for encryption / decryption.
```
liuh-80, line 64: The encrypted data is binary format, please add detail how to store it in CONFIG_DB. I think it's base64

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: Yeah, the encryption and decryption is solely based on the "admin" password only. If the password expires and it has been changed, the network admin needs to generate a new encrypted passkey using the new admin password hash present in the shadow file.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: Passkey is the mandatory field in CLI. It can not be eliminately. And the main purpose of this implementation is to hide the passkey. Thus, it will be removed from the "show tacacs" output too. 
But if you still want to get the original passkey, you can either decrypt it using decrypt API or refer "common-auth-sonic" file where the passkey is present in the plaintext format.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: It will be stored in the form of a string.

nmoray:
The code PRs for this HLD are as follows.
[Part - I](https://github.com/sonic-net/sonic-utilities/pull/3027)
[Part - II](https://github.com/sonic-net/sonic-utilities/pull/3027)

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
liuh-80: @nmoray , I think the length part need change, because when you encrypt a text to binary then convert to text, the length will usually be much longer, currently sonic already enable yang validation, if the length not change, yang validation will failed with some TACACS key.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
liuh-80: @nmoray , I think the encrypted key update should be automatically done when admin password changed.

If not, when the admin password changed, device can't connect with tacacs server anymore, which will break many automation service in production environment.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
liuh-80: According to the example in hostcfgd PR, it's a base64 string.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,87 @@
...
+<snip_from_shadow_file>
+admin:$6$YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1:19446:0:99999:7:::
+<snip_from_shadow_file>
+Here, "YTJ7JKnfsB4esnbS$5XvmYk2.GXVWhDo2TYGN2hCitD/wU9Kov.uZD8xsnleuf1r0ARX3qodIKiDsdoQA444b8IMPMOnUWDmVJVkeg1" is the encrypted version of the admin level password. 
+#### Show CLI changes
+Furthermore, aside from encrypting the passkey stored within CONFIG_DB, this infrastructure ensures that the passkey itself remains concealed from any of the displayed CLI outputs. Consequently, the passkey field has been eliminated from the "show tacacs" output, and it will now solely indicate the status whether the passkey is configured or not. For instance,
+show tacacs
+["TACPLUS global passkey configured Yes / No"]
+### DB migration
+A DB migration script will be added for users to migrate existing config_db to convert tacacs passkey plaintext to encrypted.
```
liuh-80, line 77: Can this feature be enable/disable by CONFIG_DB setting?

I'm warry about backward compatibility, if a production environment has hundreds of thousands of devices need OS upgrade, and they are installed different old version, upgrade is difficult and may cause livesite.

If there is a flag to disable this feature, then it's much easy to do the upgrade, also can keep backward compatibility.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: @madhupalu can you please update this line as "aes-128-cbc is the encryption algorithm used for encryption / decryption and the ciphertext generated is of type base64".

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: @liuh-80 Automating this would not be straight forward as we need to modify "passwd" linux command to achieve the same.

We can document a requirement that this implementation is based on the admin password present in shadow file. Thus, if the operator is planing to change the "admin" password, he / she has to propagate that change to all the devices.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: @liuh-80 We can achieve the backward compatibility by following your suggestion. We will add check in hostcfgd if the already configured passkey is in plaintext or not. If it is in plantext (same as the one in common-auth-sonic), we will skip the decryption and proceed. This way, operator doesn't need to change anything if he / she upgrades the image. The logic will allow the user to login with existing TACACS credentials.

nmoray: @liuh-80 what are your thoughts on this?

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
nmoray: @liuh-80 one question related to sonic-system-tacacs.yang model. Is it being used for pushing the TACACS configs or pulling the telemetry data related to TACACS or both?

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
Yarden-Z: So there will be added effort to adjust the hash key and perform the encryption-decryption once we change the admin password?

nmoray: @Yarden-Z only the encrypted key inside CONFIG_DB needs to be changed and it will be same across all the devices so operator simply needs to propagate it to all the devices.

Yarden-Z: > @Yarden-Z only the encrypted key inside CONFIG_DB needs to be changed and it will be same across all the devices so operator simply needs to propagate it to all the devices.

I'm unsure if this is the only action required.
Let's say we have the tacplus secret encrypted with password "123456", which is the admin password.
If a user changes the admin password - we need to first decrypt the tacacs key and then re-encrypt it with the new admin password, otherwise - this data is lost.

nmoray: @Yarden-Z why do we need to decrypt the passkey? Here passkey is not changing, the encryption password is changing.

nmoray: Passkey in plaintext would be the same right!

Yarden-Z: Yes, passkey in plaintext should remain the same. Encryption password is changing.
If we used this encryption password - then the ciphertext won't be recoverable.
Per the HLD: 
`When a user configures the TACACS passkey using the SONIC CLI, it will now be securely stored in encrypted format instead of plaintext.`

So, if we are storing it in ciphertext, and the encryption key is changed (admin password) then the ciphertext encryption needs to change as well.
Am I missing something here?

nmoray: okay the flow is like this,

When we encrypt a tacplus passkey with admin password (assume "1234"), it writes an encrypted hash in CONFIG_DB. Same time the encrypted key gets decrypted by hostcfgd and saved into "common-auth-sonic" file in plaintext format (here we don't need to do anything manually). Now let's say. operator needs to change the admin password to "4321" then he / she needs to write a new tacplus passkey in CONFIG_DB by simply running the "config tacacs passkey <str>" CLI. This way, a new encrypted key will get created in CONFIG_DB and the plaintext version of it will be stored in common-auth-sonic file. 

here, the operator only need to look after the encryption part. Decryption will happen automatically. I hope I am able to clarify. :-)

venkatmahalingam: IMO, every organization has different policy/requirements in terms of periodically changing the admin/user passwords of the devices, we should not depend on admin password for encryption key, rather follow the behavior (i.e have a separate master password configuration and default could be system MAC) of other popular NOS for SONiC.

References:
https://www.juniper.net/documentation/us/en/software/junos/user-access/topics/topic-map/master-password-configuration-encryption.html
https://www.cisco.com/c/en/us/support/docs/security-vpn/ipsec-negotiation-ike-protocols/46420-pre-sh-keys-ios-rtr-cfg.html

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
liuh-80: I think this is OK, please update design doc and code.
Also if the feature been disabled, there need some signal check, maybe hostcfgd write a warning/error syslog when render the TACACS config files.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
Yarden-Z: I agree with the suggestion of having a "master password" or some global password for encrypting sensitive data.
This can later on be re-used for other sensitive items such as radius passkey or LDAP passkey (which is coming shortly).
In order to define this "master password" - we'll need to think of a way to create, store it and manage it if required.

In addition - if we would like to scale up this solution (add it other relevant areas in the system where we need to protect sensitive information) we do not want a user performing all those extra actions in the system (currently, it's 1 place, tomorrow it will be 5-10 different configs to change).

nmoray: @venkatmahalingam @Yarden-Z okay understood. I have referred the above link and decided to go with similar implementation where, I will introduce a 'key_encrypt' flag in tacplus table to enable this feature. If enabled, while adding the passkey, it will ask for a password (after entering "config tacacs passkey <str>" command) for the encryption. This password will not be visible on to the console (while entering). It will be stored in /etc/tacplus_pass file (similar to /etc/tacplus_user or /etc/tacplus_nss.conf) in the background and will be referred during the decryption process at hostcfgd.

Additionally, to make this implementation backward compatible, I will keep 'key_encrypt' flag as optional. Like, if it is absent or explicitly marked as False, there will not be any passkey encryption. Passkey will be stored in plaintext for mat only inside CONFIG_DB. 

I hope, now we can conclude on the HLD. :-)

jeff-yin: > If enabled, while adding the passkey, it will ask for a password (after entering "config tacacs passkey " command) for the encryption.

I think @Yarden-Z 's previous comment is asking for a more global solution, and should be one that is portable to other solutions like RADIUS, LDAP, and possibly other features that require symmetrically encrypted passkeys like NTP, OSPF/BGP, etc.

@Yarden-Z  also said:
> In order to define this "master password" - we'll need to think of a way to create, store it and manage it if required.

This might warrant another HLD!

nmoray: > @venkatmahalingam @Yarden-Z @liuh-80 okay understood. I have referred the above link and decided to go with similar implementation where, I will introduce a 'key_encrypt' flag in tacplus table to enable this feature. If enabled, while adding the passkey, it will ask for a password (after entering "config tacacs passkey " command) for the encryption. This password will not be visible on to the console (while entering). It will be stored in /etc/tacplus_pass file (similar to /etc/tacplus_user or /etc/tacplus_nss.conf) in the background and will be referred during the decryption process at hostcfgd.
> 
> Additionally, to make this implementation backward compatible, I will keep 'key_encrypt' flag as optional. Like, if it is absent or explicitly marked as False, there will not be any passkey encryption. Passkey will be stored in plaintext format only inside CONFIG_DB.
> 
> I hope, now we can conclude on the HLD. :-)

@venkatmahalingam what are your thoughts?

nmoray: > > If enabled, while adding the passkey, it will ask for a password (after entering "config tacacs passkey " command) for the encryption.
> 
> I think @Yarden-Z 's previous comment is asking for a more global solution, and should be one that is portable to other solutions like RADIUS, LDAP, and possibly other features that require symmetrically encrypted passkeys like NTP, OSPF/BGP, etc.
> 
> @Yarden-Z also said:
> 
> > In order to define this "master password" - we'll need to think of a way to create, store it and manage it if required.
> 
> This might warrant another HLD!

@jeff-yin Please let us know if you have any better way for encrypting the passkey? For having a same encryption password across all the features (REDIUS, TACACS, LDAP, etc), we can maintain a separate file at /etc/ to store the password similar to tacplus_nss.conf and same password will be used to encrypt the passkey. We will not ask for any password at the time of configuring any of these feature. All feature will by default read a file from /etc/ to encrypt / decrypt the passkey. Additionally, this encryption feature will be enabled / disabled with the help of a flag and this will be there in all the relevant  tables like, TACPLUS, REDIUS etc.

nmoray: This way, we can have MASTER password for the passkey encryption and it will not have any dependency on any other module.

shdasari: The best approach would be to introduce an infra that can provide encryption/decryption as a service across SONiC which will also take care of the key management and storage. This way, all protocols can use the service without worrying about management of the key. The service must be accessible to applications across dockers and host.

venkatmahalingam: > The best approach would be to introduce an infra that can provide encryption/decryption as a service across SONiC which will also take care of the key management and storage. This way, all protocols can use the service without worrying about management of the key. The service must be accessible to applications across dockers and host.

@nmoray This design looks generic and works for all protocols (TACACS+, RADIUS, LDAP, BGP..etc) that need the encryption/decryption. Refer below diagram for some details.

image

madhupalu: @venkatmahalingam thank you, I have updated the HLD. Yes the design is generic, support all three protocols (TACACS+, RADIUS, LDAP, BGP..etc) that need the encryption/decryption. 202311 release is around the corner, could you please approve and merge the HLD, we will work on code PRs and get it done.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
liuh-80: sonic-system-tacacs.yang is used for push TACACS config, also when change TACACS config on device with CLI, the model also will be used for validation.

Increase the length of passkey is backward compatibility. No need other change.

venkatmahalingam (approved):

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -41,60 +42,62 @@ The revised data handling procedure among the modules is outlined as follows:
...
+|         +-------------+        |         +-------------+              |  HostCfg Enforcer   |
+|         |     PAM     |        |         |             |   key-store  +----------^----------+
+|         |  Libraries  |        |    +---->  Master key |     __                  |
+|         +-------------+        |    |    |   Manager   |----|r |                 | Encrypted passkey
++---------------+----------------+    |    |             |    |o |                 |
+                |                     |    +-------------+    |o |                 |
+           +----v----+                |           |           |t |         +-------+--------+
+           |         |                |           |            --          |                |
+           |   CLI   +----------------+           +------------------------>    ConifgDB    |
+           |         |                                  Encrypted passkey  |                |
```
venkatmahalingam, line 58: Fix the typo. ConifgDB

madhupalu: Fixed it. @venkatmahalingam can you merge this PR? or should I ask @zhangyanzhao ?

nmoray:
@venkatmahalingam Created a code [PR](https://github.com/sonic-net/sonic-buildimage/pull/17201) for Master key Manager.

madhupalu:
@venkatmahalingam @zhangyanzhao We have addressed all the review comments, can you merge the PR to master branch.  Code PRs are in review.

Thanks

zhangyanzhao:
@madhupalu  can you please help to add the code PRs by referring to #806 ? Currently, it is hard to find all code PRs from the comments of this HLD. BTW, I will move this feature to 202405 release.

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,105 @@
...
+           |         |                |           |            --          |                |
+           |   CLI   +----------------+           +------------------------>    ConfigDB    |
+           |         |                                  Encrypted passkey  |                |
+           +---------+                                                     +----------------+
+```
+This decryption step is crucial because the login or SSH daemon references the PAM config file to verify the TACACS secret / passkey. If it remains encrypted, the SSH daemon will be unable to recognize the passkey, leading to login failures. The depicted block diagram clearly showcases the enhanced capabilities of the existing submodules.
+
+
+### Approach
+1. A runtime flag (via config_db) for Enable / disable feature: "key_encrypt"
```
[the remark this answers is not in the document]
nmoray: @liuh-80 I have already updated the yang model in the [PR](https://github.com/sonic-net/sonic-buildimage/pull/17201). I will also add the new schema in this HLD.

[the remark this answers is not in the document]
nmoray: The passkey encryption in CONFIG_DB will be done during the configuration part (in aaa.py). 
No Liuh, we don't need to have 'key_encrypt' flag in the existing schema. It is optional. If this flag is absent, encryption will not happen and the passkey will be pushed in plain text format only inside CONFIG_DB.

After rollout, if the user wants to enable this feature, he / she needs to simply enable "key_encrypt" flag and re-configure the passkey so that it will get encrypted and further replaced the same in CONFIG_DB.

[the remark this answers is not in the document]
nmoray: When user enable the "key_encrypt" flag, he / she needs to reconfigure the passkey. This is the requirement. Two steps are required to be followed if anyone wants to use this feature. But this will certainly not hamper the back-compatibility. It is backward compatible.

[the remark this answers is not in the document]
nmoray: It is like, if we need to configure SVI, we first need to add a vlan interface and later mark it as L3 interface by assigning an IP. Two steps. :-)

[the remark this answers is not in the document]
nmoray: BTW, if we are using CLI to encrypt the passkey, it will take care of updating the same into CONFIG_DB.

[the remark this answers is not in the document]
nmoray: You want me to add CLI to enable 'key_encrypt' flag for TACACS?

nmoray:
> @madhupalu can you please help to add the code PRs by referring to #806 ? Currently, it is hard to find all code PRs from the comments of this HLD. BTW, I will move this feature to 202405 release.

@madhupalu please add below set of PRs in the description.
1. [Security Cipher module](https://github.com/sonic-net/sonic-buildimage/pull/17201)
2. [Encryption using security cipher module](https://github.com/sonic-net/sonic-utilities/pull/3027)
3. [Decryption using security cipher module](https://github.com/sonic-net/sonic-host-services/pull/81)

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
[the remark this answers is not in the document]
nmoray: @liuh-80 I have now provided an option to set the key_encrypt flag in TACPLUS table via CLI. Please check the [PR-Part I](https://github.com/sonic-net/sonic-utilities/pull/3027).  Now user need to provide -e option for enabling encryption feature. If not used, it will internally mark key_encrypt flag as false. Thus, post rollout of this feature, that flag will automatically be added in during the configuration itself. 

```
sonic/# config tacacs passkey -h
Usage: config tacacs passkey [OPTIONS] <secret_string>

  Specify TACACS+ server global passkey <STRING>

Options:
  -e, --encrypt   Enable passkey encryption feature
  -h, -?, --help  Show this message and exit.
```

And as mentioned in the previous message, hostcfgd will take care of handling the backward compatibility by doing the decryption only when key_encrypt flag is present and it is set to true. Please refer [PR - Part II](https://github.com/sonic-net/sonic-host-services/pull/81)

zhangyanzhao:
@qiluo-msft  can you please help to double check this HLD and merge this PR after your check? Thanks.

zhangyanzhao:
community review recording https://zoom.us/rec/share/7U0nSCopaMzKnOOOXjGF0tXHKwoznifSBYMDctqwwASNQGNvQob2WAGL9zwOaXwd.qiD2kqkwN0BZpQ1h

doc/TACACSPLUS_PASSKEY_ENCRYPTION.md
```diff
@@ -0,0 +1,115 @@
...
+2. CLI will ask for a encryption master key/password while configuring the TACACS passkey and at the backend it will be stored at /etc/cipher_pass file which is accessible to only root user with read only permissions
```
qiluo-msft, line 66: >master key [](http://example.com/codeflow?start=33&length=10)

This is a disruptive behavior change on CLI. From network operators' point of view, it is ideal that each device could figure out a unique and random master key and just use it internally.

Could you also check other industry NOS implementation and follow the similar practice?

ludal35: FYI : 
- Cisco devices running IOS (-XE or -XR flavors) rely on a very poor encryption (type 7) that can be easily reversed using online tools (https://packetlife.net/toolbox/type7/). AFAIK there's no master key on IOS, therefore configuration backups can be easily redeployed on another box.
- Juniper devices running Junos rely by default on an obfuscation algorithm (type 9), which is also a weak encryption for shared secrets stored in configuration. Such passwords can be decoded using online tools (https://www.m00nie.com/juniper-type-9-password-tool/) or even using a tool provided in Junos (_request system decrypt password_) so that configuration backup files can be restored on RMA'ed devices for example. Nevertheless, Juniper introduced new CLI commands to configure a system master password to provide stronger encryption for configuration secrets (the master password itself is not saved as part of the configuration). The master password is used as input to the password based key derivation function (PBKDF2) to generate an encryption key. The key is used as input to the Advanced Encryption Standard in Galois/Counter Mode (AES256-GCM). The plain text that the user enters is processed by the encryption algorithm (with key) to produce the encrypted text (cipher text). More details avaible here : https://www.juniper.net/documentation/us/en/software/junos/user-access/topics/topic-map/master-password-configuration-encryption.html#id-hardening-shared-secrets-in-junos-os

HTH.

nmoray: Thanks @ludal35 for the references from the popular NOSs.

@qiluo-msft If we want to have unique encrypted key for every device, it is possible with or without changing the password too. 

But the pain point here is, if we have to vary the password for every device, network operator has to keep track of those many passwords (in case he / she wants to change it in the future). 

Note: Here the encryption is solely based on two things, original passkey (in plaintext) and the associated password.

ludal35: Junos approach seems to be a good compromise:
- by default, password is not stored in plain text in the config file (obfuscated format) so that a config file can be redeployed on another device (in case of RMA for exemple).
- if needed, the network administrator can use a master key to encrypt the secrets (not only the TACACS server key, but also the Radius one, the OSPF/BGP neighbor passwords, IKE preshared keys...). There are no restrictions on the master password (@nmoray reusing the same master password across all devices is possible).
- Moreover, Junos can optionnaly rely on the ***TPM chip*** of the Juniper device (if available) to protects secrets such as private keys, system primary passwords, and other sensitive data by storing it in an AES256 encrypted format (instead of storing sensitive data in a clear text format).

nmoray: @ludal35 you mean single master password for all?

ludal35: @nmoray Yes, a single master password for all secrets stored in the SONiC config file (TACACS / Radius server keys and maybe the SNMP communities/secrets as well). Regarding the OSPF/BGP neighbor passwords, I assume this is handled by FRR so this is probably out of the SONiC scope.

nmoray: Yeah @ludal35 as I mentioned in the community review meeting, I intentionally kept the design flexible so that either we can have single master password for all or can keep different passwords for different features.

zhangyanzhao:
merged based on release triage

zhangyanzhao:
code PRs need be reviewed.

zhangyanzhao:
code PRs are not approved yet, move to backlog

PR #1484, opened 2023-09-22
Update sflow_hld.md with pkt drop notifications support

madhupalu:
Update sflow_hld.md with pkt drop notifications support


madhupalu:
@zhangyanzhao I have shared the HLD for wider community review. however, the HLD feature has dependencies with SAI vendors support from Marvel, EC, BRCM and Nvidia. So far I didn't receive any commitments from the vendors yet.  IMHO we should defer this feature for upcoming release. 
I will keep posted as and when the Vendors commitment to this feature.

pttch:
Do it have any other CLI command to setup the sflow for drop reason or profile can associate with drop notification ?

PR #1485, opened 2023-09-23
Warmboot Manager  HLD for warm reboot improvement 

akarshgupta25:

PR #1489, opened 2023-09-29
Add gNOI Reboot HLD.

github76543:
Initial draft of gNOI Reboot HLD.


doc/gnoi-reboot/gnoi_reboot.md
```diff
@@ -0,0 +1,252 @@
...
+_NA_
+
+
+## Architecture Design
+
+The proposal is to create a Reboot Backend (BE) daemon in a new “Framework” container to handle a reboot request and forward it to the platform via DBUS.
+
+This daemon provides a location to implement application level reboot functionality before forwarding a reboot request to the platform (in the spirit of “[Docker to Host Communications HLD](https://github.com/sonic-net/SONiC/blob/master/doc/mgmt/Docker%20to%20Host%20communication.md)”.
+
+The reboot request contains a delay parameter, method and subcomponent fields. For subcomponents we can have support for:
```
ganglyu, line 86: How do we use subcomponent to specify optic, container and etc?

doc/gnoi-reboot/gnoi_reboot.md
```diff
@@ -0,0 +1,252 @@
...
+
+
+## Requirements
+
+_NA_
+
+
+## Architecture Design
+
+The proposal is to create a Reboot Backend (BE) daemon in a new “Framework” container to handle a reboot request and forward it to the platform via DBUS.
```
ganglyu, line 82: Why do we need this backend daemon? Can we use GNOI server to invoke DBUS?

doc/gnoi-reboot/gnoi_reboot.md
```diff
@@ -0,0 +1,252 @@
...
+
+This daemon provides a location to implement application level reboot functionality before forwarding a reboot request to the platform (in the spirit of “[Docker to Host Communications HLD](https://github.com/sonic-net/SONiC/blob/master/doc/mgmt/Docker%20to%20Host%20communication.md)”.
+
+The reboot request contains a delay parameter, method and subcomponent fields. For subcomponents we can have support for:
+
+
+
+* system reboot (method = warm, cold, fast, etc)
```
qiluo-msft, line 90: >system reboot [](http://example.com/codeflow?start=2&length=13)

Could you detailed the reboot design for all the methods? There is no straightforward mapping between these methods with gNOI enum RebootMethod. We prposed below mapping:

```
enum RebootMethod {
  UNKNOWN = 0;     // Invalid default method.
  COLD = 1;        // Shutdown and restart OS and all hardware.
  POWERDOWN = 2;   // Halt and power down, if possible.
  HALT = 3;        // Halt, if possible.
  WARM = 4;        // Reload configuration but not underlying hardware.
  NSF = 5;         // Non-stop-forwarding reboot, if possible.
  // RESET method is deprecated in favor of the gNOI FactoryReset.Start().
  reserved 6;
  POWERUP = 7;     // Apply power, no-op if power is already on.
}
```
For SONiC cold reboot, we can use COLD method.
For SONiC warm reboot, we can use WARM method.
For SONiC fast reboot, we can use NSF method.
For SONiC config reload, we can use reserved method.

doc/gnoi-reboot/gnoi_reboot.md
```diff
@@ -0,0 +1,252 @@
...
+* optic reboot
+* container reboot
```
qiluo-msft, line 92: >container reboot [](http://example.com/codeflow?start=2&length=16)

Does it fit better to another gnoi API Containerz.Start?

doc/gnoi-reboot/gnoi_reboot.md
```diff
@@ -0,0 +1,252 @@
...
+
+There is no obvious container home for a reboot backend daemon: SWSS and PMON were considered. SWSS has the primary role of processing CONFIG_DB inputs (or kernel events) to APP_DB and ASIC_DB outputs. PMON, by name, exists to monitor the platform. Further, there are plans for a state verification daemon which also needs a home.
+
+The Framework container is envisioned to provide a home for foundational services that interact with multiple containers/applications.
+
+State verification daemon periodically (or by request) asks applications to verify their state. For example:
+
+
+
+* SWSS Cfgmgr confirms CONFIG DB matches APPL DB.
```
qiluo-msft, line 124: >Cfgmgr [](http://example.com/codeflow?start=7&length=6)

Why we need "State verification daemon"? Cfgmgr could check matching and write to StateDB, and Syncd could check matching and write to StateDB.

doc/gnoi-reboot/gnoi_reboot.md
github76543: That would need to be designed: for now I'm treating empty subcomponent as a system wide switch reboot.

Being explicit: "we can have support". I'm not at this time proposing/designing support for optic or container reboot in this HLD.

doc/gnoi-reboot/gnoi_reboot.md
github76543: I have no objection: on reflection I think that's cleaner/simpler.

doc/gnoi-reboot/gnoi_reboot.md
github76543: This is a bit of tangent from reboot HLD, but I mentioned it:

State verification daemon periodically and by request asks other system entities to verify their state.
- the daemon aggregates results to report a "system" state
- anyone can ask the daemon to verify and report state for the entire system

Why? I'll leave that to a forthcoming state verification HLD (authored by someone else).

doc/gnoi-reboot/gnoi_reboot.md
github76543: @bhagatgit 
@akarshgupta

doc/gnoi-reboot/gnoi_reboot.md
github76543: @akarshgupta
@bhagatgit

doc/gnoi-reboot/gnoi_reboot.md
```diff
@@ -0,0 +1,252 @@
...
+
+Where StatusCode is defined in [sonic-swss-common/common/status_code_util.h](https://github.com/sonic-net/sonic-swss-common/blob/master/common/status_code_util.h) and can be populated via swss::statusCodeToStr(StatusCode::SWSS_RC_SUCCESS).
+
+Data is JSON encoded RebootRequestResponse, RebootStatusResponse or CancelRebootResponse (from [system.proto](https://github.com/openconfig/gnoi/blob/main/system/system.proto)) indicated by the Key value.
+
+
+### CLI
+
+A cli based tool/script can write/read to/from the “Reboot_Request_Channel” and “Reboot_Response_Channel” notification channels to provide cli based reboot support.
+
```
isabelmsft, line 186: Could you provide a sample of what this CLI would look like (and how it's distinguished from the preexisting `reboot` CLI)? Including the option of specifying subcomponents?

doc/gnoi-reboot/gnoi_reboot.md
github76543: I have removed "container reboot" from the HLD.

doc/gnoi-reboot/gnoi_reboot.md
qiluo-msft: If this is not part of your design, could you remove this section, or link to anther HLD?

doc/gnoi-reboot/gnoi_reboot.md
akarshgupta25: https://github.com/sonic-net/SONiC/pull/1485 proposes daemonizing the warm-boot orchestration and thus will be the backend for warm-boot requests. Additionally, we want a common backend daemon for all types of reboots (cold, warm etc.). If we don't have this daemon then gNOI server would need to figure out the back-end based on the reboot request type. In my opinion, having a common backend provides a cleaner architecture, hence this backend daemon has been proposed.

hdwhdw:
@qiluo-msft whats the update here? I think the implementation is already finished so not sure I can do anything but approve it

PR #1490, opened 2023-10-04
FEC auto determination

longhuan-cisco:
FEC mode is a critical configuration for a port, which needs to be configured properly for the port to come up.

There are below scenarios that can end up with wrong FEC mode:
1. In DPB(Dynamic Port Breakout) case, today's DPB CLI doesn't generate FEC config for newly created ports, FEC mode is default to `none` at [SAI/SDK](https://github.com/opencomputeproject/SAI/blob/a94bbbe43242a4d9e1a4d9f70780ea9251127f5d/inc/saiport.h#L1012) layer.
2. In non-DPB case,
    - Some platforms have no FEC configured in CONFIG_DB by default. The FEC mode can be either default to `none` at SAI/SDK layer or manually configured by user who might not have enough domain knowledge.
    - Some platforms have default FEC defined in `port_config.ini`, which however might not be suitable for the specific port/optics on the system.

The feature in this document is to address the issue in both of above scenarios in a common platform-independent way, since the rule to determine FEC for a given port/optics is common for all platforms.


doc/port-fec/fec-auto-determination.md
```diff
@@ -0,0 +1,185 @@
...
+    - Some platforms have default FEC defined in `port_config.ini`, which however might not be suitable for the specific port/optics on the system.
+
+The feature in this document is to address the issue in both of above scenarios in a common platform-independent way, since the rule to determine FEC for a given port/optics is common for all platforms.
+
+### High-Level Design
+
+- Add `determine-fec` module which can determine FEC mode based on common rule for a given port with a given optics.
+    - This module provides a `determine_fec` API which can be invoked by below entities.
+- Enhance today's DPB CLI to automatically determine and configure FEC in CONFIG_DB for newly created ports, based on determine-fec module.
+- Add a user-triggered CLI `fec-auto-correct` to automatically determine and configure FEC in CONFIG_DB for existing ports, based on determine-fec module.
```
ds952811, line 49: How about having this integrated into pmon#xcvrd to automate the entire process?

i.e.
1. Introduce a platform-independent common rule file as a .json for pmon#xcvrd (e.g. sonic-xcvrd/xcvrd/data/media_autoconf.json)
2. pmon#xcvrd loads the common rules at startup and have it override if a platform-specific media_autoconf.json is present in its device folder.
3. Upon SFP insertion, automatically determine the 'Recommended FEC type' based on the transceiver type detected on a port, and then generate a per port "auto_fec=rs|fc|none" to STATE_DB (and APPL_DB)
4. Add "auto" to the admin FEC type, which means the per port FEC type in CONFIG_DB/APPL_DB will be "fec=auto|rs|fc|none"
5. When fec=auto, the OA should configure the per port FEC type based on "auto_fec" from pmon#xcvrd, and fallbacks to 'none' for NRZ port, 'rs' for PAM4 ports,  when 'auto_fec' is not present.

doc/port-fec/fec-auto-determination.md
```diff
@@ -0,0 +1,185 @@
...
+| 25         | 8               | rs   |
+| 50         | 1               | rs   |
+| 50         | 2               | rs   |
+| 50         | 4               | rs   |
+| 50         | 8               | rs   |
+| 50         | 16              | rs   |
+
+> [!NOTE]
+> If a port has matched FEC entry in both above tables, then prefers FEC entry in first table.
+> For example: A port with 100G-DR optics running in non-breakout mode (`optics_type=100G-DR, lane_speed=25, num_lane=4`) will have a match in both rules, the matched FEC entry in table 1 will be preferably choosen, which is ```none``` in this case.
```
ds952811, line 105: Could we also have the following corner included in the HLD?
i.e.
1. 100G DAC + 4x25G port breakout:  fec=none
2. 100G DAC + 2x50G port breakout:  fec=rs
3. 100G DAC + 1x100G port breakout:  fec=rs

doc/port-fec/fec-auto-determination.md
ds952811: One more thing, the *hint* from pmon#xcvrd could also be useful when we get started working on the CL73 AutoNeg integration, we'll need to adjust the FEC type advertisement based on the transceiver detected on the port

longhuan-cisco: Yea, we already mentioned xcvrd integration in this HLD:
https://github.com/sonic-net/SONiC/blob/35b5585d31bf29c83587fc4a5066c82d871e0c8a/doc/port-fec/fec-auto-determination.md?plain=1#L176
The plan is to treat this as future enhancement, which can be an incremental update(via separate PR) on this HLD. And more detail will be added at that time.

Thanks for the suggestion on the implementation detail, which will be taken into consideration.

doc/port-fec/fec-auto-determination.md
longhuan-cisco: 1x100G(optics_type=ALL_OTHER, lane_speed=25, num_lanes=4) and 2x50G(optics_type=ALL_OTHER, lane_speed=50, num_lanes=1) is already covered. Will add for others.

longhuan-cisco:
Add @shyam77git @jaganbal-a

zhangyanzhao:
community review recording https://zoom.us/rec/share/T8d5F2zsuzrStE5Sj31-cOf5dxOfT7zbS2pjfhhOuTFgwMR2UwHQjfduRIF5joY.IbX5hMETLHTs-3DJ

lguohan:
we need backward compatibility.

doc/port-fec/fec-auto-determination.md
```diff
@@ -0,0 +1,187 @@
...
+    - Some platforms have default FEC defined in `port_config.ini`, which however might not be suitable for the specific port/optics on the system.
+
+The feature in this document is to address the issue in both of above scenarios in a common platform-independent way, since the rule to determine FEC for a given port/optics is common for all platforms.
+
+### High-Level Design
+
+Add `determine-fec` module which can determine FEC mode based on common rule for a given port with a given optics. This module provides a `determine_fec` API which can be invoked in below use cases:
+1. DPB use case: Enhance today's DPB CLI to automatically determine and configure FEC in CONFIG_DB for dynamically created ports, based on determine-fec module.
+2. non-DPB use case: Add a user-triggered CLI `fec-auto-correct` to automatically determine and configure FEC in CONFIG_DB for existing ports, based on determine-fec module.
+    - Future plan: determine-fec module can be further enhanced to be integrated with xcvrd, which can be triggered automatically during transceiver insertion, without human intervention. (details TBD)
```
dgsudharsan, line 49: Please update the HLD as discussed in the community to capture the flow where FEC setting is generated by xcvrd based on media type and pushed to application DB.

doc/port-fec/fec-auto-determination.md
```diff
@@ -0,0 +1,187 @@
...
+| 50         | 8                                 | rs   |
+| 50         | 16                                | rs   |
+
+Above tables can be defined as JSON file, and be loaded by determine-fec module. Platform can also override the default FEC mapping by providing its own JSON file.
+
+> [!NOTE]
+> If a port has matched FEC entry in both above tables, then prefers FEC entry in first table. (Only exception: For `lane_speed=10, num_lane=1 or 4`, we prefer FEC entry in second table)
+
+
+### Diagram For Different Use Cases
```
dgsudharsan, line 110: Please update the diagrams capturing the flow from xcvrd. 
For the DPB use case, since xcvrd today listens to port breakout and generates settings, it would be better for xcvrd to use the same flow as during initialization to push the FEC setting for newly added ports in case of DPB

doc/port-fec/fec-auto-determination.md
```diff
@@ -0,0 +1,187 @@
...
+        config_db ->>+ syncd: notify for FEC update
+    end
+
+    determine_fec -->>- fec_correct_cli: Done
+    fec_correct_cli -->>- user: Done
+
+```
+
+> [!NOTE]
+> In the above use cases, automatically determined FEC mode is saved in running config, user needs to do ```config save``` to save it permanently.
```
dgsudharsan, line 178: As discussed in the community review, the FEC settings should be added to app_db and not config_db. In addition there should be a global configuration knob to allow xcvrd auto setting FEC. This knob should be disabled by default to honor backward compatibility.

dgsudharsan (changes requested):

doc/port-fec/fec-auto-determination.md
shyam77git: @dgsudharsan 
That's correct. We would be adding these two updates to the HLD

Also, we would update this HLD with following as well:
- Remove CLI mention from the HLD and directly mention regular use-case i.e. xcvrd is the consumer of new FEC determination API (mentioned in this HLD)
- Anytime, there is FEC config in config_db, it would supersede any existing or upcoming FEC setting. 
   For this HLD it implies: Apply FEC determination API only if there is no FEC entry/setting in config_db.

doc/port-fec/fec-auto-determination.md
```diff
@@ -0,0 +1,187 @@
...
+| 50         | 1                                 | rs   |
+| 50         | 2                                 | rs   |
+| 50         | 4                                 | rs   |
+| 50         | 8                                 | rs   |
+| 50         | 16                                | rs   |
+
+Above tables can be defined as JSON file, and be loaded by determine-fec module. Platform can also override the default FEC mapping by providing its own JSON file.
+
+> [!NOTE]
+> If a port has matched FEC entry in both above tables, then prefers FEC entry in first table. (Only exception: For `lane_speed=10, num_lane=1 or 4`, we prefer FEC entry in second table)
```
AndyRodan, line 107: First table defines the Optical media interface only. there can be on a single optical interface different types of electrical interface which define different FEC types.
for example: 100GBASE-DR (Clause 140) defines optical interface of 1 fiber lane.
Electrical interfaces that can be are (according to IEEE clause 140)
CAUI-4 (100G on 4 electrical lanes) - NO-FEC
100GAUI-2 (100G on 2 electrical lanes) - RS-FEC
100GAUI-1 (100G on 1 electrical lane) - RS FEC

This does not align with the statement in line 107 that " If a port has matched FEC entry in both above tables, then prefers FEC entry in first table".
FEC should be decided based on the Electrical interface (Second table) and not the first table in optical cable cases

AndyRodan: I agree, there are some cases where on same lane_speed you can have different FEC results.
so it should based on inputs from both tables, but not just one of them

zhangyanzhao:
@longhuan-cisco can you please help to add the code PRs by referring to #806 ? More info can be found from https://github.com/sonic-net/SONiC/wiki/Becoming-a-contributor

zhangyanzhao:
@longhuan-cisco does this feature require SAI change?

longhuan-cisco:
> @longhuan-cisco can you please help to add the code PRs by referring to #806 ? More info can be found from https://github.com/sonic-net/SONiC/wiki/Becoming-a-contributor

Some major updates need to be done in HLD as discussed in community review. And code PRs will be added after that.

> @longhuan-cisco does this feature require SAI change?

According to the current plan, it's not required.

zhangyanzhao:
code PR is not ready, move to backlog for future release

PR #1508, opened 2023-10-31
High-level design for Wake-on-LAN feature in SONiC

lizhijianrd:
High-level design for Wake-on-LAN feature in SONiC.

|Description | Repo  | PR title | State |
| Add `wol` script in sonic-utilities | sonic-utilities |  | Todo |
| Implement `SonicWolService` gNOI service | sonic-gnmi | | Todo |


doc/wol/Wake-on-LAN-HLD.md
```diff
@@ -0,0 +1,136 @@
...
+         +-----------------------------------------+
+     116 |    Password (optional, 4 or 6 bytes)    |
+         +-----------------------------------------+
+```
+
+## CLI Design
+
+The `wol` command is used to send magic packet to target device.
+
+This command requires root privilege because it'll modify the power-mode of target device.
```
venkatmahalingam, line 72: SONiC device just need to send the magic packet to target device(s), right? not sure why we need root privilege at  the SONiC device for command execution.

doc/wol/Wake-on-LAN-HLD.md
```diff
@@ -0,0 +1,136 @@
...
+- `-c count`: Count of magic packet to send. `count` must between 1 and 5. Default value is 1. This param must use with `-i`.
+- `-i interval`: Wait `interval` milliseconds between sending each magic packet. `interval` must between 0 and 2000. Default value is 0. This param must use with `-c`.
+
+### Example
+
+```
+admin@sonic:~$ wol Ethernet10 00:11:22:33:44:55
+admin@sonic:~$ wol Ethernet10 00:11:22:33:44:55 -b
+admin@sonic:~$ wol Vlan1000 00:11:22:33:44:55 -p 00:22:44:66:88:aa
+admin@sonic:~$ wol Vlan1000 00:11:22:33:44:55 -p 192.168.1.1 -c 3 -i 2000
```
venkatmahalingam, line 93: Which module is building the Magic packet for SONiC CLI and gNOI interfaces?

doc/wol/Wake-on-LAN-HLD.md
```diff
@@ -0,0 +1,136 @@
...
+- `target_mac`: MAC address of the target device.
+- `-b`: Use broadcast MAC address instead of target device's MAC address as **Destination MAC Address in Ethernet Frame Header**.
+- `-p password`: An optional 4 or 6 byte password, in ethernet hex format or quad-dotted decimal[^3].
+- `-c count`: Count of magic packet to send. `count` must between 1 and 5. Default value is 1. This param must use with `-i`.
+- `-i interval`: Wait `interval` milliseconds between sending each magic packet. `interval` must between 0 and 2000. Default value is 0. This param must use with `-c`.
+
+### Example
+
+```
+admin@sonic:~$ wol Ethernet10 00:11:22:33:44:55
```
prvattem, line 90: These commands need to be executed in root mode.
Please update the example accordingly.

doc/wol/Wake-on-LAN-HLD.md
venkatmahalingam: Also, capture the validations e.g interface not present, down, socket errors..etc.

doc/wol/Wake-on-LAN-HLD.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Overview
+
+**Wake-on-LAN** (**WoL** or **WOL**) is an Ethernet or Token Ring computer networking standard that allows a computer to be turned on or awakened from sleep mode by a network message[^1]. This document describes the Wake-on-LAN feature design in SONiC.
+
+## Background
+
+Below diagram describes a common usage of WoL on SONiC switch, this HLD will focus on the green part:
+
+![WoL Background](./wol-background.png)
```
prvattem, line 35: Please expand `SONIC Switch` in the diagram and explain the components involved in generating the `Magic Packet`

doc/wol/Wake-on-LAN-HLD.md
lizhijianrd: Yes, SONiC device just send the magic packet. As the doc mentioned, the magic packet will modify the power-mode of the target device, so it's considered as an OPERATION command and requires root.

lizhijianrd: Thanks for point this out, I think we can remove this root privilege requirement since SONiC only need to send a packet. I've update the doc.

doc/wol/Wake-on-LAN-HLD.md
lizhijianrd: Thanks for your comments!
1. Added a `Component` section with some pictures to illustrate that.
2. Added `CLI unit test plan` to cover these validation.

doc/wol/Wake-on-LAN-HLD.md
lizhijianrd: Thanks for your comment! After consideration, we removed the root privilege requirement.

doc/wol/Wake-on-LAN-HLD.md
lizhijianrd: Thanks, I added a `Component` section with some pictures to illustrate that.

zhangyanzhao:
Community review recording https://zoom.us/rec/share/Cfvbr6CAmItpfgei_yhS1yJIYs5e2qL--G8Vf7rBUfzJg6sV09lt26Sv4ScxWio.B4ytA8vMPsMCxj7M

doc/wol/Wake-on-LAN-HLD.md
```diff
@@ -0,0 +1,172 @@
...
+
+The `wol` command is used to send magic packet to target device.
+
+### Usage
+
+```
+wol <interface> <target_mac> [-b] [-p password] [-c count] [-i interval]
+```
+
+- `interface`: SONiC interface name.
```
prvattem, line 101: Are PortChannel intfs supported apart from Ethernet and Vlan?

doc/wol/Wake-on-LAN-HLD.md
```diff
@@ -0,0 +1,172 @@
...
+       6 |               Source MAC                |           HEADER
+         +------+------+---------------------------+
+      12 | 0x08 | 0x42 |   \      \      \      \
+         +------+------+------+------+------+------+------------------------------
+      14 | 0xFF | 0xFF | 0xFF | 0xFF | 0xFF | 0xFF |
+         +------+------+------+------+------+------+        ETHERNET FRAME
+      20 | Target MAC (repeat 16 times, 96 bytes)  |           PAYLOAD
+         +-----------------------------------------+
+     116 |    Password (optional, 4 or 6 bytes)    |
+         +-----------------------------------------+
```
venkatmahalingam, line 88: Wake on LAN UDP dest port -9 is not a requirement?

doc/wol/Wake-on-LAN-HLD.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+## Test Plan
+
+### Unit Tests for `wol` CLI
+| Case Description | Expected Result |
+| :- | :- |
+| Input a valid SONiC interface name. | Parameter validation pass, send magic packet |
+| Input an invalid SONiC interface name. | Parameter validation Fail |
+| Input a valid SONiC interface name, but the interface status is not `up`. | Return `Error: interface not up` |
```
rathnasabapathyv, line 157: As discussed in the community-meeting, this error can be thrown only when the input interface is ethernet-interface by validating the interface-status as UP. 

If the input interface is VLAN, then actual host can be connected on any of the VLAN-member-ports and that actual target-device connected interface status can be down at that point. But since the VLAN-interface-status can be up, user will not see any error and user might think actual target device is woken up successfully, but in reality due to transient actual interface-status down/up flap, magic-pkt might not have reached target-host. 

In this case, user initiating this cmd might not be aware they have to retry this cmd again to wake up the target-host again. May be, we can consider throwing warning, if any of the VLAN-member interface-status is down at the time of issuing this cmd and then it's upto the user to fix interface-status and reissue the cmd again.

It's better to explain, how this situation can be handled in HLD.

doc/wol/Wake-on-LAN-HLD.md
lizhijianrd: Yes, PortChannel will be supported.

doc/wol/Wake-on-LAN-HLD.md
lizhijianrd: The `wol` utility will construct and send an Ethernet frame. It doesn't require the packet is an IP packet.

doc/wol/Wake-on-LAN-HLD.md
lizhijianrd: Hi @rathnasabapathyv, thanks for your question. After consideration, I think if users input VLAN as the interface, then the `wol` utility will only ensure the packet is delivered to target VLAN, but not guarantee to reach each physical interface. So,
1. Users should have some monitor mechanism to check if the target device is wake up as expected. 
2. If users want to ensure the magic-packet be sent out from a physical interface, they should specify the physical interface name instead of vlan name as the `wol` CLI parameter.

skg-net:
@lizhijianrd Can you please update the Quality Metric (Alpha/Beta/GA) for the feature either in this PR comments or in HLD itself based on https://github.com/sonic-net/SONiC/blob/master/doc/SONiC%20feature%20quality%20definition.md

PR #1522, opened 2023-11-28
Python logger enhancement to support setting log level at real time

Junchao-Mellanox:
# Overview

SONiC provides a Python logger implementation in `sonic_py_common.logger.Logger`. This implementation is a simple wrapper of Python standard package `syslog`. It does not provide the ability to change log level at real time. Sometimes, in order to get more debug information, developer has to manually change the log level in code on a running switch and restart the Python daemon. This PR introduces an enhancement to allow user setting log level on fly.

# PR

TBD


keboliu (approved):

liat-grozovik:
@prgeor appreciate your feedback on the HLD

doc/syslog/python-logger-enhancement.md
```diff
@@ -0,0 +1,108 @@
...
+logger = Logger('<log_identifier>', enable_realtime_log_level=True)
+```
+
+To set log level to DEBUG:
+
+```
+swssloglevel -l DEBUG -c <log_identifier>
+```
+
+There interface of `sonic_py_common.logger.Logger` is completely backward compatible.
```
prgeor, line 64: @Junchao-Mellanox backward compatible with ?

doc/syslog/python-logger-enhancement.md
prgeor, line 1: @Junchao-Mellanox won't there will be conflict with set_min_log_priority_info() set by the particular daemon and user overrides with real time debug level and forgets to set the log level to default as per daemon init?

doc/syslog/python-logger-enhancement.md
Junchao-Mellanox: backward compatible with its original interface. I will re-phrase.

Junchao-Mellanox: Fixed.

doc/syslog/python-logger-enhancement.md
Junchao-Mellanox: Hi @prgeor , you are right. There could be conflict if user forgets to set the log level to default value. One possible solution is to make "set_min_log_priority_info" directly write CONFIG DB if enable_set_log_level_on_fly is True. Do you want me to change it like that? Or, do you have other suggestions?

prgeor: @Junchao-Mellanox how does writing to CONFIG_DB prevent the issue I said?

Junchao-Mellanox: Hi @prgeor , I think the conflict you mentioned is like this:
```
1. User create logger instance
2. User code set logger level to INFO (But DB has logger level DEBUG)
```

In step#2, there could be a race condition that: code wants to set logger level to INFO in main thread, DB thread wants to set logger level to DEBUG level. It depends on which code runs last. I think this is the conflict.

And if we implement   set_min_log_priority_info() like this:
```
def set_min_log_priority_info(level):
    if self.enable_set_log_level_on_fly:
        # do nothing, let DB thread handle the log level
    else:
       # directly set log level
```

In this case, we will have consistent result. What do you think?

prgeor: @Junchao-Mellanox so we need to protect `enable_set_log_level_on_fly` from race condition too?

Junchao-Mellanox: that flag is set only via __init__, no need to protect it.

liat-grozovik:
@prgeor could you please refer to the latest doc following feedback and see if additional comments available?
@keboliu can you confirm all relevant PRs are tagged on this PR description?

prgeor:
> @prgeor could you please refer to the latest doc following feedback and see if additional comments available? @keboliu can you confirm all relevant PRs are tagged on this PR description?

@liat-grozovik ack

prgeor:
@Junchao-Mellanox There is one flaw with existing implementation of Logging in `sonic_py_common/logger.py` where syslog() call impacts all the daemons running in the container. To understand what I am saying, take a look at this PR which is trying to fix this https://github.com/sonic-net/sonic-buildimage/pull/17171/files

We need log level which is local to that daemon which is logging. The existing Logger implementation pollutes the log identifier of other daemons as well.

Junchao-Mellanox:
Hi @liat-grozovik,

This PR is pending on PR https://github.com/sonic-net/sonic-buildimage/pull/17171 which is handled by @xincunli-sonic . For PR 17171, we have one open item:

- log message always contain "localhost" instead of real hostname. 

For this item, I can always reproduce on my side, but Xincun cannot reproduce on his side. So, I requested a debug session with him this week. I am waiting for his response.

liat-grozovik:
thanks @Junchao-Mellanox for keeping handling it. now that the pending PR is merged, i assume you need new PR on top of it. once you have please update the PR description with the new code PR instead of the closed one.

@prgeor can we get your approval for the HLD?

prgeor:
> thanks @Junchao-Mellanox for keeping handling it. now that the pending PR is merged, i assume you need new PR on top of it. once you have please update the PR description with the new code PR instead of the closed one.
> 
> @prgeor can we get your approval for the HLD?

@liat-grozovik reviewing

doc/syslog/python-logger-enhancement.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Overview
+
+SONiC provides two Python logger implementations: `sonic_py_common.logger.Logger` and `sonic_py_common.syslogger.SysLogger`. Both of them do not provide the ability to change log level at real time. Sometimes, in order to get more debug information, developer has to manually change the log level in code on a running switch and restart the Python daemon. This is not convenient.
+
+SONiC also provides a C/C++ logger implementation in `sonic-platform-common.common.logger.cpp`. This C/C++ logger implementation is also a wrapper of Linux standard `syslog` which is widely used by swss/syncd. It provides the ability to set log level on fly by starting a thread to listen to CONFIG DB LOGGER table change. SONiC infrastructure also provides the Python wrapper for `sonic-platform-common.common.logger.cpp` which is `swsscommon.Logger`. However, this logger implementation also has some drawbacks:
+
+
+1. `swsscommon.Logger` assumes redis DB is ready to connect. This is a valid assumption for swss/syncd. But it is not good for a Python logger implementation because some Python script may be called before redis server starting.
```
prgeor, line 22: @Junchao-Mellanox How are you overcoming this limitation? I see that you are still writing to the DB which means redis server needs to be running

prgeor (commented):
@Junchao-Mellanox to keep things simple, we an make use of config_db logger table for user to configure the log level for debugging purpose like below:-

```
 "LOGGER": {
            "xcvrd": {
                "LOGLEVEL": "DEBUG",
                "LOGOUTPUT": "SYSLOG"
            },
```

Use the Identifier specified in syslogger to fetch the current log level from config DB 
For eg. The identfier could be Xcvrd, CmisMgr, StateUpdateTask etc.

The advantage is we don't need to create extra threads for handling this

Reading redis DB is not a costly affair

This would limit the change to syslogger 

Please note, logger.py will be deprecated over time as its usage causes confusion and because of which we introduced syslogger

prgeor:
@Junchao-Mellanox one main advantage of the above approach is enabling logging at module level. For eg. Xcvrd common or main thread can have one SysLogger("Xcvrd") instance for eg. CMIS mgr may have another SysLogger("CmisMgr") instance. The logging can be enabled per module within the same thread/process

doc/syslog/python-logger-enhancement.md
Junchao-Mellanox: To overcome it, the implementation will:
1. At the beginning of the thread function, it will wait until redis is up. If the connect to redis fails, it will silently ignore the connection error and retry in 5 seconds. 
2. In the while loop of thread function, catch connection error while reading from redis

Junchao-Mellanox:
> @Junchao-Mellanox to keep things simple, we an make use of config_db logger table for user to configure the log level for debugging purpose like below:-
> 
> ```
>  "LOGGER": {
>             "xcvrd": {
>                 "LOGLEVEL": "DEBUG",
>                 "LOGOUTPUT": "SYSLOG"
>             },
> ```
> 
> Use the Identifier specified in syslogger to fetch the current log level from config DB For eg. The identfier could be Xcvrd, CmisMgr, StateUpdateTask etc.
> 
> The advantage is we don't need to create extra threads for handling this
> 
> Reading redis DB is not a costly affair
> 
> This would limit the change to syslogger
> 
> Please note, logger.py will be deprecated over time as its usage causes confusion and because of which we introduced syslogger

Hi @prgeor , this is exactly the implementation will do.

Junchao-Mellanox:
> @Junchao-Mellanox one main advantage of the above approach is enabling logging at module level. For eg. Xcvrd common or main thread can have one SysLogger("Xcvrd") instance for eg. CMIS mgr may have another SysLogger("CmisMgr") instance. The logging can be enabled per module within the same thread/process

Totally agree. And this is exactly my implementation plan.

prgeor:
> > @Junchao-Mellanox one main advantage of the above approach is enabling logging at module level. For eg. Xcvrd common or main thread can have one SysLogger("Xcvrd") instance for eg. CMIS mgr may have another SysLogger("CmisMgr") instance. The logging can be enabled per module within the same thread/process
> 
> Totally agree. And this is exactly my implementation plan.

@Junchao-Mellanox why do we need a thread?

doc/syslog/python-logger-enhancement.md
```diff
@@ -0,0 +1,217 @@
...
+
+> Note: `swssloglevel` cannot send signal to daemons running in other container because it is actually a program installed in swss container.
+
+config syslog level -c <component> -l <log_level> [--service <service_name>] [--program <program_name>] [--pid <pid>]
+
+- "-c": component name or log identifier name. It is the key of LOGGER table.
+- "-l": log level. One of "DEBUG", "INFO", "NOTICE", "WARN", "ERROR"
+- "--service": container name. Optional. The signal will be sent to that container if present.
+- "--program": program name. Optional, only applicable when --service is present. The signal will be sent to that program running in the service.
+- "--pid": process ID. Optional. If --service is present, the signal will be sent to the PID running in the service; otherwise, the signal will be send to the PID running in host side.
```
prgeor, line 155: @Junchao-Mellanox if user does not specify the optional argument --service, then also this will work?

Junchao-Mellanox: If `--service` is not present, the command will try to send the signal to host side process which has the pid equals to `--pid`

doc/syslog/python-logger-enhancement.md
```diff
@@ -0,0 +1,217 @@
...
+
+# Send SIGHUP to PID 20 running in PMON without update log level
+config syslog level --service pmon --pid 20
+
+# Send SIGHUP to PID 20 running in host without update log level
+config syslog level --pid 20
+
+# Invalid, --program must be used together with --service
+config syslog level --program xcvrd
+
```
prgeor, line 182: @Junchao-Mellanox can we have option `signal` instead of `level` to send signal?

Junchao-Mellanox: Hi @prgeor , the requirement is to have one single command which update log level and make it effect, so, I have to add these two function together.

doc/syslog/python-logger-enhancement.md
```diff
@@ -0,0 +1,217 @@
...
+    while True:
+        # child_log is created in child process, so it will register itself to
+        # the global log registry of child process
+        child_log.log_notice('log something')
+
+p = multiprocessing.Process(target=child)
+p.join()
+```
+
+2. Main process shall forward SIGHUP to child processes. There are many options to achieve this, here is an example:
```
prgeor, line 105: @Junchao-Mellanox Why we need per process handling? The signal handling is per process so child process is independ of parent process and can have its own signal handling. Probably what we need is to support single process with multiple threads.

Junchao-Mellanox: Hi @prgeor , this is just an example in case user uses it in multi-processing case. Your suggestion is also a good option and probably a more general option for multi processing case. Basically, this feature supports single process with multiple threads well.

Junchao-Mellanox: fixed

doc/syslog/python-logger-enhancement.md
```diff
@@ -0,0 +1,154 @@
...
+### High-Level Design
+
+#### SysLogger class change
+
+- A new class level variable `log_registry` shall be added to `SysLogger`class. The type is `dict`.
+- A new argument `enable_runtime_config` shall be added to `SysLogger.__init__`. Default to False.
+- SysLogger instance shall register itself to `log_registry` if `enable_runtime_config` is True.
+- SysLogger instance shall read log level configuration from CONFIG DB if `enable_runtime_config` is True and DB entry is available.
+- SysLogger instance shall update its log level configuration to CONFIG DB if `enable_runtime_config` is True and DB entry is not available.
+- A new class level method `refresh_config` shall be added to `SysLogger` class. The method shall be called while receiving SIGHUP to update log configuration for each logger instance registered to `log_registry`.
```
prgeor, line 55: @Junchao-Mellanox can we maintain the SysLogger Objects inside src/sonic-py-common/sonic_py_common/daemon_base.py so that the main thread will create the list of SysLogger objects required in `DaemonBase.logger_instances[]`? This way the main daemon process will be able to do the signal handling based upon the `identifier` assigned to each Syslogger debug level configured in the CONFIG_DB

liat-grozovik:
@prgeor could you please help to review following the recent changes based on your feedback?

liat-grozovik:
@Junchao-Mellanox is there code PR aligned now with the updated HLD? if so, can you please remove the closed PR and update the description with the relevant ones?

Junchao-Mellanox:
Hi @liat-grozovik , I will remove the closed one. I will add the code PR once it is ready for review.

doc/syslog/python-logger-enhancement.md
```diff
@@ -0,0 +1,140 @@
...
+
+Note: `sonic_py_common.logger.Logger` is not in the feature scope because it is going to be deprecated.
+
+### Architecture Design
+
+`swsscommon.Logger` depends on a thread to listen to CONFIG DB LOGGER table change. It refreshes log level for each logger instances once the thread detects a DB entry change. A thread is considered heavy in a python script, especially that there are many short and simple python scripts which also use logger. To keep python logger light weight, it uses a different design than `swsscommon.Logger`:
+
+- `SysLogger` class shall be changed to a singleton.
+- `SysLogger` instance shall load log level configuration from DB during initialization stage if DB configuration is available.
+- `SysLogger` instance shall save log level configuration to DB during initialization stage if DB configuration is not available.
```
prgeor, line 42: @Junchao-Mellanox Why syslogger need to save log level to DB? The user will configure the log level which will be updated to DB and the main process would be woken up via signal to fetch the user set log level from the DB. There is no need to save log level within SysLogger class.

doc/syslog/python-logger-enhancement.md
```diff
@@ -0,0 +1,140 @@
...
+
+![python-logger-enhancement](/doc/syslog/images/python_logger_enhancement.svg)
+
+### High-Level Design
+
+#### SysLogger class change
+
+- `SysLogger` class shall be changed to a singleton
+- A new argument `enable_runtime_config` shall be added to `SysLogger.__init__`. Default to False.
+- `SysLogger` instance shall load log level configuration from DB during initialization stage if `enable_runtime_config` is True and DB config is available.
```
prgeor, line 53: @Junchao-Mellanox why `enable_runtime_config` default is kept False? What is the intention?

doc/syslog/python-logger-enhancement.md
```diff
@@ -0,0 +1,140 @@
...
+- `SysLogger` instance shall save log level configuration to DB during initialization stage if `enable_runtime_config` is True and DB configuration is not available.
+- A new class level method `refresh_config` shall be added to `SysLogger` class. This method handles load or save log level configuration logic.
```
prgeor, line 55: @Junchao-Mellanox can we rename the method to `update_log_level` ?

doc/syslog/python-logger-enhancement.md
Junchao-Mellanox: we need to register logger itself to DB so that `swssloglevel -p` will print all the available log identifiers. Like this:
```
swssloglevel -p
SAI_API_ACL                   SAI_LOG_LEVEL_NOTICE
SAI_API_ARS                   SAI_LOG_LEVEL_NOTICE
SAI_API_ARS_PROFILE           SAI_LOG_LEVEL_NOTICE
SAI_API_BFD                   SAI_LOG_LEVEL_NOTICE
SAI_API_BMTOR                 SAI_LOG_LEVEL_NOTICE
SAI_API_BRIDGE                SAI_LOG_LEVEL_NOTICE
SAI_API_BUFFER                SAI_LOG_LEVEL_NOTICE
SAI_API_COUNTER               SAI_LOG_LEVEL_NOTICE
SAI_API_DASH_ACL              SAI_LOG_LEVEL_NOTICE
SAI_API_DASH_DIRECTION_LOOKUP SAI_LOG_LEVEL_NOTICE
SAI_API_DASH_ENI              SAI_LOG_LEVEL_NOTICE
SAI_API_DASH_INBOUND_ROUTING  SAI_LOG_LEVEL_NOTICE
SAI_API_DASH_METER            SAI_LOG_LEVEL_NOTICE
SAI_API_DASH_OUTBOUND_CA_TO_PASAI_LOG_LEVEL_NOTICE
SAI_API_DASH_OUTBOUND_ROUTING SAI_LOG_LEVEL_NOTICE
SAI_API_DASH_PA_VALIDATION    SAI_LOG_LEVEL_NOTICE
SAI_API_DASH_VIP              SAI_LOG_LEVEL_NOTICE
SAI_API_DASH_VNET             SAI_LOG_LEVEL_NOTICE
SAI_API_DEBUG_COUNTER         SAI_LOG_LEVEL_NOTICE
SAI_API_DTEL                  SAI_LOG_LEVEL_NOTICE
SAI_API_FDB                   SAI_LOG_LEVEL_NOTICE
SAI_API_GENERIC_PROGRAMMABLE  SAI_LOG_LEVEL_NOTICE
SAI_API_HASH                  SAI_LOG_LEVEL_NOTICE
SAI_API_HOSTIF                SAI_LOG_LEVEL_NOTICE
SAI_API_IPMC                  SAI_LOG_LEVEL_NOTICE
SAI_API_IPMC_GROUP            SAI_LOG_LEVEL_NOTICE
SAI_API_IPSEC                 SAI_LOG_LEVEL_NOTICE
SAI_API_ISOLATION_GROUP       SAI_LOG_LEVEL_NOTICE
SAI_API_L2MC                  SAI_LOG_LEVEL_NOTICE
SAI_API_L2MC_GROUP            SAI_LOG_LEVEL_NOTICE
SAI_API_LAG                   SAI_LOG_LEVEL_NOTICE
SAI_API_MACSEC                SAI_LOG_LEVEL_NOTICE
SAI_API_MCAST_FDB             SAI_LOG_LEVEL_NOTICE
SAI_API_MIRROR                SAI_LOG_LEVEL_NOTICE
SAI_API_MPLS                  SAI_LOG_LEVEL_NOTICE
SAI_API_MY_MAC                SAI_LOG_LEVEL_NOTICE
SAI_API_NAT                   SAI_LOG_LEVEL_NOTICE
SAI_API_NEIGHBOR              SAI_LOG_LEVEL_NOTICE
SAI_API_NEXT_HOP              SAI_LOG_LEVEL_NOTICE
SAI_API_NEXT_HOP_GROUP        SAI_LOG_LEVEL_NOTICE
SAI_API_POLICER               SAI_LOG_LEVEL_NOTICE
SAI_API_PORT                  SAI_LOG_LEVEL_NOTICE
SAI_API_QOS_MAP               SAI_LOG_LEVEL_NOTICE
SAI_API_QUEUE                 SAI_LOG_LEVEL_NOTICE
SAI_API_ROUTE                 SAI_LOG_LEVEL_NOTICE
SAI_API_ROUTER_INTERFACE      SAI_LOG_LEVEL_NOTICE
SAI_API_RPF_GROUP             SAI_LOG_LEVEL_NOTICE
SAI_API_SAMPLEPACKET          SAI_LOG_LEVEL_NOTICE
SAI_API_SCHEDULER             SAI_LOG_LEVEL_NOTICE
SAI_API_SCHEDULER_GROUP       SAI_LOG_LEVEL_NOTICE
SAI_API_SRV6                  SAI_LOG_LEVEL_NOTICE
SAI_API_STP                   SAI_LOG_LEVEL_NOTICE
SAI_API_SWITCH                SAI_LOG_LEVEL_NOTICE
SAI_API_SYSTEM_PORT           SAI_LOG_LEVEL_NOTICE
SAI_API_TAM                   SAI_LOG_LEVEL_NOTICE
SAI_API_TUNNEL                SAI_LOG_LEVEL_NOTICE
SAI_API_TWAMP                 SAI_LOG_LEVEL_NOTICE
SAI_API_UDF                   SAI_LOG_LEVEL_NOTICE
SAI_API_VIRTUAL_ROUTER        SAI_LOG_LEVEL_NOTICE
SAI_API_VLAN                  SAI_LOG_LEVEL_NOTICE
SAI_API_WRED                  SAI_LOG_LEVEL_NOTICE
buffermgrd                    NOTICE
coppmgrd                      NOTICE
fdbsyncd                      NOTICE
fpmsyncd                      NOTICE
gearsyncd                     NOTICE
intfmgrd                      NOTICE
nbrmgrd                       NOTICE
neighsyncd                    NOTICE
orchagent                     NOTICE
portmgrd                      NOTICE
portsyncd                     NOTICE
sflowmgrd                     NOTICE
syncd                         NOTICE
teammgrd                      NOTICE
teamsyncd                     NOTICE
tlm_teamd                     NOTICE
tunnelmgrd                    NOTICE
vlanmgrd                      NOTICE
vrfmgrd                       NOTICE
vxlanmgrd                     NOTICE
wjhcli                        NOTICE
wjhd                          NOTICE
```

doc/syslog/python-logger-enhancement.md
Junchao-Mellanox: Two reason here:
1. We want to keep backward compatible. Before the feature, there is no runtime config support for SysLogger.
2. Enabling runtime config will slow down the logger creation for a litter bit, we want each SysLogger instance to only enable it when need.

doc/syslog/python-logger-enhancement.md
Junchao-Mellanox: sure

prgeor (approved):

PR #1577, opened 2024-01-04
Adding the HLD for SONiC reset local users' passwords

azmy98:
This document provides general information about reset local users' passwords during init implementation in SONiC

The scope of this document is to cover definition, design and implementation of SONiC reset local users' passwords during init feature.

| Submodule | PR Title | Status |
| ---| --- | --- |
| sonic-buildimage | [Feature YANG Model & SystemD service](https://github.com/sonic-net/sonic-buildimage/pull/18874) |GitHub issue/pull request detail |
| sonic-utilities | [CLI commands](https://github.com/sonic-net/sonic-utilities/pull/3302) | GitHub issue/pull request detail |
| sonic-swss-common | [Config DB Schema](https://github.com/sonic-net/sonic-swss-common/pull/874) | GitHub issue/pull request detail|
| sonic-platform-common | [Feature Base Class Implementation](https://github.com/sonic-net/sonic-platform-common/pull/465) | GitHub issue/pull request detail |


azmy98:
This document provides general information about Long reset button press implementation in SONiC

The scope of this document is to cover definition, design and implementation of SONiC Long reset button press feature and Banner CLI.

dgsudharsan (changes requested):

doc/SONiC_long_reset_button_press_hld.md
```diff
@@ -0,0 +1,277 @@
...
+            self.platform_long_reboot_press.start()
+        else:
+            syslog.syslog(syslog.LOG_INFO, "LongRebootPress: Skipping..")
+
+
+def main():
+    LongRebootPress().start()
+```
+
+The feature is added to the init flow as long as the we set the ```ENABLE_LONG_RESET_BUTTON_PRESS``` in ```rules/config```
```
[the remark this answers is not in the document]
azmy98: yes, same as we do for:
CHANGE_DEFAULT_PASSWORD ?= y

doc/SONiC_long_reset_button_press_hld.md
```diff
@@ -0,0 +1,277 @@
...
+    }
+    /* end of top level container */
+}
+/* end of module sonic-long-reset-button-press */
+```
+
+
+# Warmboot and Fastboot Design Impact
+
+The feature will not have any impact on fast reboot or warmboot.
```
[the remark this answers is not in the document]
azmy98: This is interesting. The system should be able to determine if this reboot is either a physical press on the switch or a warm/fast reboot by calling the `get_reboot_cause` function that we are use in the script. So, I don't think we need it. @dgsudharsan, can you confirm?

zhangyanzhao:
community review recording https://zoom.us/rec/share/Mmpq0viHcC2J7QQnHzszlZX4iPR7VqYnXj4u3r7Rzavs4nPG1SImgBb2FK-fL--K.12p00bKsBDn_RCO9

doc/SONiC_long_reset_button_press_hld.md
```diff
@@ -0,0 +1,282 @@
...
+        description "First Revision";
+    }
+
+    container sonic-long-reset-button-press {
+
+        container LONG_RESET_BUTTON {
+
+            description "LONG_RESET_BUTTON part of config_db.json";
+
+            container STATE {
```
venkatmahalingam, line 243: Suggest to avoid STATE name for the CONFIG_DB data as it looks like a read-only data, please use STATUS.

doc/SONiC_long_reset_button_press_hld.md
```diff
@@ -0,0 +1,282 @@
...
+| CLI   | Сommand-line Interface                    |
+| YANG  | Yet Another Next Generation               |
+
+### 
+
+# Requirements Overview
+
+The long reset button press feature implementation shall support the following:
+
+    1. On long reboot press, the OS will restore the local users' configurations, by deleting non-default users and restoring default passwords for default users and expiring  these passwords due to security conecerns.
```
venkatmahalingam, line 54: Is the use-case mainly to reset the password of the default user? if no, why cant we login to the switch with the default user to delete the non-default users explicitly?

doc/SONiC_long_reset_button_press_hld.md
```diff
@@ -0,0 +1,282 @@
...
+    2. Each vendor can decide what is the behavior on long reset button press and add his own implementation.
```
venkatmahalingam, line 55: What if my requirement is to factory reset the switch upon pressing the long reset button? do we have a flexibility in the code to enhance for the future requirement?

zhangyanzhao:
This feature should be turned off by default.

doc/SONiC_long_reset_button_press_hld.md
```diff
@@ -0,0 +1,282 @@
...
+    4. On short reset button press, the current init flow should not be affected and should boot normally.
+    5. Feature should be enabled or diabled by setting compliation flag to true or false.
+    6. Feature can be disabled and enabled using CLI command.
+    7. The feature is enabled by default.
+
+
+### 
+
+# Architecture Design
+
```
venkatmahalingam, line 66: I think, we should start the long reset service only when there is a platform level capability for the long reset support.

doc/SONiC_long_reset_button_press_hld.md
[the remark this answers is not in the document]
azmy98: As part of the meeting discussion, I added the changes to support warmboot/fastboot.

doc/SONiC_long_reset_button_press_hld.md
azmy98: @venkatmahalingam , just to be aligned, do we need to change the leaf as well? or just the container name?

doc/SONiC_long_reset_button_press_hld.md
azmy98: yes, I changed the feature name accordingly. the fearture will be reset local users for the default users and in the default implementation this will be triggered by a long reset button

doc/SONiC_long_reset_button_press_hld.md
```diff
@@ -0,0 +1,282 @@
...
+The long reset button press feature implementation shall support the following:
+
+    1. On long reboot press, the OS will restore the local users' configurations, by deleting non-default users and restoring default passwords for default users and expiring  these passwords due to security conecerns.
+    2. Each vendor can decide what is the behavior on long reset button press and add his own implementation.
+    3. Each vendor must add support for distiguishing between short physical reboot and long physical reboot. 
+    4. On short reset button press, the current init flow should not be affected and should boot normally.
+    5. Feature should be enabled or diabled by setting compliation flag to true or false.
+    6. Feature can be disabled and enabled using CLI command.
+    7. The feature is enabled by default.
```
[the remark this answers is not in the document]
azmy98: changed to disabled :)

doc/SONiC_long_reset_button_press_hld.md
azmy98: I changed the feature name to reset local users' passwords and this triggered by default by long reset button, but you can add the triggering from any condition you see fit. Now you can add a new service that is very similar to this and do your work of reset factory but not for this one..

doc/SONiC_long_reset_button_press_hld.md
azmy98: I agree, I want to add a requirement for it.

doc/SONiC_local_users_password_reset_hld.md
```diff
@@ -0,0 +1,300 @@
...
+
+The service is dependent on the ```database.service``` for reading the configured feature state, so it needs to run after it during boot. As well as it needs to run before any connection is open and this includes all the following services: `sshd.service`, `getty.target`, `systemd-logind.service` and `serial-getty@ttyS0.service`.
+
+### 
+
+# High-Level Design
+
+This feature will be a built-in SONiC feature. There will be three main files to consider in this feature:
+
+1. ```src/sonic-platform-common/sonic_platform_base/reset_local_users_passwords_base.py``` - The default behavior implementation will reside in this file, including when to trigger the feature and how to reset the local users' configurations.
```
dgsudharsan, line 79: Why do we want to name the plugin reset_local_users_passwords? If the trigger of this flow is going to be only long press, I recommend renaming it reset_button.py or something similar.

doc/SONiC_local_users_password_reset_hld.md
```diff
@@ -0,0 +1,300 @@
...
+2. ```platform/<vendor-path>/sonic_platform/reset_local_users_passwords.py``` - The vendor specifc implementation of the feature, each vendor can decide what is the trigger cause to start the functionality and how it is implemented.
```
dgsudharsan, line 80: Same here. If the trigger is going to pressing reset button alone, lets rename it accordingly.

doc/SONiC_local_users_password_reset_hld.md
```diff
@@ -0,0 +1,300 @@
...
+3. ```src/sonic-host-services/scripts/reset-local-users-passwords``` - The python file that will be called on service start during init that imports the vendor's specific implementation.
+
+The default behavior will delete non-default users and restore the original passwords of the default users and expire them based on the content of the file of ```/etc/sonic/default_users.json``` on long reboot press.
```
dgsudharsan, line 83: Do we address the security concern raised in the sonic community meeting? This behavior will reset the passwords alone and not data or configuration and thus its a security concern if someone resets and logs in with default password and obtains all the data. Should we consider resetting configuration and other important files?

doc/SONiC_local_users_password_reset_hld.md
```diff
@@ -0,0 +1,300 @@
...
+    }
+    /* end of top level container */
+}
+/* end of module sonic-local-users-passwords-reset */
+```
+
+
+# Warmboot and Fastboot Design Impact
+
+The feature will not have any impact on fast reboot or warmboot.
```
dgsudharsan, line 282: This daemon itself shouldn't run when the boot type is warmboot or fastboot as the reset button press results only in cold boot. Warmboot and fastboot have strict timing requirements and since this flow will not happen during warm or fastreboot, lets make sure not to run the daemon. 
Please update the same in HLD as well

doc/SONiC_local_users_password_reset_hld.md
```diff
@@ -0,0 +1,300 @@
...
+            '''
+            # can inherent the base class implementation
+    ```
+
+
+The new serivce will call the python script ```src/sonic-host-services/scripts/reset-local-users-passwords``` which in return it will read the feature state from the ```LOCAL_USERS_PASSWORDS_RESET``` table in ```CONFIG_DB``` database. If feature is enabled and a long reset button press detected, the service will perfom the implementation in the ```start()``` in the vendor's specifc implementation as we can see in the following code snippet of ```src/sonic-host-services/scripts/reset-local-users-passwords```:
+
+
+```
+class LocalUsersConfigurationReset:
```
dgsudharsan, line 150: I would recommend not to add code into the HLD. the HLD should contain design and flow diagrams. The code is subject to change and will lead to confusion in the future.

doc/SONiC_local_users_password_reset_hld.md
```diff
@@ -0,0 +1,300 @@
...
+### Config command group
+**The following command set local-users-passwords-reset  feature state:**
+```bash
+config local-users-passwords-reset state <enabled|disabled>
+```
+
+### Show command
+**The following command display long-reset-button-press  configuration:**
+```bash
+root@sonic:/home/admin# show local-users-passwords-reset
```
dgsudharsan, line 232: I would prefer the show command to show the last reboot reset button state, rather than just showing the configuration. Can you please publish the last reboot reset button state to state_db and use it to display in show, so that user will be aware if a long press is detected after the current boot?

dgsudharsan (changes requested):

doc/SONiC_local_users_password_reset_hld.md
azmy98: I will update it also in the requirements

doc/SONiC_local_users_password_reset_hld.md
Yarden-Z: This is a recovery mechanism, if users would like to use this in order to recover their own passwords (which they forgot), without deleting the data on top of the switch - this is the way to do so.
Customers that are using AAA methods will not be affected by this - so let's reduce this to local users only.
In case of a local user, if we have a PW reset, then the default users will be available and might pose a security risk. This is why customers that do not want a recovery mechanism should disable this functionality.
There is indeed a tradeoff between functionality and user experience to security in this feature, but it is manageable.
Also, in order to execute this attack and attack will have to have physical access to the switch, and only after a reboot (which can be tracked since the system will experience downtime)

doc/SONiC_local_users_password_reset_hld.md
dgsudharsan: If show reboot-cause can cover it, I am fine with it

doc/SONiC_local_users_password_reset_hld.md
dgsudharsan: I think the concern was raised by @lguohan . I think another scenario is when a switch is RMA, the switch's data could be easily accessed if the long reset was enabled. This might be the motivation for the security concern. @lguohan  Can you please see if Yarden's comment addresses your concern?

dgsudharsan (approved):

qiluo-msft:
@liuh-80 Could you help review?

doc/SONiC_local_users_password_reset_hld.md
```diff
@@ -0,0 +1,285 @@
...
+
+# High-Level Design
+
+This feature will be a built-in SONiC feature. There will be three main files to consider in this feature:
+
+1. ```src/sonic-platform-common/sonic_platform_base/reset_local_users_passwords_base.py``` - The default behavior implementation will reside in this file, including when to trigger the feature and how to reset the local users' configurations.
+2. ```platform/<vendor-path>/sonic_platform/reset_local_users_passwords.py``` - The vendor specifc implementation of the feature, each vendor can decide what is the trigger cause to start the functionality and how it is implemented.
+3. ```src/sonic-host-services/scripts/reset-local-users-passwords``` - The python file that will be called on service start during init that imports the vendor's specific implementation.
+
+The default behavior will delete non-default users and restore the original passwords of the default users and expire them based on the content of the file of ```/etc/sonic/default_users.json``` on long reboot press.
```
liuh-80, line 84: @azmy98 , can you give more detail about how to restore default password?

According to this HLD: https://github.com/sonic-net/SONiC/blob/master/doc/reset_factory/ResetFactoryHLD.md?plain=1

The "/etc/sonic/default_users.json" only contains **hashed** passwords. so, it's not possible restore default user password from this file.

liat-grozovik:
@azmy98 when do you expect to share the code PRs?

liat-grozovik:
it was decided to remove this HLD for SONiC. PR is closed.

doc/SONiC_local_users_password_reset_hld.md
azmy98: @liuh-80, It is possible sir. You can use the `chpasswd -e` shell command for this purpose. Please refer to an example in the `files/image_config/reset-factory/reset-factory` file under `reset_users()` function.

liat-grozovik:
I see no further comments and all seems to be addressed.
@liuh-80 @venkatmahalingam  FYI the PR is going to be merged soon.

Appreciate if you can also review the code PRs and provide your feedback.

liat-grozovik:
@venkatmahalingam would you mind to provide code PR review feedback as well?

zhangyanzhao:
YANG PR will be reviewed in the YANG workgroup meeting. Other PR review is on-going.

zhangyanzhao:
@liat-grozovik  will bring this to TSC to get guidance how to handle such PR which needs 2nd reviewers.

fastiuk:
@liat-grozovik / @zhangyanzhao can we merge this PR?

doc/SONiC_local_users_password_reset_hld.md
liuh-80: #closed

liuh-80 (approved):

fastiuk:
@liuh-80 could you please merge it?

fastiuk:
@liuh-80 / @zhangyanzhao / @liat-grozovik could you please merge it? It was approved a month ago

zhangyanzhao:
code PRs are not merged, move to backlog

PR #1586, opened 2024-01-10
[smartswitch] Update IP assignment flow

Pterosaur:
Update the configuration workflow of the midplane network:
1. sonic-systemd-generator will generate the configuration for the midplane network
2. systemd-networkd will configure the midplane network according to these configuration
3. Refactor the service dependency to require `database.service` after `midplane-netowrk.service`.
5. Move the smart switch configuration from hwsku.json to platform.json

| PR title                                                 | State                                                                                                              |
|----------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------|
| https://github.com/sonic-net/sonic-buildimage/pull/17161 | GitHub issue/pull request detail |
|  https://github.com/sonic-net/sonic-buildimage/pull/17443   | GitHub issue/pull request detail  |
|                                                          |                                                                                                                    |
|  https://github.com/sonic-net/sonic-buildimage/pull/18178   | GitHub issue/pull request detail  |
|                                                          |                                                                                                                    |


prgeor:
@Pterosaur can you add description?

oleksandrivantsiv:
@Pterosaur, @prgeor 

I agree that we need to change the flow on the DPU side to configure the midplane interface before the DB container is running.

However, I see no benefits in changing the flow for the NPU side to have a midplane bridge created with no interfaces before the DB starts. It doesn't affect the overall IP address assignment flow as it still requires a DHCP server container to start. Why do we need to introduce a new service on the NPU?

Pterosaur:
> @Pterosaur, @prgeor
> 
> I agree that we need to change the flow on the DPU side to configure the midplane interface before the DB container is running.
> 
> However, I see no benefits in changing the flow for the NPU side to have a midplane bridge created with no interfaces before the DB starts. It doesn't affect the overall IP address assignment flow as it still requires a DHCP server container to start. Why do we need to introduce a new service on the NPU?

Hi @oleksandrivantsiv , There is no hard requirement for this changing in NPU, but it will simplify the service graph of NPU.

As you know, the database service is a very early booting process that is required by other services, like SWSS GNMI, and so on.  In the NPU, a predictable scenario is the GNMI needs to W/R the DPU database instances in the NPU. So, these database service needs to be started before GNMI. And there may be the same scenario for SWSS and other components.

If we follow the current design, the midplane bridge is created and configured at interface.service, we have to redesign the service dependencies graph like, 
`database.service->interface.service->database_dpu.service->(swss.service, gnmi.service )...`, this design will spread the modification from a new service, midplane-network.service , to other existing services.  
IMHO, it complicates the implementation and may cause a fault if someone adds a new service who didn't notice this dependencies graph.

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
```diff
@@ -349,11 +344,10 @@ The feature has minimal impact on the memory consumption. Overall it requires le
 
 ### Unit Test cases ###
 
-1. Add new YANG model tests for the new MID_PLANE_BRIDGE and DPUS table.
-2. Extend existing YANG model tests to cover MGMT_INTERFACE table changes.
-3. Add new tests to cover the handling of MID_PLANE_BRIDGE and DPUS tables in sonic-cfggen utility.
-4. Add new tests to cover DHCP server configuration generation for the Smart Switch in sonic-cfggen utility.
-5. Extend interface-config.service tests to verify midplane bridge configuration generation.
+1. Extend existing YANG model tests to cover MGMT_INTERFACE table changes.
```
oleksandrivantsiv, line 347: Can you please review this test case also? There are no MGMT_INTERFACE changes required anymore.

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
```diff
@@ -149,14 +175,8 @@ Based on the preset `t1-smartswitch` default topology the configuration generate
...
     },
-    "MID_PLANE_BRIDGE" {
-        "GLOBAL": {
-            "bridge" : "bridge_midplane",
-            "address": "169.254.200.254/24"
-        }
-    },
     "DHCP_SERVER_IPV4": {
-        "bridge_midplane": {
+        "bridge-midplane": {
```
oleksandrivantsiv, line 179: The name of the bridge here doesn't match the name in `platform.json` midplane_network table:
"midplane_bridge" vs "bridge-midplane"

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
```diff
@@ -149,14 +175,8 @@ Based on the preset `t1-smartswitch` default topology the configuration generate
             "subtype": "SmartSwitch",
         }
     },
-    "MID_PLANE_BRIDGE" {
```
oleksandrivantsiv, line 152: I have a concern with removing the MID_PLANE_BRIDGE table from the config DB. The DHCP server implementation for the Smart Switch depends on this table. Please check this PR: https://github.com/sonic-net/sonic-buildimage/pull/17338
We still can generate the MID_PLANE_BRIDGE table configuration based on the information in the platform.json file.

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
Pterosaur: Fixed.

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
Pterosaur: I feel it's strange if we still leave them in the config DB. Because in my current design, we will not support to change the midplane bridge from config DB anymore. If we want to change it, we need to edit the platform.json. A table of Config DB cannot be configured, it's confused.
But I agree to leave it in this PR and I can find folks to update their current implementation if this HLD is approved and merged.

oleksandrivantsiv: To have midplane bridge information in the config DB gives us a more convenient way to read it. However, the configuration should be treated as a read-only.

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
```diff
@@ -199,53 +222,123 @@ The DHCP_SERVER_IPV4_PORT table includes the following:
...
+WantedBy=multi-user.target
+```
+
+``` text
+# database.service
+
+[Unit]
+...
+Requires=midplane-network.service
+After=midplane-network.service
```
oleksandrivantsiv, line 295: It is not required to modify the `database.service`.
To make sure that `midplane-network.service` runs before `database.service` it is possible to use the `Before` section:
```
[Unit]
Description=Midplane network service
Requires=systemd-networkd.service
After=systemd-networkd.service
Before=database.service

[Service]
Type=oneshot
User=root
ExecStart=/usr/lib/systemd/systemd-networkd-wait-online -i bridge-midplane
```
This will allow us to make the `midplane-network.service` atomic and disable or remove it without affecting the `database.service`

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
```diff
@@ -88,22 +92,27 @@ To implement the switch side functionality the following changes shall be done:
 - Config DB schema should be extended to include the following information:
   - Support of the new `SmartSwitch` subtype in DEVICE_METADATA config DB table. The new subtype shall allow to identify when the SONiC is running on the Smart Switch.
   - Support of the new `MID_PLANE_BRIDGE` table with the following information
-    - Midplane bridge configuration with the bridge IP address and list of the DPU PCIe interfaces that should be added to the bridge for the Smart Switch case.
+    - Midplane bridge configuration with the list of the DPU PCIe interfaces that should be added to the bridge for the Smart Switch case.
```
prgeor, line 95: @Pterosaur can we update line 89 to indicate its not hwsku.json but platform.json. @oleksandrivantsiv FYI

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
prgeor: @oleksandrivantsiv I think both midplane-network and database services can come up independently without any dependency. @Pterosaur Can you explain why we need explicit dependency?

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
```diff
@@ -199,53 +222,123 @@ The DHCP_SERVER_IPV4_PORT table includes the following:
...
+
+[Network]
+Address=169.254.200.254/24
 ```
 
-### IP assignment flow ###
+``` text
+# midplane-network.network
+[Match]
+Name=dpu0 dpu1
```
prgeor, line 252: @Pterosaur How is this file generated for `N` DPUs in a platform?

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
```diff
@@ -199,53 +222,123 @@ The DHCP_SERVER_IPV4_PORT table includes the following:
...
+
+In the DPU side, the steps are almost similar to NPU. But the generated configuration is different but simpler.
+
+``` text
+# midplane-network.network
+[Match]
+Name=eth0-midplane
+
+[Network]
+DHCP=yes
```
prgeor, line 310: @Pterosaur are we using SONiC specific DHCP client?

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
```diff
@@ -199,53 +222,123 @@ The DHCP_SERVER_IPV4_PORT table includes the following:
...
+[Unit]
+Description=Midplane network service
+Requires=systemd-networkd.service
+After=systemd-networkd.service
 
-To trigger the DPU SONiC to query the IP address configuration from the DHCP server the `MGMT_INTERFACE` in the config DB MUST remain empty. This also ensures that the DPU SONiC image can boot and bring the management interface up automatically with the default configuration.
+[Service]
+Type=oneshot
+User=root
+ExecStart=/usr/lib/systemd/systemd-networkd-wait-online -i eth0-midplane --timeout=600
```
prgeor, line 324: @Pterosaur systemd-networkd already knows the interfaces it is managing and hence will wait for a default timeout for the interface to be ready. Why this explicit timeout for DPU but NOT for NPU case where you are not specifying similar wait for dpu0, dpu1, ..interfaces?

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
Pterosaur: This file will be generated/edited by [sonic-systemd-generator.](https://github.com/sonic-net/sonic-buildimage/tree/master/src/systemd-sonic-generator). And this process will get the DPU information from platform.json.

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
Pterosaur: In the NPU side, we don't mind when these DPU PCIe interfaces are created/added into the midplane bridge. So, in the NPU side, we just need to wait for the bridge-midplane.

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
Pterosaur: According to my experiment and a link: https://askubuntu.com/a/1047654/1708404 .I feel the systemd-networkd will use an in-built DHCP client by default.
Do you have any concerns about this behavior? Why we need to use the SONiC specific DHCP client for midplane network?

doc/smart-switch/ip-address-assigment/smart-switch-ip-address-assignment.md
Pterosaur: > It is not required to modify the `database.service`. To make sure that `midplane-network.service` runs before `database.service` it is possible to use the `Before` section:
> 
> ```
> [Unit]
> Description=Midplane network service
> Requires=systemd-networkd.service
> After=systemd-networkd.service
> Before=database.service
> 
> [Service]
> Type=oneshot
> User=root
> ExecStart=/usr/lib/systemd/systemd-networkd-wait-online -i bridge-midplane
> ```
> 
> This will allow us to make the `midplane-network.service` atomic and disable or remove it without affecting the `database.service`

Done

Pterosaur: > @oleksandrivantsiv I think both midplane-network and database services can come up independently without any dependency. @Pterosaur Can you explain why we need explicit dependency?

1. To the DPU side, I believe I don't need to explain much detail. Because we have some components, including database.service, need to leverage the midplane network IP address to identify its DPU ID.
2. To the NPU side, I feel we need to bind the DPU Redis instance on the midplane bridge. If that IP is assigned after the Redis instance, I don't know whether the redis can delay bounding. Even if we suppose the redis support the delay bounding, but based on the current implementation, other components, like swsscommon, cannot support that. https://github.com/sonic-net/sonic-swss-common/blob/253ceb61916fabbf8f261aa9c7507e33d8628312/common/dbconnector.cpp#L536-L550 , So, other critical service may be crashed.

prgeor (approved):

oleksandrivantsiv (approved):

PR #1613, opened 2024-02-14
Add Multi ASICs support for GCU HLD.

xincunli-sonic:
This PR introduces Multi-ASIC support for SONiC's Generic Configuration Update and Rollback feature. It enhances the existing design to handle configurations across multiple ASICs and the host namespace. 

1. The changes make configuration management tools namespace-aware, allowing them to apply configuration patches specifically to global, ASIC, or host scopes.
2. The JSON Patch format is extended to include namespace identifiers, enabling targeted configuration updates. 
3. This PR also updates CLI options for namespace specification and improves validation, verification, and testing for Multi-ASIC scenarios, ensuring reliable configuration management in complex SONiC deployments.

Code change PRs
1. apply-patch: https://github.com/sonic-net/sonic-utilities/pull/3249


doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -871,6 +872,62 @@ show rollback log [exec | verify | status]
...
+The SONiC utilities for configuration management (apply-patch, checkpoint, and rollback) will be enhanced to become namespace-aware. These utilities will determine the target namespace for each operation from the configuration patch itself or from user input for operations that involve checkpoints and rollbacks.
+
+#### 3.2.3.3 JSON Patch Format Extension
+
+The JSON Patch format will be extended to include the namespace identifier for each operation. Operations that target the host namespace will be marked with a "localhost" identifier, while those intended for a specific ASIC will include an "asicN" identifier, where N denotes the ASIC number.
+
+```json
+
+{
+  "asic0": [
```
wen587, line 894: Not sure if this aligns with json patch diff.
When we generate host and asics config, it is represented as a single json.
If we use json patch to generate the diff, it will show in below format:
```
[
    {"op": "remove", "path": "/asic0/DSCP_TO_TC_MAP"},
    {"op": "add", "path": "/asic0/POLICER", "value": { /* Policer configuration */ }}
    {"op": "add", "path": "/localhost/SYSLOG_SERVER", "value": { /* Syslog server configuration */ }}
 ]
```
May need to compare these two format pro and cons.

xincunli-sonic: The pushed jsonpatch will be like 

```json
[
    {"op": "remove", "path": "/asic0/DSCP_TO_TC_MAP"},
    {"op": "add", "path": "/asic0/POLICER", "value": { /* Policer configuration */ }}
    {"op": "add", "path": "/localhost/SYSLOG_SERVER", "value": { /* Syslog server configuration */ }}
 ]
```

But for reuse current single ASIC GCU logic, we need categorize the change per ASIC which includes localhost as well. Then it would be change to like below one as the design proposal:

```json

{
  "asic0": [
    {"op": "remove", "path": "/DSCP_TO_TC_MAP"},
    {"op": "add", "path": "/POLICER", "value": { /* Policer configuration */ }}
  ],
  "localhost": [
    {"op": "add", "path": "/SYSLOG_SERVER", "value": { /* Syslog server configuration */ }}
  ]
}
```

wen587: LGTM.

wen587 (approved):

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -871,6 +872,62 @@ show rollback log [exec | verify | status]
...
+
+In Multi-ASIC SONiC platforms, configurations can be applied either globally, affecting all ASICs, or individually, targeting a specific ASIC or the host. The configuration management tools must, therefore, be capable of identifying and applying configurations based on their intended scope.
+
+#### 3.2.3.2 Namespace-aware Configuration Management
+
+The SONiC utilities for configuration management (apply-patch, checkpoint, and rollback) will be enhanced to become namespace-aware. These utilities will determine the target namespace for each operation from the configuration patch itself or from user input for operations that involve checkpoints and rollbacks.
+
+#### 3.2.3.3 JSON Patch Format Extension
+
+The JSON Patch format will be extended to include the namespace identifier for each operation. Operations that target the host namespace will be marked with a "localhost" identifier, while those intended for a specific ASIC will include an "asicN" identifier, where N denotes the ASIC number.
```
isabelmsft, line 889: Is `localhost` the same as `global` namespace referenced in this doc? https://github.com/sonic-net/SONiC/blob/master/doc/database/multi_namespace_db_instances.md

xincunli-sonic: Good point, in the implementation, we will use `sonic-cfggen` to retrieve the localhost and other asics' configurations, the localhost namespace would be [empty string](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-config-engine/portconfig.py#L171), we can do convert to `empty string` when anytime calls apply-patch cli with `localhost` namespace.

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -871,6 +872,62 @@ show rollback log [exec | verify | status]
...
+}
+```
+
+#### 3.2.3.4 Applying Configuration Changes
+
+When applying a configuration patch, the system will:
+
+    Parse the extended JSON Patch to identify the target namespaces.
+    Apply the operations intended for the host namespace directly to the host's configuration database.
+    Apply ASIC-specific operations to the respective ASIC's configuration database.
```
wen587, line 910: Previously there is a service validator that only design for single asic. We should check if there could be service restart need for each namespace. Current supported services are listed here.  https://github.com/sonic-net/sonic-utilities/blob/master/generic_config_updater/gcu_services_validator.conf.json

xincunli-sonic:
@qiluo-msft, can you help merge, Thanks!

doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md
```diff
@@ -871,6 +872,62 @@ show rollback log [exec | verify | status]
...
+{
+  "asic0": [
+    {"op": "remove", "path": "/DSCP_TO_TC_MAP"},
+    {"op": "add", "path": "/POLICER", "value": { /* Policer configuration */ }}
+  ],
+  "localhost": [
+    {"op": "add", "path": "/SYSLOG_SERVER", "value": { /* Syslog server configuration */ }}
+  ]
+}
+```
```
isabelmsft, line 902: Could we have a comparison between multi-asic and single-asic patch and CLI output after patch application for failure and success scenarios?

qiluo-msft:
> | 1         | Add a new table. |

I think most testcases could be applicable to multi-asic. And there should be some special testcases for multi-asic, like "add a new ASIC config subtree", "add a new ASIC with empty config", etc.

---
Refers to: doc/config-generic-update-rollback/SONiC_Generic_Config_Update_and_Rollback_Design.md:1028 in 38f70d2. [](commit_id = 38f70d2d1e1a12b9f41c96313d70da81e2596eee, deletion_comment = False)

qiluo-msft (dismissed):
LGTM. Please also check with other reviewers.

qiluo-msft:
Please resolve conflict.

xincunli-sonic:
Used forked repo to recreate a new PR: https://github.com/sonic-net/SONiC/pull/1671

qiluo-msft (approved):
Please check with other reviewers.

arlakshm (approved):

PR #1621, opened 2024-02-22
[HLD] bmp for monitoring SONiC BGP info

FengPan-Frank:
[HLD] bmp for monitoring SONiC BGP info

|Description | Repo  | PR link| State |
|Yang model definition| sonic-buildimage |https://github.com/sonic-net/sonic-buildimage/pull/18530|GitHub issue/pull request detail|
|Add BMP_STATE_DB dedicated instance|sonic-buildimage|https://github.com/sonic-net/sonic-buildimage/pull/19016| GitHub issue/pull request detail|
|Add BMP_STATE_DB support in swss-common|sonic-swss-common|https://github.com/sonic-net/sonic-swss-common/pull/879| GitHub issue/pull request detail|
|bmpcfgd service implementation|sonic-buildimage|https://github.com/sonic-net/sonic-buildimage/pull/18940| GitHub issue/pull request detail|
|config cli implementation|sonic-utility|https://github.com/sonic-net/sonic-utilities/pull/3286|  GitHub issue/pull request detail|
|show cli implementation|sonic-utility|https://github.com/sonic-net/sonic-utilities/pull/3289|GitHub issue/pull request detail|
|openbmpd service implementation|sonic-bmp|https://github.com/sonic-net/sonic-bmp/pull/7| GitHub issue/pull request detail|
|Add sonic-bmp submodule|sonic-buildimage|https://github.com/sonic-net/sonic-buildimage/pull/20780| GitHub issue/pull request detail|
|Build bmp container into sonic-buildimage|sonic-buildimage|https://github.com/sonic-net/sonic-buildimage/pull/18946| GitHub issue/pull request detail|
|Add bmp dynamic feature switch in frr side config.|sonic-buildimage|https://github.com/sonic-net/sonic-buildimage/pull/20895| GitHub issue/pull request detail|


doc/bmp/bmp.md
```diff
@@ -0,0 +1,354 @@
...
+
+- BMP will support both config and CLI, which can enable database population per table type population by default and in runtime.
+
+- User could use StreamingTelemetry path to monitor bgp table via [The GNMI protobuf](#https://github.com/openconfig/gnmi/blob/5473f2ef722ee45c3f26eee3f4a44a7d827e3575/proto/gnmi/gnmi.proto#L309 )
+
+## 2.2 OpenBMP
+[OpenBMP](#https://www.openbmp.org/) is open source BGP monitor framework, which supports collecting, aggregating, data persisting, as well as dashboard, etc. For this project, we will fork this project with some internal code changes which populates the required data into Redis for our usage.
+
+## 2.3 BGP/FRR update
+Current [BGP](#https://github.com/sonic-net/sonic-frr) container has been forked from [FRR](#https://github.com/FRRouting/frr), which supports BMP protocol from version frr/7.2
```
qiluo-msft, line 92: Current [BGP] container has been forked from [FRR] -> Current BGP container uses a implementation (sonic-frr) forked 
 from FRR

FengPan-Frank: Fixed.

zhangyanzhao:
community review recording https://zoom.us/rec/share/t7kdlkY0uSCaf9fVifFaeS2rY8RSlWrQlIcO7CaZsaE0HU1bnp8-orK5IN81SXup.QdSvaujeNaOagbml

doc/bmp/bmp.md
```diff
@@ -0,0 +1,373 @@
...
+|DB name    |   DB No. |     Description|
+|  ----     |:----:| ----|
+|APPL_DB    |  specific ASIC indexed DB per BGP neighbor |   Application running data |
+
+### BGP Neighbor table schema
+
+This table will capture BGP Neighbor attributes supported on BGP neighbor, including capability it support.
+
+```
+admin@str-7050qx-32s-acs-04:~$ redis-cli -n 0 HGETALL BGP_NEIGHBOR_TABLE:10.0.0.61
```
venkatmahalingam, line 137: https://github.com/sonic-net/SONiC/pull/1621/files#diff-0649e404e5d4496d9d6af8f6b9411cb483e3a6d376a927d8660baf476465d7b3R138
"SONiC to stream events occurring in SONiC switch (e.g BGP flap, process not running, ...) in real time via gNMI subscription model.
The events are defined using versioned YANG schema."

We have the above YANG model for streaming the BGP states today and looks like BGP_NEIGHBOR_TABLE supports extensive state information, hope we'll migrate to use one streaming method i.e BMP or events infra.

doc/bmp/bmp.md
```diff
@@ -0,0 +1,373 @@
...
+
+# 2 Architecture Design
+
+The following are the architecture changes. 
+
+## 2.1 High-Level Architecture
+As above context mentioned, we need to bring up new container as bmp on SONiC, after change the relevant component architecture is as below:
+<img src="images/architecture_diagram.png" alt="Architecture Diagram" width="800">  
+
+- Add new bmp container, which has limited resource control.
```
praveenraja1, line 85: Do we new docker for running the "openBMPd" daemon? 

I understand that the plan is to have, 

1) config "feature" command to enable/disable this.
2) Limit CPU resource usage of the new docker . 

Even if we leverage the existing frr container to run openBMPd, there are alternate ways to achieve the above, for instance #1 we could have a "config bmp <enable|disable>"  which could internally start/stop the openBMPd process. And for #2 we could use command such as "nice/cpulimit" to set process usage limits. 

From what i have seen adding a new container (base docker, with no overhead app) increases RAM usage by 20 to 30 MB, so just want to ensure if we really need a new container here.

doc/bmp/bmp.md
```diff
@@ -0,0 +1,373 @@
...
+```
+   127.0.0.1:6379[4]> keys FEATURE|bmp*
+   1) "FEATURE|bmp"
+
+```
+
+Create below config items list for enabling and disabling different table.
+
+```
+   127.0.0.1:6379[4]> HGETALL FEATURE|bmp
```
rathnasabapathyv, line 256: Since FRR does support bmp enable/disable as a cmd-line arg (`/usr/lib/frr/bgpd -A 127.0.0.1 -M bmp`), to support per table level enable/disable as shown below, it's better to discuss this with FRR-community and introduce table level enable/disable support via FRR CLI knob, so that zebra/bgpd doesn't have to spend any cycle in sending change notifications for those tables to BMP ; Otherwise zebra/bgpd will always send notifications to BMP for the disabled tables and bmp ignores it later based on this config-db.

Please discuss with FRR community on this, as discussed in the HLD review community meeting (March-12-2024)

doc/bmp/bmp.md
```diff
@@ -0,0 +1,373 @@
...
+| OpenBMP                  | Open Source BGP Monitoring Protocol (OpenBMP) Collection Framework    |https://github.com/SNAS/openbmp|
+| BGP/FRR                  | SONiC BGP container and bgpd daemon which will collborate with BMP here   |https://github.com/sonic-net/sonic-frr|
+| GNMI                     | gRPC Network Management Interface   |https://github.com/openconfig/gnmi|
+
+
+# 1 Requirements Overview
+
+## 1.1 Functional requirements
+
+Now SONiC supports standard BGP session established with BGP neighbors, but there's only limited BGP info readable in database, which makes some cases difficult to debug, we need to dump more BGP data, meanwhile we're also expecting some live debug option which could dump BGP related info on demand and improve the debuggability.
```
venkatmahalingam, line 52: VTYSH show commands provide extensive dumps today for the debuggability, trying to understand the new information that we support part of openBMP.

On-change subscription for BGP data is clearly an use-case, I would like to see more filtering options (e.g RIB-in or our per neighbor) in the HLD so that we can limit the number entries going into the DB.

doc/bmp/bmp.md
```diff
@@ -0,0 +1,373 @@
...
+
+```
+! 
+  bmp mirror buffer-limit 4294967214 
+! 
+  bmp targets test 
+    bmp stats interval 1000 
+    bmp monitor ipv4 unicast pre-policy 
+    bmp monitor ipv6 unicast pre-policy 
+    bmp connect 127.0.0.1 port 5000 min-retry 1000 max-retry 2000
```
venkatmahalingam, line 111: One option is to connect to the local collector and other option is to connect to the remote collector, if we have an use-case not to store into the redis DB but to only stream into the remote collector, are we planning to give such an option?

doc/bmp/bmp.md
```diff
@@ -0,0 +1,373 @@
...
+### Daemon parameter
+From version frr/7.2, bgpd daemon supports parameter to enable bmp functionality, we need to adjust BGP docker init relevant script and turn on bmp collection option.
+
+```
+/usr/lib/frr/bgpd -A 127.0.0.1 -M bmp 
+```
+
+## 2.4 Database Design
+
+BMP will continually populate existing redis table APPL_DB, just with different keys defined to cover the data so that functional requirement could be supported.
```
venkatmahalingam, line 124: Please consider pushing this state information into STATE_DB or new DB instance if that makes our life easy.

doc/bmp/bmp.md
```diff
@@ -0,0 +1,373 @@
...
+   - Description: Show BGP prefixes received total count.  
+   - Result: Query BGP_RIB_IN_TABLE table and show calculated result in human readable format.
+
+```
+
+
+# 4 Resource usage and Test plan
+
+### Resource usage
+Since BMP is a sidecar daemon for improve the SONiC debuggability and monitor efficiency, we don't expect it take too much CPU usage, and prefer it could be able to work under low priority but without break. Now FRR to openbmp is working under push model, if limit cpu of bmp into 5%, it just operate slow, but without missing update, when we flap the bgp several times, all the flap updates will pushed to bmp finally. Thus we should be able to restrist CPU of BMP to lower priority without
```
venkatmahalingam, line 340: There was an exercise done optimize memory usage in DASH WG. Since we may store lot of RIB information into the DB, may be it's ideal to use protobuf format rather than text format for DB entries (refer the below link for more information).
https://github.com/sonic-net/DASH/blob/main/documentation/general/data/AppDBMemoryEstimation.xlsx

doc/bmp/bmp.md
FengPan-Frank: Updated.

doc/bmp/bmp.md
FengPan-Frank: Yes, will make some comparison for this, using pb should be feasible just the data will not be readable if we use redis-cli

doc/bmp/bmp.md
FengPan-Frank: Correct, that will require dedicated BMP service being deployed and maintained, for now we're not planning of this.

doc/bmp/bmp.md
FengPan-Frank: Are you asking smaller granularity control like per neighbor? instead of per table?.

doc/bmp/bmp.md
FengPan-Frank: Sure, currently FRR just post data to BMP following standard BGP protocol, thus neighbor table will be sent in BGP_MSG_OPEN message (#https://datatracker.ietf.org/doc/html/rfc4271#page-13) and all other tables including rib/ls_state... will be sent in BGP_MSG_UPDATE. A better option is to update in FRR side and controlled by the config/cli together, so that no extra data is transmitted between FRR and BMP. Will follow up with FRR community.

doc/bmp/bmp.md
venkatmahalingam: I'm asking per neighbor level filtering at least and if possible, per table level filtering is good as well.

doc/bmp/bmp.md
venkatmahalingam: yes, most database contents are in text format, we may need to check the possibility of memory optimization as we may push lot of RIB information into the DB.

doc/bmp/bmp.md
rathnasabapathyv: Thanks. Let's discuss with FRR community and enhance FRR capability to achieve this, as part of bringing in this feature to community-sonic.

doc/bmp/bmp.md
FengPan-Frank: Sure, we will support table level filtering first, later could append neighbor level filtering if specific requirement asks.

doc/bmp/bmp.md
FengPan-Frank: Thanks, it's indeed feasible as either way, and some tradeoff just exists as consideration.
The advantage of using dedicated container is that we can avoid packaging bmp feature onto some slim images if DUT resource is small and expensive relatively.

doc/bmp/images/architecture_diagram.png
forrestchu, line 1: I have some concerns about this architecture design:
1）Why not BMP send route to external OpenBMP servers, which may keep more detailed BGP history based on Kafka.
2)  Can BMP module in FRR write route to Redis directly? It is more straightforward.

FengPan-Frank: 1> That's will require us to maintain dedicated OpenBMP service, which is not in current scope, maybe later we can do that, if so it will be configurable simply from bmp perspective as well.
2> I think it will introduce unnecessary complexities into FRR, if like that FRR can persist Rib info directly instead of BMP, right? We choose BMP path now because BMP was natively supported in FRR in ASYNC mode, and not much extra cost added into FRR.

forrestchu: OK, Got the concept behind your design. Thanks~

qiluo-msft (approved):

PR #1642, opened 2024-03-26
[Dash] Sync dash-sonic hld from 'dash' repo

prsunny:
Sync with dash repo for the sonic-hld - https://github.com/sonic-net/DASH/blob/main/documentation/general/dash-sonic-hld.md


prsunny:
@KrisNey-MSFT for viz

r12f (approved):

PR #1646, opened 2024-03-28
[DASH] Fix Typos for vnet peering, mac address

albgonzpri:
…r differnet destinations (3) different PAs for different destinations

### DIFF
--- a/doc/dash/dash-sonic-hld.md
+++ b/doc/dash/dash-sonic-hld.md
@@ -1006,7 +1006,8 @@ Refer DASH documentation for the test plan.
     {
         "DASH_VNET_TABLE:Vnet1": {
             "vni": "45654",
-            "guid": "559c6ce8-26ab-4193-b946-ccc6e8f930b2"
+            "guid": "559c6ce8-26ab-4193-b946-ccc6e8f930b2",
+            "peer_list": "Vnet200"
         },
         "OP": "SET"
     },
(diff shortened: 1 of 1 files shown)


doc/dash/dash-sonic-hld.md
```diff
@@ -1006,7 +1006,8 @@ Refer DASH documentation for the test plan.
     {
         "DASH_VNET_TABLE:Vnet1": {
             "vni": "45654",
-            "guid": "559c6ce8-26ab-4193-b946-ccc6e8f930b2"
+            "guid": "559c6ce8-26ab-4193-b946-ccc6e8f930b2",
+            "peer_list": ["Vnet200"]
```
prsunny, line 1010: I don't think peer_list here matters. This was added initially but no longer used. Peering vnet can be just mentioned in the Route table

albgonzpri: Makes sense. The DASH_ROUTE_GROUP_TABLE allows to effective link via routing peered vnets
Will adjust this and remove all references to peer_list in the doc

doc/dash/dash-sonic-hld.md
```diff
@@ -318,7 +318,6 @@ DASH_VNET_TABLE:{{vnet_name}}
     "vni": {{vni}} 
     "guid": {{"string"}}
     "address_spaces": {{[list of addresses]}} (OPTIONAL)
-    "peer_list": {{vnet_name_list}} (OPTIONAL)
```
prsunny, line 321: Alberto, sorry for the confusion. The entry is still present in schema as optional. However it is not used for the current scenarios. So its good to leave it here for consistency, but not to add to examples.

prsunny (approved):

PR #1282, opened 2023-03-14
Add HLD for IPv4 port based DHCP server in SONiC

yaqiangz:
This document describes the design details of ipv4 port based DHCP server in SONiC.

|Description | Repo  | PR title | State |
| ------ | ------ | ------- | ----- |
|Add yang model | sonic-buildimage | [[yang][dhcp_server] Add dhcp_server_ipv4 yang model](https://github.com/sonic-net/sonic-buildimage/pull/16327) | GitHub issue/pull request detail |
|Add dhcp_server container |sonic-buildimage|[[dhcp_server] Add dhcp_server container](https://github.com/sonic-net/sonic-buildimage/pull/14031) |GitHub issue/pull request detail|
|~~Add support for dhcp_server in minigraph parser~~ |~~sonic-buildimage~~|~~[[minigraph] Add support for ipv4 dhcp_server in minigraph parser](https://github.com/sonic-net/sonic-buildimage/pull/16313)~~|GitHub issue/pull request detail|
|Add dhcpservd |sonic-buildimage|[[dhcp_server] Add dhcpservd to dhcp_server container](https://github.com/sonic-net/sonic-buildimage/pull/16547)|GitHub issue/pull request detail|
|Enable dhcp_server container for BmcMgmtToRROuter in j2 template|sonic-buildimage||To be done|
|Update DHCPv4 relay start parameter when dhcp_server feature enabled|sonic-buildimage||To be done|
|Restart dhcpv4 relay when dhcp_server feature entry update|sonic-buildimage||To be done|
|Publish test plan|sonic-mgmt||To be done|
|Add config add/del CLI|sonic-buildimage||To be done|
|Add config update CLI|sonic-buildimage||To be done|
|Add config range add/del CLI|sonic-buildimage||To be done|
|Add config option add/del CLI|sonic-buildimage||To be done|
|Add config ip bind/unbind CLI|sonic-buildimage||To be done|
|Add config option bind/unbind CLI|sonic-buildimage||To be done|
|Add show info CLI|sonic-buildimage||To be done|
|Add show range/option CLI|sonic-buildimage||To be done|
|Add show lease CLI|sonic-buildimage||To be done|

PR #1651, opened 2024-04-02
DHCP DOS Mitigation HLD in SONiC

muhammadalihussnain:
This Design Provides DHCP DOS Attack Mitigation Support in HLD

| Repo  | PR title | State |
| ---------------------------------------------------| ----------------------------------------------- | ---------------------------------------|
| sonic-swss | [SwSS Chnages for DHCP DoS Mitigation Feature](https://github.com/Azure/sonic-swss/pull/3130 ) | ![GitHub issue/pull request detail] (https://img.shields.io/github/pulls/detail/state/Azure/sonic-swss/3130 [img.shields.io])|
| sonic-buildimage |  [YANG Support for DHCP  DoS Mitigation](https://github.com/Azure/sonic-buildimage/pull/18873)| ![GitHub issue/pullrequest detail] (https://img.shields.io/github/pulls/detail/state/Azure/sonic-buildimage/18873) [img.shields.io] | 
| sonic-utilities | [Utilities Changes for DHCP DoS Mitigation Feature](https://github.com/sonic-net/sonic-utilities/pull/3301 ) | ![GitHub issue/pull request detail] (https://img.shields.io/github/pulls/detail/state/Azure/sonic-utilities/3301 [img.shields.io]) |
| sonic-buildimage |   [DHCP DoS Logger for DHCP DoS Mitigation Feature](https://github.com/Azure/sonic-buildimage/pull/18947)| ![GitHub issue/pullrequest detail] (https://img.shields.io/github/pulls/detail/state/Azure/sonic-buildimage/18947) [img.shields.io]
| sonic-mgmt | [Test Plan and Test Case for DHCP DOS Mitigation Feature](https://github.com/sonic-net/sonic-mgmt/pull/12869)| ![GitHub issue/pullrequest detail] (https://img.shields.io/github/pulls/detail/state/Azure/sonic-mgmt/12869) [img.shields.io]


doc/Dhcp_Mitigation/DHCP Mitigation.md
```diff
@@ -0,0 +1,216 @@
...
+It can be seen in the screenshot below that a legitimate client (PC1) is unable to be serviced by the DHCP server. After the continuous bombardment of DHCP DISCOVER packets, the DORA process fails to complete, indicating a potential Denial of Service (DoS) situation.
+
+
+![image](../../images/dhcp_mitigation_hld/DORA.png)
+
+
+### DHCP DoS Mitigation
+
+
+To prevent possible DHCP DoS attacks in SONiC, we suggest using rate-limiting with Linux Traffic Control (TC).
```
Blueve, line 114: Can existing COPP rate-limiting policy mitigate this issue?
With COPP policy, the packets will not even trapped to CPU when exceeding pre-defined rate-limit.

+ @yaqiangz @kellyyeh to comment

yaqiangz: Yes, existing COPP can mitigate this issue. But I think advantage of this proposal is that it can configure per interface to narrow down the limit granularity

doc/Dhcp_Mitigation/DHCP Mitigation.md
```diff
@@ -0,0 +1,216 @@
...
+- No SAI API change or addition is needed for this HLD. 
+
+
+### CLI/YANG model Enhancements 
+
+##### CLI  Configuration Commands
+
+Proposed SONiC CLI commands (OPTION 1 - sonic-buildimage)
+- config dhcp_relay mitigation-rate add [interface] [number of packets]
+- config dhcp_relay mitigation-rate delete [interface] [number of packets]
```
yaqiangz, line 180: From background cmd, the rate limit would take effect on UDP dport 67, which would also effect packet to dhcp_server (https://github.com/sonic-net/SONiC/pull/1282), it would cause some confusion or misuse if it's under `dhcp_relay` sub-command.
`sudo tc filter add dev [Interface] protocol ip parent ffff: prio 1 u32 match ip protocol 17 0xff match ip dport 67 0xffff police rate [Byte Rate] burst [Byte Rate] conform-exceed drop`

doc/Dhcp_Mitigation/DHCP Mitigation.md
```diff
@@ -0,0 +1,216 @@
...
+
+
+![image](../../images/dhcp_mitigation_hld/add-seq.png)
+
+
+### Sequence Diagram to Del rate-limit 
+
+
+
+![image](../../images/dhcp_mitigation_hld/del-seq.png)
```
yaqiangz, line 166: In add-seq, if ingress qdisc does not exist, it would add new. Would it be better to align del to it that if there is not filter in qdisc, then delete this qdisc?

doc/Dhcp_Mitigation/DHCP Mitigation.md
muhammadalihussnain: As suggested, under dhcp_relay subcommand it will cause confusion and misuse, we will consider the rate-limit directly on the interface.

doc/Dhcp_Mitigation/DHCP Mitigation.md
muhammadalihussnain: Deleting the qdisc when there are no filters present seems like a logical step to maintain consistency. However, from a practical standpoint, unnecessarily creating and deleting qdiscs when adding and deleting rate limits can introduce unnecessary overhead and complexity.
In our opinion, keeping the qdisc even when empty is better than constantly deleting and recreating it whenever needed. It's simpler and saves computational tasks in the long run.

doc/Dhcp_Mitigation/DHCP Mitigation.md
muhammadalihussnain: Currently, COPP also resolves rate limiting but not specific focus for per interface. The proposed solution provides changes on per interface.

doc/Dhcp_Mitigation/DHCP Mitigation.md
```diff
@@ -0,0 +1,203 @@
...
+
+##### CLI  Configuration Commands
+
+Proposed SONiC CLI commands (sonic-utilities)
+-   config interface dhcp-mitigation-rate add [interface] [number of packets]
+-   config interface dhcp-mitigation-rate delete [interface] [number of packets]
+-   show interface dhcp-mitigation-rate
+
+Background Linux TC commands
+-   sudo tc qdisc add dev [Interface] handle ffff: ingress
```
yaqiangz, line 176: Current SONiC has integrated tc. If I understand correctly, this feature won't involve other dependencies or modules, only add cli in sonic-utilities?

muhammadalihussnain: Yes, We are only proposing changing in CLI

yaqiangz (approved):
LGTM

zhangyanzhao:
Community review recording https://zoom.us/rec/share/kZGlzxdhqBbPVIIwEDk_XJZ7mQMW9zj0V3ivIhvYV9GRizRwsKtiNL_wt11NisOe.Cbuvg7vUT2a9mKOu

ridahanif96:
@sudhanshukumar22  We have explored rate limit via ACL. Currently ACL doesn't have capability of rate limit. COPP doesn't rate-limit at interface level.
can you pls provide your suggestions, regarding implementation of rate limit at interface level for ASIC.

philo-micas:
A large number of DHCPv6 messages may also cause the resources of the DHCP Server to be consumed. Whether to consider supporting the rate limit of DHCPv6 messages?

muhammadalihussnain:
Hi @dgsudharsan @prsunny @venkatmahalingam @lguohan We have modified design. We are proposing change in Behavior to work with this design by removing DHCP rate limit of 300 packets per second implemented through CoPP and replacing it with port-level rate limits implemented in the kernel via Linux Traffic Control.
Please review and suggest should we bring it again to Community review.  We are open for design modifications/suggestions

ridahanif96:
@zhangyanzhao Hi Yanzhao, can you assign reviewer for this design. We have updated HLD as per the comments/suggestions during HLD Community review.

doc/Dhcp_Mitigation/DHCP Mitigation.md
```diff
@@ -0,0 +1,252 @@
...
+DHCP traffic can be filtered and rate-limited by dropping all packets that exceed a user-defined rate, allowing legitimate clients to be serviced by a DHCP server despite an ongoing attack. This design provides a mechanism for DHCP rate-limiting on a specified port. Applying DHCP rate limit on a specific port requires two parameters:
+
+- port
+  The port on which the DHCP rate limit is to be applied.
+- rate
+ (packets per second)An integer specifying a DHCP packet rate in packets per second. 
+
+Since traffic control(TC) only supports rates in the form of bytes per second, this value is multiplied by 406 (number of bytes that make up a DHCP discover packet).
+Upon running this command, an ingress queuing discipline is created on the specified port via traffic control(TC). Next, a traffic control(TC) filter is added to filter DHCP discover packets on protocol 17 (UDP) and destination port 67 (port used by DHCP) and a dropping action is applied to filtered incoming traffic. Incoming DHCP discover packets that exceed the rate are dropped to stop the attack from overwhelming the DHCP server.
+
```
Yarden-Z, line 150: Incoming dropped packets are logged in the ipfilter log?
Maybe we can add something that if something passes over this threshold we should perform some extra action like raise a warning?

doc/Dhcp_Mitigation/DHCP Mitigation.md
```diff
@@ -0,0 +1,252 @@
...
+     leaf dhcp_rate_limit {
+   	        description: "dhcp rate limit ;
+   				  type uint32 {
+   					   range 0..8000;
+   				 }
+   			 }
+
+
+### Testing Requirements/Design
```
Yarden-Z, line 233: Where will the test cases to check functionality be defined? Here there are only CLI test definitions

doc/Dhcp_Mitigation/DHCP Mitigation.md
```diff
@@ -0,0 +1,252 @@
...
+No SAI API change or addition is needed for this HLD. 
+
+
+### CLI/YANG model Enhancements 
+
+##### CLI  Configuration Commands
+
+Proposed SONiC CLI commands (sonic-utilities)
+-   config interface dhcp-mitigation-rate add [port] [packet-rate]
+-   config interface dhcp-mitigation-rate delete [port] [packet-rate]
```
Yarden-Z, line 213: Why do we need the [packet-rate] here? If we delete it then I assume we do not need to specify the packet rate value

doc/Dhcp_Mitigation/DHCP Mitigation.md
asraza07: Hi @Yarden-Z , yes the test cases mentioned here are for CLI which will be implemented in the sonic-utilities repo. Other than these, we plan on contributing test plans and test case code to the sonic-mgmt repo for functionality testing of this feature.

doc/Dhcp_Mitigation/DHCP Mitigation.md
asraza07: Hi @Yarden-Z ,
Yes, logically we do not need to specify the packet rate to delete it in this case, however, we implemented the command this way because this is the standard convention used in other SONiC commands (where the value to be removed is always specified in the command, eg. vlan del, interface IP remove, etc). We wanted to keep the command in line with the current CLI behavior of SONiC to maintain consistency.

doc/Dhcp_Mitigation/DHCP Mitigation.md
muhammadalihussnain: Hi @Yarden-Z , No, Incoming dropped packets are not  logged in the ipfilter log for  Linux systems. iptables can do the task but then we have to mark those packets with traffic control utility and then using iptables we can log about those marked packets " drop by tc" . but tc does not have the ability to log. 
Combination of tc and iptables can be used for logs but in this approach we can not drop the packets, which is the reqiurement to mitigate DHCP DoS Attack.

Yarden-Z: Then if this occurrence happens - how is the user notified of this?
If this is only done and not logged/notified - then a whole attack scenario might be missed.
We will block it, but we will not allow the users to have any information and data about their network and about a potential attack.
In addition - if this happens due to legitimate reasons (large cluster bring up) this might fail as well but without any info to the user.

muhammadalihussnain: User can be notifiy via show command. In the show command, user can see the number of dropped packets on a specific interface.

asraza07:
> A large number of DHCPv6 messages may also cause the resources of the DHCP Server to be consumed. Whether to consider supporting the rate limit of DHCPv6 messages?

Hi @philo-micas , currently our design supports DHCPv4 only. In the future, we plan on adding support for DHCPv6 rate-limiting.

asraza07:
Hi @yaqiangz @Yarden-Z , We have updated the HLD according to suggestions. Kindly help review it.

ridahanif96:
@Yarden-Z Hi Yarden, We have updated HLD PR. can you please help review updated HLD and linked code PRs. Thanks in advance.

ridahanif96:
@yaqiangz Hi Yaqiang, We have updated HLD PR. can you please help review updated HLD and linked code PRs. Thanks in advance.

doc/Dhcp_Mitigation/DHCP Mitigation.md
```diff
@@ -0,0 +1,296 @@
...
+
+- port
+  The port on which the DHCP rate limit is to be applied.
+- rate
+ (packets per second)An integer specifying a DHCP packet rate in packets per second. 
+
+Since traffic control(TC) only supports rates in the form of bytes per second, this value is multiplied by 406 (number of bytes that make up a DHCP discover packet).
+Upon running this command, an ingress queuing discipline is created on the specified port via traffic control(TC). Next, a traffic control(TC) filter is added to filter DHCP discover packets on protocol 17 (UDP) and destination port 67 (port used by DHCP) and a dropping action is applied to filtered incoming traffic. Incoming DHCP discover packets that exceed the rate are dropped to stop the attack from overwhelming the DHCP server.
+
+A custom program (daemon) has been developed to monitor network traffic. This daemon continuously checks all network ports (tc qdisc) for signs of packet drops. If it detects increase in dropped packets on a specific port, it logs a message to a designated file. This log message serves as an alert for the user, notifying them about potential congestion and packet loss issues on the affected port.
```
yaqiangz, line 151: If there is a copp configuration exist, I guess to user, the min value of limit by tc and copp would take effect. If copp limit is lower than tc, what would the daemon perform? 2 scenarios I am curious: 1) copp limit <packets rate < tc limit. 2) copp limit < tc limit < packets rate

doc/Dhcp_Mitigation/DHCP Mitigation.md
```diff
@@ -0,0 +1,296 @@
...
+    "PORT":
+      {
+        "Ethernet0": {
+            "admin_status": "up",
+            "alias": "fortyGigE0/0",
+            "index": "0",
+            "lanes": "25,26,27,28",
+            "mtu": "9100",
+            "speed": "40000",
+            "dhcp_rate_limit" : "300"
```
yaqiangz, line 196: Whether the unit of rate same as copp? If in COPP configure the cir and cbs are 300, are they mean same rate limit with your configuration?

asraza07: Hi @yaqiangz , yes it is the same rate unit. In COPP, the default rate limit is 300 packets/sec which is the same rate limit in our configuration. Since we plan on replacing default COPP with our configuration, we chose to give a default value of 300 packets/sec to ensure backward compatibility.

doc/Dhcp_Mitigation/DHCP Mitigation.md
asraza07: Hi @yaqiangz , yes you are right, if both TC and COPP are configured, the minimum value will take effect. For now, the monitor daemon only reads packets dropped due to TC. If a lower COPP rate is configured, then packets will be dropped by COPP, not TC. In this case the monitor daemon will not report dropped packets.
#### Scenario 1: copp limit < packets rate < tc limit
In this scenario, only COPP will drop packets and the TC limit will go unused as packets rate is less than TC limit, so the monitor daemon will not report packets dropped by COPP.
#### Scenario 2: copp limit < tc limit < packets rate
In this scenario, since the COPP limit is lower than TC and packets go through COPP first (since it is in hardware) only COPP limit will take effect, so the monitor daemon will not report packets dropped by COPP.

To avoid this, our design is aimed to replace the default COPP DHCP rate limit of 300 so that by default COPP and TC do not exist at the same time. We did this because TC allows for more precise rate limiting at an interface level so it holds an advantage over COPP in the event of a DoS attack.

If you have any suggestions on how we can modify the monitor daemon to also record COPP dropped packets, then please share, we will try to add it to our design.

Kindly also review the linked code PRs to this HLD, thanks in advance.

zhangyanzhao:
late coming features with multiple code PRs, defer to next release per release triage

doc/Dhcp_Mitigation/DHCP Mitigation.md
yaqiangz: I checked change you added in db_migrater https://github.com/sonic-net/sonic-utilities/pull/3301/files. If upgraded from a version that doesn't support/configure tc rate limit, it would be configured as 300. You mentioned that `by default COPP and TC do not exist at the same time.` But from current PRs they will exist at the same time? In this scenario, copp is set to 300 or 100, and tc rate limit is 300. https://github.com/sonic-net/sonic-buildimage/blob/74317df1b07f778d8b7b21159afc92a7583c1760/files/image_config/copp/copp_cfg.j2#L37. TC configuration and log wouldn't take effect. Would it be better to keep it empty when migrating if previous OS didn't contain this value?

Based on above, if we agree to keep empty by default for this configure. I wonder whether the log daemon process run all the time even if there is not any tc limit exist?

doc/Dhcp_Mitigation/DHCP Mitigation.md
yaqiangz: I checked change you added in db_migrater https://github.com/sonic-net/sonic-utilities/pull/3301/files. If upgraded from a version that doesn't support/configure tc rate limit, it would be configured as 300. You mentioned that by default COPP and TC do not exist at the same time. But from current PRs they will exist at the same time? In this scenario, copp is set to 300 or 100, and tc rate limit is 300. https://github.com/sonic-net/sonic-buildimage/blob/74317df1b07f778d8b7b21159afc92a7583c1760/files/image_config/copp/copp_cfg.j2#L37. TC configuration and log wouldn't take effect. Would it be better to keep it empty when migrating if previous OS didn't contain this value in the scenario that DHCP copp exist? And add this configuration on demand

Based on above, if we agree to keep empty by default for this configure. I wonder whether the log daemon process run all the time even if there is not any tc limit exist?

asraza07: Hi @yaqiangz , we had planned to depreciate the default DHCP copp configuration as mentioned in our HLD document. We have now made the PR and linked it to this HLD. Our design ensures that by default only TC rate limits will be configured so that both TC and COPP do not exist at the same time. In this scenario, the log daemon process will run properly with no complications.

asraza07:
Hi @Yarden-Z @zhangyanzhao , kindly help review this PR. Thanks in advance!

yaqiangz:
> Hi @Yarden-Z @yaqiangz , kindly help review this HLD. Thanks in advance!

Hi @asraza07, I want to confirm that with COPP existing, this proposal would only take effect in the scenario: [tc rate limit < actual packet rate < COPP rate limit]. And in my understanding, current proposal cannot mitigate ddos attack that [actual packet rate > COPP rate limit], correct? I wonder whether you have plan or insights for this scenario?

asraza07:
> > Hi @Yarden-Z @yaqiangz , kindly help review this HLD. Thanks in advance!
> 
> Hi @asraza07, I want to confirm that with COPP existing, this proposal would only take effect in the scenario: [tc rate limit < actual packet rate < COPP rate limit]. And in my understanding, current proposal cannot mitigate ddos attack that [actual packet rate > COPP rate limit], correct? I wonder whether you have plan or insights for this scenario?

Hi @yaqiangz , you are correct, DHCP DoS mitigation is only possible in the scenario where actual packet rate is less than the COPP rate limit. We have designed this feature on the premise that COPP is not intended to mitigate DHCP DoS attacks; it is there to regulate the flow of control plane and ensure control plane stays intact. Our feature is there to mitigate DHCP DoS attacks (only possible via the kernel). Without removing COPP, we can say that DHCP DoS mitigation is only possible when packet rate does not exceed COPP limit, because otherwise COPP starts dropping DHCP packets including legitimate clients.

Yarden-Z (approved):
Looks ok

yaqiangz (approved):
LGTM

ridahanif96 (approved):

zhangyanzhao:
not all code PRs are merged, move to backlog

ridahanif96:
> not all code PRs are merged, move to backlog

@zhangyanzhao 
Code PRs are done since long and approved but still awaiting for merge. We are already pushing hard for repo owner to merge it.

ridahanif96:
> > not all code PRs are merged, move to backlog
> 
> @zhangyanzhao 
> Code PRs are done since long and approved but still awaiting for merge. We are already pushing hard for repo owner to merge it.

@qiluo-msft kindly help merge PRs.

zhangyanzhao:
Two code PRs are still open. Please help to review and merge. Thanks.

muhammadalihussnain:
@yaqiangz please help us to get 2 related PRs meged. release is very near. thanks

zhangyanzhao:
One code PR is not merged yet, move the feature status to "MoveToBacklog" @anilkpandey

PR #1659, opened 2024-04-15
Improve Route performance with ZMQ

liuh-80:
Improve Route performance with ZMQ


stepanblyschak:
@liuh-80 RouteOrch also publishes route to APPL_STATE_DB - https://github.com/sonic-net/sonic-swss/blob/master/orchagent/routeorch.cpp#L3162 for route programming feedback to zebra.
This is using Redis Pipeline, do you plan to move it to ZMQ?

qiluo-msft:
> @liuh-80 RouteOrch also publishes route to APPL_STATE_DB - https://github.com/sonic-net/sonic-swss/blob/master/orchagent/routeorch.cpp#L3162 for route programming feedback to zebra. This is using Redis Pipeline, do you plan to move it to ZMQ?

No plan to move RouteOrch->APPL_STATE_DB to ZMQ.

PR #1687, opened 2024-04-30
Update Copp Manager Redesign test plan

JibinBao:
Update Copp Manager Redesign test plan
Add detailed information for traffic validation according to the change of https://github.com/sonic-net/sonic-mgmt/pull/11819

### DIFF
--- a/doc/copp/Copp_Manager_Redesign_test_plan.md
+++ b/doc/copp/Copp_Manager_Redesign_test_plan.md
@@ -80,9 +80,11 @@ The test will be supported on ptf32, ptf64, t1 and t2.
 ## Test cases
 ### Test cases #1 -  Verify 4 traps'default configuration
 1. Verify the 4 traps(arp, lacp, udld, ip2me) are installed  by sending traffic with the default config
+   (For traffic validation, please see the Notes)
 
 ### Test cases #2 - Add a new trap
 #### Pre-condition: The tested trap should be uninstalled on dut.
+#### Remove ip2me because bgp traffic can fall back to ip2me trap then interfere following traffic tests
 1. Set always_enabled of bgp to true with following cmds:
    
   + generate copp config file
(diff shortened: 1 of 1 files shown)


nhe-NV (approved):

Blueve (dismissed):

doc/copp/Copp_Manager_Redesign_test_plan.md
```diff
@@ -80,9 +80,11 @@ The test will be supported on ptf32, ptf64, t1 and t2.
 ## Test cases
 ### Test cases #1 -  Verify 4 traps'default configuration
 1. Verify the 4 traps(arp, lacp, udld, ip2me) are installed  by sending traffic with the default config
+   (For traffic validation, please see the Notes)
 
 ### Test cases #2 - Add a new trap
 #### Pre-condition: The tested trap should be uninstalled on dut.
+#### Remove ip2me because bgp traffic can fall back to ip2me trap then interfere following traffic tests
```
Blueve, line 87: BRCM ASIC based platform has an invisible default trap which will cause traffic been trapped even we remove ip2me. I suggest you update the statement and just mention the ip2me as an example.

They commit to update test case to adapt them ASIC behavior.

JibinBao: Change as below. Is it ok?
Since different vendor's ASIC might have some different invisible default traps will cause traffic been trapped, we alsoe need to remove them.
For example: Remove ip2me because bgp traffic can fall back to ip2me trap then interfere following traffic tests

JibinBao: Change as below. Is it ok?
Since different vendor's ASIC might have some different invisible default traps will cause traffic been trapped, we alsoe need to remove them. For example:  Remove ip2me because bgp traffic can fall back to ip2me trap then interfere following traffic tests

JibinBao: @Blueve 
Fixed.
Can you review it again?
Thanks

Blueve (approved):

prabhataravind:
Hi @JibinBao, my apologies for the delayed review.  I have seen cases where packets get trapped to default trap group (for eg: when TTL for incoming packet is 1 as is usually the case with BGP packets) in some platforms after removing both bgp and ip2me traps. Please make sure this sceanrio/behavior is accounted for in the redesign.

PR #1707, opened 2024-06-04
SRv6 Policy HLD 

eddieruan-alibaba:


eddieruan-alibaba:
https://github.com/eddieruan-alibaba/SONiC/blob/eruan-srv6p/doc/srv6/srv6_policy.md

HLD link

doc/srv6/srv6_policy.md
```diff
@@ -0,0 +1,511 @@
...
+127.0.0.1:6380> hgetall "PIC_CONTEXT_TABLE:8657"
+1) "nexthop"
+2) "2064:100::1d,2064:100::1e"
+3) "vni_label"
+4) "na,na"
+5) "vpn_sid"
+6) "fd00:201:201:fff1:11::,fd00:202:202:fff2:22::"
+7) "seg_src"
+8) "2064:100::1f,2064:100::1f"
+```
```
kperumalbfn, line 236: Does fpmsyncd create NHGroup table based on PIC_CONTEXT_TABLE?

doc/srv6/srv6_policy.md
```diff
@@ -0,0 +1,511 @@
...
+
+## Candidate path's state machine
+## Candidate path without bfd configuration
+In this case, the candidate path's default state is UP. Pathd would not be able to manage cpath's state. It would be external controller's job to manage cpath's state.
+
+## Candidate path with bfd configuration
+In this case, the default state is DOWN. bfd notification would be used to update cpaths' state. 
+
+### Policy with endpoint
+bfd packet would be constructed with SRv6 header which would go via both specified cpath and engpoint. The protection path would be via specified SID notes and the end point node. This protection would not cover link if the SID list is not hop by hop.
```
kperumalbfn, line 325: Does it make use of h/w offloaded BFD?

doc/srv6/srv6_policy.md
```diff
@@ -0,0 +1,511 @@
...
+
+## Candidate path with bfd configuration
+In this case, the default state is DOWN. bfd notification would be used to update cpaths' state. 
+
+### Policy with endpoint
+bfd packet would be constructed with SRv6 header which would go via both specified cpath and engpoint. The protection path would be via specified SID notes and the end point node. This protection would not cover link if the SID list is not hop by hop. 
+
+### Policy without endpoint, a.ka. Color Only Policy
+
+bfd packet would be constructed with SRv6 header which would go via the specified cpath.
```
kperumalbfn, line 329: Hardware offloaded BFD with SRV6 header will require SAI BFD object changes.

doc/srv6/srv6_policy.md
eddieruan-alibaba: PIC_CONTEXT_TABLE is for VPN SID, not for TE/ Policy, which is discussed in https://github.com/eddieruan-alibaba/SONiC/blob/eruan-pic/doc/pic/bgp_pic_edge.md#high-level-design

doc/srv6/srv6_policy.md
eddieruan-alibaba: Both software and hw offload, depending on BFD timer and scale.

doc/srv6/srv6_policy.md
shuaishang: Yes，the SAI changes has been merged.
https://github.com/opencomputeproject/SAI/pull/1759

BorisH2018:
Hi,
Eddie , here are few more comments additionally to those which I made on Thursday's call.

1) Binding SID (BSID) is not ,mentioned at all in the HLD. RFC 9256  (SR Policy Architecture) 6.1 says: "Each candidate path MAY be defined with a BSID."  So in forwarding plane BSID can be mapped into an another Segment List (NHG). IMO, this functionality is quite important. Will it be included in some later version?

2) FRR SRv6 Policy Configuration section
1. Add "ipv6-address" as segment type for segment list for SRv6. Currently, FRR CLI only has "mpls" and "nai" as the segment type.

May be would be better to say explicitly that Type B  segment type (item 4. RFC 9256) is supported in FRR)?

3)  HLD mentions BFD, why can't we also use S-BFD?

4) I also have not seen in configuration the discriminator of CP (2.5, RFC 9256). Is it in roadmap?
5) Also to align with RFC 9256, would be better to mention that HLD currently focuses on explicit CPs.
6) Forwarding information NHG's schema
"Zebra maintains a SRv6 Policy NHG in two levels. The first level is for ECMP over multiple polices, and the second level is for ECMP over multiple candidate paths." -> It is not clear, why do we need ECMP over multiple policies? Do you mean a composite CP (2.2 RFC 9256) ? It is quite rare implementation case and would be good to clarify it. Also according to the configuration a 1st level of an ECMP is an ECMP over SLs.

Thank you.

SY,
Boris

PR #1717, opened 2024-06-17
Smartswitch Dash PA Validation offload to NPU

kperumalbfn:
Offload DASH PA Validation rules to NPU

[schema] add a set of SmartSwitch related tables  - https://github.com/sonic-net/sonic-swss-common/pull/947
[zmq] add proxy mode to the ZmqServer - https://github.com/sonic-net/sonic-swss-common/pull/948
[DASH] add DASH offload manager and PA validation offload - https://github.com/sonic-net/sonic-swss/pull/3358
[dash] add zmq_dpu_proxy_address_base parameter to telemetry.go https://github.com/sonic-net/sonic-gnmi/pull/324
[DASH] enable offload manager on Nvidia SmartSwitch https://github.com/sonic-net/sonic-buildimage/pull/20714


doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,92 @@
...
+- Provide PA validation support in NPU when this feature is not supported in DPU pipeline model.
+- Use existing ACL tables and ACL rules design in NPU.
+- New orchagent(DashOffloadOrch) is added to support this feature and future offloads to NPU.
+
+## 1.2 Scale Requirements
+
+# 2 Modules Design
+
+## 2.1 STATE_DB changes(per-DPU)
+New table in STATE_DB(DASH_OFFLOAD_STATE_TABLE) is added to inform NPU about DPU SAI capability and whether NPU offload is required for certain feature. During SAI initialization, DashOrchagent queries SAI API(SAI_API_DASH_PA_VALIDATION) and sets the state_db table with the NPU offload state.
```
ganglyu, line 47: Can we put this table in DPU_APPL_STATE_DB? Is this on NPU or DPU?

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,92 @@
...
+
+| Table             | Key              | Field                   | Description
+|:-----------------:|:----------------:|:-----------------------:|:--------------:|
+|DASH_OFFLOAD_STATE_TABLE |
+|                   | DPU_ID           |                         |                |
+|                   |                  | pa_validation           |  "true" - NPU offload required|
+|                   |                  |                         | "false" - NPU offload not required|
+
+
+## 2.2 APPL_STATE_DB
```
ganglyu, line 57: This database should be DPU_APPL_STATE_DB.

kperumalbfn:
@qiluo-msft @ganglyu @liuh-80 Please review the GNMI/ZMQ changes for Dash offload

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,149 @@
...
+<img src="images/DashOffloadManager.svg">
+
+Once the offload is required, the DashOffloadManager will start designated orch (e.g. PAValidationOffloadOrch) that will subscribe to the configuration and do the offload logic.
+
+<img src="images/DashOffloadManagerWithConsumer.svg">
+
+The Dash Offload Manager is disabled by default and only enabled for specific platforms that require its functionality.
+
+### 2.1.1 SONiC GNMI server changes
+A new parameter `zmq_dpu_proxy_address_base` is added to the telemetry.go to enable the GNMI Server -> DASH Offload Manager ZMQ connection. If specified, the GNMI server will use `zmq_dpu_proxy_address_base + dpu_index:zmq_port` as a destination for DPU ZMQ connection. For example, on the system with 2 DPUs and base address 127.0.0.10, the DASH Offload Manager's ZMQ server runs on 127.0.0.10:8100 for DPU0 and on 127.0.0.11:8100 for DPU1.
```
liuh-80, line 62: Currently GNMI already get DPU address by ConfigDB "DPUS" table.
Why make this change to get DPU by base address?

https://github.com/sonic-net/sonic-gnmi/blob/e8449250bec233c81cf4b6aa324c26da7fea5732/sonic_data_client/mixed_db_client.go#L110

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,149 @@
...
+- Provide PA validation support in NPU when this feature is not supported in DPU pipeline model.
+- Use existing ACL tables and ACL rules design in NPU.
+
+## 1.2 Scale Requirements
+
+# 2 Modules Design
+
+## 2.1 Dash Offload Manager
+The new orchagent application DashOffloadManager will be responsible for DASH offloading logic. It will collect all the needed information for offloading and perform all the relevant configurations.
+To get the DASH configuration that should be offloaded, the DashOffloadManager will act as a transparent ZMQ proxy between the GNMI server and the DPU swss, forwarding all the configuration and intercepting the tables that should be offloaded.
```
ganglyu, line 50: Why is the ZMQ proxy necessary? Could you subscribe to DPU_APPL_DB instead?

Yakiv-Huryk: The idea is to have control over what is sent to the DPU, which is a more flexible tool for implementing different kinds of offload-related functionality.

In case of PA validation offload, we don't forward the PA Validation table entries to the DPU (since DPU has no use of it). This saves some NPU<->DPU bandwidth and allows to keep the DPU side simple (no entries - nothing to create). 

In the future, we can extend this infrastructure also to alter the content of the configuration (if such a need arises for some other offload feature)

ganglyu: GNMI client<->GMMI server<->ZMQ<->DPU orchagent<->NPU APPL_STATE_DB

The DPU orchagent should update the results to the DPU_APPL_STATE_DB on the NPU. If we don't forward the PA Validation table entries to the DPU, the client won't know if this configuration failed or was intercepted.

Yakiv-Huryk: Are you referring to the flow described here https://github.com/sonic-net/SONiC/pull/1759 ?
As I understand, the feedback design is not finalized yet. Once it is, I'll have to add the same logic to the offload flow to also fill the result into the DPU_APPL_STATE_DB for the offloaded (intercepted) PA validation entries. Is it ok?

ganglyu: Could you please include this in the design?

ganglyu:
@qiluo-msft 
This design change is expected to impact our ZMQ performance. Can we accept this impact?

ganglyu:
@Yakiv-Huryk 
This design change is expected to impact our ZMQ performance. Could you please evaluate the ZMQ performance before and after the change?

Yakiv-Huryk:
> @Yakiv-Huryk This design change is expected to impact our ZMQ performance. Could you please evaluate the ZMQ performance before and after the change?

I did the following test:

Sending 10k DASH_PREFIX_TAG_TABLE entries (each having 100 IPs). 
The DASH_PREFIX_TAG_TABLE has the fastest processing on the DPU (the entries are just saved into the memory). This way we make sure that the test is impacted as little as possible by the end consumer of the config (DPU swss)
I've used the py_gnmicli.py from https://github.com/lguohan/gnxi as a client.

I measured the following:
* the time it takes to send a config (to see if there is any back pressure from gnmi server)
* the time it takes between the first and the last entry to be processed on the DPU
* the cpu/dram usage for the gnmi process (telemetry.go) and zmq-proxy (dashoffloadmanager)

**Test without proxy**
* GNMI client config apply time: ~16sec (+/-1sec)
* DPU difference between first and last entry processed: ~16sec (+/-1sec)

Telemetry.go CPU/RAM:
telemetry-no-proxy2


**Test with proxy**

* GNMI client config apply time: ~16sec (+/-1sec) (!)
* DPU difference between first and last entry processed: ~16sec (+/-1sec) (!)

Telemetry.go CPU/RAM:
telemetry-with-proxy2

DashOffloadManager(zmq-proxy) CPU/RAM:
proxy2


The bottom line is that there is no measurable difference (at least via this test methodology).
The main difference is that with proxy, there is another process (another stage in the gnmi_client->gnmi_server->zmq_proxy->DPU SWSS) which doesn't impact the overall bandwidth of the config we can process.
The proxy does however use some CPU, in the extreme case (during a burst of config received) it will use 100% of a single CPU core.

To explain the behavior, I want to emphasize that the config goes through the pipeline (gnmi_client->gnmi_server->zmq_proxy->DPU SWSS), and the bandwidth of the pipeline is limited by its slowest stage. The proxy is the fastest stage here since it doesn't process the data (simply sending the raw data into zmq socket).

If you have any testing ideas/scenarios you want me to do, please share.

kperumalbfn:
@Yakiv-Huryk could you update all the Sonic swss PRs in this HLD description.

ganglyu:
@r12f would you please review this PR?

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,172 @@
...
+| 0.1 |  06/10/2024 |     Kumaresh Perumal  | Initial version                  |
+| 0.2 |  08/21/2024 |          Yakiv Huryk  | Added DashOffloadManager         |
+
+
+# About this Manual
+This document provides general information about offloading PA validation feature to NPU in Smartswitch.
+
+# Definitions/Abbrevations
+
+|                          |                                          |
```
r12f, line 29: missing table headers

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
r12f: I am having the same concern here.... This seem to be complicated to implement, because of a few things:

1. The DASH offload manager  or the NPU side DASH orchs need to provide the exact same feedback loop as the swss. If anything changed in swss, they need to be changed as well, which can be easily missed and causing problem.
2. In order to support independent DPU upgrades, each DPU will needs to have its own DASH offload manager and all the DASH orchs in the NPU side.
3. Dependency and object handling can also be a problem. This solution is trying to provide a generic way to handle all the DASH object offloading in the future, but I feel it doesn't really do the job. Explicit PA validation rule might be the simpliest case, where the only thing that we needs to do is to redirect the rules into the NPU side. However, other DASH objects can have dependencies, e.g., Implicit PA validation rules are coming from VNET + CA-PA mappings. In this case, we cannot simply redirect the rules into the NPU side, but have to copy it, because they are also used in the outbound pipeline. 

Overall, I feel Gang is correct. The other way that Gang proposed there is actually much more cleaner and maintainable. All we need is just a if case in the swss, and every other things can be reused, such as feedback loops.

r12f: @ganglyu , please let us know your thought.

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,172 @@
...
+To get the DASH configuration that should be offloaded, the DashOffloadManager will act as a transparent ZMQ proxy between the GNMI server and the DPU swss, forwarding all the configuration and intercepting the tables that should be offloaded.
+To simplify the management of the DPU offload and achieve optimal performance, every DPU is handled by a separate instance of a ZMQ Proxy (pair of ZMQ Server and Client)
+
+<img src="images/DashOffloadManager.svg">
+
+Once the offload is required, the DashOffloadManager will start designated orch (e.g. PAValidationOffloadOrch) that will subscribe to the configuration and do the offload logic.
+
+<img src="images/DashOffloadManagerWithConsumer.svg">
+
+The Dash Offload Manager is disabled by default and only enabled for specific platforms that require its functionality.
```
r12f, line 69: do we have any logic to update the gnmi server to point to the proxy or the dpu?

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,172 @@
...
+# 2 Modules Design
+
+## 2.1 Dash Offload Manager
+The new orchagent application DashOffloadManager will be responsible for DASH offloading logic. It will collect all the needed information for offloading and perform all the relevant configurations.
+To get the DASH configuration that should be offloaded, the DashOffloadManager will act as a transparent ZMQ proxy between the GNMI server and the DPU swss, forwarding all the configuration and intercepting the tables that should be offloaded.
+To simplify the management of the DPU offload and achieve optimal performance, every DPU is handled by a separate instance of a ZMQ Proxy (pair of ZMQ Server and Client)
+
+<img src="images/DashOffloadManager.svg">
+
+Once the offload is required, the DashOffloadManager will start designated orch (e.g. PAValidationOffloadOrch) that will subscribe to the configuration and do the offload logic.
```
r12f, line 65: where is the data being stored in NPU side? is it in NPU database or DPU database? If in NPU database, the schema will need to be changed, but I think it is not covered in the spec.

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+<img src="images/DashOffloadAcl.svg">
+
+### 2.2.1 PA Validation Offload GNMI feedback
+To preserve the GNMI feedback behavior, the offload logic must also create an entry in the DPU's APPL_STATE_DB.
+For each PA validation processed, the PaValidationOffloadOrch creates the following entry:
+
+```
+DASH_PA_VALIDATION_TABLE:{{vni}}
```
r12f, line 154: does this mean the NPU side PAValidationOffloadOrch will write into the DPU database?

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,172 @@
...
+    "result": {{result}}
+```
+
+The result is 0 for success, or > 0 for error code
+
+Please refer to https://github.com/sonic-net/SONiC/pull/1759 for more details regarding GNMI feedback requirements and behavior.
+
+## 2.3 DPU Shut/Restart
+When DPU goes down/restarts, the ACL configuration should be cleaned. It's done by the Dash Offload Manager which listens to the ChassisStateDB DPU_STATE Table. When it detects that the DPU is down (dpu_control_plane_state is down), the PaValidationOffloadOrch is deinitialized, leading to ACL configuration cleanup and ZMQ proxy subscription removal.
```
r12f, line 163: we might also need to cover the DPU upgrade case, because when chassis db is updated, this orch might not be running at all.

@ganglyu  to confirm.

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,172 @@
...
+The offloading is achieved by using a User-Defined EGRESS ACL table bound to the NPU<->DPU port. The table is created/managed per DPU.
+
+```
+{
+    "ACL_TABLE_TYPE": {
+        "DASH_PA_VALIDATION": {
+            "MATCHES": [
+                "TUNNEL_VNI",
+                "SRC_IP",
+                "SRC_IPV6"
```
r12f, line 89: it will be better to add the destination IP in here as well. The reason is because SmartSwitch lives in T1 and can receive other traffic in the same VNET, which is not sent to the DPU, but to other VMs. Adding the SmartSwitch data plane VIP as destination IP will be safer and more future-proof.

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
ganglyu: I'm not sure if this design is necessary. If the DPU doesn't need the PA validation table and only the NPU requires it, we can configure this table directly for the NPU.

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
ganglyu: Theoretically, we need to upgrade this orch when we upgrade the DPU.

r12f:
and.... this design looks already implemented....

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
```diff
@@ -0,0 +1,172 @@
...
+  - 64 ACL drop rules (one per each unique VNI in the configuration)
+
+# 2 Modules Design
+
+## 2.1 Dash Offload Manager
+The new orchagent application DashOffloadManager will be responsible for DASH offloading logic. It will collect all the needed information for offloading and perform all the relevant configurations.
+To get the DASH configuration that should be offloaded, the DashOffloadManager will act as a transparent ZMQ proxy between the GNMI server and the DPU swss, forwarding all the configuration and intercepting the tables that should be offloaded.
+To simplify the management of the DPU offload and achieve optimal performance, every DPU is handled by a separate instance of a ZMQ Proxy (pair of ZMQ Server and Client)
+
+<img src="images/DashOffloadManager.svg">
```
r12f, line 63: I assume this is after the gnmi splitter, will be better to make it more clear how it works with the independent dpu upgrade changes.

doc/smart-switch/PA-Validation/SmartSwitchPAValidationOffload.md
kperumalbfn: @r12f These ACLs are attached to DPU connected ports in the Egress direction. So, forwarding rule(based on DPU VIP route or ENI-ACL based forwarding) makes sure that only DPU destined traffic are egressing NPU towards DPU. So, dest_ip is not required. It will also increase the TCAM key width and may impact the scale numbers for this ACL and also other T1 specific Egress ACLs.

r12f: Got it! Egress ACL should work!

KrisNey-MSFT:
hi @r12f - do we still need this PR if it's already been implemented?  Just checking for tracking purposes...thx!

r12f:
Hi Kristina, yes. We had a meeting to sync on this, and we will do a presentation in the smart switch group meeting to get everyone on the same page soon. It was causing some surprises in the discussions.

KrisNey-MSFT:
[surprised]     Kristina Moore reacted to your message:

KrisNey-MSFT:
> Hi Kristina, yes. We had a meeting to sync on this, and we will do a presentation in the smart switch group meeting to get everyone on the same page soon. It was causing some surprises in the discussions.

Sounds like a plan

KrisNey-MSFT:
hi @r12f - was this one presented in the SmartSwitch meeting, by chance?

KrisNey-MSFT:
hi @r12f - was this one presented in the SmartSwitch meeting?  Just checking to see if it still needs to be open.  TY...

KrisNey-MSFT:
hi @kperumalbfn - - was this one presented in the SmartSwitch meeting? Just checking to see if it still needs to be open. TY...

KrisNey-MSFT:
Per @kperumalbfn Hi Kristina Moore,keep this PR open. We have already presented this in our WG.  The functionality is not required now as Nvidia will change their P4 pipeline to add PA validation functionality in DPU. But this HLD will be useful for other offload functionalities that come in future.

KrisNey-MSFT:
hi @kperumalbfn - is NVidia still working on PA validation?  Just checking in to see if we want to keep this one open still.  TY!

KrisNey-MSFT:
hi @kperumalbfn - checking back please, do we need to keep this one open?  TY!

prsunny:
@KrisNey-MSFT , as i see, the linked PRs are closed. We probably would need to revisit this in future when this enhancement is needed.

KrisNey-MSFT:
Thanks very much - I saw that the linked ones were closed; I just don't want to close any authors' PR w/o asking them first, in case they are using it somehow.   I appreciate you responding!

KrisNey-MSFT:
Per @prsunny closing PR: revisit this in future when this enhancement is needed.

KrisNey-MSFT:
hi @kperumalbfn - per @prsunny comment, would you please close this PR?

PR #1719, opened 2024-06-20
Update for Supermicro platform

puffc:
Add Supermicro SSE-T7132S 32x400G switch


puffc:
@kannankvs Would you please review this change?

PR #1778, opened 2024-08-09
[SmartSwitch]: Fix typo on ip address assignment doc

Pterosaur:
According to the yang model, the subtype should be used to indicate that this is SmartSwitch.


wen587 (approved):
lgtm

PR #1784, opened 2024-08-23
Added 'last_sync_time' field to STORAGE_INFO schema in STATE_DB

assrinivasan:
**What I did:**

This change adds a new field `last_sync_time` per storage device to `STORAGE_INFO|<storage_device_name>`
in `STATE_DB`. More context about the schema is [here](https://github.com/sonic-net/SONiC/blob/master/doc/storagemond/storagemond-hld.md#31-storage-info-state_db-schema).

**Why I did it**

This would provide additional context in case of repeated daemon crash scenario which leads to storagemon daemon status =  
`FATAL     Exited too quickly (process log may have details).`

**Signed-off-by:** 
Ashwin Srinivasan [assrinivasan@microsoft.com](mailto:assrinivasan@microsoft.com)

### DIFF
--- a/doc/storagemond/storagemond-hld.md
+++ b/doc/storagemond/storagemond-hld.md
@@ -371,19 +371,19 @@ key                     = STORAGE_INFO|<disk_name>  ; This key is for informatio
 
 ; field                 = value
 
-device_model            = STRING                    ; Describes the Vendor information of the disk                                           (Static)
-serial                  = STRING                    ; Describes the Serial number of the disk                                                (Static)
-temperature_celsius     = STRING                    ; Describes the operating temperature of the disk in Celsius                             (Dynamic)
-total_fsio_reads        = STRING                    ; Describes the total number of filesystem reads completed successfully                  (Dynamic)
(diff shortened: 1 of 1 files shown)


doc/storagemond/storagemond-hld.md
```diff
@@ -382,6 +383,7 @@ disk_io_writes          = STRING                    ; Describes the total number
 reserved_blocks         = STRING                    ; Describes the reserved blocks count of the SSD                                         (Dynamic)
 firmware                = STRING                    ; Describes the Firmware version of the SSD                                              (Dynamic)
 health                  = STRING                    ; Describes the overall health of the SSD as a % value based on several SMART attrs      (Dynamic)
+last_sync_time          = STRING                    ; The latest successful sync time of disk attribtes to STATE_DB in UNIX timestamp format (Dynamic)
```
prgeor, line 386: @assrinivasan can you update the time stamp format to MM:DD:YYYY format instead of epoch time which is not user convenient

assrinivasan: Done in latest

prgeor (approved):

PR #1789, opened 2024-08-28
HLD for host access control

QupinghaoNN:
This High-Level Design (HLD) document explains the hosts access feature in SONiC. The aim is to use the hosts_access feature of the Linux system to control client access to the host. This time, the main goal is to restrict clients from accessing the host via SSH.


amazor (commented):
Consider adding a log whenever a user attempts to access the daemon remotely.
To do so use the optional `shell_command` in the hosts.allow/host.deny file syntax:
`daemon_list : client_list [ : shell_command ]`

For example, the shell command can be a written as follows to allow logging:

`spawn (echo "DENIED: access from %c" | logger -t [%d])` 

Extra details can be extracted based on the daemon, such as username (`%u`).

More details on syntax can be found here: https://linux.die.net/man/5/hosts.allow#:~:text=and%20incompatible%20way.-,%25%20EXPANSIONS,-The%20following%20expansions

QupinghaoNN:
> Consider adding a log whenever a user attempts to access the daemon remotely. To do so use the optional `shell_command` in the hosts.allow/host.deny file syntax: `daemon_list : client_list [ : shell_command ]`
> 
> For example, the shell command can be a written as follows to allow logging:
> 
> `spawn (echo "DENIED: access from %c" | logger -t [%d])`
> 
> Extra details can be extracted based on the daemon, such as username (`%u`).
> 
> More details on syntax can be found here: https://linux.die.net/man/5/hosts.allow#:~:text=and%20incompatible%20way.-,%25%20EXPANSIONS,-The%20following%20expansions

I followed your advice and tried to modify and verify it.
     `sshd: 55.170.50.0/24 : spawn  ( /bin/echo "DENIED\: access from %c (host\: %h, user\: %u)" | /usr/bin/logger -t %s -p user.info ) `

A log entry will be generated in the  `/var/log/syslog ` file：
image

However, for SSH, access logs are already recorded in `/var/log/auth.log`. So, is it still necessary for us to add logs?
image

amazor:
I believe it is still useful since the `show logging` CLI command only looks at `/var/log/syslog` and other daemons may not have their own obvious default logging location. The log in `/var/log/auth.log` can give more detailed information on why the connection failed (or succeeded) since it comes from ssh itself.

amazor:
It would be useful to also add hostname access/deny. Both for FQDN (test.example.com) and last match patterns (.example.com)
You can see these here: https://linux.die.net/man/5/hosts.allow#:~:text=A%20string%20that%20begins%20with%20a%20%27.%27

QupinghaoNN:
> r FQDN

ok, I will add it to the  j2 file.

QupinghaoNN:
> I believe it is still useful since the `show logging` CLI command only looks at `/var/log/syslog` and other daemons may not have their own obvious default logging location. The log in `/var/log/auth.log` can give more detailed information on why the connection failed (or succeeded) since it comes from ssh itself.

OK， I will add it  to  the j2  file.

QupinghaoNN:
> It would be useful to also add hostname access/deny. Both for FQDN (test.example.com) and last match patterns (.example.com) You can see these here: https://linux.die.net/man/5/hosts.allow#:~:text=A%20string%20that%20begins%20with%20a%20%27.%27

I think that even without supporting hostname, it can still meet most usage scenarios. We won't adapt to the hostname for this time. We can add it later if needed. What do you think?

amazor:
I think that's okay, we can leave it open for the community. Great job!

PR #1795, opened 2024-09-01
[doc]: High frequency telemetry HLD

Pterosaur:
High frequency telemetry HLD initial version

PRs:
https://github.com/sonic-net/sonic-swss-common/pull/1006
https://github.com/sonic-net/sonic-sairedis/pull/1573


doc/stream-telemetry/stream-telemetry-hld.md
```diff
@@ -0,0 +1,930 @@
...
+
+### Revision  
+
+### Scope  
+
+This section describes the scope of this high-level design document in SONiC.
+
+### Definitions/Abbreviations 
+
+This section covers the abbreviation if any, used in this high-level design document and its definitions.
```
r12f, line 58: are these sections empty? if so, it will be better to put some other word instead of these "auther-guidance-words". this is very confusing, especially all section headers doesn't have any numbered prefix, it is very easy to lose when reading the doc.

doc/stream-telemetry/stream-telemetry-hld.md
```diff
@@ -0,0 +1,930 @@
...
+The purpose of this section is to give an overview of high-level design document and its architecture implementation in SONiC. 
+
+### Requirements
+
+This section list out all the requirements for the HLD coverage and exemptions (not supported) if any for this design.
+
+### Architecture Design 
+
+This section covers the changes that are required in the SONiC architecture. In general, it is expected that the current architecture is not changed.
+This section should explain how the new feature/enhancement (module/sub-module) fits in the existing architecture.
```
r12f, line 71: same here. these "guidance" words are better to be removed. very confusing to read.

doc/stream-telemetry/stream-telemetry-hld.md
```diff
@@ -0,0 +1,930 @@
...
+    dma_engine --IPFIX record--> netlink_module
+    netlink_module --IPFIX record--> counter_syncd
+    counter_syncd --telemetry message--> gnmi
+```
+
+### High-Level Design ###
+
+#### Modules ####
+
+##### Netlink Module #####
```
r12f, line 131: i know the numbered prefix are not there in the template, but the doc will be much more readable with it. we should consider add them here.

doc/stream-telemetry/stream-telemetry-hld.md
```diff
@@ -0,0 +1,930 @@
...
+32-63: "Export Timestamp: Second"
+64-95: "Sequence Number = 0, start from 0 and incremental sequence counter modulo 2^32"
+96-127: "Observation Domain ID = 0, always 0"
+
+```
+
+##### IPFIX template #####
+
+``` mermaid
+
```
r12f, line 180: this graph failed rendering...

doc/stream-telemetry/stream-telemetry-hld.md
```diff
@@ -0,0 +1,930 @@
...
+        genl_multicast_group: "stats"
+}
+
+```
+
+Ring buffer model
+
+Pin CPU?
+
+![netlink_dma_channel](netlink_dma_channel.drawio.svg)
```
r12f, line 150: are we still uses the chunk size concept in our API?

doc/stream-telemetry/stream-telemetry-hld.md
Pterosaur: Sorry, Yes, there are many template "guidance" words I haven't removed them.

doc/stream-telemetry/stream-telemetry-hld.md
Pterosaur: Yes, we have.

Pterosaur: Or leverage the existing solution, time-based reporting?

doc/stream-telemetry/stream-telemetry-hld.md
```diff
@@ -0,0 +1,986 @@
...
+192-192: "1"
+193-207: "Element ID = Object index for the stats N"
+208-223: "Field Length = 8 bytes"
+224-255: "Enterprise Number = SAI TYPE ID + SAI STATS ID for the stats N"
+
+```
+
+- For high-frequency counters, the native IPFIX timestamp unit of seconds is insufficient. Therefore, we introduce an additional element, `observationTimeNanoseconds`, for each record to meet our requirements.
+- The enterprise bit is always set to 1 for stats records.
+- The element ID of IPFIX is derived from the object index. For example, for `Ethernet5`, the element ID will be `0x5 | 0x8000 = 0x8005`.
```
marian-pritsak, line 231: Having a variable element ID per SAI object will blow up the number of templates in the report (separate template ID per port, queue etc.).

I suggest having a fixed element ID for each stat, and introduce a new element `objectID`, similar to `observationTimeNanoseconds`, to put the object ID of a port or a queue in a data set itself.

marian-pritsak (changes requested):

doc/stream-telemetry/stream-telemetry-hld.md
```diff
@@ -0,0 +1,986 @@
...
+
+#### Netlink Module and DMA Engine
+
+These two modules need to be provided by vendors. This document proposes a ring buffer communication model to support all expected TAM configurations as follows.
+
+![netlink_dma_channel](netlink_dma_channel.drawio.svg)
+
+### Data format
+
+We will use IPFIX as the report format, with all numbers in the IPFIX message in network-order (Big-endian).
```
marian-pritsak, line 178: You cannot make the assumption that data will be Big- or Little-endian. If it is collected by the ASIC and DMAd to a ring buffer, then kernel driver would need  to correct the endianness of each data set. We need to have a bit in the element ID or Enterprise number that would indicate endianness.

marian-pritsak (changes requested):

doc/stream-telemetry/stream-telemetry-hld.md
```diff
@@ -0,0 +1,986 @@
...
+chunk_size         = uint32 ; number of stats groups in a telemetry message.
+cache_size         = uint32 ; number of chunks that can be cached.
+```
+
+#### STREAM_TELEMETRY_GROUP
+
+```
+STREAM_TELEMETRY_GROUP|{{profile_name}}|{{group_name}}
+    "object_names": {{list of object name}}
+    "object_counters": {{list of stats of object}}
```
marian-pritsak, line 368: How do you plan to handle the update of the counter list or object list from the report POV? is it going to be a new template ID?

marian-pritsak (changes requested):

doc/stream-telemetry/stream-telemetry-hld.md
Pterosaur: In my mind, I assume all stats(per port, queue etc) in one profile will be encoded into ONE template. I will generate a unique template ID for a profile.

doc/stream-telemetry/stream-telemetry-hld.md
Pterosaur: Yes, I assume we will use a new template ID.
We treat any configuration update as a new profile. The previous session will be interrupted. The SAI report object will get a new SAI_TAM_REPORT_ATTR_REPORT_IPFIX_TEMPLATE_ID

doc/stream-telemetry/stream-telemetry-hld.md
Pterosaur: Are you worried that endian conversion will consume too many CPU clocks?

doc/stream-telemetry/stream-telemetry-hld.md
zhangyanzhao: > Hi @zhangyanzhao , Do you have any suggestions or concerns about adding a number prefix to each title in the template?

Good suggestion. Updated the template. Github markdown does not support auto number yet, need community member to update it manually when creating HLD doc with the template.

Pterosaur:
Hi @r12f , Please help to review or approve it if everything is good to you

r12f:
Looks good! But somehow I could not approve this due to a GitHub bug...

r12f (approved):
Ok, the approval works now. There is an old comment somehow getting stuck and blocking the sign off. It  works now.

doc/stream-telemetry/stream-telemetry-hld.md
```diff
@@ -0,0 +1,647 @@
+# Stream telemetry high level design <!-- omit in toc -->
```
qiluo-msft, line 1: "Stream telemetry" is a term already used by sonic gnmi based telemetry implementation. Suggest you replace it with another term like "high frequency counters telemetry" (just an example).

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
```diff
@@ -0,0 +1,706 @@
...
+```
+
+- If the number of stats in a group is small, multiple snapshots may be encoded into a single IPFIX message.
+- If the number of stats in a group exceeds 8K, the group must be split across multiple IPFIX messages.
+
+The IPFIX template should be provided by vendors. This document does not restrict how to split or concatenate snapshots, but each separated snapshot must include its own `observationTimeNanoseconds`.
+
+#### 7.2.4. Netlink message
+
+We expect all control messages and out-of-band information to be transmitted by the SAI. Therefore, it is unnecessary to read the attribute header of netlink and the message header of Genetlink from the socket. Instead, we can insert a bulk of IPFIX recordings as the payload of the netlink message. The sample code for building the message from the kernel side is as follows:
```
rck-innovium, line 394: Please specify who registers this new genetlink family: SDK kernel module?

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
```diff
@@ -0,0 +1,706 @@
...
+
+struct genl_multicast_group stel_mcgrps[] = {
+    { .name = "ipfix" },
+};
+
+// Family definition
+static struct genl_family stel_family = {
+    .name = "sonic_stel",
+    .version = 1,
+    // ...
```
rck-innovium, line 406: Please specify if any genetlink attribute is to be defined for the IPFix message. Providing a attribute will help us to extend the genetlink to provide additional info other than IPFix messages.

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
```diff
@@ -0,0 +1,706 @@
...
+160-191: "..."
+192-192: "1"
+193-207: "Element ID = Object index for the stats N"
+208-223: "Field Length = 8 bytes"
+224-255: "Enterprise Number = SAI TYPE ID + SAI STATS ID for the stats N"
+
+```
+
+- For high-frequency counters, the native IPFIX timestamp unit of seconds is insufficient. Therefore, we introduce an additional element, `observationTimeNanoseconds`, for each record to meet our requirements.
+- The element ID of IPFIX is derived from the object index. For example, for `Ethernet5`, the element ID will be `0x5 | 0x8000 = 0x8005`, where `0x8000` indicates that the enterprise bit is set to 1.
```
rck-innovium, line 289: Please describe this in detail. Some objects like buffer_pool, buffer_profile etc. do not have a user visible index.
Please also describe how this element-id to object-id translation is to be done in _counter syncd_ and if there are any information shared between __orchagent__ and _counter syncd_ for this.

rck-innovium (changes requested):

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
Pterosaur: There is no any genetlink attribute in this interface. This design is intentional, because we want all extensions to go through SAI and not other channels.

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
Pterosaur: Thanks for your constructive suggestion and reminder. I didn't notice there is object, such as buffer_pool, without a user visible index.
So, I re-design this part. Please check the commit: https://github.com/sonic-net/SONiC/pull/1795/commits/f3aef298b71b176879f5cf417ad8283d282189d3

In our new design, the counter syncd can map from element ID to object name via session tables in state db. And if we want to get the mapping between object name to SAI OID, there is already a name map in counter db. We can directly leverage it.
image
image

rck-innovium (approved):

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
```diff
@@ -0,0 +1,714 @@
...
+HIGH_FREQUENCY_TELEMETRY_PROFILE|{{profile_name}}
+    "stream_state": {{enabled/disabled}}
+    "poll_interval": {{uint32}}
+```
+
+```
+key                = HIGH_FREQUENCY_TELEMETRY_PROFILE|profile_name a string as the identifier of high frequency telemetry
+; field            = value
+stream_state       = enabled/disabled ; Enabled/Disabled stream.
+poll_interval      = uint32 ; The interval to poll counter, unit microseconds.
```
r12f, line 512: hi Ze, should we add a remote write endpoint in this profile?

Pterosaur: What kind of endpoint? for OpenTelemetry collector or outside service like InflexDB?

Pterosaur: Done， please check it

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
```diff
@@ -0,0 +1,720 @@
...
+576-639: "Port 2: SAI_PORT_STAT_IF_IN_ERRORS = 0"
+640-703: "Port 3: SAI_PORT_STAT_IF_IN_ERRORS = 6"
+
+704-719: "Set ID = 256"
+720-735: "Set Length = 36 bytes"
+736-799: "observationTimeNanoseconds = 30000"
+800-863: "Port 1: SAI_PORT_STAT_IF_IN_ERRORS = 20"
+864-927: "Port 2: SAI_PORT_STAT_IF_IN_ERRORS = 0"
+928-991: "Port 3: SAI_PORT_STAT_IF_IN_ERRORS = 8"
+
```
bendrapubalareddy, line 385: It would be helpful and clear to give sample snapshot of IPFIX message format for Elephant Flow counters with details of flow vector containing 5 tuple vaules(SIP, DIP, sport, dport, protocol), flow start time, flow end time, flow counter values.

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
```diff
@@ -0,0 +1,720 @@
...
+    config_db ->> hft_orch: HIGH_FREQUENCY_TELEMETRY_PROFILE
+    config_db ->> hft_orch: HIGH_FREQUENCY_TELEMETRY_GROUP
+    port_orch ->> hft_orch: Port/Queue/Buffer ... object
+
+    hft_orch ->> syncd: Config TAM objects
+
+    syncd ->> dma_engine: Config stats
+    syncd ->> hft_orch: Config was applied in the ASIC
+    syncd ->> hft_orch: Query IPFIX template
```
bendrapubalareddy, line 612: Could you please elaborate the flow sequence on how IPFIX template is programmed to DMA Engine/Netlink kernel modules?

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
```diff
@@ -0,0 +1,720 @@
...
+N/A
+
+### 8.2. CLI/YANG model Enhancements
+
+#### 8.2.1. Config CLI
+
+``` shell
+
+# Add a new profile
+sudo config high_frequency_telemetry profile add $profile_name --stream_state=$stream_state --poll_interval=$poll_interval --chunk_size=$chunk_size --chunk_count=$chunk_count
```
bendrapubalareddy, line 667: "otel_endpoint": {{string of endpoint}} (Optional)
    "otel_certs": {{string of path}} (Optional)
should these optional attributes be added to high_frequency_telemetry profile config or they should be part of another config?

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
```diff
@@ -0,0 +1,720 @@
...
+
+    alt Is stream status enabled?
+
+        hft_orch ->> syncd: Start telemetry stream
+
+        loop Push stats until stream disabled
+            loop collect a chunk of stats
+                dma_engine ->> asic: query stats from asic
+                asic --) dma_engine: stats
+                dma_engine ->> netlink_module: Push stats
```
bendrapubalareddy, line 626: As per the "title: High frequency telemetry architecture" flow, dma_engine shall push the stats in IPFIX format to netlink module. If so, should we change "dma_engine ->> netlink_module: Push stats" to "dma_engine ->> netlink_module: Push stats in IPFIX format" ?

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
Pterosaur: Hi @bendrapubalareddy , the dma engine and netlink module will both be provided by the vendor, we don't know when the IPFix template is programmed to the DMA Engine/Netlink kernel modules. So, we have a callback from SAI to notify the template is ready in SAI.
Here is its state machine, hope it can answer your question:
https://github.com/opencomputeproject/SAI/blob/master/doc/TAM/SAI-Proposal-TAM-stream-telemetry.md#tam-telemetry-type-state-machine

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
Pterosaur: Thanks for your point, we shuold support this new options in CLI.

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
Pterosaur: Good suggestion. will fix it.

bendrapubalareddy:
@Pterosaur If the counters are to be streamed from PIE port or front panel port of the ASIC directly, without sending to control plane sonic applications, that would be different config and this HLD does NOT want to have that option in it's config model. Is that understanding correct?

doc/high-frequency-telemetry/high-frequency-telemetry-hld.md
```diff
@@ -0,0 +1,720 @@
...
+48-63: "Number of Fields = 3"
+64-79: "Element ID=325"
+80-95: "Field length = 4 bytes"
+96-96: "1"
+97-111: "Element ID = 2 (port index)"
+112-127: "Field Length = 8 bytes"
+128-159: "Enterprise Number = 0x00010004"
+160-160: "1"
+161-175: "Element ID = 5 (port index)"
+176-191: "Field Length = 8 bytes"
```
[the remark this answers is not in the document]
Pterosaur: It's always expected to be 8 bytes. Because we want to follow the design of SAI. The type of stats in SAI is uint64.

Pterosaur:
> @Pterosaur If the counters are to be streamed from PIE port or front panel port of the ASIC directly, without sending to control plane sonic applications, that would be different config and this HLD does NOT want to have that option in it's config model. Is that understanding correct?

We don't cover this scenario in this HLD, at least not in this version of HLD.

Pterosaur:
> This document is generally light on SONIC-SAI workflow. Can you add some text describing the same ? Specifically -
> 
> 1. Are the counter capabilities retrieved and used ? What are the expectations ?
Yes, we have query_stats_st_capability API in SAI, if the capability isn't supported, we will raise an error from syslog.
> 2. If a counter reports as not supporting clear-on-read how does SONIC clear the counters ?
It's a typical problem of SONiC. According to the current SONiC solution, we will store the current snapshot of stats when we call clear action from SONiC side. And the UI will show the difference value of the stats.
> 3. How does SONIC handle the wrap-around conditions ?
Since we are using the 8bit(Uint64) for the counter, I believe that most of the time, it won't happen. 
> 4. how does SONIC deal with the cases where HostCPU is too slow to empty the circular buffer ?
Drop the new stats. That's why we propose the ring buffer as the data structure.
> 5. If IPFIX template supplies a length (other than 8) for the counter, how is it handled ?
It's always expected to be 8 bytes. Because we want to follow the design of SAI. The type of stats in SAI is uint64.
> 6. How does SONIC use the SAI_TAM_TEL_TYPE_ATTR_MODE ?
SONiC prefers SAI_TAM_TEL_TYPE_MODE_SINGLE_TYPE.  We may use SAI_TAM_TEL_TYPE_MODE_MIXED_TYPE in the future for better efficiency. But due to the limitation of some vendors, we prefer SINGLE TYPE.
> 7. General overview of the design for SAI_SWITCH_ATTR_TAM_TEL_TYPE_CONFIG_CHANGE_NOTIFY ?
We have a state machine: https://github.com/opencomputeproject/SAI/blob/master/doc/TAM/SAI-Proposal-TAM-stream-telemetry.md#tam-telemetry-type-state-machine
> 8. How are the config and data threads kept synchronised ?
In the phase 1, if we want to update the config, we have to delete the existing stream and create a new stream, this should be guarantee by SONiC side(orchagent) . And we leverage the tam telemetry type state machine to submit the config into the SAI level.

@bandaru-viswanath

r12f (approved):

JaiOCP (approved):

r12f:
Sync'ed with Bandaru offline and got his approval via email before the merge as well.

PR #1801, opened 2024-09-05
Updated YANG guidelines to use sonic-net instead of Azure in namespace

arista-hpandya:
I was adding a yang model and noticed that we now use "sonic-net" instead of "Azure" in the namespace field of a yang model. However, this conflicted with the guidelines posted in the documentation. This PR aims to make the doc consistent with the conventions in place.


arista-hpandya:
Fixes  #1802

wen587 (approved):
lgtm

PR #1813, opened 2024-09-25
HLD for OpenConfig Path Translation and Root Expansion in UMF

ndas7:
High-level design to add a plugin-based path translation layer between the UMF gNMI server and external data clients to translate both requests and responses for interoperability between multiple clients and server implementations.

### DIFF
--- a/doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
+++ b/doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
@@ -1,0 +1,338 @@
+# OpenConfig Path Translation HLD #
+
+## Table of Contents
+- [Table of Contents](#table-of-contents)
+- [Revision](#revision)
+- [Scope](#scope)
+- [Definition/Abbreviation](#definitionsabbreviations)
+- [Overview](#overview)
+- [Requirements](#requirements)
+- [Architecture Design](#architecture-design)
+- [High Level Design](#high-level-design)
+- [SAI API](#sai-api)
+- [Configuration and Management](#configuration-and-management)
+- [Warmboot and Fastboot Design Impact](#warmboot-and-fastboot-design-impact)
+- [Memory Consumption](#memory-consumption)
+- [Restrictions/Limitations](#restrictionslimitations)
+- [Testing Requirements/Design](#testing-requirementsdesign)
+- [Open/Action items - if any](#openaction-items---if-any)
+
+### Revision
+| Rev  | Rev Date   | Author(s)          | Change Description |
+|------|------------|--------------------|--------------------|
+| v0.1 | 09/17/2024 | Neha Das (Google)   | Initial version |
+
+### Scope
+
+This document describes the high-level design to add a plugin-based path translation layer between the UMF gNMI server and external data clients to translate both requests and responses for interoperability between multiple clients and server implementations.
+
+### Definitions/Abbreviations
+
+- [gNMI](https://github.com/openconfig/gnmi) : gRPC Network Management Interface
+- API : Application Programming Interface
+- SONiC: Software for Open Networking in the Cloud
+- UMF : Unified Management Framework
+- FE: Frontend
+- BE: Backend
+
+### Overview
+
+Unified Management Framework (UMF) is a SONIC management framework which provides gNMI and other common north bound interfaces to manage configuration and status of SONIC based switches.
+
+The difference in paths and formats used by existing clients and multiple sever implementations, including UMF, can lead to interoperability issues. Some of the gaps that we have encountered include:
+1. Support for gNMI requests for config and telemetry at the Single root path which is currently not supported in UMF. Single root means the xpath `/openconfig`. UMF supports gNMI requests modeled in OpenConfig subtrees such as `/openconfig/interfaces`, `/openconfig/qos`, etc but not at the `/openconfig` root level.
+2. Difference in path format in UMF's gNMI implementation `Origin:<empty>, Elements: /openconfig-interfaces:interfaces/interface[name=<>]` versus OpenConfig standard path `Origin: openconfig, Elements: /interfaces/interface[name=<>]`. In general, UMF uses `openconfig-${module-name}:${root-container-name}` as the first element of its path, whereas Openconfig uses `openconfig` as the origin and `${root-container-name}` for the first path element without module name. Some legacy clients/tools use the following format as well: `Origin: <empty>, Elements: /openconfig/interfaces/interface[name=<>]`
+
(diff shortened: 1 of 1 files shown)


doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
```diff
@@ -0,0 +1,338 @@
...
+- [Definition/Abbreviation](#definitionsabbreviations)
+- [Overview](#overview)
+- [Requirements](#requirements)
+- [Architecture Design](#architecture-design)
+- [High Level Design](#high-level-design)
+- [SAI API](#sai-api)
+- [Configuration and Management](#configuration-and-management)
+- [Warmboot and Fastboot Design Impact](#warmboot-and-fastboot-design-impact)
+- [Memory Consumption](#memory-consumption)
+- [Restrictions/Limitations](#restrictionslimitations)
```
mbalachandar, line 16: Limitation section is empty, are there not limitations?

doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
```diff
@@ -0,0 +1,338 @@
...
+```
+```
+Translated SetResponse
+prefix:<origin:"openconfig">
+response:<timestamp:123456 path:<>op:Update>
+response:<timestamp:123456 path:<>op:Replace>
+```
+
+##### SubscribeRequest
+For a single root SubscribeRequest, the translation is similar to a GetRequest. It will be translated into a SubscribeRequest with multiple top-container paths.
```
mbalachandar, line 266: what will be the response if one of the top container path returns error?

doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
```diff
@@ -0,0 +1,338 @@
...
+
+#### Config DB Enhancements
+
+This HLD does not include any changes to the Config DB.
+
+### Warmboot and Fastboot Design Impact
+
+This HLD does not have any impact on Warmboot and Fastboot.
+
+### Memory Consumption
```
mbalachandar, line 291: by supporting the root level path for both get / subscribe may cause out of memory issue in large scale setup, we may need more info?

doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
```diff
@@ -0,0 +1,338 @@
...
+
+### Restrictions/Limitations
```
mbalachandar, line 293: Limitation section is empty, are there no limitations?

mbalachandar (commented):
comments

doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
```diff
@@ -0,0 +1,338 @@
...
+- UMF : Unified Management Framework
+- FE: Frontend
+- BE: Backend
+
+### Overview
+
+Unified Management Framework (UMF) is a SONIC management framework which provides gNMI and other common north bound interfaces to manage configuration and status of SONIC based switches.
+
+The difference in paths and formats used by existing clients and multiple sever implementations, including UMF, can lead to interoperability issues. Some of the gaps that we have encountered include:
+1. Support for gNMI requests for config and telemetry at the Single root path which is currently not supported in UMF. Single root means the xpath `/openconfig`. UMF supports gNMI requests modeled in OpenConfig subtrees such as `/openconfig/interfaces`, `/openconfig/qos`, etc but not at the `/openconfig` root level.
```
faraazbrcm, line 43: Are you planning to support Get/SET/Subscribe at this level? If so, do you have benchmark data on how it scales with more models and data?.

doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
```diff
@@ -0,0 +1,338 @@
...
+prefix:<origin:"openconfig">path:<>
+```
+
+```
+Translated GetRequest
+prefix:<>path:<elem:<name:"openconfig-interfaces:interfaces">>path:<elem:<name:"openconfig-lacp:lacp">>path:<elem:<name:"openconfig-platform:components">>path:<elem:<name:"openconfig-qos:qos">>path:<elem:<name:"openconfig-sampling:sampling">>path:<elem:<name:"openconfig-system:system">>
+```
+
+##### GetResponse
+Since the single root GetRequest is translated into multiple top-container paths, the JSON_IETF response will receive multiple notifications (one for each translated path). These multiple notifications need to be combined into one single JSON_IETF value and sent back to the client.
```
faraazbrcm, line 183: Instead of translating at the gnmi_server, should we add native support in translib to handle single root requests? Aggregation seems like a costly operation, in my opinion

doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
```diff
@@ -0,0 +1,338 @@
...
+```
+
+##### SetRequest
+For a Single root SetRequest with JSON_IETF encoding, the JSON value is in the same format as in the translated GetResponse above.
+A mix of single root and non-single root SET is not supported. So for delete/update/replace, either there is only one root delete/update/replace, or non-single root ones.
+
+To translate the SetRequest:
+
+###### Single Root Delete
+It will be translated into multiple delete paths as show below:
```
faraazbrcm, line 210: If I understand correctly, this will result in multiple CAS transactions in translib, right? If so, it's better to handle it with a single CAS (check and save, Redis MULTI) to reduce the chance of errors due to transactional failure

doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
```diff
@@ -0,0 +1,338 @@
...
+Original Delete
+prefix:<origin:"openconfig">delete:<>
+```
+```
+Translated Delete
+prefix:<>delete:<elem:<name:"openconfig-interfaces:interfaces">>delete:<elem:<name:"openconfig-lacp:lacp">>delete:<elem:<name:"openconfig-platform:components">>delete:<elem:<name:"openconfig-qos:qos">>delete:<elem:<name:"openconfig-sampling:sampling">>delete:<elem:<name:"openconfig-system:system">>
+```
+
+###### Single Root Update/Replace
+For update and replace, the JSON payload needs to be parsed and the payload has to be separated into multiple JSON payloads according to the top-container names, and then placed into individual updates/replaces.
```
sachinholla, line 221: Translator implementation should also ensure order of updates?

doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
ndas7: If one of the top container path returns an error, the error will be logged and the response ignored for the Get/Subscribe operations. The error will be raised for a Set operation.

doc/mgmt/OpenConfig_Path_Translation_Root_Expansion_HLD.md
ndas7: Yes, that's right - the order of the deletes, replace, updates is still maintained.

PR #1823, opened 2024-10-04
Fine-Grained Hashing for dynamic nexthop groups.

ashutosh-agrawal:
Added prefix-based match mode that allows list of nexthops IP to be discovered from route updates instead of requiring users to configure them statically. The new mode will add support for consistent hashing as traditionally implemented in other commercial NOSes.

Code PRs:
[Yang Model](https://github.com/sonic-net/sonic-buildimage/pull/21545)
[SWSS](https://github.com/sonic-net/sonic-swss/pull/3488)
[SONiC Utilities](https://github.com/sonic-net/sonic-utilities/pull/3740)


doc/ecmp/fine_grained_next_hop_hld.md
```diff
@@ -115,11 +120,14 @@ Warm restart is required for this design since the solution would benefit from h
...
-Following new tables will be added to Config DB. Unless otherwise stated, the attributes are mandatory.
+Following new tables will be added to Config DB. Unless otherwise stated, the attributes are mandatory. FG_NHG_MEMBER 
+table is not used or required for dynamic-nhg mode.
 ```
 FG_NHG|{{fg-nhg-group-name}}:
-    "bucket_size": {{hash_bucket_size}}
-    "match_mode" : {{route-based/nexthop-based}}
+    "bucket_size"   : {{hash_bucket_size}}
+    "match_mode"    : {{route-based/nexthop-based}}
+    "nhg_mode:      : {{static-nhg/dynamic-nhg}} (Optional)
```
anish-n, line 129: can we instead re-use match_mode itself?
route-based = filter by full route, prefix + nexthop IP
nexthop-based = filter by next-hop. All prefixes which match the next-hop IPs get fine grained ecmp.
prefix-based = filter by ip prefix in the route. All prefixes which match the specified prefix get fine grained ecmp regardless of the next-hop IPs.

ashutosh-agrawal: Hi Anish, I was initially thinking along the same lines but didn't want to redefine the functionality associated with route-based in the current HLD for backword compatibility reasons. In the following excerpt from the current HLD, route-based in defined as matching on the prefix only
`route-based looks to prefix to filter routes, and uses fine grained ecmp when the route prefix matches the FG_NHG_PREFIX prefix.
`

anish-n: The proposal to reuse match_mode will not redefine functionality and will also not cause any backward compatibility issues, because a brand new prefix-based match_mode being defined which does not impact the other matchmodes.

The other reason to keep using match_mode and not define a nhg_mode is: nexthop-based is also a type of dynamic-nhg, it dynamically learns BGP routes and matches based on the next-hop IP.

ashutosh-agrawal: Hi Anish, I understand your point but we were tripped by the existing definition of route-based in the yang model.

`            enum route-based {
               description
                 "Fine grained ecmp is used when the route prefix matches the FG_NHG_PREFIX prefix";
            }
`
In both the existing use case and the new one added by us, route prefix needs to match FG_NHG_PREFIX. So, we added another attribute to differentiate between the existing case where the list of nexthop IPs is statically configured vs our enhancement where nexthops IPs don't need to be specified though config.

ashutosh-agrawal: @anish-n As you originally suggested, I've modified the proposal to add prefix-based as an another enum value of match-mode to for dynamically learnt nexthop IPs. Please let me know if you have any other comments.

anish-n (approved):

ashutosh-agrawal:
@zhangyanzhao Please let me know if anything else needs to be done to merge this PR.

prsunny (approved):

prsunny:
> @prsunny can you please help to take a look at this HLD PR and see if you are ok to merge? The review has been done.

merged

PR #1829, opened 2024-10-14
HLD Port FEC BER

vincentpcng:
**What I did**

Add design proposal for Port FEC BER feature

**Summary of Changes:**

Enhance the collect and compute the FEC BER on each port. 
Modify the port_rates.lua , portstat.py and netstat.py

| PR Title                                                                   | State                            | Context                            |
|----------------------------------------------------------------------------|----------------------------------|------------------------------------|
| [[sonic-utilities] Add port FEC BER show changes](https://github.com/sonic-net/sonic-utilities/pull/3607)                 | [PR Status](https://img.shields.io/github/pulls/detail/state/sonic-net/sonic-utilities/3607) | [PR Checks](https://img.shields.io/github/status/contexts/pulls/sonic-net/sonic-utilities/3607)|
| [[sonic-swss] Add port FEC BER feature swss](https://github.com/sonic-net/sonic-swss/pull/3363)  | [PR Status](https://img.shields.io/github/pulls/detail/state/sonic-net/sonic-swss/3363) | [PR Checks](https://img.shields.io/github/status/contexts/pulls/sonic-net/sonic-swss/3363)|
| [[sonic-mgmt] Add the port FEC BER mgmt test](https://github.com/sonic-net/sonic-mgmt/pull/15481)  | [PR Status](https://img.shields.io/github/pulls/detail/state/sonic-net/sonic-sairedis/15481) | [PR Checks](https://img.shields.io/github/status/contexts/pulls/sonic-net/sonic-mgmt/15481)|

### DIFF
--- a/doc/port_fec-ber/port_fec_ber.md
+++ b/doc/port_fec-ber/port_fec_ber.md
@@ -156,14 +156,14 @@ Step 3 : look up link serdes speed
 
 Step 4 : assume statistical average error rate as 1e-8 per frame
 
-     rs_average_frame_error = 1e-8 
+     rs_average_frame_ber = 1e-8 
 
 Step 5: calcuate BER
 
     interval = PORT_STATE poll interval which is 1000 ms currently
 
     Pre FEC BER = (SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS - SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS_last) / (serdes * lanes_count * interval / 1000)
-    Post FEC BER = (SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES - SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES_last) * rs_average_frame_error / (serdes * lanes_count * interval / 1000)
+    Post FEC BER = (SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES - SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES_last) * rs_average_frame_ber / (serdes * lanes_count * interval / 1000)
 
 
 Step 6: the following data will be updated and its latest value stored in the COUNTER_DB, RATES table after each iteraction
--- a/doc/port_fec-ber/port_fec_ber.md
+++ b/doc/port_fec-ber/port_fec_ber.md
@@ -154,20 +154,16 @@ Step 3 : look up link serdes speed
         serdes = 0
     end
 
-Step 4 : frame size
+Step 4 : assume statistical average error rate as 1e-8 per frame
 
-     /* NRZ encoding Rs-528  or PAM4 encoding Rs-544*/
-    if user_port_lane_speed <= 25G
-      frame_size = 5280
-    else
-      frame_size = 5440
+     rs_average_frame_error = 1e-8 
 
 Step 5: calcuate BER
 
     interval = PORT_STATE poll interval which is 1000 ms currently
 
     Pre FEC BER = (SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS - SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS_last) / (serdes * lanes_count * interval / 1000)
-    Post FEC BER = (SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES - SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES_last) * "frame size" / (serdes * lanes_count * interval / 1000)
+    Post FEC BER = (SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES - SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES_last) * rs_average_frame_error / (serdes * lanes_count * interval / 1000)
 
 
 Step 6: the following data will be updated and its latest value stored in the COUNTER_DB, RATES table after each iteraction
--- a/doc/port_fec-ber/port_fec_ber.md
+++ b/doc/port_fec-ber/port_fec_ber.md
@@ -90,10 +90,8 @@ There are no changes in the current Sonic Architecture.
 
 SAI provide access to each interface the following attributes
 - SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS
-  - monotonically increasing
   - return not support if its not working for an interface
 - SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES
-  - monotonically increasing
   - return not support if its not working for an interface
 
 
@@ -105,7 +103,7 @@ The following redis DB entries will be access for the BER calculations
 |--------------|-------------|------------------|--------|----------------|----------------|  
 |COUNTER_DB |COUNTERS |SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS |R |number |Total number bits corrected</sub>|
 |COUNTER_DB |COUNTERS |SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES |R |number |Tota number uncorrectable frame |
-|COUNTER_DB |COUNTERS_PORT_NAME_MAP |SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES|R |number |Oid to name mapping |  
+|COUNTER_DB |COUNTERS_PORT_NAME_MAP | name & oid  |R |name  |Oid to name mapping |  
 |COUNTER_DB |RATES |SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS_last | New, RW|number |Last corrected bits counts |
 |COUNTER_DB |RATES |SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES_last |New, RW|number |Last uncorrctedble frame counts |  
 |COUNTER_DB |RATES |FEC_PRE_BER |New, RW| floating |calculated pre fec BER |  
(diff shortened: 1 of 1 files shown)


doc/port_fec-ber/port_fec_ber.md
```diff
@@ -0,0 +1,205 @@
...
+
+Step 6: the following data will be updated and its latest value stored in the COUNTER_DB, RATES table after each iteraction
+
+    Pre FEC BER , Post FEC BEC, SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS_last and SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES_last
+
+```
+## 5 Sample Output
+```
+root@ctd615:/usr/local/lib/python3.11/dist-packages/utilities_common#  portstat -f
+      IFACE    STATE    FEC_CORR    FEC_UNCORR    FEC_SYMBOL_ERR    FEC_PRE_BER    FEC_POST_BER
```
qinchuanares, line 181: The FEC_CORR seems to show the corrected number of frames, is it possible to add corrected number of bits as an extra column?

doc/port_fec-ber/port_fec_ber.md
```diff
@@ -0,0 +1,205 @@
...
+    if user_port_lane_speed <= 25G
+      frame_size = 5280
+    else
+      frame_size = 5440
+
+Step 5: calcuate BER
+
+    interval = PORT_STATE poll interval which is 1000 ms currently
+
+    Pre FEC BER = (SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS - SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS_last) / (serdes * lanes_count * interval / 1000)
```
eddyk-nvidia, line 169: why is the division by 1000 at the end of the calculation ?
BER in interval = (# errors / # bits) in interval
#bits = serdes rate * lane * interval

doc/port_fec-ber/port_fec_ber.md
```diff
@@ -0,0 +1,205 @@
...
+    end
+
+Step 4 : frame size
+
+     /* NRZ encoding Rs-528  or PAM4 encoding Rs-544*/
+    if user_port_lane_speed <= 25G
+      frame_size = 5280
+    else
+      frame_size = 5440
+
```
eddyk-nvidia, line 164: Why the post FEC BER errors are multiplied by the frame_size ?  This is the worst case for amount of errors.
For any uncorrectable FEC block, the bit errors can be between 8 to 5280 for <=25G and 16 to 5440 for >=50G.
The lowest possability is the most likely statistically ( 8 or 16)

doc/port_fec-ber/port_fec_ber.md
vincentpcng: The initial request is to provide the "FEC PRE BER" and " FEC POST BER". If additional information or fields are needed, we can handle the request as a future enhancement

doc/port_fec-ber/port_fec_ber.md
vincentpcng: The polling is done through the port_rates.lua script. The script were run at 1000ms interval as default. The interval were pass to the script as an argument which is in ms... that is 1 sec were pass in as 1000.
Now in order to account for the total number of bit ingress.. we need to multiple the serdes bps  with the total duration is sec, hence  the reason for divided by 1000

doc/port_fec-ber/port_fec_ber.md
vincentpcng: This is a questions of choice.  Since we have no idea the "number of bits" which were bad in this case, we need to take a guess, either the best case as you suggested or the worst case which we were using at this moment.
The way I will suggest how to interpret this value should be  anything other than zero, we have issue.

qinchuanares:
When the port is down or flapping, or even postFEC error > 0, preFEC BER calculation does not make sense.

After link flap or link down, when link is stably up, the preFEC BER calculation uses counters reset from 0 at link up.

doc/port_fec-ber/port_fec_ber.md
AndyRodan: I agree that the bits passed are multiplied by the interval number.
My lack of understanding is why we have both multiplication of interval variable and division of 1000.
the total amount of bits passed in 1000ms interval are: serdes * lanes_count * interval
Dividing by another 1000 is redundant no?.

doc/port_fec-ber/port_fec_ber.md
AndyRodan: Taking the worst-case scenario (5440) is the most unlikely statistically and might present a false picture.
for a random error distribution, with a typical post fec average BER of 1e-12 it's 1 error for evert 1e12 bits which is ~ 1 error per ~1e8 Fec blocks.

doc/port_fec-ber/port_fec_ber.md
vincentpcng: it is because the interval which were passing is equal to 1000,  and can be changed with a cli as needed and the value it took is in ms. Hence when we multiple by interval we need to divide it by 1000.

doc/port_fec-ber/port_fec_ber.md
vincentpcng: look like folks prefer  the statistical average when calculate the post fec,  will make the changes

vincentpcng:
> When the port is down or flapping, or even postFEC error > 0, preFEC BER calculation does not make sense.
> 
> After link flap or link down, when link is stably up, the preFEC BER calculation uses counters reset from 0 at link up.

vincentpcng:
hit the wrong pull and accidental closed the PR, reopen it.

vincentpcng:
> When the port is down or flapping, or even postFEC error > 0, preFEC BER calculation does not make sense.
> 
> After link flap or link down, when link is stably up, the preFEC BER calculation uses counters reset from 0 at link up.

Agree, the measurement could be compromised if the link is not stable, like  flapping occur, etc

doc/port_fec-ber/port_fec_ber.md
vincentpcng: doc updated to use average for post fec BER.

Junchao-Mellanox (approved):

keboliu (approved):

Junchao-Mellanox:
The feature only capture the realtime port FEC BER. Is it possible to capture any history port FEC BER issue so that we will be able to get some debug information when an issue only happens in a short period?

doc/port_fec-ber/port_fec_ber.md
```diff
@@ -0,0 +1,201 @@
...
+   + netstat.py :
+
+     Add new format support to display the BER in a floating point format
+
+
+### 4.1 Assumptions
+
+SAI provide access to each interface the following attributes
+- SAI_PORT_STAT_IF_IN_FEC_CORRECTED_BITS
+  - monotonically increasing
```
eddyk-nvidia, line 93: after link status change the counter isn't reliable. After offline discussion with Microsoft it was decided to remove this requirement

doc/port_fec-ber/port_fec_ber.md
```diff
@@ -0,0 +1,201 @@
...
+  - return not support if its not working for an interface
+- SAI_PORT_STAT_IF_IN_FEC_NOT_CORRECTABLE_FRAMES
+  - monotonically increasing
```
eddyk-nvidia, line 96: after link status change the counter isn't reliable. After offline discussion with Microsoft it was decided to remove this requirement

vincentpcng:
The following code review PR is for the implementation of this HLD
(1) https://github.com/sonic-net/sonic-swss/pull/3363
(2) https://github.com/sonic-net/sonic-utilities/pull/3607
(3) https://github.com/sonic-net/sonic-mgmt/pull/15481

vincentpcng:
Review feedback , the following key points were mentions

(1) can we add min , max, average BER
(2) what about future serdes , can we use user data rate when compute the BER. 
(3) what about other devices support

response:
(1) we can add the min, max, and average to the show command... this can be done as a future enhancement
(2) unless IEEE defined the enoding, otherwise we dont know (i) what the future serdes speed, (ii) what type of encoding were use in it, hence can;t be done. 
One suggestion is to use the user data rate for computing. This is a possible trade off. However, the resulted BER will become not  a true reflection. The community future suggest can we add more columns and use the "user data rate".
This will be left as future enhancement discussion
(3)  This feature is as stated, require the SAI support. If SAI layer can provide the counter, than we should be able to support the devices.

prgeor (commented):
@vincentpcng can you please link all the code PRs to the HLD?

prgeor (approved):

vincentpcng:
I added the code PR(s)  links before the meeting minute comments...

Let me include it again here

The following code review PR is for the implementation of this HLD
(1) https://github.com/sonic-net/sonic-swss/pull/3363
(2) https://github.com/sonic-net/sonic-utilities/pull/3607
(3) https://github.com/sonic-net/sonic-mgmt/pull/15481

doc/port_fec-ber/port_fec_ber.md
```diff
@@ -0,0 +1,199 @@
...
+        serdes = 53.125e+9
+    elseif port_data_rate == 100000 then
+        serdes = 106.25e+9
+    else
+        serdes = 0
+    end
+
+Step 4 : assume statistical average error rate as 1e-8 per frame
+
+     rs_average_frame_ber = 1e-8
```
prgeor, line 157: @AndyRodan, @vincentpcng can you please provide the reference for this?

vincentpcng: Andy provide some reference base on the 802.3 ck and the 1e-8 statictiscal average were base on Andy feedback.

doc/port_fec-ber/port_fec_ber.md
vincentpcng: requirement removed from the doc

doc/port_fec-ber/port_fec_ber.md
vincentpcng: requirement removed from the doc

qinchuanares (approved):

PR #1446, opened 2023-08-17
Virtual Router Redundancy Protocol Adaptation HLD

micas-net:
Hld for VRRP adaptation to SONiC, as mentioned in issue: https://github.com/sonic-net/SONiC/issues/1366:

| Repo  | PR title | State |
| ----------------- | ----------------- | ----------------- |
| sonic-buildimage | https://github.com/sonic-net/sonic-buildimage/pull/18617, https://github.com/sonic-net/sonic-buildimage/pull/20383 |
| sonic-swss | https://github.com/sonic-net/sonic-swss/pull/3106, https://github.com/sonic-net/sonic-swss/pull/3313, https://github.com/sonic-net/sonic-swss/pull/3315 |
| sonic-swss-common | https://github.com/sonic-net/sonic-swss-common/pull/813, https://github.com/sonic-net/sonic-swss-common/pull/927 |
| sonic-utilities | https://github.com/sonic-net/sonic-utilities/pull/2949 |

PR #1834, opened 2024-10-23
Revert "renaming vrrpmgr into macvlanmgr"

micas-net:
Thanks for @venkatmahalingam 's review on https://github.com/sonic-net/sonic-swss/pull/3106#discussion_r1808035256 . It would be more appropriate to use VRRPMGRD here.

HLD:https://github.com/sonic-net/SONiC/pull/1446


doc/vrrp/VRRP_Adaptation_HLD.md
```diff
@@ -458,7 +458,7 @@ admin@sonic:~$ redis-cli -n 1 hgetall "ASIC_STATE:SAI_OBJECT_TYPE_ROUTER_INTERFA
 
 ![VRRP ADD/DEL VRRP instance Flows](images/VRRP_Config_Instance_Flow.png)
 
-macvlanmgrd:
+vrrpmgrd:
   - Listens to VRRP create, delete and parameter change in CONFIG DB
   - Upon change
     - Add/del VRRP instance corresponding Macvlan device to kernel with IPs and state.
```
venkatmahalingam, line 464: [doc/vrrp/images/VRRP_Architecture_Diagram.png](https://github.com/sonic-net/SONiC/pull/1834/files#diff-9b3f93693b3201f4f82520d7ba8fa786d9c36792712124e16c5c2828395372a6)
Why do we have vrrpmgrd part of bgp container and vrrpsyncd part of swss container? sorry missed to comment on this earlier.
I thought, like other modules, we need to have vrrpmgrd part of swss and vrrpsyncd part of bgp container, agree?

balanokia:
vrrpmgrd still holds good since the scope of these macvlan devices are specific to VRRP. macvlan devices may be used by any other protocols as well in SONiC. Generically, macvlanmgr as consumer of user configuration would add complexity when there are needs to demultiplex to other than VRRP.

PR #1840, opened 2024-10-29
Improve gnmi full config update

ganglyu:


doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -441,18 +441,42 @@ For incremental configuration, 'config apply-patch' should not restart gNMI, bgp
...
-
-<img src="images/full_rpc_gnoi.svg" alt="full-gnmi-gnoi" width="800px"/>
-
-The full configuration request will be overwritten by subsequent full configuration request or incremental configuration request.
-
-<img src="images/mixed requests.svg" alt="overwritten-config" width="800px"/>
+We will use single gnmi request to support full configuration update.
+1. The host service will perform YANG validation and return an error if it fails.
+2. The host service will mask the gNMI service, preventing it from restarting after a config reload.
+3. The host service will execute a config reload using the input configuration.
```
mbalachandar, line 447: is there any chance that switch will not be in desired state after config reload, if so will it be handled here? like rollback?

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -441,18 +441,42 @@ For incremental configuration, 'config apply-patch' should not restart gNMI, bgp
...
-<img src="images/full_rpc.svg" alt="full-gnmi" width="800px"/>
-
-* Use gNMI request and gNOI request together to implement full configuration update.
-
-<img src="images/full_rpc_gnoi.svg" alt="full-gnmi-gnoi" width="800px"/>
-
-The full configuration request will be overwritten by subsequent full configuration request or incremental configuration request.
-
-<img src="images/mixed requests.svg" alt="overwritten-config" width="800px"/>
+We will use single gnmi request to support full configuration update.
```
mbalachandar, line 444: during the full configuration update, do we block any config update request from other interfaces like REST, CLI?

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -441,18 +441,42 @@ For incremental configuration, 'config apply-patch' should not restart gNMI, bgp
...
+1. The host service will perform YANG validation and return an error if it fails.
```
mbalachandar, line 445: not sure if the yang validation address the validation based on the state data, like TCAM ACL limit? If not then yang validation will succeed, but config reload may fail.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: yes, we will use "config reload -f" to rollback

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: We do not block other configuration update requests. All configuration updates made before the 'config reload' will be overwritten, but those made after the 'config reload' will still function.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: Good point! Currently, SONiC yang does not check the ACL limit, but we can add this limitation.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
```diff
@@ -441,18 +441,42 @@ For incremental configuration, 'config apply-patch' should not restart gNMI, bgp
...
+2. The host service will mask the gNMI service, preventing it from restarting after a config reload.
+3. The host service will execute a config reload using the input configuration.
+4. The host service will execute a config save if step 3 is successful.
+5. The host service will perform a config reload to recover if step 3 fails.
+6. The host service will unmask the gNMI service.
```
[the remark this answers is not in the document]
sachinholla: I see this is listed as a limitation below.. Is there any gNOI interface for clients to force restart the server?

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
faraazbrcm, line 445: In cases where we need to enforce custom validation rules that go beyond the constraints of the Yang model, how does the current design plan to handle these? translib based infrastructure provides this mechanism.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
ganglyu: We can use gnoi reboot interface to restart device and apply all the configurations.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: You can continue using the current translib infrastructure, and this change will not impact it. We are utilizing the origin field of the gnmi request, and only those requests with the 'sonic-db' origin field can use this gnmi full config update.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
sachinholla: I think reboot will be not so practical. There will be a significant downtime due to config reload already. Reboot requires another downtime.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
ganglyu: I agree, we don't have a good solution at the moment. Let's keep this limitation, as gNMI can't be used to change the management network or gNMI server settings.

doc/mgmt/gnmi/SONiC_GNMI_Server_Interface_Design.md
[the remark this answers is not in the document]
ganglyu: I agree, we don't have a good solution at the moment. Let's keep this limitation, as gNMI can't be used to change the management network or gNMI server settings.

PR #1847, opened 2024-11-07
Minor fixes for reboot HLD

vvolam:
- Add more details for MODULE_REBOOT_SMARTSWITCH reboot type
- Update DPU and Switch reboot sequence with PCI detach/reattach() vendor APIs
- Remove module name from module APIs
- Remove the tests which were missed removing from first HLD PR.


doc/smart-switch/reboot/reboot-hld.md
```diff
@@ -125,30 +126,30 @@ will be set to false in the RebootStatusResponse. Until the services are termina
...
+pci_detach(self):
 ```
 Detaches the DPU PCI device from the NPU. In the case of non-smart-switch chassis, no action is taken.
 
 Returns:
     True
 ```
 
-pci_reattach(self, module_name):
+pci_reattach(self):
```
prgeor, line 144: @vvolam nit pci_rescan() ?

doc/smart-switch/reboot/reboot-hld.md
```diff
@@ -125,30 +126,30 @@ will be set to false in the RebootStatusResponse. Until the services are termina
...
 ```
 Rescans the PCI bus and attach the DPU back to NPU. In the case of non-smart-switch chassis, no action is taken.
 
 Returns:
     True
 ```
 
-get_bus_info(self, module_name):
+get_bus_info(self):
```
prgeor, line 152: @vvolam suggest get_pci_bus_info()

doc/smart-switch/reboot/reboot-hld.md
vvolam: @prgeor done. Thanks!

doc/smart-switch/reboot/reboot-hld.md
vvolam: @prgeor done. Thanks!

doc/smart-switch/reboot/reboot-hld.md
vvolam: > @vvolam nit pci_rescan() ?

@prgeor, Though I modified, wanted to rename back as pci_reattach, as pci_rescan is Linux kernel standard function name. To avoid confusion and conflicts better to avoid same names.

rameshraghupathy (approved):

prgeor:
@vvolam consider a case where NPU was rebooted do we have check in place to detect whether NPU actually rebooted once the NPU comes up?

vvolam:
> @vvolam consider a case where NPU was rebooted do we have check in place to detect whether NPU actually rebooted once the NPU comes up?

@prgeor Once the system issues the reboot command, it is exiting the context based on my understanding. We can only check the switch state in our test scripts. Please correct me, if I missing any thing here?

prgeor:
> > @vvolam consider a case where NPU was rebooted do we have check in place to detect whether NPU actually rebooted once the NPU comes up?
> 
> @prgeor Once the system issues the reboot command, it is exiting the context based on my understanding. We can only check the switch state in our test scripts. Please correct me, if I missing any thing here?



> > @vvolam consider a case where NPU was rebooted do we have check in place to detect whether NPU actually rebooted once the NPU comes up?
> 
> @prgeor Once the system issues the reboot command, it is exiting the context based on my understanding. We can only check the switch state in our test scripts. Please correct me, if I missing any thing here?

@vvolam you don't need current context to check but there are different ways in which you can do.
1. NPU can put a marker for each DPU and check if the marker exist while NPU comes up
2. NPU can check the uptime of DPU before rebooting and know whether it rebooted

I think we can probably keep this for future

vvolam:
> > > @vvolam consider a case where NPU was rebooted do we have check in place to detect whether NPU actually rebooted once the NPU comes up?
> > 
> > 
> > @prgeor Once the system issues the reboot command, it is exiting the context based on my understanding. We can only check the switch state in our test scripts. Please correct me, if I missing any thing here?
> 
> > > @vvolam consider a case where NPU was rebooted do we have check in place to detect whether NPU actually rebooted once the NPU comes up?
> > 
> > 
> > @prgeor Once the system issues the reboot command, it is exiting the context based on my understanding. We can only check the switch state in our test scripts. Please correct me, if I missing any thing here?
> 
> @vvolam you don't need current context to check but there are different ways in which you can do.
> 
> 1. NPU can put a marker for each DPU and check if the marker exist while NPU comes up
> 2. NPU can check the uptime of DPU before rebooting and know whether it rebooted
> 
> I think we can probably keep this for future
@prgeor Sure, I will keep a note on this to enhance the HLD for this. Thank you!

doc/smart-switch/reboot/reboot-hld.md
```diff
@@ -89,12 +90,12 @@ DPUs are internally connected to the NPU via PCI-E bridge. Below is the reboot s
...
 
-* Subsequently, the NPU detaches the DPU PCI device using the sysfs interface by executing `echo 1 > /sys/bus/pci/devices/XXXX:XX:XX.X/remove`. This detachment is currently performed via sysfs, with future plans to implement a vendor-specific API when the DPU bus information cannot have a fixed value.
+* Subsequently, the NPU detaches the DPU PCI device via the pci_detach() vendor API, or through the sysfs interface by executing echo 1 > /sys/bus/pci/devices/XXXX:XX:XX.X/remove, if the vendor API is unavailable.
 
 * Next, the NPU triggers a platform vendor reboot API to initiate the reboot process for the DPU. If the DPU is stuck or unresponsive, the DPU reboot platform API should
 attempt a cold boot or power cycle to recover it.
 
-* The NPU either immediately rescans the PCI bus upon return or after a specified timeout period. This rescan is performed via the sysfs interface by echoing '1' to /sys/bus/pci/rescan.
+* The NPU either immediately rescans the PCI bus upon return or after a specified timeout period. This rescan is performed via the pci_reattach() vendor API or, if the vendor API is unavailable, by echoing '1' to /sys/bus/pci/rescan.
```
prgeor, line 98: @vvolam how is the timeout specified?

prgeor (commented):
@vvolam in the Test Plan section there is this rendered text with ### is that intentional?

``###Graceful boot/reboot###``

doc/smart-switch/reboot/reboot-hld.md
vvolam: @prgeor There is no timeout here. Rescan happens after reboot API returns the handle. I will reword HLD accordingly. Thank you!

vvolam:
> @vvolam in the Test Plan section there is this rendered text with ### is that intentional?
> 
> `###Graceful boot/reboot###`

It was a syntax error. Fixed it. Thank you!

prgeor (approved):

PR #1858, opened 2024-11-22
OpenConfig Platform Transceiver HLD

nakano-omw:
# OpenConfig support for Platform Transceiver.

Add support for following features:
- serial-no
- connector-type
- vendor
- vendor-part
- date-code
- supply-voltage
- laser-temperature
- output-power
- input-power
- laser-bias-current


nakano-omw:
This HLD was reviewed by the UMF WG on November 27, 2025.

doc/mgmt/OpenConfig_Platform.md
```diff
@@ -0,0 +1,454 @@
...
+- connector-type
+- vendor
+- vendor-part
+- date-code
+- supply-voltage
+- laser-temperature
+- output-power
+- input-power
+- laser-bias-current
+
```
jeff-yin, line 114: What about high/low warning/critical thresholds for temperature and voltage? These should also be encoded in transceiver DOM values.

doc/mgmt/OpenConfig_Platform.md
```diff
@@ -0,0 +1,454 @@
...
+3.	Verify that operations supported for gNMI/REST works fine for oc-transceiver:vendor.
+4.	Verify that operations supported for gNMI/REST works fine for oc-transceiver:vendor-part.
+5.	Verify that operations supported for gNMI/REST works fine for oc-transceiver:serial-no.
+6.	Verify that operations supported for gNMI/REST works fine for oc-transceiver:date-code.
+7.	Verify that operations supported for gNMI/REST works fine for supply-voltage of transceiver.
+8.	Verify that operations supported for gNMI/REST works fine for oc-transceiver:laser-temperature.
+9.	Verify that operations supported for gNMI/REST works fine for oc-transceiver:output-power.
+10.	Verify that operations supported for gNMI/REST works fine for oc-transceiver:input-power.
+11.	Verify that operations supported for gNMI/REST works fine for oc-transceiver:laser-bias-current.
+
```
jeff-yin, line 449: Add test cases for breakout as well. If a transceiver is in a breakout mode, like 4x100G, need to make sure the display is handled correctly.

doc/mgmt/OpenConfig_Platform.md
```diff
@@ -0,0 +1,454 @@
...
+## 1.2 Design Overview
+### 1.2.1 Basic Approach
+SONiC already supports telemetry for parameters like packet counters with methods such as Get via REST and gNMI. This feature adds support for telemetry for states of transceiver components with gNMI and YANG models of openconfig-transceiver.yang and with REST based on translib infra.
+
+### 1.2.2 Container
+The code changes for this feature are part of *Management Framework* container which includes the REST server and *gnmi* container for gNMI support in *sonic-mgmt-common* repository.
+
+### 1.2.3 The name of components
+openconfig-platform.yang does not hierarchically define their components.
+This design uses the name comprised of a prefix of " transceiver_" and the interface name managed by SONiC to specify the path of the target transceiver components for REST/gNMI without the the name conflicts(e.g., “transceiver_Ethernet0”).
```
jeff-yin, line 124: Discuss what happens when a breakout port is specified. It's a logical port, not necessarily physical, so what happens? (e.g., if I have a 100G port broken out into 4x25G and they're called Ethernet0-3, displaying transceiver_Ethernet2 will display what?)
- Display channel value(s) for that logical port only?
- Display channel values for all physical channels for the transceiver that includes that logical port? 

Consider breakout modes that cover multiple channels as well (e.g., 2x50G, 4x100G, etc.).

jeff-yin (changes requested):

doc/mgmt/OpenConfig_Platform.md
slicking: Given that the component is a physical transceiver it seems intuitive that there would be one name for each module.  If multiple names map to the same module (e.g. four logical interfaces in a 4x100g breakout) it may be difficult to perform operations like "query all transceivers" since duplicates would be returned.
Also note that transceivers are field replaceable units and often do need to be replaced.  There is value in having a naming scheme which can be directly tied to the labeling of cages/connectors on the front of a switch which is usually consecutive numbering (e.g. 1..32 or 0..63).  The fewer translations and re-naming steps required to go between telemetry for a transceiver and finding that transceiver on a switch, the fewer opportunities for mistakes there will be.

jeff-yin: Ok, so it would be:
- Return nothing, since we don't support non-physical ports. 

That's fine with me, so please clarify that behavior in this document.

sugisono-omw: @jeff-yin @slicking 
Thank you for your comments.
I'll report what happens when we show the information on interfaces in breakout mode. 

As a result of our investigation on the current implementation of SONiC, SONiC does not have the function to check which physical channel is allocated to the interface we want to check.
That means if a transceiver has 4 physical channels (e.g., phy0-3) and we want to allocate each channel to an interface(e.g., Ethenet0-3), we have no way to know which physical channel the interface (Ethernet0) grasps, although we can get the information on each interface. 
So, if we improve our code as it works for breakout mode, we need to implement the function to find out the physical channel each interface uses.

If we miss such functions, we would appreciate it if someone could give us the information.
We agree that we add the notation to the HLD that our code does not show the information on the interfaces in the breakout mode as the behavior of our code.

doc/mgmt/OpenConfig_Platform.md
sugisono-omw: @jeff-yin 
As I wrote the reply to the comment related to the break mode, we need to implement the function to map physical channels to the corresponding interfaces. 
We would be glad if this request could be handled for the function's future development instead of in this PR.

doc/mgmt/OpenConfig_Platform.md
sugisono-omw: Thank you for your suggestion.
We found that we can fetch the value of thresholds.
As a first step, we are thinking to implement the functions to get the value of the following thresholds. 
components/component[transceiver_XXX]/transceiver/thresholds/threshold[CRITICAL]/state/output-power-upper	 components/component[transceiver_XXX]/transceiver/thresholds/threshold[CRITICAL]/state/output-power-lower	 components/component[transceiver_XXX]/transceiver/thresholds/threshold[CRITICAL]/state/input-power-upper	 components/component[transceiver_XXX]/transceiver/thresholds/threshold[CRITICAL]/state/input-power-lower	 components/component[transceiver_XXX]/transceiver/thresholds/threshold[CRITICAL]/state/laser-bias-current-upper	 components/component[transceiver_XXX]/transceiver/thresholds/threshold[CRITICAL]/state/laser-bias-current-lower	
components/component[transceiver_XXX]/transceiver/thresholds/threshold[WARNING]/state/output-power-upper	 components/component[transceiver_XXX]/transceiver/thresholds/threshold[WARNING]/state/output-power-lower	 components/component[transceiver_XXX]/transceiver/thresholds/threshold[WARNING]/state/input-power-upper	 components/component[transceiver_XXX]/transceiver/thresholds/threshold[WARNING]/state/input-power-lower	 components/component[transceiver_XXX]/transceiver/thresholds/threshold[WARNING]/state/laser-bias-current-upper	 components/component[transceiver_XXX]/transceiver/thresholds/threshold[WARNING]/state/laser-bias-current-lower

nakano-omw:
@zhangyanzhao I would like to see it added to SONiC 202505 Release

nakano-omw:
@jeff-yin I have updated the HLD with the changes related to additional support for CRITICAL/WARNING threshold parameters. In addition, I have also submitted a code PR for the support. I would appreciate it if you could review the code. Thank you!
https://github.com/sonic-net/sonic-mgmt-common/pull/160

zhangyanzhao:
@nakano-omw  I added this feature to 202505 release. Can you please add the code PRs by referring to #806 ? Thanks.

PR #1860, opened 2024-12-04
Add Srv6 static config HLD

BYGX-wcr:
This PR proposes to add a High-level Design document for changes in SONiC to support static configuration of Segment-routing over IPv6. Besides, a YANG model specification for the new table in CONFIG_DB is also defined.

Implementation PRs:
| Module | PR link |
| :--------: | :-------: |
| sonic-bgpcfgd | https://github.com/sonic-net/sonic-buildimage/pull/21156 |
| FRR | https://github.com/FRRouting/frr/pull/16894 |
| sonic-yang-models | https://github.com/sonic-net/sonic-buildimage/pull/21175 |


doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,174 @@
...
+; New table
+; holds local SID to behavior mapping, allow 1:1 or n:1 mapping
+
+key = SRV6_MY_SID_TABLE|ipv6address
+; field = value
+block_len = blen             ; bit length of block portion in address, default 32
+node_len = nlen              ; bit length of node ID portion in address, default 16
+func_len = flen              ; bit length of function portion in address, default 16
+arg_len = alen               ; bit length of argument portion in address, default 0
+action = behavior            ; behaviors defined for the SID
```
abdosi, line 103: can we make default uN function ?

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,174 @@
...
+vrf = VRF_TABLE.key          ; Optional, VRF name for decapsulation actions
```
abdosi, line 104: Let add comment applicable only for UDT4,DT6,DT46 ?

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,174 @@
...
+module: sonic-srv6
+  +--rw sonic-srv6
+     +--rw SRV6_MY_SID
+     |  +--rw SRV6_MY_SID_LIST* [ip-address]
+     |     +--rw ip-address    inet:ipv6-address
+     |     +--rw block_len?    uint8
+     |     +--rw node_len?     uint8
+     |     +--rw func_len?     uint8
+     |     +--rw arg_len?      uint8
+     |     +--rw action?       string
```
abdosi, line 150: we can define explicit string enum here ?  for our current usecase we can just have enum of uN/A, uDT46 ?

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,174 @@
...
+
+key = SRV6_MY_SID_TABLE|ipv6address
+; field = value
+block_len = blen             ; bit length of block portion in address, default 32
+node_len = nlen              ; bit length of node ID portion in address, default 16
+func_len = flen              ; bit length of function portion in address, default 16
+arg_len = alen               ; bit length of argument portion in address, default 0
+action = behavior            ; behaviors defined for the SID
+vrf = VRF_TABLE.key          ; Optional, VRF name for decapsulation actions
+adj = address,               ; Optional, list of adjacencies for cross-connect actions
```
abdosi, line 105: we can remove this as we are not planning to test it anyways. This can be extended. in future as needed.

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: you mean uN right?

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: removed

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,174 @@
...
+     +--rw SRV6_MY_SID
+     |  +--rw SRV6_MY_SID_LIST* [ip-address]
+     |     +--rw ip-address    inet:ipv6-address
+     |     +--rw block_len?    uint8
+     |     +--rw node_len?     uint8
+     |     +--rw func_len?     uint8
+     |     +--rw arg_len?      uint8
+     |     +--rw action?       string
+     |     +--rw vrf?          -> /vrf:sonic-vrf/VRF/VRF_LIST/name
+     |     +--rw adj*          inet:ipv6-address
```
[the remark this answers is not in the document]
BYGX-wcr: removed

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: changed, uA needs to be removed per your comment.

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: added

doc/srv6/srv6_static_config_hld.md
abdosi: yes

doc/srv6/sonic-srv6.yang
```diff
@@ -0,0 +1,79 @@
...
+            leaf arg_len {
+                type uint8 {
+					range "1..128";
+				}
+
+                default 0;
+            }
+
+            leaf action {
+                type enumeration {
```
dgsudharsan, line 65: This might restrict the static configuration only for the 3 types below. Is the plan to support only these 3 types for now and later extend it?

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,162 @@
...
+        "FCBB:BBBB:20:F2::" : {
+           "action": "uDT46",
+           "vrf":  "VRF-1001"
+        },
+    }
+```
+
+
+## 3.2 Bgpcfgd changes
+
```
dgsudharsan, line 123: Can you please capture the FRR configuration sample. It would be good if you can provide sonic field to FRR configuration mapping.

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,162 @@
...
+ 
+
+# 3 Feature Design
+
+At the time of writing this document, FRR has been able to program the SRv6 related tables in APPL_DB through fpmsyncd.
+However, there is still one gap preventing SONiC being utilized for SRv6 SDN deployment.
+Specifically, there is no mechamism in SONiC allowing SDN controllers or users to directly add configuration for SRv6 without involving BGP.
+
+In this document, we first define a new **SRV6_MY_SID_TABLE** table in CONFIG_DB that serves as the configuration source of SRv6 in SONiC.
+Then, we design a new SRv6 Manager module in bgpcfgd to subscribe to the **SRV6_MY_SID_TABLE** table and compile changes in CONFIG_DB to changes in the configurations of FRR.
```
dgsudharsan, line 76: Need to mention about the sonic specific patch to support action field in FRR as by default FRR doesn't suppor it.

dgsudharsan (changes requested):

doc/srv6/sonic-srv6.yang
BYGX-wcr: Yes, currently we only plan to support these 3 types of action.

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: There is still some ongoing discussion on the details of the new FRR configuration CLI. We'd better add the config sample and mapping later.

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: I added a note

abdosi:
@dgsudharsan : Regarding supporting below requirement for DT46 End behavior do we need something like this in Config DB schema for MY_SID_TABLE ? We need OA to read this to create tunnel with the correct attribute. What do you suggest ?
  
                            TTL/DSCP/ECN Decap tunnel mode(pipe, uniform).

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,162 @@
...
+        },
+        "FCBB:BBBB:20:F2::" : {
+           "action": "uDT46",
+           "vrf":  "VRF-1001"
+        },
+    }
+```
+
+
+## 3.2 Bgpcfgd changes
```
venkatmahalingam, line 122: Are we not using frrcfgd?

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,162 @@
...
+If it gets an invalid configuration input, it should log the event in the syslog and not compile the configuration into FRR.
+
+## 3.3 YANG Model
+The simplified version of the YANG model is defined below.
+```
+module: sonic-srv6
+  +--rw sonic-srv6
+     +--rw SRV6_MY_SID
+     |  +--rw SRV6_MY_SID_LIST* [ip-address]
+     |     +--rw ip-address    inet:ipv6-address
```
venkatmahalingam, line 136: Can't we have same ipv6 address exist across VRFs? should use VRF as a key instead of field in the table?

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,162 @@
...
+| BGP  | Border Gateway Protocol |
+| SID  | Segment Identifier  |
+| SRv6 | Segment Routing IPv6  |
+| SDN | Software Defined Network |
+| uSID | Micro Segment |
+| VRF  | Virtual Routing and Forwarding  |
+
+# About this Manual
+
+This document provides general information about the design of the enhancements in SONiC to support static configuration of Segmentation Routing over IPv6 protocol, which is crucial for SRv6 SDN deployment (without usage of BGP).
```
venkatmahalingam, line 42: Are we adding support only for end-point behaviors? how about steering the traffic to SRv6 at the head-end?

doc/srv6/sonic-srv6.yang
```diff
@@ -0,0 +1,79 @@
...
+                    enum uN;
```
venkatmahalingam, line 66: This means that you support only uSID not regular end.dt6, end.dt46..etc, correct, we should mention clearly in the HLD that static SRv6 supports only uSID based handling.

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: Hi Venkatsen,

The primary reason is that we don't use frrcfgd in our daily operations. Since bgpcfgd has been leveraged to do static route configuration, we think it is not inappropriate to use bgpcfgd for static SRv6 configuration.

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: This table defines the MY_SID_LIST which should only have one instance in the current system. The VRF field is used to specify which VRF should be used for look up after the SRv6 functions decapsulates the packet. It is not used to specify which VRF the SID belongs to.

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: Yes. Right now, we only intend to add support for end-point behaviors because the traffic steering is done on the NICs in our use scenarios.

doc/srv6/srv6_static_config_hld.md
venkatmahalingam: @eddieruan-alibaba FYI.

eddieruan-alibaba: Should we discuss it in Routing WG? @BYGX-wcr  @venkatmahalingam @lguohan

doc/srv6/sonic-srv6.yang
BYGX-wcr: I added a table for the current list of supported behaviors. Feel free to extend the list after you add the necessary support and tests.

doc/srv6/srv6_static_config_hld.md
venkatmahalingam: Yes.

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,172 @@
...
+
+# 1 Introuduction and Scope
+
+This document describes the high-level design of the new features in SONiC to support SRv6 SDN.
+The new features include the addtion of a new table in CONFIG_DB to allow static configuration of SRv6 and the enhancement of bgpcfgd to program FRR with input from CONFIG_DB.
```
ahsalam, line 47: Hi @BYGX-wcr, thanks for the HLD to support SRv6 uSID static configuration. 

The static SRv6 uSID configuration model in FRR as follows: 
-	Configure an SRv6 locator and identify its parameters (name, prefix, block-len, node-len, func-bits). The CLI for this is already in FRR mainline. 
-	Configure a static SRv6 SID under the SRv6 locator and identify its parameters (behavior, vrf, etc.). The CLI for this is being added in this FRR PR (https://github.com/FRRouting/frr/pull/16894)

In the current HLD, the configuration model is different and hence can’t be mapped to the current SRv6 uSID configuration model in FRR. 

To resolve this, I would propose the following: 
-	Create a new table (SRV6_MY_LOCATOR) for the SRv6 locator config and its parameters (name, prefix, block-len, node-len, func-bits).
-	In the SRV6_MY_SID_TABLE, you add the locator as new parameter for the SID. The locator parameter points to one of the locators in SRV6_MY_LOCATOR via the locator’s name. This way you can map the config to FRR config.

FRR inherits the (block-len, node-len, func-bits) of the locator and install the static SID into SONiC APPL_DB via fpm-sonic-dplane. This is already mainline in FRR and SONiC.

eddieruan-alibaba: The locator PR is added via https://github.com/sonic-net/sonic-buildimage/pull/20954 to frrcfgd. 

The sample config db json could be found at 

https://github.com/sonic-net/sonic-mgmt/blob/master/ansible/vars/configdb_jsons/7nodes_cisco/PE3.json#L435

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: Azure Networking has an event tonight starting at 6pm so I am unable to join this week's routing WG meeting. May we discuss this in next week's routing WG meeting? I can start an email thread about this topic. @venkatmahalingam, may you share your email address with me?

venkatmahalingam: venkatesan_mahalinga@dell.com

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,172 @@
...
+
+key = SRV6_MY_SID_TABLE|ipv6address
+; field = value
+block_len = blen             ; bit length of block portion in address, default 32
+node_len = nlen              ; bit length of node ID portion in address, default 16
+func_len = flen              ; bit length of function portion in address, default 16
+arg_len = alen               ; bit length of argument portion in address, default 0
+action = behavior            ; behaviors defined for the SID, default uN
+vrf = VRF_TABLE.key          ; Optional, VRF name for decapsulation actions, only applicable to "uDT6" and "uDT46" by now, default "default"
+dscp_mode = dscp_decap_mode  ; Optional, the parameter that specifies how the node should handle DSCP bits when it performs decapsulation, default "uniform"
```
abdosi, line 105: can you add constraint that it applies only to : uDT46

doc/srv6/srv6_static_config_hld.md
abdosi: @ahsalam: I understand the logic being used by SRV6_MY_LOCATOR. However our current config db schema is good enough to program FRR. We do not want indirection logic using 2 different config tables. If you see in below example we have all the information to program frr.  We will parse ipv6 address and extract the opcode `prefix` to FRR.  Keeping one table in synch with APP_DB format is easier for us.

abdosi: @ahsalam / @eddieruan-alibaba : We also planning to remove `vrf` from config db schema as our current use case is to use `SAI_MY_SID_ENTRY_ATTR_TUNNEL_ID` for uDT46 usecase. Currently Tunnel Decap Schema https://github.com/sonic-net/SONiC/blob/56a689c40f3d68536a5ed06b592080003dbd8c7c/doc/decap/subnet_decap_HLD.md?plain=1#L149 does not support vrf for overlay interface so their is no point of specifying VRF as needed by `SAI_MY_SID_ENTRY_ATTR_VRF`.

abdosi: DT46 will use default VRF.  Our use case is not a typical L3VPN use case.

doc/srv6/srv6_static_config_hld.md
BYGX-wcr: Added

ahsalam (approved):
LGTM. Discussed the PR offline with @abdosi and @BYGX-wcr

cscarpitta (approved):

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,165 @@
...
+
+|Test Cases | Test Result |
+| :------ | :----- |
+|add config for a SID with uN action in CONFIG_DB | verify the locator config entry is created in FRR config|
+|add config for a SID with uDT46 action in CONFIG_DB | verify the opcode config entry is created in FRR config with default VRF|
+|(Negative case) add config for a SID without action in CONFIG_DB | verify that the configuration did not get into FRR config |
+|(Negative case) add config for a SID with an unsupported action in CONFIG_DB | verify that the configuration did not get into FRR config |
+|delete config for a SID with uN action in CONFIG_DB | verify the locator config entry is deleted in FRR config|
+|delete config for a SID with uDT46 action in CONFIG_DB | verify the opcode config entry for the uDT46 action is deleted in FRR config|
+
```
abdosi, line 161: @BYGX-wcr : can we add some failure UT case also. What will happen if bgpcfgd gets exception and exited ? I assume bgp/swss/syncd docker restart and everything should get reprommed but lets add the test case to confirm.

BYGX-wcr: Do you mean when bgpcfgd restarts, the SRv6 manager should retrieve all existing MY_SID config from CONFIG_DB and program them to FRR?

GaladrielZhao (approved):

zice312963205:
LGTM

zice312963205 (approved):

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,189 @@
...
+            +--rw block_len?    uint8
+            +--rw node_len?     uint8
+            +--rw func_len?     uint8
+            +--rw arg_len?      uint8
+            +--rw vrf?          union
+        +--rw SRV6_MY_SIDS_LIST* [ip_address]
+            +--rw ip_address    inet:ipv6-address
+            +--rw locator       -> /srv6:sonic-srv6/SRV6_MY_SIDS/SRV6_MY_SIDS_LIST/locator_name
+            +--rw action?       enumeration
+            +--rw vrf?          union
```
abdosi, line 170: do we need VRF here ? we already have define here: https://github.com/sonic-net/SONiC/pull/1860/files#diff-609c9acbc8e836a9c903d3f657cffcad26870bcdc9b961ab4bd6ff4c42a1a7e8R165

BYGX-wcr: Yes, they have different semantics.

eddieruan-alibaba:
This PR has been discussed on 12/20 in Routing WG. The meeting minutes could be found at https://lists.sonicfoundation.dev/g/sonic-wg-routing/wiki/38324. 

Changrong has already updated this document based on the conclusion from this meeting.

eddieruan-alibaba (dismissed):

venkatmahalingam (approved):

venkatmahalingam (changes requested):
Please see outstanding comments,

venkatmahalingam:
If FRR configuration looks the following,

        locators
            locator loc1
                prefix fd00:201:201::/48 block-len 32 node-len 16 func-bits 16
                    opcode ::fff1:11:0:0:0 end-dt46 vrf Vrf1
                    opcode ::fff1:12:0:0:0 end-dt46 vrf Vrf2
                    opcode ::fff1:14:0:0:0 end-dt4  vrf Vrf3
                    opcode ::fff1:16:0:0:0 end-dt6  vrf Vrf3
                    opcode ::fff1:18:0:0:0 end
                    
We should better represent them in the YANG model, I would suggest the following in config-db similar to ACL_TABLE and ACL_RULE tables.

**SRV6_MY_LOCATORS|locator_name**
  prefix:
  block_len:
  node_len:
  func_len:
  arg_len:
  sids = SRV6_MY_SIDS_TABLE.key

**SRV6_MY_SIDS_TABLE|sid_name**
  sid_name: 

**SRV6_MY_SIDS_OPCODE|sid_name|ipv6-address**
   sid_name: SRV6_MY_SIDS_TABLE.key
   action: uN/uDT46
   vrf:
   dscp_mode: uniform/pipe

abdosi:
> If FRR configuration looks the following,
> 
> ```
>     locators
>         locator loc1
>             prefix fd00:201:201::/48 block-len 32 node-len 16 func-bits 16
>                 opcode ::fff1:11:0:0:0 end-dt46 vrf Vrf1
>                 opcode ::fff1:12:0:0:0 end-dt46 vrf Vrf2
>                 opcode ::fff1:14:0:0:0 end-dt4  vrf Vrf3
>                 opcode ::fff1:16:0:0:0 end-dt6  vrf Vrf3
>                 opcode ::fff1:18:0:0:0 end
> ```
> 
> We should better represent them in the YANG model, I would suggest the following in config-db similar to ACL_TABLE and ACL_RULE tables.
> 
> **SRV6_MY_LOCATORS|locator_name** prefix: block_len: node_len: func_len: arg_len: sids = SRV6_MY_SIDS_TABLE.key
> 
> **SRV6_MY_SIDS_TABLE|sid_name** sid_name:
> 
> **SRV6_MY_SIDS_OPCODE|sid_name|ipv6_address** sid_name: SRV6_MY_SIDS_TABLE.key ipv6_address: action: uN/uDT46 vrf: dscp_mode: uniform/pipe

@venkatmahalingam : we had discussed this in WG meeting that it is similar to ACL model but we concluded to use the two table model that @BYGX-wcr has define. Doing above make it complex handling and that is not needed for both MSFT and @eddieruan-alibaba use case.

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,189 @@
...
+            +--rw locator_name  string
+            +--rw prefix        inet:ipv6-address
+            +--rw block_len?    uint8
+            +--rw node_len?     uint8
+            +--rw func_len?     uint8
+            +--rw arg_len?      uint8
+            +--rw vrf?          union
+        +--rw SRV6_MY_SIDS_LIST* [ip_address]
+            +--rw ip_address    inet:ipv6-address
+            +--rw locator       -> /srv6:sonic-srv6/SRV6_MY_SIDS/SRV6_MY_SIDS_LIST/locator_name
```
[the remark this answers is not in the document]
BYGX-wcr: Thanks. I have updated the YANG model spec.

doc/srv6/srv6_static_config_hld.md
jefftant: @eddieruan-alibaba 
Has static SRv6 meeting happened? Is there a recording

venkatmahalingam:
> > If FRR configuration looks the following,
> > ```
> >     locators
> >         locator loc1
> >             prefix fd00:201:201::/48 block-len 32 node-len 16 func-bits 16
> >                 opcode ::fff1:11:0:0:0 end-dt46 vrf Vrf1
> >                 opcode ::fff1:12:0:0:0 end-dt46 vrf Vrf2
> >                 opcode ::fff1:14:0:0:0 end-dt4  vrf Vrf3
> >                 opcode ::fff1:16:0:0:0 end-dt6  vrf Vrf3
> >                 opcode ::fff1:18:0:0:0 end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We should better represent them in the YANG model, I would suggest the following in config-db similar to ACL_TABLE and ACL_RULE tables.
> > **SRV6_MY_LOCATORS|locator_name** prefix: block_len: node_len: func_len: arg_len: sids = SRV6_MY_SIDS_TABLE.key
> > **SRV6_MY_SIDS_TABLE|sid_name** sid_name:
> > **SRV6_MY_SIDS_OPCODE|sid_name|ipv6_address** sid_name: SRV6_MY_SIDS_TABLE.key ipv6_address: action: uN/uDT46 vrf: dscp_mode: uniform/pipe
> 
> @venkatmahalingam : we had discussed this in WG meeting that it is similar to ACL model but we concluded to use the two table model that @BYGX-wcr has define. Doing above make it complex handling and that is not needed for both MSFT and @eddieruan-alibaba use case.

IMO, the current schema is complex because for a given locator, if you want to find/map a SIDs (opcode config), you need to literally loop all MY_SID entries or have a separate cache for locator to SIDs mapping before pushing the configs to FRR. This mapping could very well be accepted from the user if we know that opcode configs are always unique under a locator, as asked by @dgsudharsan if you could mention how you are mapping config-db static srv6 config to FRR command, that would be helpful.

 locators
        locator **loc1**
            prefix **fd00:201:201::/48** block-len 32 node-len 16 func-bits 16
                opcode ::fff1:11:0:0:0 end-dt46 vrf Vrf1
                opcode ::fff1:18:0:0:0 end
         locator **loc2**
            prefix **fd00:201:202::/48** block-len 32 node-len 16 func-bits 16
                opcode ::fff1:11:0:0:0 end-dt46 vrf Vrf1
                opcode ::fff1:18:0:0:0 end
                
I think, MY_SID (e.g opcode commands above) is unique under a locator, if that's the case, how would you use the tables SRV6_LOCATORS_LIST and SRV6_MY_SIDS_LIST for above configuration?

https://datatracker.ietf.org/doc/html/draft-ietf-spring-srv6-yang#section-3.4.1

IETF draft also talks about having a local SID under a given locator not outside of locator.
**Allows static (explicit) configuration for local-SIDs under a given locator.**

doc/srv6/srv6_static_config_hld.md
venkatmahalingam: https://lists.sonicfoundation.dev/g/sonic-wg-routing/wiki/38324

venkatmahalingam: Recording: https://zoom.us/rec/play/4sa90QBJt9cpcjO1jAM6X-9BRH-ERgMQSDAk8LZEXsRRHmn9AIs_QnO7WB02cil2Hdr0Cn3AfdaZbyRT.-S6Tl6oAPV-kFHF5?canPlayFromShare=true&from=share_recording_detail&continueMode=true&componentName=rec-play&originRequestUrl=https%3A%2F%2Fzoom.us%2Frec%2Fshare%2FDnFbtO1WyAmfKtTn0S68frJvAKMoK3dBAzUfc0_TAsX_AaolkaBLfil_6639MG_o.2LvhKP2AmovknzBz

BYGX-wcr:
Hi @venkatmahalingam,

For the FRR config example that you showed, the CONFIG_DB tables will be as follows:
```
"SRV6_MY_LOCATORS": {
    "loc1": {
        "prefix": "fd00:201:201::"
     }
    "loc2": {
        "prefix": "fd00:201:202::"
    }
}
"SRV6_MY_SIDS": {
    "default|fd00:201:201:fff1:11::": {
        "action": "uDT46",
        "decap_vrf": "Vrf1"
    }
    "default|fd00:201:201:fff1:18::":  {
        "action": "uN"
    }
    "default|fd00:201:202:fff1:11::":  {
        "action": "uDT46",
        "decap_vrf": "Vrf1"
    }
    "default|fd00:201:202:fff1:18::": {
        "action": "uN"
    }
}
```

It is true that if you want to iterate through all SIDs of a given locator, you need to scan the whole SRV6_MY_SIDS table.  Yet, the expected way in which bgpcfgd works is that the SRv6 manager only programs FRR config when there are changes in SRV6_MY_SIDS table. The SRV6_MY_LOCATORS table is mainly used as a reference table for validation.

In Alibaba's use case where they start collecting config with locators, scanning the whole SRV6_MY_SIDS table is necessary anyway. I don't see any potential problem.

venkatmahalingam:
> Hi @venkatmahalingam,
> 
> For the FRR config example that you showed, the CONFIG_DB tables will be as follows:
> 
> ```
> "SRV6_MY_LOCATORS": {
>     "loc1": {
>         "prefix": "fd00:201:201::"
>      }
>     "loc2": {
>         "prefix": "fd00:201:202::"
>     }
> }
> "SRV6_MY_SIDS": {
>     "default|fd00:201:201:fff1:11::": {
>         "action": "uDT46",
>         "decap_vrf": "Vrf1"
>     }
>     "default|fd00:201:201:fff1:18::":  {
>         "action": "uN"
>     }
>     "default|fd00:201:202:fff1:11::":  {
>         "action": "uDT46",
>         "decap_vrf": "Vrf1"
>     }
>     "default|fd00:201:202:fff1:18::": {
>         "action": "uN"
>     }
> }
> ```
> 
> It is true that if you want to iterate through all SIDs of a given locator, you need to scan the whole SRV6_MY_SIDS table. Yet, the expected way in which bgpcfgd works is that the SRv6 manager only programs FRR config when there are changes in SRV6_MY_SIDS table. The SRV6_MY_LOCATORS table is mainly used as a reference table for validation.
> 
> In Alibaba's use case where they start collecting config with locators, scanning the whole SRV6_MY_SIDS table is necessary anyway. I don't see any potential problem. @eddieruan-alibaba , correct me if I am wrong.

My understanding is LOCATOR table is not just for reference; we have to use both the tables LOCATOR & MY_SID (to carry only opcode related info.) for FRR configuration, when we have the whole SID in the MY_SID table, not sure why do we need prefix in the LOCATOR table. 

Having said that, I still believe that you can simply use the existing below config-DB schema that Alibaba defined (may not be perfect from YANG perspective) for your use-case as well instead of creating yet another new schema (SRV6_MY_SID_TABLE) in the config-db, the below table exactly maps with the FRR commands to avoid any confusion as to when to use SRV6_LOCATOR that Alibaba defined vs SRV6_MY_SID_TABLE Msft proposed for static SRv6.

  "SRV6_LOCATOR": {
        "lsid1": {
            "argu_len": "48",
            "block_len": "32",
            "func_len": "32",
            "node_len": "16",
            "opcode_prefix": [
                "::fff1:1:0:0:0",
                "::fff1:11:0:0:0"
            ],
            "opcode_act": [
                "end-dt46",
                "end-dt46"
            ],
            "opcode_data": [
                "vrf Vrf1",
                "vrf Vrf2"
            ],
            "prefix": "fd00:201:201::/48"
        }
    },
    
    
    I see what you are doing in the bgpcfgd (https://github.com/sonic-net/sonic-buildimage/pull/21156/files) i.e mapping SID FCBB:BBBB:20:F1:: to locator commands, IMO, we can simply follow SRV6_LOCATOR that Alibaba defined for your case as well, let me know if I'm missing something that makes SRV6_MY_SID_TABLE table in the config-db as a better option.
    
    def test_uN_add():
    mgr = constructor()

    op_test(mgr, 'SET', ("FCBB:BBBB:20:F1::", {
        'action': 'uN'
    }), expected_ret=True, expected_cmds=[
        'segment-routing',
        'srv6',
        'locators',
        'locator FCBB:BBBB:20:: block-len 32 node-len 16 func-bits 16',
        'prefix FCBB:BBBB:20::/48',
        'sid FCBB:BBBB:20:F1::/64 uN'
    ])

BYGX-wcr:
> > Hi @venkatmahalingam,
> > For the FRR config example that you showed, the CONFIG_DB tables will be as follows:
> > ```
> > "SRV6_MY_LOCATORS": {
> >     "loc1": {
> >         "prefix": "fd00:201:201::"
> >      }
> >     "loc2": {
> >         "prefix": "fd00:201:202::"
> >     }
> > }
> > "SRV6_MY_SIDS": {
> >     "default|fd00:201:201:fff1:11::": {
> >         "action": "uDT46",
> >         "decap_vrf": "Vrf1"
> >     }
> >     "default|fd00:201:201:fff1:18::":  {
> >         "action": "uN"
> >     }
> >     "default|fd00:201:202:fff1:11::":  {
> >         "action": "uDT46",
> >         "decap_vrf": "Vrf1"
> >     }
> >     "default|fd00:201:202:fff1:18::": {
> >         "action": "uN"
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It is true that if you want to iterate through all SIDs of a given locator, you need to scan the whole SRV6_MY_SIDS table. Yet, the expected way in which bgpcfgd works is that the SRv6 manager only programs FRR config when there are changes in SRV6_MY_SIDS table. The SRV6_MY_LOCATORS table is mainly used as a reference table for validation.
> > In Alibaba's use case where they start collecting config with locators, scanning the whole SRV6_MY_SIDS table is necessary anyway. I don't see any potential problem. @eddieruan-alibaba , correct me if I am wrong.
> 
> My understanding is LOCATOR table is not just for reference; we have to use both the tables LOCATOR & MY_SID (to carry only opcode related info.) for FRR configuration, when we have the whole SID in the MY_SID table, not sure why do we need prefix in the LOCATOR table.
> 
> Having said that, I still believe that you can simply use the existing below config-DB schema that Alibaba defined (may not be perfect from YANG perspective) for your use-case as well instead of creating yet another new schema (SRV6_MY_SID_TABLE) in the config-db, the below table exactly maps with the FRR commands to avoid any confusion as to when to use SRV6_LOCATOR that Alibaba defined vs SRV6_MY_SID_TABLE Msft proposed for static SRv6.
> 
> "SRV6_LOCATOR": { "lsid1": { "argu_len": "48", "block_len": "32", "func_len": "32", "node_len": "16", "opcode_prefix": [ "::fff1:1:0:0:0", "::fff1:11:0:0:0" ], "opcode_act": [ "end-dt46", "end-dt46" ], "opcode_data": [ "vrf Vrf1", "vrf Vrf2" ], "prefix": "fd00:201:201::/48" } },
> 
> ```
> I see what you are doing in the bgpcfgd (https://github.com/sonic-net/sonic-buildimage/pull/21156/files) i.e mapping SID FCBB:BBBB:20:F1:: to locator commands, IMO, we can simply follow SRV6_LOCATOR that Alibaba defined for your case as well, let me know if I'm missing something that makes SRV6_MY_SID_TABLE table in the config-db as a better option.
> 
> def test_uN_add():
> mgr = constructor()
> 
> op_test(mgr, 'SET', ("FCBB:BBBB:20:F1::", {
>     'action': 'uN'
> }), expected_ret=True, expected_cmds=[
>     'segment-routing',
>     'srv6',
>     'locators',
>     'locator FCBB:BBBB:20:: block-len 32 node-len 16 func-bits 16',
>     'prefix FCBB:BBBB:20::/48',
>     'sid FCBB:BBBB:20:F1::/64 uN'
> ])
> ```

The addition of prefix field in the SRV6_MY_LOCATORS is for the completeness of locator definition and being consistent with FRR.

The reason why we decided to abandon Alibaba's single table design was because people complaint about the implicit semantics in the "opcode_data" field and difficulty in YANG validation. You can revisit the recording of Routing WG discussion to confirm that.

venkatmahalingam:
@BYGX-wcr 
The addition of prefix field in the SRV6_MY_LOCATORS is for the completeness of locator definition and being consistent with FRR.

The reason why we decided to abandon Alibaba's single table design was because people complaint about the implicit semantics in the "opcode_data" field and difficulty in YANG validation. You can revisit the recording of Routing WG discussion to confirm that.

> Yes, I'm aware of what we discussed in the last WG meeting, after looking at the pull requests you have created for this feature, I'm thinking we should not over-complicate the existing schema unless it provides some value, 
I have created following options; we can briefly discuss this if needed in the next WG call.

I was one the mentioning about "the difficulty in YANG validation" in the WG meeting, either we can continue to use the existing SRV6_LOCATOR table (Alibaba proposed) for static SRV6 as well (option1) and clean up later for easy YANG validation (option2) or do the option2 now itself.

**Option1: (Short term)** - Only if we think, Option 2 is not possible now given the timeline for this static SRv6 activity (yes, we are aware of the YANG validation issues, if needed, we can try custom YANG validation for time being)
 "SRV6_LOCATOR": {
        "lsid1": {
            "argu_len": "48",
            "block_len": "32",
            "func_len": "32",
            "node_len": "16",
            "opcode_prefix": [
                "::fff1:1:0:0:0",
                "::fff1:11:0:0:0"
            ],
            "opcode_act": [
                "end-dt46",
                "end-dt46"
            ],
            "opcode_data": [
                "vrf Vrf1",
                "vrf Vrf2"
            ],
            "prefix": "fd00:201:201::/48"
        }
    },
    
 **Option2: (Long term)**   -> This is aligned with IETF YANG (https://datatracker.ietf.org/doc/html/draft-ietf-spring-srv6-yang#section-3.4.1) and we can do YANG validation better. We can choose appropriate table names if the below ones are not correct.
     "SRV6_LOCATOR": {
        "lsid1": {
            "argu_len": "48",
            "block_len": "32",
            "func_len": "32",
            "node_len": "16",
            "sids":  sid1
            "prefix": "fd00:201:201::/48"
        },
    },
    SRV6_MY_SIDS_TABLE|sid1: {
    },
    "SRV6_MY_SIDS_OPCODE_TABLE": {
       "sid1|::fff1:1:0:0:0:" {
           "action" : end-dt46
           "vrf" : Vrf1
        },
       "sid1|::fff1:11:0:0:0" {
          "action" : end-dt46
          "vrf" : Vrf2
     },
   }
    
Option3: 
My understanding from the last WG meeting,
"SRV6_LOCATOR": {
        "lsid1": {
            "argu_len": "48",
            "block_len": "32",
            "func_len": "32",
            "node_len": "16",
            "prefix": "fd00:201:201::/48"
        },
    },
    "SRV6_MY_SIDS_TABLE": {
       "::fff1:1:0:0:0:" {
           "locator": lsid1
            "action" : end-dt46
           "vrf" : Vrf1
        },
       ":fff1:11:0:0:0:" {
           "locator": lsid1
            "action" : end-dt46
           "vrf" : Vrf2
        },
   }

but I see the below updates in the HLD, I see a problem (in general for option3) when any changes to the lsid1 entry can cause to remove the entire locator from FRR and re-add with updated locator and all its opcode commands (we may need to loop entire SRV6_MY_SIDS_TABLE as locator is not the key, also when anyone removes lsid1, we should make sure no stale entries in SRV6_MY_SIDS_TABLE pointing to lsid1, this is not possible with direct YANG validation, we may need custom validation). However, this is not a cleaner solution IMO. I prefer option2, if not, option1 for time being. Option3, not only inconsistent with current LOCATOR table (Alibaba proposed), also not a good long-term solution for a proper YANG validation.
"SRV6_LOCATOR": {
        "lsid1": {
            "argu_len": "48",
            "block_len": "32",
            "func_len": "32",
            "node_len": "16",
            "prefix": "fd00:201:201::/48"
        },
    },
    "SRV6_MY_SIDS_TABLE": {
       "fd00:201:201::fff1:1:0:0:0:" {
           "locator": lsid1
            "action" : end-dt46
           "vrf" : Vrf1
        },
       "fd00:201:201:fff1:11:0:0:0:" {
           "locator": lsid1
            "action" : end-dt46
           "vrf" : Vrf2
        },
   }

GaladrielZhao:
> @BYGX-wcr The addition of prefix field in the SRV6_MY_LOCATORS is for the completeness of locator definition and being consistent with FRR.
> 
> ## The reason why we decided to abandon Alibaba's single table design was because people complaint about the implicit semantics in the "opcode_data" field and difficulty in YANG validation. You can revisit the recording of Routing WG discussion to confirm that.
> Yes, I'm aware of what we discussed in the last WG meeting, after looking at the pull requests you have created for this feature, I'm thinking we should not over-complicate the existing schema unless it provides some value, I have created following options; we can briefly discuss this if needed in the next WG call.
> 
> I was one the mentioning about "the difficulty in YANG validation" in the WG meeting, after looking at the changes you have made, I think, either we can continue to use the existing SRV6_LOCATOR table (Alibaba proposed) for static SRV6 as well (option1) and clean up later for easy YANG validation (option2) or do the option2 now itself.
> 
> **Option1: (Short term)** - Only if we think, Option 2 is not possible now given the timeline for this static SRv6 activity (yes, we are aware of the YANG validation issues, if needed, we can try custom YANG validation for time being)
> 
> ```
>  "SRV6_LOCATOR": {
>         "lsid1": {
>             "argu_len": "48",
>             "block_len": "32",
>             "func_len": "32",
>             "node_len": "16",
>             "opcode_prefix": [
>                 "::fff1:1:0:0:0",
>                 "::fff1:11:0:0:0"
>             ],
>             "opcode_act": [
>                 "end-dt46",
>                 "end-dt46"
>             ],
>             "opcode_data": [
>                 "vrf Vrf1",
>                 "vrf Vrf2"
>             ],
>             "prefix": "fd00:201:201::/48"
>         }
>     },
> ```
> 
> **Option2: (Long term)** -> This is aligned with IETF YANG (https://datatracker.ietf.org/doc/html/draft-ietf-spring-srv6-yang#section-3.4.1) and we can do YANG validation better. We can choose appropriate table names if the below ones are not correct.
> 
> ```
> "SRV6_LOCATOR": {
>        "lsid1": {
>            "argu_len": "48",
>            "block_len": "32",
>            "func_len": "32",
>            "node_len": "16",
>            "sids":  sid1
>            "prefix": "fd00:201:201::/48"
>        },
>    },
>    SRV6_MY_SIDS_TABLE|sid1: {
>    },
>    "SRV6_MY_SIDS_OPCODE_TABLE": {
>       "sid1|::fff1:1:0:0:0:" {
>           "action" : end-dt46
>           "vrf" : Vrf1
>        },
>       "sid1|::fff1:11:0:0:0" {
>          "action" : end-dt46
>          "vrf" : Vrf2
>     },
>   }
> ```
> 
> Option3: My understanding from the last WG meeting,
> 
> ```
> "SRV6_LOCATOR": {
>         "lsid1": {
>             "argu_len": "48",
>             "block_len": "32",
>             "func_len": "32",
>             "node_len": "16",
>             "prefix": "fd00:201:201::/48"
>         },
>     },
>     "SRV6_MY_SIDS_TABLE": {
>        "::fff1:1:0:0:0:" {
>            "locator": lsid1
>             "action" : end-dt46
>            "vrf" : Vrf1
>         },
>        ":fff1:11:0:0:0:" {
>            "locator": lsid1
>             "action" : end-dt46
>            "vrf" : Vrf2
>         },
>    }
> ```
> 
> but I see the below updates in the HLD. So, I see a problem (in general for option3) when any changes to the lsid1 entry can cause to remove the entire locator from FRR and re-add with updated locator and all its opcode commands (we may need to loop entire SRV6_MY_SIDS_TABLE as locator is not the key, also when anyone removes lsid1, we should make sure no stale entries in SRV6_MY_SIDS_TABLE pointing to lsid1, this is not possible with direct YANG validation, we may need custom validation). However, this is not a cleaner solution IMO. I prefer option2, if not, option1 for time being. Option3, not only inconsistent with current LOCATOR table (Alibaba proposed), also not a good long-term solution for a proper YANG validation.
> 
> ```
> "SRV6_LOCATOR": {
>         "lsid1": {
>             "argu_len": "48",
>             "block_len": "32",
>             "func_len": "32",
>             "node_len": "16",
>             "prefix": "fd00:201:201::/48"
>         },
>     },
>     "SRV6_MY_SIDS_TABLE": {
>        "fd00:201:201::fff1:1:0:0:0:" {
>            "locator": lsid1
>             "action" : end-dt46
>            "vrf" : Vrf1
>         },
>        "fd00:201:201:fff1:11:0:0:0:" {
>            "locator": lsid1
>             "action" : end-dt46
>            "vrf" : Vrf2
>         },
>    }
> ```

Hi @venkatmahalingam ,

We(Alibaba) actually concluded in our last discussion not to use the "opcode" field anymore, but instead to use the full SID. This way, there will be no "implicit semantics", making the configuration clearer and easier to validate with YANG.

If we use the full SID under the locator, the FRR configuration would be like:
```
locators
    locator loc1
        prefix fcbb:bbbb:1::/48 block-len 32 node-len 16 func-bits 16
        sid     fcbb:bbbb:1:fe01:: behavior uDT6 vrf Vrf1
        sid     fcbb:bbbb:1:fe02:abcd:: behavior uDT4 vrf Vrf1
        sid     fcbb:bbbb:1:fe03:abcd:abcd:: behavior uDT46 vrf Vrf2
    locator loc2
        prefix fcdd:dddd:2::/48 block-len 32 node-len 16 func-bits 16
        sid     fcdd:dddd:2:fe11:abcd:: behavior uDT6 vrf Vrf2
        sid     fcdd:dddd:2:fe12:: behavior uDT46 vrf Vrf3
```
And the schema would be (just an example for SID format):
```
"SRV6_LOCATOR": {
    "lsid1": {
        "argu_len": "48",
        "block len": "32",
        "func len": "32",
        "node_len": "16",
        "sid": [
            "fd00:201:201:fff1:1:0:0:0/80",
            "fd00:201:201:fff1:11:0:0:0/80"
        ],
        "opcode_act": [
            "end-dt46",
            "end-dt46"
        ],
        "opcode_data": [
            "vrf Vrf1",
            "vrf Vrf2"
        ],
        "prefix": "fd00:201:201::/48"
    }
}
```

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,192 @@
...
+```
+; New table
+; holds local SID to behavior mapping, the keys are full IPv6 addresses of the SIDs
+
+key = SRV6_MY_SIDS|vrf|ip_address
+; field = value
+locator = locator_name       ; the name of the locator that the SID belongs to
+action = behavior            ; behaviors defined for the SID, default uN
+decap_vrf = VRF_TABLE.key          ; Optional, VRF name for decapsulation actions, default "default", only applicable to uDT4/uDT46/uDT6 actions
+dscp_mode = dscp_decap_mode  ; Optional, the parameter that specifies how the node should handle DSCP bits when it performs decapsulation, default "uniform", only applicable to uDT4/uDT46/uDT6 actions
```
abdosi, line 125: rename it to decap_dscp_mode

BYGX-wcr: done

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,192 @@
...
+
+For example:
+    "SRV6_MY_SIDS" : {
+        "default|FCBB:BBBB:20::" : {
+           "action": "uN"
+        },
+        "default|FCBB:BBBB:20:F1::" : {
+           "action": "uDT46",
```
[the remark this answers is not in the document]
BYGX-wcr: added

dgsudharsan:
> @BYGX-wcr The addition of prefix field in the SRV6_MY_LOCATORS is for the completeness of locator definition and being consistent with FRR.
> 
> ## The reason why we decided to abandon Alibaba's single table design was because people complaint about the implicit semantics in the "opcode_data" field and difficulty in YANG validation. You can revisit the recording of Routing WG discussion to confirm that.
> Yes, I'm aware of what we discussed in the last WG meeting, after looking at the pull requests you have created for this feature, I'm thinking we should not over-complicate the existing schema unless it provides some value, I have created following options; we can briefly discuss this if needed in the next WG call.
> 
> I was one the mentioning about "the difficulty in YANG validation" in the WG meeting, after looking at the changes you have made, I think, either we can continue to use the existing SRV6_LOCATOR table (Alibaba proposed) for static SRV6 as well (option1) and clean up later for easy YANG validation (option2) or do the option2 now itself.
> 
> **Option1: (Short term)** - Only if we think, Option 2 is not possible now given the timeline for this static SRv6 activity (yes, we are aware of the YANG validation issues, if needed, we can try custom YANG validation for time being)
> 
> ```
>  "SRV6_LOCATOR": {
>         "lsid1": {
>             "argu_len": "48",
>             "block_len": "32",
>             "func_len": "32",
>             "node_len": "16",
>             "opcode_prefix": [
>                 "::fff1:1:0:0:0",
>                 "::fff1:11:0:0:0"
>             ],
>             "opcode_act": [
>                 "end-dt46",
>                 "end-dt46"
>             ],
>             "opcode_data": [
>                 "vrf Vrf1",
>                 "vrf Vrf2"
>             ],
>             "prefix": "fd00:201:201::/48"
>         }
>     },
> ```
> 
> **Option2: (Long term)** -> This is aligned with IETF YANG (https://datatracker.ietf.org/doc/html/draft-ietf-spring-srv6-yang#section-3.4.1) and we can do YANG validation better. We can choose appropriate table names if the below ones are not correct.
> 
> ```
> "SRV6_LOCATOR": {
>        "lsid1": {
>            "argu_len": "48",
>            "block_len": "32",
>            "func_len": "32",
>            "node_len": "16",
>            "sids":  sid1
>            "prefix": "fd00:201:201::/48"
>        },
>    },
>    SRV6_MY_SIDS_TABLE|sid1: {
>    },
>    "SRV6_MY_SIDS_OPCODE_TABLE": {
>       "sid1|::fff1:1:0:0:0:" {
>           "action" : end-dt46
>           "vrf" : Vrf1
>        },
>       "sid1|::fff1:11:0:0:0" {
>          "action" : end-dt46
>          "vrf" : Vrf2
>     },
>   }
> ```
> 
> Option3: My understanding from the last WG meeting,
> 
> ```
> "SRV6_LOCATOR": {
>         "lsid1": {
>             "argu_len": "48",
>             "block_len": "32",
>             "func_len": "32",
>             "node_len": "16",
>             "prefix": "fd00:201:201::/48"
>         },
>     },
>     "SRV6_MY_SIDS_TABLE": {
>        "::fff1:1:0:0:0:" {
>            "locator": lsid1
>             "action" : end-dt46
>            "vrf" : Vrf1
>         },
>        ":fff1:11:0:0:0:" {
>            "locator": lsid1
>             "action" : end-dt46
>            "vrf" : Vrf2
>         },
>    }
> ```
> 
> but I see the below updates in the HLD. So, I see a problem (in general for option3) when any changes to the lsid1 entry can cause to remove the entire locator from FRR and re-add with updated locator and all its opcode commands (we may need to loop entire SRV6_MY_SIDS_TABLE as locator is not the key, also when anyone removes lsid1, we should make sure no stale entries in SRV6_MY_SIDS_TABLE pointing to lsid1, this is not possible with direct YANG validation, we may need custom validation). However, this is not a cleaner solution IMO. I prefer option2, if not, option1 for time being. Option3, not only inconsistent with current LOCATOR table (Alibaba proposed), also not a good long-term solution for a proper YANG validation.
> 
> ```
> "SRV6_LOCATOR": {
>         "lsid1": {
>             "argu_len": "48",
>             "block_len": "32",
>             "func_len": "32",
>             "node_len": "16",
>             "prefix": "fd00:201:201::/48"
>         },
>     },
>     "SRV6_MY_SIDS_TABLE": {
>        "fd00:201:201::fff1:1:0:0:0:" {
>            "locator": lsid1
>             "action" : end-dt46
>            "vrf" : Vrf1
>         },
>        "fd00:201:201:fff1:11:0:0:0:" {
>            "locator": lsid1
>             "action" : end-dt46
>            "vrf" : Vrf2
>         },
>    }
> ```

@venkatmahalingam  Can you clarify why yang can't support back ref? Currently when locator table is removed the yang validation for my_sid table will fail as it points to a locator without reference. I believe option#3 aligns with other tables ACL rule has a reference to ACL table https://github.com/sonic-net/sonic-buildimage/blob/cb85ddb33f0a9e54e0e6a11635e04af489ee4d9e/src/sonic-yang-models/yang-templates/sonic-acl.yang.j2#L56

dgsudharsan:
@BYGX-wcr @abdosi @venkatmahalingam The concern that Venkat raises whether the key of SRV6_MY_SID_TABLE is a unique key across locators or not. If not unique, it may lead to problems that you cannot have two locators pointed by the same my_sid entry. Please clarify if this is a valid use case or not. 
If valid it makes sense to use option #2 but if this is not a valid use case option #3 makes sense.

@eddieruan-alibaba

BYGX-wcr:
Hi @dgsudharsan @venkatmahalingam ,

I have fixed the inconsistencies in the document per our discussions. I also added an example for the MY_LOCATORS table.

dgsudharsan (approved):

venkatmahalingam (approved):

eddieruan-alibaba:
@BYGX-wcr @venkatmahalingam @dgsudharsan I saw all parties have approved this PR. Do we still need to discuss three options listed in this week's WG?

eddieruan-alibaba:
> @BYGX-wcr @abdosi @venkatmahalingam The concern that Venkat raises whether the key of SRV6_MY_SID_TABLE is a unique key across locators or not. If not unique, it may lead to problems that you cannot have two locators pointed by the same my_sid entry. Please clarify if this is a valid use case or not. If valid it makes sense to use option #2 but if this is not a valid use case option #3 makes sense.
> 
> @eddieruan-alibaba

MY_SID is used similar as vpn label in MPLS logically. it should be unique at the local device.

venkatmahalingam:
> @BYGX-wcr @venkatmahalingam @dgsudharsan I saw all parties have approved this PR. Do we still need to discuss three options listed in this week's WG?

No, we're good @eddieruan-alibaba

eddieruan-alibaba (approved):

abdosi:
@dgsudharsan : Please add OA PR in this PR description.

Yakiv-Huryk:
> @dgsudharsan : Please add OA PR in this PR description.

The PRs:
https://github.com/sonic-net/sonic-swss-common/pull/962
https://github.com/sonic-net/sonic-swss/pull/3414
https://github.com/sonic-net/sonic-swss/pull/3443

doc/srv6/srv6_static_config_hld.md
rlhui: Are we ready to add FRR config now?

doc/srv6/srv6_static_config_hld.md
```diff
@@ -0,0 +1,204 @@
...
+The current list of supported SRv6 behaviors allowed to be define in CONFIG_DB is as follows:
+
+| Alias | SRv6 Behaviors |
+| :------ | :----- |
+| uN | End with NEXT-CSID |
+| uDT46 | End.DT46 with CSID |
+
+## 3.2 Bgpcfgd changes
+
+To enable automatic programming SRv6 configurations from CONFIG_DB to FRR, we need to add a new module in bgpcfgd to watch changes in **SRV6_MY_LOCATORS** and **SRV6_MY_SIDS** and compile the corresponding changes in FRR's configurations.
```
rlhui, line 161: Can we also add the code flow from FRR to app_db, which table in app_db is used for my SID to be programmed to HW?

BYGX-wcr: @ahsalam , @cscarpitta , can you help with this issue?

zhangyanzhao:
mark this feature as done for 202505 release since HLD and all code PRs are merged.

PR #1863, opened 2024-12-09
Support config policer and bind to interface or ACL table

baozhen-H3C:
The current design does not support configuring traffic policers through the command line, and they cannot be bound to interfaces or ACL tables. This document provides recommendations for supporting this function.

### DIFF
--- a/doc/qos/Configure_policer_and_bind_to_interface_or_ACL_table.md
+++ b/doc/qos/Configure_policer_and_bind_to_interface_or_ACL_table.md
@@ -1,0 +1,416 @@
+# Configure Policer and bind it to interface or ACL table HLD #
+
+## 1.Table of Content 
+- [Policer configure HLD](#Policer-configure-hld)
+  - [1.Table of Content](#1table-of-content)
+    - [1.1. Revision](#11-revision)
+    - [1.2. Scope](#12-scope)
+    - [1.3. Definitions/Abbreviations](#13-definitionsabbreviations)
+    - [1.4. Overview](#14-overview)
+    - [1.5. Requirements](#15-requirements)
+    - [1.6. Architecture Design](#16-architecture-design)
+    - [1.7. High-Level Design](#17-high-level-design)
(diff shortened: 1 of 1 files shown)


doc/qos/Configure_policer_and_bind_to_interface_or_ACL_table.md
```diff
@@ -0,0 +1,416 @@
...
+	      type enumeration {
+	          enum L2;
+	          enum L3;
+	          enum L3V6;
+	          enum MIRROR;
+	          enum MIRRORV6;
+	          enum MIRROR_DSCP;
+	          enum CTRLPLANE;
+	          enum MIRROR;
+	          enum POLICER;
```
Griffin-micas, line 312: Perhaps it's not necessary to introduce a separate POLICER type. The POLICER_ACTION can be treated as an independent action and applied to rules in other types of ACL tables.

doc/qos/Configure_policer_and_bind_to_interface_or_ACL_table.md
```diff
@@ -0,0 +1,416 @@
...
+
+- Notes:
+  - <mode\> could be "sr_tcm/tr_tcm".
+  - <meter_type\> could be "packets/bytes".
+  - <color_source\> could be "blind/aware". "Blind" indicates that the policer uses a color-blind mode to mark traffic packets, while "aware" indicates that the policer uses a color-aware mode to mark traffic packets.
+  - <green_packet_action\> could be "drop/forward/copy/copy_cancel/trap/log/deny/transit". The same applies to <yellow_packet_action\> and <red_packet_action\>.
+
+- Example:
+  ```
+  admin@sonic:~$ sudo config policer add policer_1 -mode sr_tcm -mtype packets -csource blind -cir 1024 -cbs 2048
```
Griffin-micas, line 111: Perhaps you could further explain, if there's a need to revise the attributes of a policer, whether it's updated directly or the existing policer is removed and recreated.

PR #1873, opened 2024-12-18
Make smart switch HA detailed design doc less ambiguous

erer1243:
1. fixed a weird artifact of the mermaid graph where arrows were overlapping
2. Change all instances of empty string value to `none` to make remove ambiguity of "" being an empty list or a single element list of the element "". It may make sense for some of these to be called `auto` or `default` instead - eg `outbound_eni_mac_lookup`. Let me know your thoughts.
3. Make some field names more consistent


r12f (approved):
lgtm! Thanks Oliver!

PR #1893, opened 2025-01-17
Change the key of SRV6_MY_SIDS table from ipv6-address type to ipv6-prefix type

BYGX-wcr:
Summary: This is a small fix to enhance the original HLD to make it more aligned with the FRR's CLI.

Originally, we thought that the key of the SRV6_MY_SIDS table should be an ipv6-address type and the prefix of the SID should be determined by block_len + node_len + func_len. However, with current FRR's configuration model, the length of a SID is actually flexible as long as it falls under the locator's prefix. Thus, in the CONFIG_BD, we should specify the SID as an ipv6 prefix instead of an address so that the user can define the length of the static SIDs.

### DIFF
--- a/doc/srv6/srv6_static_config_hld.md
+++ b/doc/srv6/srv6_static_config_hld.md
@@ -127,23 +127,25 @@ Schema:
 ; New table
 ; holds local SID to behavior mapping, the keys are the locator name plus the full IPv6 addresses of the SIDs
 
-key = SRV6_MY_SIDS|locator|ip_address
+key = SRV6_MY_SIDS|locator|ip_prefix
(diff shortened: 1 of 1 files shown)


abdosi:
please add some description in PR for the reason of change for example to align config model with FRR configuration.

abdosi (approved):

r12f (approved):

BYGX-wcr:
> please add some description in PR for the reason of change for example to align config model with FRR configuration.

Updated

eddieruan-alibaba (approved):
Yes. That is FRR's current behavior.

PR #1897, opened 2025-01-22
EVPN_VXLAN_HLD: update to allow recording for p2mp remote vteps

bradh352:
Complimentary ticket to https://github.com/sonic-net/sonic-swss/pull/3480 in order to allow p2mp vteps to be tracked.  This will mostly aid in debugging.  The HLD specifically states this when talking about P2P vs P2MP:

"For an external user, there will be no changes from usability perspective since the schema is unchanged."

However this is not entirely true since the state table wasn't recording p2mp remote vteps and therefore `show vxlan remotevtep` would return no results making a user believe the vxlan tunnel was not properly established.  This as well as the aforementioned PR corrects this oversight.

Signed-off-by: Brad House (@bradh352)

### DIFF
--- a/doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
+++ b/doc/vxlan/EVPN/EVPN_VXLAN_HLD.md
@@ -76,6 +76,7 @@
 | 0.9 | | Nikhil Kelhapure | Warm Reboot Section added |
 | 1.0 | | Sudharsan D.G | Using P2MP Tunnel for Layer2 functionality |
 | 1.1 | | Adam Yeung | Update Fastboot limitation |
+| 1.2 | 2025-01-22 | Brad House | Record p2mp vxlan remote vteps in state table |
 
 
 # Definition/Abbreviation
@@ -521,7 +522,7 @@ key             = VXLAN_TUNNEL:name    ; tunnel name
 SRC_IP          = ipv4                 ; tunnel sip
 DST_IP          = ipv4                 ; tunnel dip 
 tnl_src         = "CLI"/"EVPN"  
-operstatus      = "oper_up"/"oper_down"  
+operstatus      = "oper_up"/"oper_down"/"oper_p2mp"
 ```
 
 ### 4.2.4 COUNTER_DB changes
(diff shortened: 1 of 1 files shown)


bradh352:
@dgsudharsan mind reviewing this change?  You seem to have been the one to contribute the P2MP vs P2P information.

dgsudharsan:
@bradh352 Unlike P2P where we have individual tunnel objects, P2MP implies there is only one tunnel in the local device.  Since there is only one tunnel object, state_db was intentionally  not populated.

bradh352:
@dgsudharsan
> @bradh352 Unlike P2P scenario where we have individual tunnel objects, P2MP implies there is only one tunnel in the local device. Since there is only one tunnel object, state_db was intentionally not populated and mentioned so in the HLD as well

Sure, but that's more of an ***internal*** design aspect, right?  The HLD specifically states that there are no usability changes, but there is infact a difference ... even though the user has no knowledge of if the underlying implementation is p2p vs p2mp.  

You can't expect a user to know that when they log into their Dell S5248F switch they will see `show vxlan remotevtep` peers, but when they log into their Mellanox SN2100 switch they won't.  

As far as I know, there's no way to simply show the remote vteps on P2MP without this patch (you can get it via vtysh obviously).

What users see should be the same from silicon to silicon for common features.

Are you completely against this change, or would you rather it done a different way?

bradh352:
any further comments?

dgsudharsan:
> any further comments?

There is still some changes between the two approaches and the user needs to be aware. This CLI will wrongly imply that there are multiple P2P tunnels where the table says SIP, DIP. For P2MP scenario, there is only one tunnel. This The remote flooding is done through L2MC group with remote end points. 

To avoid misleading user this CLI is intentionally left blank.Having this will imply platform will support P2P tunnels. As you mentioned any user who wants to know would use FRR CLI to figure it out.

bradh352:
> > any further comments?
> 
> There is still some changes between the two approaches and the user needs to be aware. This CLI will wrongly imply that there are multiple P2P tunnels where the table says SIP, DIP. For P2MP scenario, there is only one tunnel. This The remote flooding is done through L2MC group with remote end points.
> 
> To avoid misleading user this CLI is intentionally left blank.Having this will imply platform will support P2P tunnels. As you mentioned any user who wants to know would use FRR CLI to figure it out.

The command is `show vxlan remotevtep` not something like `show vxlan tunnels`, so the proposed change doesn't conflict with the spirit of the command.  And the `oper_status` now shows p2mp instead of up/down for p2mp tunnels for clarification.  

I don't think someone should need to evaluate FRR output for something like this, especially when using frr-management-framework where they're not directly configuring FRR.  You seem to imply this change would cause confusion, I fail to see what confusion you're referring to since the command being changed is to output known remote vteps which were blank with p2mp tunnels.

PR #1900, opened 2025-01-22
Update SONiC User Manual with migration details

jeff-yin:
Addresses https://github.com/sonic-net/sonic-buildimage/issues/21253

- Added notes about minigraph.xml being deprecated
- Added a new section about migrating SONiC versions
- Fixed some broken links


FireStormOOO:
This looks good to me and would've answered my questions that prompted the linked issue.

PR #1927, opened 2025-02-17
add active-standby dualtor HLD

zjswhhh:
As it has been requested by many, submitting this PR to publish the active-standby dualtor feature design. 

We have been doing small incremental changes to the design, which were not properly documented in existing internal doc. Hence there are still some `TODO` in the markdown. Suggest we check in the current version as a start point.

Tried my best to
1. eliminate options we didn't choose at the end
2. update terminologies that are out-of-date
3. update changes we made in the design

But please help review in case I missed anything


zjswhhh:
Hi @manamand2020 - this is the doc you were requesting.

prsunny (approved):
lgtm

PR #1943, opened 2025-03-07
CoPP Neighbor Miss Trap and Other Enhancements HLD

rminnikanti:
HLD for CoPP Neighbor Miss Trap type support and Other CoPP Enhancements

What we did:
- Added support for new Neighbor Miss trap type and its default CoPP configuration. 
- Introduced SAI enum capability query in orchagent to identify SAI neighbor miss trap support and publish the supported trap types to STATE_DB for application visibility
-  Introduced CoPP CLI support to display trap CoPP configuration and their status.


prsunny:
@prabhataravind , please review as this should cover subnet trap.

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
```diff
@@ -0,0 +1,289 @@
...
+
+## 7. High-Level Design
+
+### 7.1 SWSS Design
+
+To maintain backward compatibility and considering that support for newly introduced trap types in SAI is vendor-dependent, an enum capability query for all trap types has been implemented in orchagent, with intelligent exceptions.
+
+During initialization, Copporch in Orchagent will perform an SAI enum capability query to identify supported trap types and maintain them in a local list called `supported_trap_ids`.
+
+Additionally, Copporch will maintain a list called `default_supported_trap_ids`, which contains currently supported trap IDs in copporch in orchagent. If the SAI does not support the enum capability query API, Copporch will assume that the `default_supported_trap_ids` are supported for backward compatibility.
```
selva-nexthop, line 81: the existing  _default_supported_trap_ids_ is vendor specific?

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
```diff
@@ -0,0 +1,289 @@
...
+          "always_enabled": "true"
+    }
+}
+
+```
+
+### 7.4 Schema Changes
+
+State DB will have a new table called COPP_TRAP_CAPABILITY_TABLE which will have the supported trap types. 
+Introducing new value pair "oper_status" field in existing COPP_TRAP_TABLE, to track the operational status of the trap type.
```
selva-nexthop, line 129: What happens if there is any change happens in SAI capabilities (e..g firmware upgrade)?  Will it trigger any daemon restart?  how this list will be updated for such changes?

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
```diff
@@ -0,0 +1,289 @@
...
+A new trap group queue1_group3 will be added to the default configuration with the following settings:
+
+```json
+  "queue1_group3": {
+          "trap_action":"trap",
+          "trap_priority":"1",
+          "queue": "1",
+          "meter_type":"packets",
+          "mode":"sr_tcm",
+          "cir":"200",
```
selva-nexthop, line 110: How the cir, cbs, and red_action values are derived?  These values might be strict for high-traffic environments causing unnecessary packet drops.

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
```diff
@@ -0,0 +1,289 @@
...
+
+* New CLI UT cases will be added to verify the CoPP CLI show command. Test cases will verify the output of the show command with input from mock_copp_cfg.json and mock CoPP tables in CONFIG_DB and STATE_DB test data.
+
+## 13. System Test Plan
+
+### 13.1. Test Cases
+
+An existing test case in the sonic-mgmt PTF framework tests VLAN subnet route traffic policing under the default CoPP configuration.
+
+A new test case will be introduced to verify the neighbor miss trap configuration with its default settings in copp_cfg.json using VLAN subnet route traffic. This test case will also update/empty/NULL configuration for neighbor miss in CONFIG_DB to override the default settings and verify the operational status of neighbor miss in the CLI output accordingly.
```
selva-nexthop, line 289: can we add real-world ARP traffic stress tests ?

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
```diff
@@ -0,0 +1,289 @@
...
+![CoPP Trap WorkFlow](images/CoPP_Trap_WorkFlow.png "Figure 1: CoPP Trap WorkFlow")
+
+### 7.3 CoPP Default Configuration Changes
+
+A default configuration for neighbor miss trap type will be added to copp_cfg.j2. Neighbor miss is default trap type with always_enabled set to true.
+
+A new trap group queue1_group3 will be added to the default configuration with the following settings:
+
+```json
+  "queue1_group3": {
```
selva-nexthop, line 104: Rationale behind using queue1_group3? 
If queue1_group3 is congested, Neighbor Miss traffic may be dropped while other less critical packets get through.

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
```diff
@@ -0,0 +1,289 @@
...
+## 9. Yang Model
+
+No changes. Yang model for CoPP is defined in [sonic-copp.yang](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-copp.yang).
+
+## 10. CLI Design
+
+### 10.1 Show Commands
+
+A new CLI command, `show copp configuration`, is being introduced. The purpose of this command is to display the CoPP configuration applied to SAI. The command will retrieve configuration from copp_cfg.json and CONFIG_DB. Additionally, the command will display the operational status of the trap types retrieved from STATE_DB.
+
```
selva-nexthop, line 195: Can we measure the performance of this CLI in scale/stress environment?  It is retrieving from multiple sources and it might take several seconds in large-scale switches.

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
```diff
@@ -0,0 +1,289 @@
...
+CBS......................... 200
+CIR......................... 200
+Yellow Action............... forward
+Green Action................ forward
+Red Action.................. drop
+```
+
+## 11. Warmboot and Fastboot Design Impact
+
+No warmboot and fastboot impact is expected for this feature.
```
selva-nexthop, line 267: how CoPP changes apply during warmboot?

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
rminnikanti: When syncd docker restarts, it triggers a restart of swss. During the initialization of CoppOrch, an enum capability query is performed. This occurs in both warmboot and coldboot scenarios. This process is similar to the flow in other Orch's, such as aclorch.cpp
Can you elaborate the use case?

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
rminnikanti: Neighbor miss traffic does not police ARP traffic. It is to police IP traffic that is trapped to the CPU when the L2 lookup has no entry for the neighbor. ARP traffic has a separate copp configuration in copp_cfg.json.

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
```diff
@@ -0,0 +1,298 @@
...
+ ; field               = value
+ trap_ids              = STRING
+ ```
+
+oper_status field in COPP_TRAP_TABLE
+
+```
+ key                   = "COPP_TRAP_TABLE|<trap_name>"
+ ; field               = value
+ state                 = STRING
```
prsunny, line 155: Suggest use the existing schema model for STATE_DB state. For e.g `"state"-"enabled/disabled"`

dgsudharsan:
Can you please test trap flow counters with this feature?

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
rminnikanti: "state" is an existing field of COPP_TRAP_TABLE table which is written by coppmgrd. Its value is "ok". No changes are being made to this field.

This HLD adds only "oper_status" : "installed" field to this table.

rminnikanti: If your suggestion pertains to the "oper_status" field, please confirm so I can make the necessary adjustments. 

HLD proposes "oper_status" field be set to "installed" if the trap is configured successfully. If the trap is not configured, "oper_status" field will be absent.

prsunny: So whats the difference with state and oper_status? if its not installed, we still keep the state as "ok"? May be we should treat this as "nok"?

rminnikanti: "state" is added by coppmgrd.
"oper_status" is added by orchagent.
After the configuration is processed and pushed to APPL_DB, coppmgrd sets the "state" to "ok", indicating feature resolution is done and configuration is processed.
Once the trap is configured to SAI, orchagent sets "oper_status" to "installed".

prsunny: ok. May I suggest the following:
1. If using `oper_status`, then the value should be `up/down` to be consistent with the other oper_status 
2. Alternatively, you can used `hw_status `and have `installed/not-installed `as values.

rminnikanti: Fixed. Changed to option 2 as suggested.

prsunny (approved):

rminnikanti:
> Can you please test trap flow counters with this feature?

trap counters tested.

prsunny:
@rminnikanti , can you please follow the description template and update all the associated PRs?

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
```diff
@@ -0,0 +1,299 @@
...
+
+## 7. High-Level Design
+
+### 7.1 SWSS Design
+
+To maintain backward compatibility and considering that support for newly introduced trap types in SAI is vendor-dependent, an enum capability query for all trap types has been implemented in orchagent, with intelligent exceptions.
+
+During initialization, Copporch in Orchagent will perform an SAI enum capability query to identify supported trap types and maintain them in a local list called `supported_trap_ids`.
+
+Additionally, Copporch will maintain a list called `default_supported_trap_ids`, which contains currently supported trap IDs in copporch in orchagent. If the SAI does not support the enum capability query API, Copporch will assume that the `default_supported_trap_ids` are supported for backward compatibility.
```
dgsudharsan, line 82: I don't support adding default_supported_trap_ids. If the query API is not present, we should treat it as any configuration and push to the SAI. This will ensure the backward compatibility as well

dgsudharsan (changes requested):

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
rminnikanti: Hi @dgsudharsan,

The default_supported_trap_ids list ensures backward compatibility. It includes all currently known traps used by orchagent, except for "neighbor_miss". This ensures that all existing trap configurations are pushed to the SAI, even if the query API isn't available. It also prevents orchagent from throwing exceptions in cases where the vendor SAI does not support "neighbor_miss" and using a different method to handle neighbor_miss traffic.

In summary, all current traps are pushed to SAI even if query API is not available only neighbor_miss and future traps need query API availability.

rminnikanti:
> @rminnikanti , can you please follow the description template and update all the associated PRs?

Updated.

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
rminnikanti: @dgsudharsan, The `default_supported_trap_ids` list and its values ensure backward compatibility while not mandating query API or neighbor_miss trap support from the vendor SAI. I see that, other Orch's in orchagent have similar default values if the query API is not supported.

Please suggest.

doc/copp/Copp_Neighbor_Miss_Trap_And_Enhancements.md
rminnikanti: Neighbor miss takes the same flow as other trap types. After warm reboot all the trap configurations are removed and pushed to SAI again.

prsunny:
@dgsudharsan , would you approve?

selva-nexthop (approved):

prsunny:
@prabhataravind , Merging the HLD. Please review #3624

Issue #357, opened 2019-04-09
CRM didn't delete ACL_TABLE which has been removed


JaredLius:
1. load the acl config below,bind acl_table 'aaa' to Ethernet0
{
    "ACL_TABLE": {
        "aaa": {
            "type": "L3",
            "ports": "Ethernet0" 
        }
    },
    "ACL_RULE": {
        "aaa|rule_0": {
            "PRIORITY": "55",
            "PACKET_ACTION": "DROP",
            "L4_SRC_PORT": "0" 
        },
        "aaa|rule_1": {
            "PRIORITY": "55",
            "PACKET_ACTION": "DROP",
            "L4_SRC_PORT": "1" 
        }
    }
}

2. check crm acl table resources
admin@SONIC:~/acl_scal$ crm show resources acl table

Table ID         Resource Name      Used Count    Available Count
---------------  ---------------  ------------  -----------------
0x70000000003b4  acl_entry                   2               3070 <----bind to Ethernet0's Table ID is 0x70000000003b4
0x70000000003b4  acl_counter                 2               3070

Table ID         Resource Name      Used Count    Available Count
---------------  ---------------  ------------  -----------------
0x7000000000363  acl_entry                   2               6142

3. change the acl config: set bind port to Ethernet1 and load the config again.
4. check crm resources
admin@SONIC:~/acl_scal$ crm show resources acl table

Table ID         Resource Name      Used Count    Available Count
---------------  ---------------  ------------  -----------------
0x70000000003b4  acl_entry                   0               3070 <---this table oid has been deleted by aclorch, I think it need be deleted from crm too.
0x70000000003b4  acl_counter                 0               3070

Table ID         Resource Name      Used Count    Available Count
---------------  ---------------  ------------  -----------------
0x70000000003bb  acl_entry                   2                  0
0x70000000003bb  acl_counter                 2                  0

Table ID         Resource Name      Used Count    Available Count
---------------  ---------------  ------------  -----------------
0x7000000000363  acl_entry                   2               6142

5. 
#asic_db has deleted the acl_table oid 0x70000000003b4 but crm counter db didn't delete.

admin@SONIC:~/acl_scal$ redis-cli -n 2 keys *ACL*
 1) "CRM:ACL_TABLE_STATS:0x70000000003b4"    --->still exist.
 2) "CRM:ACL_STATS:EGRESS:VLAN" 
 3) "CRM:ACL_STATS:INGRESS:PORT" 
[...]

PR #1951, opened 2025-03-27
Configuration Replacement via gNOI

karuppiah-dharmaraj:
This feature allows user to replace current running configuration with the known configuration file in the serialization of DB format. It also supports copying files to and from remote/local locations via gNOI


kwangsuk (approved):

doc/mgmt/gnmi/Config_Replace_gNOI.md
```diff
@@ -0,0 +1,226 @@
...
+* SCP
+* FTP
+* HTTP
+
+Otherwise, It will get the file from below local space
+* config (/etc/sonic)
+* home (home directory of respective user)
+* USB
+
+If the input file is in compressed format(.tar), It will decompress the file to get the JSON format file. JSON file will be kept under temporary directory.
```
qiluo-msft, line 97: >If the input file is in compressed format(.tar) [](http://example.com/codeflow?start=0&length=47)

I believe we do not need to invent a new gRPC protocol for this design purpose. The official gNMI protocol has "replace" operation defined, and when request "replace" with a path of "root" will be exactly what you need.

ref: [https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#344-modes-of-update-union-replace-replace-and-update](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#344-modes-of-update-union-replace-replace-and-update)

doc/mgmt/gnmi/Config_Replace_gNOI.md
```diff
@@ -0,0 +1,226 @@
...
+## Revision
+
+| Rev | Date       | Author | Change Description |
+| --- | ---------- | ------ | ------------------ |
+| 0.1 | 03/21/2025 | Karuppiah Dharmaraj | Initial version    |
+
+
+## Scope
+
+This document outlines the high-level design of supporting Configuration Replacement via gNOI
```
qiluo-msft, line 35: >gNOI [](http://example.com/codeflow?start=89&length=4)

gNOI is for operations, like "reboot". gNMI is for configuration management. Is it better to use gNMI protocol and gNMI client?

anders-nexthop:
A related proposal was made in sonic-gnmi [Issue #357](https://github.com/sonic-net/sonic-gnmi/issues/357). An implementation was proposed as well (PR is [here](https://github.com/sonic-net/sonic-gnmi/pull/369) and test fix is [here](https://github.com/sonic-net/sonic-mgmt/pull/17886)).

This approach seems to be very in line with what is being outlined here.

doc/mgmt/gnmi/Config_Replace_gNOI.md
```diff
@@ -0,0 +1,226 @@
...
+
+### 1. Overview 
+
+The Configuration Replacement feature provides the capability to replace the current running configuration, i.e., CONFIG_DB, with a known good configuration file saved in the serialization of DB format, i.e. config_db.json. 
+This feature is designed to limit the impact to features/services that are being replaced, with minimum service disruption. 
+
+### 2. Requirements
+#### Phase 1: 
+- Add YANG constraints and custom validation callbacks in SONIC yang specific to the features
+- Config replacement can be requested via gNOI management interfaces.
```
hdwhdw, line 65: Can you clarify what is gNOI management interface? configuration management is the scope of gNMI and not gNOI. gNOI is for operation such as rebooting or removing a file. See https://github.com/openconfig/gnoi/tree/main

doc/mgmt/gnmi/Config_Replace_gNOI.md
```diff
@@ -0,0 +1,226 @@
...
+* FTP
+* HTTP
+
+Otherwise, It will get the file from below local space
+* config (/etc/sonic)
+* home (home directory of respective user)
+* USB
+
+If the input file is in compressed format(.tar), It will decompress the file to get the JSON format file. JSON file will be kept under temporary directory.
+Config Replacement component loads a desired config file from the temporary directory and compares against CONFIG_DB to generate the diff results tagged with operation, table, keys and fields. The resultant diff data set will be sorted by operation - DELETE, CREATE and UPDATE in this order, and subsequently sorted by table orders learnt from CVL to apply the changes as a single transaction to CVL/DB layer. Just before updating the CONFIG_DB, CVL libyang makes sure that when and must constraints are validated. Additional ad-hoc validation will be done if CVL custom validation callbacks are mentioned in sonic-yang file
```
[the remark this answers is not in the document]
kwangsuk: The proposed config replace will leverage the CVL layer within sonic-mgmt-common to ensure proper ordering, validation, and database updates with enhanced throughput. It will not be routed through GCU.

doc/mgmt/gnmi/Config_Replace_gNOI.md
kwangsuk: The file copy operation will be defined as gNOI, enabling the config replace to be executed if a destination is set to the running-config.

doc/mgmt/gnmi/Config_Replace_gNOI.md
kwangsuk: "The file copy operation will be defined as gNOI, enabling the config replace to be executed if a destination is set to the running-config." quoted from above.

doc/mgmt/gnmi/Config_Replace_gNOI.md
```diff
@@ -0,0 +1,226 @@
...
+|JSON                      | JavaScript Object Notation                                         |
+|running-<br>configuration | The configuration controlling the device. a.k.a running-datastore,<br>CONFIG_DB	|
+|SCP			   | Secure Copy Protocol								|
+|startup-<br>configuration | The configuration that is loaded when the device boots up,<br>typically stored in a persistent datastore |
+|USB			   | Universal Serial Bus								|
+
+
+### 1. Overview 
+
+The Configuration Replacement feature provides the capability to replace the current running configuration, i.e., CONFIG_DB, with a known good configuration file saved in the serialization of DB format, i.e. config_db.json.
```
xincunli-sonic, line 59: >replace [](http://example.com/codeflow?start=65&length=7)

Please make the clarification on "Replace" Semantics (Section 1 & 4): The term "replace" could be interpreted as wiping the entire CONFIG_DB and applying the new file. However, Section 4 describes generating a diff (DELETE, CREATE, UPDATE) based on comparison with the current CONFIG_DB. This implies a more targeted update rather than a full wipe-and-load. Please clarify the exact mechanism. Does it only touch the tables/keys present in the input config_db.json file, calculating the necessary adds/deletes/modifies compared to the current state of those specific elements in CONFIG_DB? Or does it truly replace everything except the explicitly preserved defaults? Explicitly defining this behavior is crucial.

doc/mgmt/gnmi/Config_Replace_gNOI.md
```diff
@@ -0,0 +1,226 @@
...
+- A syslog message will be generated in case of an error, including the table and attribute information causing the error.
+- Copying files to and from local directories and remote locations
+
+  | **Source**                 | **Destination**                        				|
+  |--------------------------|--------------------------------------------------------------------|
+  |config:<br>coredump:<br>ftp:<br>home:<br>http:<br>log:<br>running-configuration<br>scp:<br>startup-configuration<br>tech-support:<br>usb:<br>	   |config:<br>ftp:<br>home:<br>http:<br>running-configuration<br>scp:<br>startup-configuration<br>usb:<br>	|
+- Sorting, validation support for shallow and deep difference between new and current configuration.
+- Config replace operation should be performed with minimum disruption to the device, i.e. no container restart or system reboots expected.
+- Default configuration, e.g. copp_config.json, not to be saved to a serialized DB file should be preserved in the CONFIG_DB across config replacement operation.
```
xincunli-sonic, line 75: >Default configuration [](http://example.com/codeflow?start=2&length=21)

This said default configurations (like copp_config.json) should be preserved. However, the design doesn't explicitly state how this preservation is achieved during the replace operation. Is there a predefined list of tables/keys that are ignored during the diff/apply process? Does the diff algorithm specifically exclude certain structures? Detailing this mechanism in Section 4 would strengthen the design.

doc/mgmt/gnmi/Config_Replace_gNOI.md
```diff
@@ -0,0 +1,226 @@
...
+    }
+```
+#### 6.3. Config DB Enhancements  
+No Config DB enhancement is done for this feature	
+
+### 7. Warmboot and Fastboot Design Impact  
+This feature doesnt impact warmboot/fastboot. 
+
+### 8. Memory Consumption
+No memory consumption is expected when the feature is used
```
xincunli-sonic, line 197: Loading a potentially large configuration file, holding it in memory, comparing it with the existing CONFIG_DB, generating and storing the diff set, and interacting with CVL will all consume memory, even if transiently. This section should provide a realistic estimate of the potential peak memory usage (e.g., dependent on the size of the input config or the diff) and confirm it's within acceptable limits for the target platform(s).

doc/mgmt/gnmi/Config_Replace_gNOI.md
```diff
@@ -0,0 +1,226 @@
...
+
+### 9. Restrictions/Limitations  
+Config Replacement won't be supported in click-cli
+
+### 10. Testing Requirements/Design  
+None
```
xincunli-sonic, line 203: System-level testing is critical for a feature like configuration replacement that affects the live state of the device. System tests should be defined to cover scenarios such as:

Replacing configuration impacting critical services (e.g., BGP, LACP, ACLs) and verifying minimal service disruption.
Performing replacement under system load.
Testing rollback on validation failure with a complex configuration.
Testing interaction with concurrent configuration changes via other interfaces (though ideally, locking might prevent this).
Replacing large/complex configurations.
Testing replacement using compressed .tar files.

doc/mgmt/gnmi/Config_Replace_gNOI.md
```diff
@@ -0,0 +1,226 @@
...
+       |  +---w source?            filename-uri-type
+       |  +---w destination?       filename-uri-type
+       |  +---w (operation)?       enumeration
+       +--ro output
+          +--ro status?          int32
+          +--ro status-detail?   string
+```
+```python
+YANG file: openconfig-file-mgmt-private.yang
+    typedef filename-uri-type {
```
xincunli-sonic, line 132: >filename-uri-type [](http://example.com/codeflow?start=12&length=17)

The typedef for filename-uri-type uses pattern statements for running-configuration and startup-configuration. These are specific keywords, not really patterns like ftp:.*. This might unintentionally allow strings like running-configuration-abc. Consider defining these specific keywords separately or using a different type structure (perhaps a union of an enumeration for keywords and a string with the URI pattern) for better clarity and stricter validation.

doc/mgmt/gnmi/Config_Replace_gNOI.md
karuppiah-dharmaraj: The end result is like input JSON replacing the running configuration(CONFIG_DB). This feature is trying to achieve Config Replace with minimum service disruption. Comparison between the input JSON and CONFIG_DB occurs in an object-by-object or leaf-by-leaf manner. Table/leaf is considered to be deleted, if it is not present in input JSON file but in CONFIG_DB. Table/leaf is considered to be added, if it is present in input JSON file but not in CONFIG_DB. leaf is considered to be modified, if the value is different. If there is no change in the table, respective service/application won't be notified to avoid disruption

doc/mgmt/gnmi/Config_Replace_gNOI.md
karuppiah-dharmaraj: Done

doc/mgmt/gnmi/Config_Replace_gNOI.md
kwangsuk: Yes, I agree that the system test is crucial. As outlined in the HLD, we will have phased deliverables for this feature. Phase 1 will focus on establishing the infrastructure using sFlow as an exemplar case. Following this, Phase 2 will be planned to increase the coverage of features by importing SONiC YANGs from sonic-mgmt-yang, which will require semantic validation checks by CVL custom hooks and YANG model constraints. We can conduct the system test once we reach the necessary level of maturity.

kwangsuk:
> A related proposal was made in sonic-gnmi [Issue #357](https://github.com/sonic-net/sonic-gnmi/issues/357). An implementation was proposed as well (PR is [here](https://github.com/sonic-net/sonic-gnmi/pull/369) and test fix is [here](https://github.com/sonic-net/sonic-mgmt/pull/17886)).
> 
> This approach seems to be very in line with what is being outlined here.

The proposed config replace will define OpenConfig RPC to enable both gNOI and REST PUT operation and leverage the CVL layer within sonic-mgmt-common. It will not be routed through GCU as used in (https://github.com/sonic-net/sonic-mgmt/pull/17886)

adyeung:
@anand-kumar-subramanian pls help review the change

PR #1958, opened 2025-04-08
Local ARS HLD

VladimirKuk:
This document provides high level design for the feature local Adaptive Routing and Switching (ARS).

PRs:
| Repository | PR Title | State | Context |
| ---------- | -------- | ----- | ------- |
| sonic-buildimage | [`Local ARS (Adaptive Routing and Switching)`](https://github.com/sonic-net/sonic-buildimage/pull/22264) | [PR Status](https://img.shields.io/github/pulls/detail/state/sonic-net/sonic-buildimage/22264) | [PR Checks](https://img.shields.io/github/status/contexts/pulls/sonic-net/sonic-buildimage/22264) |
| sonic-swss | [`Local ARS (Adaptive Routing and Switching)`](https://github.com/sonic-net/sonic-swss/pull/3597) | [PR Status](https://img.shields.io/github/pulls/detail/state/sonic-net/sonic-swss/3597) | [PR Checks](https://img.shields.io/github/status/contexts/pulls/sonic-net/sonic-swss/3597) |
| sonic-swss-common | [`Local ARS (Adaptive Routing and Switching)`](https://github.com/sonic-net/sonic-swss-common/pull/1002) | [PR Status](https://img.shields.io/github/pulls/detail/state/sonic-net/sonic-swss-common/1002) | [PR Checks](https://img.shields.io/github/status/contexts/pulls/sonic-net/sonic-swss-common/1002) |
| sonic-mgmt | [`Local ARS (Adaptive Routing and Switching test plan)`](https://github.com/sonic-net/sonic-mgmt/pull/21237) | [PR Status](https://img.shields.io/github/pulls/detail/state/sonic-net/sonic-mgmt/21237) | [PR Checks](https://img.shields.io/github/status/contexts/pulls/sonic-net/sonic-mgmt/21237) |
| sonic-mgmt | [`Local ARS (Adaptive Routing and Switching test plan)`](https://github.com/sonic-net/sonic-mgmt/pull/24845) | [PR Status](https://img.shields.io/github/pulls/detail/state/sonic-net/sonic-mgmt/24845) | [PR Checks](https://img.shields.io/github/status/contexts/pulls/sonic-net/sonic-mgmt/24845) |


Yubin-Li:
Hello, why ARS NHG key is prefix? how to handle if there have so many route needs point to ARS ECMP
I think we need to set per port ARS enable bit in Orch
So that if ECMP created by route orch and it's alias belong to enable list, 
We bind ars to this ECMP

Yubin-Li:
Why this schema used prefix for Key? dose it means that only some prefix can used with AR ECMP?

VladimirKuk:
> Why this schema used prefix for Key? dose it means that only some prefix can used with AR ECMP?

Yes, that is to support use-case for L3 traffic sent to NHG.
Also, currently, there is no other way to reference/identify NHG via CONFIG_DB.

Yubin-Li:
> > Why this schema used prefix for Key? dose it means that only some prefix can used with AR ECMP?
> 
> Yes, that is to support use-case for L3 traffic sent to NHG. Also, currently, there is no other way to reference/identify NHG via CONFIG_DB.

If so, how to handle if there have 100 prefixes but point to same ECMP, we want to used AR for this ECMP. It's very hard to set dynamic route point to AR ecmp

I think may be we can define some AR ECMP include port list and trip threshold
for example, I define [port1,2,3,4,5] with AR ecmp, and trip threshold = 4
once we create ECMP with [1,2,3,4] [2,3,4,5], [1,2,3,4,5], we can create with AR

VladimirKuk:
> If so, how to handle if there have 100 prefixes but point to same ECMP, we want to used AR for this ECMP. It's very hard to set dynamic route point to AR ecmp
> 
> I think may be we can define some AR ECMP include port list and trip threshold for example, I define [port1,2,3,4,5] with AR ecmp, and trip threshold = 4 once we create ECMP with [1,2,3,4] [2,3,4,5], [1,2,3,4,5], we can create with AR


There is already ARS-enabled interface table, but it could include all interfaces, so I guess shouldn't be used for this. However, to create another table with ARS-enabled interfaces, feels very wasteful.

The problematic scenario is this : suppose you have two NHGs : NHG1 (NH1,NH2,NH3) and NHG2 (NH1,NH2,NH4).
NH3 and NH4 are not reliable and going up and down.
If you rely on ECMP updates to determine whether this is ARS NHG, you could mistake NHG1 with NHG2, since their content will be the same  (NH1,NH2).
Prefix (while it is not ideal) is also, specifies the path you want to enable ARS for.

Also, the use-case was to enable static configuration, To enable dynamic NHG identification, the ARS/NHG ids should, probably, be derived from/imported by some protocol (i.e. BGP).

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+- Tunnel interface as egress interface is not supported
+
+### Requirements
+
+* Phase 1
+1. Support different ARS modes:
+    - Flowlet-based port selection
+    - Per packet port selection
+2. Support enabling ARS over NHG
+3. Support path quality configuration
```
eddyk-nvidia, line 87: Please make sure that the design supports 2 mode of operation regarding creation of NHG: done by Route OA and done by FRR when it decides on creation and push the NHG to APP DB

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+
+1. Nexthop group creation
+    * RouteOrch receives route from APPL_DB, triggering the process.
+    * Nexthop group is created
+2. Notification
+    * RouteOrch component will notify ArsOrch of NHG update.
+3. Nexthop group's SAI attributes handling:
+    * If nexthop group's ARS object ID, "set" capability is supported:
+        - The nexthop group is updated with the ARS object ID.
+    * If "set" capability is not supported - implement via "create" functionality:
```
eddyk-nvidia, line 197: what is exactly the trigger for creation of ARS ECMP for both cases - NHG is created either by Route OA and by FRR/Zebra ? Match of the prefix for a created NHG with configured in ARS_NEXTHOP_GROUP table ? On fabric init along the convergence when links go up the Route OA creates multiple NHG (deleting old one and creating a new one) incrementally adding new NHs. So, each NHG will have its own SAI OID. Each one of NHG object will need to be recreated with ARS OID ?

What about check if a port with RIF on that used as NH is ARS-enabled ? You can create now the NHG as ARS-capable with NHs (using RIFs over non-ARS enabled ports). You can think about logic for decision of enabling ARS on NHG considering whether all ports (RIF/NHs) involved in NHG are ARS-enabled or not. In this case non-ARS enabled port (RIF/NH) can't be added to ARS NHG

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+    * Nexthop group is updated with new member
+2. Notification
+    * RouteOrch component will notify ArsOrch of NHG update.
+4. Nexthop group's SAI attributes handling:
+    * If an alternative path is defined
+        - Set the relevant nexthops as part of the alternative path.
+
+__Figure 9:ARS LAG table creation flow__
+![](images/ars_lag_create.png)
+
```
eddyk-nvidia, line 218: Please make sure that ARS on LAG is wrapped with the SAI capability check. Some vendors can support ARS for NHG but not for LAG

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+                description "ARS Profile Name";
+                type string;
+            }
+
+            leaf algorithm {
+                description "ARS quality algorithm";
+                type enumeration {
+                    enum ewma {
+                        description "Exponentially Weighted Moving Average algorithm";
+                    }
```
eddyk-nvidia, line 304: Can potentially expanded with algos supported by other vendors

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+```
+; New container ARS_PROFILE_TABLE
+; ARS global configuration
+
+key                     = ARS_PROFILE|profile_name
+
+;field                  = value
+
+algorithm               = "ewma"        ;Path quality calculation algorithm
+max_flows               = uint32        ;Maximum number of flows that can be maintained for ARS
```
eddyk-nvidia, line 595: Not the only algo used by vendors. Probably will require other vendors to introduce changes in SAI API

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+
+scaling_factor          = uint32                          ;Port speed normalization
+
+Configuration example:
+
+"ARS_INTERFACE": {
+    "Ethernet0" : {
+        "scaling_factor": "100"
+    },
+    "Ethernet8" : {}
```
eddyk-nvidia, line 644: without parameter will be used with default value for "scaling factor" or without it (if not used by other vendor's algo) ?

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+Based on the port's speed, PortsOrch determines the appropriate scaling factor.
+The scaling factor is then set in the SAI layer through the syncd component to adjust the port's behavior accordingly.
+
+__Figure 6: ARS Nexthop group table creation flow__
+![](images/ars_nhg_create.png)
+
+1. Add ARS_NEXTHOP_GROUP Table
+    * The CONFIG_DB adds the ARS_NEXTHOP_GROUP table, triggering the process.
+2. Create SAI ARS Object
+    * The ArsOrch component receives the update and creates the SAI ARS object.
```
eddyk-nvidia, line 172: does design propose the creation of ARS Object per entry in ARS_NEXTHOP_GROUP table ? How to re-use the ARS Object ? SAI APIs doesn't limit the re-use

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+
+"ARS_NEXTHOP_GROUP": {
+    "default|192.168.0.100/32" : {
+        "profile_name": "ars_profile",
+        "assign_mode" : "per_flowlet_quality",
+        "flowlet_idle_time" : "256",
+        "max_flows" : "512",
+        "primary_path_threshold" : "100",
+        "alternative_path_cost": "250",
+        "alternative_path_members": {"1.1.1.1", "2.2.2.2"}
```
eddyk-nvidia, line 674: how do you in advance which NHs can be used when NHG is created from BGP by Route OA or FRR/Zebra ?

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+
+- Mandatory:
+
+1. ARS Profile Creation
+    * Verify that ARS profiles are created successfully with valid parameters.
+    * Test error handling for invalid profile configurations.
+
+2. Enabling ARS on Ports
+    * Ensure ARS is enabled on specified ports.
+    * Validate behavior when enabling ARS on unsupported interfaces.
```
eddyk-nvidia, line 859: via enabling ARS on NHG when it is defined on not-supported interfaces ? Missing logic in the design

rajendrat:
> > Why this schema used prefix for Key? dose it means that only some prefix can used with AR ECMP?
> 
> Yes, that is to support use-case for L3 traffic sent to NHG. Also, currently, there is no other way to reference/identify NHG via CONFIG_DB.

@VladimirKuk : Is there an option/way to enable ARS for all NH Groups  or enable only for matching the "vrf/ip-prefix"?

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+        }
+        /* end of list ARS_INTERFACE_LIST */
+    }
+    /* end of container ARS_INTERFACE */
+```
+
+##### ARS_NEXTHOP_GROUP
+
+```
+    container ARS_NEXTHOP_GROUP {
```
ashutosh-agrawal, line 403: ARS_PREFIX could be a better name since key is an IP Prefix and multiple prefixes can share the same nexthop group.

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+            }
+
+            leaf flowlet_idle_time {
+                type uint16 {
+                    range 2..2047;
+                }
+                description  "Idle duration in microseconds. This duration is to classifying a flow-let in a macro flow.";
+            }
+
+            leaf max_flows {
```
ashutosh-agrawal, line 453: max_flows is a field in the ars_profile table as well. Is this a duplicate or the intent here is to create a separate limit per ip-prefix?

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+            key "if_name";
+
+            leaf if_name {
+                type leafref {
+                    path "/port:sonic-port/port:PORT/port:PORT_LIST/port:name";
+                }
+                description "ARS-enabled interface name";
+            }
+
+            leaf scaling_factor {
```
ashutosh-agrawal, line 388: In figure 5 above, scaling_factor is set by the PortOrch when the port comes up. Is this configuration expected to override the scaling_factor set by PortOrch?

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,892 @@
...
+#### Sequence diagrams
+
+__Figure 4: Initilization flow__
+![](images/init_seq.png)
+
+1. Bind to Events
+The ArsOrch component binds to relevant events to start monitoring and processing ARS-related interfaces.
+
+2. Get ARS SAI Capabilities
+ArsOrch retrieves ARS capabilities from the SAI layer and saves them in STATE_DB.
```
ashutosh-agrawal, line 147: Please add a schema for STATE_DB to list what all capabilities are being saved in the DB.

VladimirKuk:
> > > Why this schema used prefix for Key? dose it means that only some prefix can used with AR ECMP?
> > 
> > 
> > Yes, that is to support use-case for L3 traffic sent to NHG. Also, currently, there is no other way to reference/identify NHG via CONFIG_DB.
> 
> @VladimirKuk : Is there an option/way to enable ARS for all NH Groups or enable only for matching the "vrf/ip-prefix"?

Change matching using nexthop set instead of prefix-based. 
If match-all is introduced, then all the NHGs will be using the same ARS OID, which is not the intended behavior.

doc/ARS/Local_ARS_HLD.md
VladimirKuk: This should known to user

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,1052 @@
...
+
+There are no new SAI APIs or attributes required.
+
+Following table lists SAI usage and supported attributes with division to phase implementation:
+
+* Phase 1
+
+| SAI api | Supported SAI attribute 
+| ------- | ----------------------- 
+|create_ars_profile | SAI_ARS_PROFILE_ATTR_PORT_LOAD_PAST<br>SAI_ARS_PROFILE_ATTR_LOAD_PAST_MIN_VAL<br>SAI_ARS_PROFILE_ATTR_LOAD_PAST_MAX_VAL<br>SAI_ARS_PROFILE_ATTR_ENABLE_IPV4<br>SAI_ARS_PROFILE_ATTR_ENABLE_IPV6<br>SAI_ARS_PROFILE_ATTR_MAX_FLOWS<br>SAI_ARS_PROFILE_ATTR_ALGO<br>SAI_ARS_PROFILE_ATTR_PORT_LOAD_FUTURE<br>SAI_ARS_PROFILE_ATTR_PORT_LOAD_PAST_WEIGHT<br>SAI_ARS_PROFILE_ATTR_PORT_LOAD_FUTURE_WEIGHT<br>SAI_ARS_PROFILE_ATTR_SAMPLING_INTERVAL<br>SAI_ARS_PROFILE_ATTR_ECMP_ARS_MAX_GROUPS<br>SAI_ARS_PROFILE_ATTR_ECMP_ARS_MAX_MEMBERS_PER_GROUP
```
venkatmahalingam, line 312: How's the ARS profile binding to a switch handled? refer the section 6.2 in ARS SAI spec.
https://github.com/opencomputeproject/SAI/blob/master/doc/ARS/Adaptive-Routing-and-Switching.md#62-bind-ars-profile-object-to-switch-object

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,1052 @@
...
+            leaf nhg_name {
+                type string;
+                description "ARS nexthop group name";
+            } 
+
+            leaf profile_name {
+                description "ARS profile Name";
+                mandatory true;
+                type leafref {
+                    path "/sars:sonic-ars/sars:ARS_PROFILE/sars:ARS_PROFILE_LIST/sars:profile_name";
```
venkatmahalingam, line 480: Why do we need this reference? is this mentioned in ARS SAI spec?

Yubin-Li:
> > If so, how to handle if there have 100 prefixes but point to same ECMP, we want to used AR for this ECMP. It's very hard to set dynamic route point to AR ecmp
> > I think may be we can define some AR ECMP include port list and trip threshold for example, I define [port1,2,3,4,5] with AR ecmp, and trip threshold = 4 once we create ECMP with [1,2,3,4] [2,3,4,5], [1,2,3,4,5], we can create with AR
> 
> There is already ARS-enabled interface table, but it could include all interfaces, so I guess shouldn't be used for this. However, to create another table with ARS-enabled interfaces, feels very wasteful.
> 
> The problematic scenario is this : suppose you have two NHGs : NHG1 (NH1,NH2,NH3) and NHG2 (NH1,NH2,NH4). NH3 and NH4 are not reliable and going up and down. If you rely on ECMP updates to determine whether this is ARS NHG, you could mistake NHG1 with NHG2, since their content will be the same (NH1,NH2). Prefix (while it is not ideal) is also, specifies the path you want to enable ARS for.
> 
> Also, the use-case was to enable static configuration, To enable dynamic NHG identification, the ARS/NHG ids should, probably, be derived from/imported by some protocol (i.e. BGP).

I think for our customer it's very hard to control by using prefixes, 
And I think bgp will not doing this.

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,1159 @@
...
+**Important Consideration**:
+
+Configuration Guidelines
+- Interfaces participating in ARS should be properly configured in the ARS_INTERFACE table.
+- For ARS functionality to operate consistently, all interfaces that belong to the same NHG or LAG must be associated with a same ARS object.
+
+List of scenario that will not trigger ARS re-selection
+ - Removal of interfaces from the ARS_INTERFACE table.
+ - Modification of the ARS_OBJECT associated with ARS interfaces.
+ - Once NHG is enabled through ARS, any changes to VLAN members—such as additions, deletions, or MAC address movements—will no longer trigger ARS re-selection.
```
dgsudharsan, line 235: Can you explain how ARS will behave in ECMP acceleration scenario? Here is the code of validate nexthop which will trigger on interface up
https://github.com/sonic-net/sonic-swss/blob/1f97afb3c6c81de755d7045aae275b3b8af610ec/orchagent/routeorch.cpp#L451

You can refer to https://github.com/sonic-net/sonic-swss/pull/351

apannerselva: hi Sudharasan,
The ECMP acceleration behavior is addressed in https://github.com/sonic-net/sonic-swss/pull/3597 by Vladimir.
A new SUBJECT_TYPE_NEXTHOP_GROUP_CHANGE is introduced to notify ARS for any change in nexthop.
please check.
Thanks

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,1155 @@
...
+                        description "Uses list of configured interfaces for ARS object selection.";
+                    }
+                    enum nexthop {
+                        description "Uses list of configured nexthops for ARS object selection.";
+                    }
+                } 
+                default "interface";
+            }
+
+            leaf ars_lag_path_selector_mode {
```
venkatmahalingam, line 449: Can we keep only supported fields in the YANG for phase-1?

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,1155 @@
...
+2. Set Port Scaling Factor
+    * Based on the port's speed, PortsOrch determines the appropriate scaling factor. <br>The scaling factor is then set in the SAI layer through the syncd component to adjust the port's behavior accordingly.
+
+##### __Figure 6: Nexthop group table created by RouteOrch flow__
+![](images/nhg_create_by_ro.png)
+
+1. Add ROUTE_TABLE Table
+    * The APPL_DB adds the ROUTE_TABLE table, triggering the process.
+2. Notification
+    * RouteOrch component will notify ArsOrch of NHG change.
```
venkatmahalingam: We talked about removing the notifications between orchagents rather call APIs for any handling? is this not taken care yet?

doc/ARS/Local_ARS_HLD.md
apannerselva: In HLD, we will keep this YANG field but in implementation will cover two phases.

doc/ARS/Local_ARS_HLD.md
apannerselva: We have already taken care the API invocation in the workflow instead of notification and we will update the writeup as well.

venkatmahalingam: OK

apannerselva:
@venkatmahalingam Can you please help in reviewing the HLD and provide your approval.
thanks.

baxia-lan:
Is this HLD ready to merge?

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,1151 @@
...
+1. NHG created by RouteOrch. APPL_DB adds entry to ROUTE_TABLE with explicit set of nexthops.
+
+2. NHG created by NhgOrch. APPL_DB first adds an entry to the NEXTHOP_GROUP_TABLE and later adds an entry to the ROUTE_TABLE with a reference to the previously created NHG. This mode of operation will be supported only when the Nexthop group's SAI ARS object id attribute "set" capability is supported, to prevent NHG recreations when multiple routes point to the updated NHG.
+
+**ARS selection for Nexthop group**
+
+The ars_nhg_path_selector_mode field in ARS_PROFILE defines the scope at which ARS selection for NH groups is applied.
+The system supports three distinct modes:
+
+**1. Global**:
```
baxia-lan, line 197: Is there such knob in SAI to enable ARS globally and avoid individual ARS interfaces/objects creation?

apannerselva: There is no SAI attribute available to explicitly enable ARS global configuration. However, the SONiC NOS will support a global ARS mode that implicitly manages the configuration of the SAI ARS_INTERFACE attribute. In this global mode, the user is expected to configure the ARS_OBJECT, which will govern ARS behavior system-wide.

apannerselva:
> Is this HLD ready to merge?

@venkatmahalingam please help in reviewing and approving the HLD.
Thanks

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,1151 @@
...
+    * ARS checks the state of the LAG members interface's ARS capability and proceeds only if ARS is enabled.
+    * If LAG was created:
+        - It is updated with the ARS object ID.
+
+##### __Figure 9: LAG creation flow__
+![](images/lag_create.png)
+
+1. LAG creation
+    * PortsOrch receives LAG from APPL_DB, triggering the process.
+2. Notification
```
venkatmahalingam: Do we still have event notification between orchagent modules? as discussed in the meeting earlier, we can simply make function calls among orchagent modules. please fix it.

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,1151 @@
...
+**Table interactions**:
+
+1. ARS_PROFILE defines the ARS global configuration parameters. ArsOrch uses this to create the ARS profile in SAI. 
+   Currently, only a single profile is supported. If multiple profiles are required, the SAI implementation must also be enhanced to accommodate this functionality.
+    
+2. ARS_OBJECT table used to specify ARS object configuration parameters. ArsOrch uses this to create the ARS SAI object.
+
+3. ARS_INTERFACES holds list of L2 egress interfaces which are ARS-enabled. NHG will be considered ARS-enabled when all of its associated nexthop interfaces are present in the configured ARS_INTERFACES when the match mode configured in ARS_PROFILE is interface-based.
+
+4. ARS_NEXTHOPS holds list of nexthops which are used for identifying ARS-enabled NHG. This table is applicable only when ars_nhg_path_selector_mode in ARS_PROFILE is nexthop.
```
sudheer-nexthop, line 181: From the example section:
"ARS_NEXTHOPS": {
    "default|1.1.1.10" : {
        "ars_obj_name": "ars_obj_name"
    },
    "default|2.2.2.20" : {
        "ars_obj_name": "ars_obj_name",
        "role": "alternative_path"
    },
    "default|3.3.3.30" : {
        "ars_obj_name": "ars_obj_name"
    }
}

I understand, this, as well as the ARS_INTERFACES are designed to apply ARS per nexthop/interface. Does SAI Support this? What does it mean to say that I have ARS operating in 'flowlet_mode' on interface A, and 'packet spray mode' on interface B?
Also, the implementation has ARS_NEXTHOP_GROUP table being used to match the route prefix. Is that still the primary mechanism to identify the NexthopGroup for ARS?

I am looking at a way where an ARS Object can be applied to a NexthopGroup that contains a given set of interfaces. Does your HLD address that? I can see this comment mentioning support for the same, but HLD doesn't reflect it: https://github.com/sonic-net/SONiC/pull/1958#issuecomment-2873742125

apannerselva: I understand, this, as well as the ARS_INTERFACES are designed to apply ARS per nexthop/interface. Does SAI Support this? What does it mean to say that I have ARS operating in 'flowlet_mode' on interface A, and 'packet spray mode' on interface B?
```
This is the SONiC config model which support ARS object selection based on Global/Interface/Nexthops mode.
There is NO equivalent SAI attribute for this.
```

Also, the implementation has ARS_NEXTHOP_GROUP table being used to match the route prefix. Is that still the primary mechanism to identify the NexthopGroup for ARS?
`Latest HLD doesn't have ARS_NEXTHOP_GROUP table with route prefix. As mentioned above ARS object selection is based on ars_nhg_path_selector_mode.
`

I am looking at a way where an ARS Object can be applied to a NexthopGroup that contains a given set of interfaces. Does your HLD address that? I can see this comment mentioning support for the same, but HLD doesn't reflect it
```
When ars_nhg_path_selector_mode is set to interface, the nexthop group containing the
specified set of interfaces will be configured as an ARS-enabled NHG in SAI, using the
ARS object associated with that interfaces.
```

sudheer-nexthop: Thanks for the clarification, @apannerselva . I have a follow-up on that, mainly to understand the expected behavior. Lets say NHG-1 contains interfaces (Eth1, Eth2 and Eth3) and NHG-2 contains (Eth2, Eth3, Eth5). From the config model, it looks like I can only specify interfaces as one big list as:
"ARS_INTERFACES": {
    "Ethernet1": {
         "ars_obj_name"          : "ars_A"
     },
     "Ethernet2": {
         "ars_obj_name"          : "ars_A"
     },
     "Ethernet3": {
         "ars_obj_name"          : "ars_A"
     },
     "Ethernet5": {
         "ars_obj_name"          : "ars_B"
     }
 }
 
 In this case, NHG-1 will get ars_A assigned unambiguously. However, for NHG-2, how do we pick the ars object?
 Is there a way, we can 'group' interfaces and assign ARS object to that group?

apannerselva: In this example, NHG-1 will be associated with ars_A, making it an ARS-capable NHG.
However, NHG-2 will not be ARS-capable because the ARS object linked to its interface has a different object name.

In High-Level Design, I have mentioned about the configuration guideline - 
```
**Important Consideration**:

Configuration Guidelines
- Interfaces participating in ARS should be properly configured in the ARS_INTERFACE table.
- For ARS functionality to operate consistently, all interfaces that belong to the same NHG or LAG must be associated with a same ARS object.
```

doc/ARS/Local_ARS_HLD.md
```diff
@@ -0,0 +1,1151 @@
...
+            key "if_name";
+
+            leaf if_name {
+                type leafref {
+                    path "/port:sonic-port/port:PORT/port:PORT_LIST/port:name";
+                }
+                description "ARS-enabled interface name";
+            }
+
+            leaf scaling_factor {
```
baxia-lan, line 674: the definition of "scaling_factor" differs from SAI_PORT_ATTR_ARS_PORT_LOAD_SCALING_FACTOR and brings ambiguity. 

`scaling_factor`: 10000
`SAI_PORT_ATTR_ARS_PORT_LOAD_SCALING_FACTOR`:40 # for 400G

Why not just set the value of SAI_PORT_ATTR_ARS_PORT_LOAD_SCALING_FACTOR to be value of scaling_factor?

`scaling_factor`: 40
`SAI_PORT_ATTR_ARS_PORT_LOAD_SCALING_FACTOR`:40 # for 400G

apannerselva: Previous changes handled cases where the speed was in MB. The current updates address review comments by using the SAI attribute instead.

doc/ARS/Local_ARS_HLD.md
apannerselva: @venkatmahalingam have addressed it the latest HLD

venkatmahalingam (approved):

lihuay (approved):

PR #1961, opened 2025-04-09
Inner mac rewrite hld

anish-n:
Add support for rewriting a VXLAN packet's inner Src Mac using ACL


prsunny (approved):
@bingwang-ms for viz

PR #1963, opened 2025-04-12
Dynamic BGP peer update and delete support

anish-n:


doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
```diff
@@ -0,0 +1,144 @@
...
+  * [Use case](#use-case)
+
+  * [Definitions/Abbreviation](#definitionsabbreviation)
+ 
+  * [1 Requirements Overview](#1-requirements-overview)
+    * [1.1 Functional requirements](#11-functional-requirements)
+    * [1.2 Scalability requirements](#12-scalability-requirements)
+  * [2 Modules Design](#2-modules-design)
+    * [2.1 Config DB](#21-config-db)
+    * [2.2 State DB](#22-state-db)
```
StormLiangMS, line 19: could you clarify what's the usage of state db here?

anish-n: Added, thanks

doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
```diff
@@ -0,0 +1,145 @@
...
+
+# 3 Test Plan
```
lipxu, line 135: If there is some issue or limitation in new feature, is it possible to disable it at runtime?

doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
```diff
@@ -0,0 +1,145 @@
...
+## 1.1 Functional requirements
+This section describes the SONiC requirements for supporting enhanced CRUD operations for dynamic bgp peers:
+- Configurable in VNET/VRFs, but equally applicable to default VRF
+- Dynamic peers' listen ranges must be modifiable at runtime such that ranges can be added and removed(changing today's behavior of being a create only attribute)
+- Dynamic peers should be deletable at runtime(changing today's behavior where deletes can only occur upon a full device/config reload)
+- Dynamic peer updates/deletes should allow for changes to associated configurations like route-maps, peer-groups, prefix-lists etc
+- Ability to track the configuration of a dynamic BGP peer via a State DB entry which is queryable by the SDN controller
+- CLI commands to show VRF/VNET BGP neighbors
+
+## 1.2 Scalability requirements
```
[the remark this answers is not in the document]
anish-n: None in terms of CPU and memory, just high scale of BGP sessions itself and associated peer group etc config

doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
anish-n: We will need to remove the template file, upon removing the template file associated with update/delete, the code will not be activated.

doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
```diff
@@ -0,0 +1,144 @@
...
+    "src_address": {{Src IP to initiate session}} (Optional)
+```
+Unless otherwise stated, the attributes are mandatory.
+
+While the above shows the State DB schema for a dynamic peer, the schema for static peers bear the same table name format of ```BGP_PEER_CONFIGURED_TABLE|{{VRF/VNET-name}}|{{Peer-name}}```, but the key value pairs under the table will be those which are used to configure static peers.
+
+## 2.3 CLI
+The following CLIs will be added, each of these will have similar CLI outputs to their ```default VRF``` counterparts which are widely in use today:
+```
+1. show ip bgp vrf {vrf_name} summary
```
arlakshm, line 94: Iso of adding new command can we add an option existing command `show ip bgp summary --vrf <vrf_name>` this way existing command `show ip bgp summary` will work for default VRF and can user can pass the vrf_name to get output per vrf.

doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
arlakshm: Can we add a config flag to disable this feature?

doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
```diff
@@ -0,0 +1,144 @@
...
+
+## 2.4 Bgpcfgd
+The following modifications will be made to bgpcfgd to support the new scenario:
+1. BGPPeerMgrBase will check, at init time, if an update template file exists as follows ```"bgpd/templates/" + self.constants["bgp"]["peers"][peer_type]["template_dir"] + "/" + update.conf.j2```.  
+   The `update.conf.j2` file is a Jinja2 template used to define the configuration updates for dynamic BGP peers.
+2. If one exists, then we will consider this device as supporting updates to the peer_type in question, and we append the update template to the list of templates which can be rendered:
+```
+        self.templates["update"] = self.fabric.from_file(update_template_path)
+```
+3. If an update template is supported and the peer_type is dynamic, we will identify the set of ip_ranges which are added, and those which are deleted as part of the operation, and pass in those as additional kwargs and render the update template, sample code:
```
arlakshm, line 110: > we will identify the set of ip_ranges which are added, and those which are deleted as part of the operation

Is this done in bgpcfgd? how is this done? Can you add more details?

doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
anish-n: oh the existing command show ip bgp summary will still work for default vrf, intention is to have all existing commands work as-is, and support the extra vrf option in a way that it matches vtysh. Doing it this way will match exactly how the command is structured in vtysh, making it helpful for everyone who is used to that structure of cli.

doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
anish-n: Yes it is done in bgpcfgd, it is done by querying vtysh to find the current set of ranges, and diff-ing that against the newly configured set of ranges, the added/deleted ranges are then passed into an update.j2 template to render the required config changes. PR: https://github.com/sonic-net/sonic-buildimage/pull/22261

I will update the HLD PR to add some more details on how it is done in bgpcfgd

prsunny (approved):

prsunny:
@StormLiangMS , @arlakshm , would you approve this?

doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
```diff
@@ -0,0 +1,145 @@
...
+- Dynamic peers' listen ranges must be modifiable at runtime such that ranges can be added and removed(changing today's behavior of being a create only attribute)
+- Dynamic peer updates/deletes should allow for changes to associated configurations like route-maps, peer-groups, prefix-lists etc
+- Ability to track the configuration of a dynamic BGP peer via a State DB entry which is queryable by the SDN controller
+- CLI commands to show VRF/VNET BGP neighbors
+
+## 1.2 Scalability requirements
+###### Table 2: Scalability
+| Component                | Expected value              |
+|--------------------------|-----------------------------|
+| Number of Dynamic BGP Peers each with route-map, prefix-list and peer-group|2k|
```
StormLiangMS: @anish-n Could you clarify what the "2K" refers to in this context? A scale of 2,000 is significantly larger compared to 64 neighbors and could potentially lead to performance issues or even system crashes. I recommend explicitly defining what the 2K and 4K entries represent in the table to avoid confusion.

doc/BGP/Bgpcfgd-dyn-peer-modification-support.md
```diff
@@ -0,0 +1,145 @@
...
+
+# 3 Test Plan
+Bgpcfgd, CLI and template changes will have corresponding UTs.
+
+New SONiC-mgmt tests will be introduced to test the following:
```
StormLiangMS, line 138: Hi @anish-n Do we have a more detailed test plan for this feature?

1. Scale testing – 2K peers is non-trivial, especially considering the scaled device.
2. Stress testing – including reloads and reboots.
3. Dynamic update/delete scenarios – what operations are supported? Could we add one section in this doc to explicitly describe what is supported and what is not? For example:
a. Updating the IP range of an existing dynamic peer.
b. Adding a new dynamic peer with route-map/prefix-list configurations.
c. Deleting an existing dynamic peer.
d. Updating the route-map/prefix-list of an existing dynamic peer.

And could you provide an update on the progress of this feature and clarify which release it is planned for?

anish-n: @StormLiangMS,
Added high level tests which will be added for scale and stress test. The intent of this HLD is supporting dynamic bgp peer changes, detailed stress test plan can be added to sonic-mgmt if required. 

In terms of operations supported, this list is present in the requirements section, from config db perspective there is no change in operations beyond allowing a change in ip_ranges passed in. We handle the change in ip range and then FRR templates will render associated change in route map/prefix list as present in template, exactly like how the add peer/delete peers function today, so operationally it would be identical.

StormLiangMS (approved):
LGTM

StormLiangMS:
hi @prsunny @arlakshm @lipxu  would you approve this one?

PR #1993, opened 2025-05-16
add doc for vnet local endpoint forwarding

zjswhhh:
Add doc for vnet local endpoint forwarding, required by ha.


doc/smart-switch/high-availability/vnet_local_endpoint_forwarding.md
```diff
@@ -0,0 +1,108 @@
...
+    acl_rule_table --> AclOrch
+    IntfOrch --> |get local endpoint intf alias| VnetOrch
+    AclOrch --> SAI/SDK
+```
+
+To distinguish the local endpoints from the endpoint list, (so we can avoid creating tunnel terminated ACL for all scenarios) a new field in `VNET_ROUTE_TUNNEL_TABLE` is added: 
+
+```
+VNET_ROUTE_TUNNEL_TABLE:{{vnet_name}}:{{prefix}}
+    "local_endpoint"  = /{/{true|false/},/{true|false/},.../}  (OPTIONAL)
```
prsunny: why we need two fields to be introduced? Can we leverage` check_directly_connected`

zjswhhh: Updated. Code PR updated accordingly as well.

PR #2008, opened 2025-06-11
Single-ASIC VOQ proposal

lakshmi-nexthop:
Details for supporting VOQ mode in single ASIC SONiC switch systems.

### DIFF
--- a/doc/voq/single_asic_voq.md
+++ b/doc/voq/single_asic_voq.md
@@ -1,0 +1,159 @@
+# **Single ASIC VOQ SONiC**
+
+# **High Level Design Document**
+
+### **Rev 1.0**
+
+# **Table of Contents**
+
+[Single ASIC VOQ SONiC](#single-asic-voq-sonic)
+
+[High Level Design Document](#high-level-design-document)
+
+[Rev 1.0](#rev-1.0)
+
+[Table of Contents](#table-of-contents)
+
+[Revision](#revision)
+
+[About this Manual](#about-this-manual)
+
+[Scope](#scope)
+
+[1 Requirements Overview](#1-requirements-overview)
+
+[1.1 Functional Requirements](#1.1-functional-requirements)
+
+[1.2 Configuration requirements](#1.2-configuration-requirements)
+
+[1.3 Agent requirements](#1.3-agent-requirements)
+
+[1.3.1 Orchagent](#1.3.1-orchagent)
+
+[1.3.2 Bgpconfd](#1.3.2-bgpconfd)
+
+[1.3.3 Sonic-utilities](#1.3.3-sonic-utilities)
+
+[1.3.4 Sonic-host-services Caclmgrd](#1.3.4-sonic-host-services-caclmgrd)
+
+[2 Modules Design](#2-modules-design)
+
+[2.1 Switch Type “VOQ” For Both Chassis and Single-ASIC cases](#2.1-switch-type-“voq”-for-both-chassis-and-single-asic-cases)
+
+[2.1.1 Orchagent, Bgpconfd Changes](#2.1.1-orchagent,-bgpconfd-changes)
+
+[2.1.2 Sonic-utilities, Sonic-host-services/Caclmgrd Changes](#2.1.2-sonic-utilities,-sonic-host-services/caclmgrd-changes)
+
+[2.1.3 CLI Changes](#2.1.3-cli-changes)
+
+[2.1.4 Config Generation Changes](#2.1.4-config-generation-changes)
+
+[2.2 Separate Switch Type for Single ASIC VOQ (“single-asic-voq”)](#2.2-separate-switch-type-for-single-asic-voq-\(“single-asic-voq”\))
+
+[2.2.1 Orchagent, Bgpconfd Changes](#2.2.1-orchagent,-bgpconfd-changes)
+
+[2.2.2 Sonic-utilities, Sonic-host-services/Caclmgrd Changes](#2.2.2-sonic-utilities,-sonic-host-services/caclmgrd-changes)
+
+[2.2.3 CLI Changes](#2.2.3-cli-changes)
+
+[2.2.4 Config Generation Changes](#2.2.4-config-generation-changes)
+
+###### **Revision**
+
+| Rev | Date | Author | Change Description |
+| ----- | ----- | ----- | ----- |
+| 1.0 | 06/10/2025 | Eswaran Baskaran[Lakshmi Yarramaneni](mailto:lakshmi@nexthop.ai) | Initial public version |
+
+# **About this Manual**
+
+This document describes the design details for supporting SONiC on a single-ASIC VOQ System. 
+
+# **Scope**
+
+This specification focussed on how to support VOQ switch functionality on single-ASIC VOQ systems. And understand the impact on various Sonic modules in single-ASIC vs. chassis based.
+
+# **1 Requirements Overview**
+
+## **1.1 Functional Requirements**
+
+The single-ASIC VOQ implementation shall support VOQ mode without relying on the presence of Chassis DB.
+
+## **1.2 Configuration requirements**
+
+iBGP configuration that was generated for chassis-based VOQ systems is not needed in single-ASIC VOQ systems. QoS configuration that was generated for system ports continues to be needed for single-ASIC VOQ systems. 
+
+## **1.3 Agent requirements**
+
+### **1.3.1 Orchagent**
+
+- Support VOQ and single-ASIC VOQ systems  
+  - Interface with Chassis DB in Chassis VOQ system but not in a single-ASIC VOQ system  
+  - Enable and manage fabric ports for single ASIC VOQ
+
+  ### **1.3.2 Bgpconfd**
+
+- Spawn off ChassisDbMgr only on Chassis VOQ system but not on single-ASIC VOQ system  
+- Handle TSA only on chassis VOQ and not on single-ASIC VOQ
+
+  ### **1.3.3 Sonic-utilities**
+
+- Support line card extensions only on Chassis VOQ system but not on single-ASIC VOQ  
+- Differentiate between internal and external BGP neighbors on Chassis VOQ only  
+- Fabric port status should only be retrieved from Chassis DB on Chassis VOQ systems but not on single-ASIC VOQ system  
+- Multi ASIC checks must evaluate to false on single ASIC VOQ
+
+  ### **1.3.4 Sonic-host-services Caclmgrd**
+
+- Should support midplane traffic only on Chassis VOQ system
+
+# **2 Modules Design**
+
+There are 2 approaches to support VOQ mode in single ASIC systems. They are described in sections 2.1 and 2.2 below.
+
+### **2.1 Switch Type “VOQ” For Both Chassis and Single-ASIC cases**
+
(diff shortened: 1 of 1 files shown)


doc/voq/single_asic_voq.md
```diff
@@ -0,0 +1,159 @@
...
+
+# **1 Requirements Overview**
+
+## **1.1 Functional Requirements**
+
+The single-ASIC VOQ implementation shall support VOQ mode without relying on the presence of Chassis DB.
+
+## **1.2 Configuration requirements**
+
+iBGP configuration that was generated for chassis-based VOQ systems is not needed in single-ASIC VOQ systems. QoS configuration that was generated for system ports continues to be needed for single-ASIC VOQ systems.
```
arlakshm: I think even for single ASIC voq devices will need the Qos configuration on the system ports, the idea is to have consistent config and SAI programing model for all voq devices

doc/voq/single_asic_voq.md
```diff
@@ -0,0 +1,159 @@
...
+###### **Revision**
+
+| Rev | Date | Author | Change Description |
+| ----- | ----- | ----- | ----- |
+| 1.0 | 06/10/2025 | Eswaran Baskaran
[Lakshmi Yarramaneni](mailto:lakshmi@nexthop.ai) | Initial public version |
+
+# **About this Manual**
+
+This document describes the design details for supporting SONiC on a single-ASIC VOQ System.
```
arlakshm: I think we need to change this is single ASIC voq fixed system device to avoid confusion with single asic voq chassis linecards

doc/voq/single_asic_voq.md
lakshmi-nexthop: That is correct. Unlike iBGP configs, QOS configuration that is generated for VOQ Chassis is needed for single ASIC VOQ also.

doc/voq/single_asic_voq.md
lakshmi-nexthop: sure, I can clarify.

arlakshm (approved):

deepak-singhal0408:
@arlakshm could you sign-off this HLD? thanks,

arista-nwolfe (commented):
The HLD doesn't talk about what to do with the existing internal ports for single-asic VOQ systems.
I.E. `Ethernet-IB0` and `Ethernet-Rec0`
I believe neither of these ports are needed for single-asic VOQ and some changes will be required in `mirrororch.cpp` to avoid requiring `Ethernet-Rec0` for everflow.

lakshmi-nexthop:
> The HLD doesn't talk about what to do with the existing internal ports for single-asic VOQ systems. I.E. `Ethernet-IB0` and `Ethernet-Rec0` I believe neither of these ports are needed for single-asic VOQ and some changes will be required in `mirrororch.cpp` to avoid requiring `Ethernet-Rec0` for everflow.

updated the orchagent section.

doc/voq/single_asic_voq.md
```diff
@@ -0,0 +1,149 @@
...
+- Support VOQ and single-ASIC VOQ modes  
+    
+  - Interface with Chassis DB in Chassis VOQ system but not in a single-ASIC VOQ  
+  - Enable and manage fabric ports for single ASIC VOQ case also
+
+  ### **1.4.2 Bgpconfd**
+
+- Spawn off ChassisDbMgr only on Chassis VOQ system but not on single-ASIC VOQ  
+    
+- Handle TSA (Traffic Shift Away) only on chassis VOQ and not on single-ASIC VOQ
```
kenneth-arista: Why doesn't TSA apply to single ASIC VOQ?

doc/voq/single_asic_voq.md
```diff
@@ -0,0 +1,149 @@
...
+
+In the platform configuration file *platform.env*, there is a *disaggregated\_chassis* flag which indicates the presence of a database-chassis container. We do not intend to use this flag as we will not be using chassis db at all.
+
+API *is\_voq\_chassis* will check for the presence of the *chassisdb.conf* file.
+
+### **2.1.1 Orchagent Changes**
+
+- Orchagent will handle VOQ functionality the same way i.e. creation of system ports. But connect to Chassis DB only if chassis DB is supported in the sonic system.  
+    
+- Given the inband and recirculation ports are not created, we need to make sure mirrororch, neighorch, etc are updated to not look for these ports in single-asic-voq mode.
```
kenneth-arista: It is still debatable whether or not recirculation is needed for everflow. As discussed in https://github.com/sonic-net/sonic-buildimage/issues/23617 a limitation exists if a recycle port is not used.

doc/voq/single_asic_voq.md
```diff
@@ -0,0 +1,149 @@
...
+
+- Support midplane traffic only on Chassis VOQ system
+
+# **2 Modules Design**
+
+### **2.1 Switch Type “VOQ” For Both Chassis and Single-ASIC cases**
+
+This method reuses the *voq* switch type for non chassis systems as well. This means the agents, utilities and CLI config tools will need to differentiate between chassis system vs non chassis in voq mode. Chassis configuration file *chassisdb.conf* can indicate if the voq system is a chassis system or not.
+
+In the platform configuration file *platform.env*, there is a *disaggregated\_chassis* flag which indicates the presence of a database-chassis container. We do not intend to use this flag as we will not be using chassis db at all.
```
kenneth-arista, line 123: You're right that the flag `disaggregated_chassis` today is used to gate the creation of CHASSIS_APP_DB. However, conceptually it is non-ideal to diverge architectures based on whether or not the platform is single versus multiple ASICs.

doc/voq/single_asic_voq.md
lakshmi-nexthop: TSA does apply to single ASIC case also. This was original mis conception. will update the doc.

doc/voq/single_asic_voq.md
lakshmi-nexthop: thanks Ken. I am also running into issue with recirc port. will discuss details.

doc/voq/single_asic_voq.md
lakshmi-nexthop: I understand your comment, but in single asic FS case, presence of chassis-database container is not valid to me and breaks the fundamental concept of FS.

eswaran-nexthop: @kenneth-arista do you mean that we have to keep the database-chassis for single-chip systems as well, or something else?

There are some pros to keeping the same structure - database-chassis container, CHASSIS_APP_DB, etc. There are some operational implications like allocating LAG IDs by hand, etc, that could be issues. This is worth a discussion in the workgroup meeting.

abdosi:
1. Egress Everflow do we need Recirc Port ==> Waiting for BRCM
2. Do recirc port just for base use case of Regular Ingress Mirroring 
     2.1 Regular Ingress mirroring does not need Recirc Port
     2.2 Regular Ingress Mirroring +  Policer  needs Recirc Port . 

@arlakshm : Arista/Nexthop wants to starts with Recirc Port (maintain backward compatibility) and later can be removed as optimization. Please help close on this.

abdosi:
Please add code PR links here.

doc/voq/single_asic_voq.md
lakshmi-nexthop: updated the doc to reflect that we keep recirc ports.

lakshmi-nexthop:
code PR: 
- https://github.com/sonic-net/sonic-swss/pull/3847
- https://github.com/sonic-net/sonic-buildimage/pull/24163
- https://github.com/sonic-net/sonic-mgmt/pull/20467
- https://github.com/sonic-net/sonic-mgmt/pull/20872

abdosi:
> 2\. Do recirc port just for base use case of Regular Ingress Mirroring
>     2.1 Regular Ingress mirroring does not need Recirc Port
>     2.2 Regular Ingress Mirroring +  Policer  needs Recirc Port .

Nexthop confirmed both Ingress and Egress Everflow works fine with Recirc Port.  Need to check for Policier.

abdosi:
@arista-nwolfe / @arlakshm please sign-off to merge this.

doc/voq/single_asic_voq.md
```diff
@@ -0,0 +1,147 @@
...
+
+API *is\_voq\_chassis* will check for the presence of the *chassisdb.conf* file.
```
arista-nwolfe, line 125: The `chassisdb.conf` file also lists the number of LAGs supported in the system:
```
lag_id_start=1
lag_id_end=1023
```

If single-asic VOQ doesn't have a `chassisdb.conf` do we need to store these variables in a different SKU file?

doc/voq/single_asic_voq.md
```diff
@@ -0,0 +1,147 @@
...
+
+### **2.1.1 Orchagent Changes**
+
+- Orchagent will handle VOQ functionality the same way i.e. creation of system ports. But connect to Chassis DB only if chassis DB is supported in the sonic system.  
+    
+- Given the inband ports are not created, we need to make sure mirrororch, neighorch, etc are updated to not look for these ports in single-asic-voq mode.
```
arista-nwolfe: What does `mirrororch` use the inband ports for?

doc/voq/single_asic_voq.md
lakshmi-nexthop: lag_id_start, lag_id_end are only applicable to chassis (and multi asic cases) as the lag ids need to be coordinated. In single asic case, no coordination needed and so we are skipping these variables.

arista-nwolfe: Correct LagId allocation doesn't need to be coordinated over multiple slices in single-asic VOQ systems but something still needs to allocate a LagId from a pool of LagIds and make sure we don't go beyond `1024` LagIds in the system. Previously that was coordinated by LUA script that depended on these values in CHASSIS_APP_DB
https://github.com/sonic-net/sonic-swss/blob/1b3f738f34fdaad69d7c1b8815023ae8f4864893/orchagent/lagids.lua#L15

I'm wondering who is doing this LagId allocation once single-asic VOQ isn't running CHASSIS_APP_DB.

lakshmi-nexthop: lagid allocation is done by the system just like it is done in a non-voq system.
you can see below code in portsorch.cpp
sai_object_id_t lag_id;
sai_status_t status = sai_lag_api->create_lag(&lag_id, gSwitchId, static_cast<uint32_t>(lag_attrs.size()), lag_attrs.data());

lag_id_start/end is basically for the system port lag (SAI_LAG_ATTR_SYSTEM_PORT_AGGREGATE_ID) which we dont need in voq fixed system case.
I hope it is clear.

arista-nwolfe: Ah okay I see, so in the case of single-asic VOQ we let the SAI allocate the LAG_ID.
I didn't realize this SAI attribute `SAI_LAG_ATTR_SYSTEM_PORT_AGGREGATE_ID` was optional
https://github.com/opencomputeproject/SAI/blob/401bd1f2c5a0332d7ea9d2356ab10db34a168427/inc/sailag.h#L154-L168

Thanks for the explanation.

sridhartalari:
Can you please provide explanation on difference between regular single asic fixed system and single ASIC fixed sytem with VOQ. also the reason we need it . Please let us know if we have fabric ports even if  it is single ASIC system

PR #2016, opened 2025-06-13
Add HLD for Reboot support BlockingMode in SONiC

lityu:
We are trying to introduce blocking mode for reboot script, which currently is in non-blocking mode. This will make automation system easier to identify whether the reboot operation is success.

Code and test PRs:
[Add BlockingMode for Reboot script by lityu · Pull Request #3958 · sonic-net/sonic-utilities](https://github.com/sonic-net/sonic-utilities/pull/3958)
[Add test plan for reboot blocking mode by lityu · Pull Request #19600 · sonic-net/sonic-mgmt](https://github.com/sonic-net/sonic-mgmt/pull/19600)

### DIFF
--- a/doc/reboot/Reboot_BlockingMode_HLD.md
+++ b/doc/reboot/Reboot_BlockingMode_HLD.md
@@ -23,7 +23,7 @@ Currently the reboot script uses linux command `systemctl reboot` to reboot the 
 
 As the result, the output from user batch side will be unpredictable. It displays as 2 kinds of behavior randomly:
 
-- [A] The user bash stopped by systemd before `systemctl reboot` returns. Then the automation system will see extra user prompt. The time sequence is as following:
+- [A] The user bash stopped by systemd after `systemctl reboot` returns. Then the automation system will see extra user prompt. The time sequence is as following:
 
 ![MAN Reboot](./img/background-A.png)
 
@@ -38,7 +38,7 @@ admin@sonic:~$ Connection to 10.150.22.134 closed by remote host.
 Connection to 10.150.22.134 closed.
 ```
 
-- [B] The user bash stopped by systemd after `systemctl reboot` returns. Then the automation system will not see extra user prompt. The time sequence is as following:
+- [B] The user bash stopped by systemd before `systemctl reboot` returns. Then the automation system will not see extra user prompt. The time sequence is as following:
 
 ![MAN Reboot](./img/background-B.png)
 


doc/reboot/Reboot_BlockingMode_HLD.md
```diff
@@ -0,0 +1,117 @@
...
+
+We are trying to introduce blocking mode for reboot script, which currently is in non-blocking mode. This will make automation system easier to identify whether the reboot operation is success.
+
+## Background
+
+Currently the reboot script uses linux command `systemctl reboot` to reboot the SONIC system. And the linux command is asynchronous. For detail you can refer `reboot` section in [Linux Man Doc for systemctl](https://www.man7.org/linux/man-pages/man1/systemctl.1.html).
+
+As the result, the output from user batch side will be unpredictable. It displays as 2 kinds of behavior randomly:
+
+- [A] The user bash stopped by systemd before `systemctl reboot` returns. Then the automation system will see extra user prompt. The time sequence is as following:
```
lizhijianrd: The user bash stopped by systemd **after** `systemctl reboot` returns?

doc/reboot/Reboot_BlockingMode_HLD.md
```diff
@@ -0,0 +1,117 @@
...
+admin@sonic:~$ sudo reboot
+Error response from daemon: Container e6a136aee372584ae6e64d1c55ae49352d842d76daff13a782a4ecfa0262d0e5 is not running
+/var/log: 3.9 GiB (4138283008 bytes) trimmed on /dev/loop1
+/host: 8 GiB (8547295232 bytes) trimmed on /dev/sda2
+Fri 20 Oct 2023 06:03:33 AM UTC Issuing OS-level reboot ...
+admin@sonic:~$ Connection to 10.150.22.134 closed by remote host.
+Connection to 10.150.22.134 closed.
+```
+
+- [B] The user bash stopped by systemd after `systemctl reboot` returns. Then the automation system will not see extra user prompt. The time sequence is as following:
```
lizhijianrd: And this line should be "before"?

bmridul:
Do we have code PRs for the changes proposed in this HLD? Can you add them to the PR description.

doc/reboot/Reboot_BlockingMode_HLD.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Blocking Mode
+
+We don't want to make a break change to SONIC reboot command, so we introduce 2 options as the following for user to identify that we are trying to run on blocking-mode. Blocking-mode will be enable if any of the option is enabled.
+
+- Parameter: Use command optional parameter `-b` to enable blocking mode for reboot script. The whole command will looks like `reboot -b`.
+
+- Config file: The reboot script will check if the `blocking_mode` is enabled in `reboot.conf` config file. More details can be check in the [reboot.conf](#rebootconf) section.
+
+If the systemctl reboot failed, the deadloop might block the console thread. So we introduce the timeout config to exit when reboot takes too long. The command line don't needs this because user can manual use `Ctrl + C` to interrupt the script.
```
bmridul, line 70: Is it possible that the bash user is killed by the time timeout expires? Will the console be available after timeout in that case?

Blueve: Yes. The user might be kicked out if the reboot too much time.
In current implementation, it doesn't consider it. 
If the SSH timeout/console session timeout was set to a smaller number than reboot timeout setting, then user will be logout first.

Blueve:
note: CONFIG_DB could be helpful in ZTP scenario and user might not want to touch file system --- from Brad

Blueve:
--- 
Jing (MSFT) : 

The conf is mainly for clarifying the default behavior of reboot script.
You can treat it as a series of default parameter to a CLI

## gnoi scenario:
The gnoi API need to be well defined. It is not good to rely on config_db to define its behavior as it will create more confusion and complexity. e.g. 
```
style 1:
/system/ping
{
   ...
   count = 3; // number of pkts
   ...
}

style 2:
good:
/system/ping
//// read [count] from CONFIG_DB/PING_CONF/COUNT
```
I would prefer the style 1 which is clearer to me


## ztp scenario:
If ZTP been used, then CLI reboot is not supposed to be used as normal operation approach.
But the admin might hope config the default reboot CLI behavior through ZTP and then make the operator's CLI operations more robust. It could be a valid scenario, but the benefit is less as the user can always perform what they want to perform in CLI mode.

doc/reboot/Reboot_BlockingMode_HLD.md
lityu: Add code PR link and test PR link:
[Add BlockingMode for Reboot script by lityu · Pull Request #3958 · sonic-net/sonic-utilities](https://github.com/sonic-net/sonic-utilities/pull/3958)
[Add test plan for reboot blocking mode by lityu · Pull Request #19600 · sonic-net/sonic-mgmt](https://github.com/sonic-net/sonic-mgmt/pull/19600)

zhangyanzhao:
@yxieca  @Blueve  can you please help to merge this PR? Thanks.

PR #2018, opened 2025-06-17
HLD for persistent local user management

manoharan-nexthop:
Currently the local users that are created are not persistent across upgrades. This HLD proposes to manage the users across upgrades by storing the configuration in CONFIG_DB and a daemon to manage them in Linux.

Implementation Pull requests:
* https://github.com/sonic-net/sonic-buildimage/pull/23923
* https://github.com/sonic-net/sonic-host-services/pull/309
* https://github.com/sonic-net/sonic-utilities/pull/4053

### DIFF
--- a/doc/User_Management/User_Management.md
+++ b/doc/User_Management/User_Management.md
@@ -1,0 +1,535 @@
+# Declarative Local User Management
+
+## Revision
+
+| Rev | Date           | Author      | Change Description                                                                                                                                   |
+|:---:|:--------------:|:-----------:|:----------------------------------------------------------------------------------------------------------------------------------------------------:|
+| 1.0 | June 16, 2025  | Manoharan Sundaramoorthy   | Initial HLD                                                                                                                                          |
+
+
+## 1. Scope
+This document describes the high-level design for a new feature that provides a declarative and persistent method for managing local user accounts on a SONiC device. This feature allows administrators to define local users, including their roles, password hashes, SSH keys, and **security policies** including **login attempt limits**.
+
+This ensures that local user accounts are consistently applied with robust security postures and persist across reboots and upgrades.
+
+## 2. Definitions/Abbreviations
+
+| Abbreviation | Definition                                                                |
+|:------------:|:--------------------------------------------------------------------------|
+| **`userd`** | The new User Daemon proposed in this design.                              |
+| **PAM** | Pluggable Authentication Modules                                          |
+
+
+## 3. Overview
+This feature introduces a new dedicated daemon, **`userd`**, which manages the full lifecycle of local users based on definitions in `CONFIG_DB`. It simplifies management by providing a user-friendly CLI, mapping abstract roles (`administrator`, `operator`) to specific Linux groups, and enforcing security policies. This provides a solution for managing secure, persistent local user accounts.
+
+**Implementation Language:** The `userd` daemon will be implemented in **C++** to ensure optimal memory usage and performance, consistent with other SONiC system daemons.
+
+## 4. Requirements
+### 4.1 Functional Requirements
+1.  The system must allow an administrator to define a local user account declaratively.
+2.  The user definition must support:
+    * Username and a pre-hashed password.
+    * A role, limited to either **`administrator`** or **`operator`**.
+    * Authorized SSH keys (statically defined).
+3.  The system must map roles to underlying Linux groups:
+    * `administrator`: members of `sudo`, `docker`, `redis` and `admin` groups.
+    * `operator`: members of a standard, non-privileged and would belong to `users` group.
+4.  The system will **auto-generate** a unique UID for each new user.
+5.  There should be atleast one user configured in the role of `administrator`
+6.  **Admin User Requirement:**
+    * The **`admin`** user (or the user specified during compilation) must NOT be included in the `init_cfg.json` to avoid overwriting existing configurations.
+    * Before enabling the feature, users must run `config user import-existing` to import existing Linux users (including admin) into CONFIG_DB.
+    * The feature **cannot** be enabled unless at least one administrator user exists in CONFIG_DB.
+    * The admin user can be modified or disabled or deleted through standard CONFIG_DB operations once managed.
+7.  **Security Policy Requirements:**
+    * **Login Attempts:** The system must support configuring a global maximum number of failed login attempts per role before accounts are temporarily locked.
+8.  User accounts and their configurations must persist across system reboots and upgrades.
+9.  **System Consistency:** On startup, the system must perform a consistency check to ensure Linux users match CONFIG_DB definitions and automatically remove any users that were added directly to Linux (bypassing CONFIG_DB).
+
+## 5. Architecture Design
+The architecture centers on the new `userd` daemon. This daemon will now interact with several core Linux subsystems to enforce the configured security policies.
+
+**Feature Control:**
+* **Disabled by Default:** The declarative local user management feature is **disabled by default** to ensure no impact on existing SONiC deployments.
+* **Feature Flag:** The feature is controlled by the `local_user_management` field in `DEVICE_METADATA|localhost` (values: `enabled` or `disabled`).
+* **Daemon Lifecycle:** The `userd` daemon only starts when the feature is explicitly enabled, ensuring zero impact when disabled.
+
+**`userd`'s Points of Interaction (when enabled):**
+1.  **CONFIG_DB:** Single source of truth for user configuration and global security policies.
+2.  **Core User Files:** `/etc/passwd`, `/etc/shadow`, `/etc/group` for basic user identity.
+3.  **PAM Configuration (`/etc/security/faillock.conf`):** To configure failed login attempt thresholds that lock individual user accounts via `pam_faillock`.
+
+## 6. High-Level Design
+
+### 6.1 `userd` Daemon
+The `userd` daemon's logic will be expanded to manage security configurations idempotently.
+
+**Startup Consistency Check:**
+* **System Reconciliation:** On startup, `userd` will perform a consistency check to ensure that all local users in the Linux system (`/etc/passwd`, `/etc/shadow`, `/etc/group`) match the definitions in CONFIG_DB.
+* **Cleanup of Unmanaged Users:** Any users found in the Linux system that are not defined in CONFIG_DB (except for system users like `root`, `daemon`, etc.) will be automatically removed to maintain consistency. Note: Only local users that are managed will be cleaned up. This is achieved by making the managed users part of a special group `local_mgd`, ensuring that other users are not intentionally removed.
+* **CONFIG_DB as Source of Truth:** This ensures that CONFIG_DB remains the single source of truth for user management and prevents configuration drift.
+
+**Feature Enablement Process:**
+* **Initial State:** When the feature is disabled (default), existing SONiC user management continues unchanged.
+* **Required Workflow:**
+  1. Run `config user import-existing --dry-run` to preview what will be imported
+  2. Run `config user import-existing` to import existing users into CONFIG_DB (preserving passwords and configurations)
+  3. Review imported users with `show user` command
+  4. Verify at least one administrator user exists in CONFIG_DB
+  5. Enable the feature with `config user feature enabled`
+  6. The `userd` daemon starts and begins managing the imported users
+* **Prerequisite:** The feature **cannot** be enabled unless at least one administrator user exists in CONFIG_DB. The `config user import-existing` command must successfully import at least one user with administrator role (users in sudo/admin groups) before feature enablement is allowed.
+
+**Admin User Requirement:**
+* **Prerequisite for Enablement:** At least one administrator user must exist in CONFIG_DB before the feature can be enabled.
+* **Import Process:** Run `config user import-existing` to import existing users (including admin) into CONFIG_DB, preserving their current configurations (passwords, SSH keys, etc.).
+* **Standard Management:** Once imported into CONFIG_DB, the admin user is managed through the same CONFIG_DB interface as other users.
+
+**User Management Process:**
+* **Group Membership:** All users managed by `userd` are automatically added to the special group `local_mgd` to identify them as managed users.
+* **Lifecycle Management:** When users are created, modified, or imported, they are added to the `local_mgd` group to ensure proper tracking and cleanup operations.
+
+**Configuration File Hierarchy:**
+* **`init_cfg.json`:** Contains `DEVICE_METADATA|localhost` with `local_user_management` field set to `disabled` by default. Does NOT include any LOCAL_USER entries to avoid overwriting existing user configurations (e.g., changed passwords) during feature enablement.
+* **`golden_config_db.json`:** Optional site-specific configuration file that can enable the feature and provide initial user configurations when used with `config load_minigraph --override_config`.
+* **Configuration Precedence:** When both files are present, `golden_config_db.json` takes precedence over `init_cfg.json` during minigraph reload operations with override enabled.
+* **Feature Enablement Workflow:** To enable the feature, users must first import existing users with `config user import-existing`, then enable the feature with `sudo config user feature enabled`. This ensures existing user configurations (including password changes) are preserved.
+
+**User Account Management:**
+* **Account Status Control:** `userd` will monitor the `enabled` attribute for each user. When `enabled` is `false`, it will change the user's shell to `/usr/sbin/nologin` to completely disable the account. When `enabled` is `true`, it will set the shell to `/bin/bash`.
+* **User Directory and Shell Configuration:** Users have home directory `/home/<username>/` and default shell `/bin/bash`.
+* **Home Directory Management:** When users are deleted, home directories are preserved (standard `userdel` behavior). Administrators can manually remove home directories if needed.
+
+**Migration and Import Logic:**
+
+* **User Discovery and Import:**
+    * `userd` will scan `/etc/passwd` for users within the configurable UID range (default: 1000-60000)
+    * For each discovered user, it will read the password hash from `/etc/shadow`
+    * Role assignment is determined by group membership analysis:
+      - Users in `sudo`, `docker`, `redis`, or `admin` groups are assigned `administrator` role
+      - All other users are assigned `operator` role
+    * SSH keys are imported from `~/.ssh/authorized_keys` if present
+    * System users (UID < 1000) and users already in CONFIG_DB are skipped
+
+* **Migration Safety Features:**
+    * **Dry-run Mode:** Preview import results without making changes
+    * **UID Range Filtering:** Only import users within specified UID ranges
+    * **Conflict Detection:** Skip users already managed by CONFIG_DB
+    * **Validation:** Ensure at least one administrator user exists after import
+    * **Rollback Capability:** Import can be undone by disabling the feature
+
+**New Logic for Security Policies:**
+
+* **To enforce Login Attempts:**
+    * `userd` will manage the PAM configuration file at `/etc/security/faillock.conf`.
+    * For role-based limits (e.g., administrators with limit of 5), it will configure different thresholds based on user group membership.
+    * When a user exceeds their role's failed attempt limit, `pam_faillock` will lock that specific user account.
+    * `userd` will be responsible for ensuring the PAM stack is configured to use `pam_faillock` with role-based policies.
+    * **PAM vs SSH Configuration:** The `pam_faillock` configuration in `/etc/security/faillock.conf` differs significantly from SSH's `MaxAuthTries` in `/etc/ssh/sshd_config`:
+
+      **SSH MaxAuthTries (`/etc/ssh/sshd_config`):**
+      - Controls attempts **per SSH connection session** before dropping the connection
+      - Default value is typically 6 attempts
+      - Only affects SSH connections, not console or other login methods
+      - Connection-scoped: user can immediately reconnect and get another set of attempts
+      - No persistent account lockout - just drops the current connection
+
+      **PAM faillock (`/etc/security/faillock.conf`):**
+      - Controls **total failed attempts across all login methods** (SSH, console, serial, etc.)
+      - Provides **persistent account lockout** that survives connection drops and reboots
+      - Account remains locked until administrator intervention or timeout expires
+      - System-wide enforcement regardless of login method
+      - Can be configured with different policies per user group/role
+
+      **How they work together:**
+      - SSH may drop a connection after `MaxAuthTries` failed attempts
+      - But `pam_faillock` tracks total failures for each individual user across all connections and methods
+      - Once `pam_faillock` threshold is reached, that specific user account is locked even for new SSH connections
+      - This provides layered security: connection-level (SSH) + account-level (PAM) protection
+
+## 7. SAI API
+No SAI API changes are required.
+
+## 8. Configuration and Management
+### 8.1 Config DB Enhancements
+The `LOCAL_USER` table schema is defined for individual user accounts, and a new `LOCAL_ROLE_SECURITY_POLICY` table is added for global role-based security policies.
+
+**Schema:**
+```json
+// Example for CONFIG_DB
+{
+    "FEATURE": {
+        "LOCAL_USER_MANAGEMENT": {
+            "state": "disabled"
+        }
+    },
+    "LOCAL_USER": {
+        "admin": {
+            "role": "administrator",
+            "password_hash": "$6$salt$hash_of_YourPaSsWoRd",
+            "ssh_keys": ["ssh-rsa AAA..."],
+            "enabled": true
+        },
+        "newadmin": {
+            "role": "administrator",
+            "password_hash": "hashed_password",
+            "ssh_keys": ["ssh-rsa BBB..."],
+            "enabled": true
+        },
+        "showuser": {
+            "role": "operator",
+            "password_hash": "hashed_password",
+            "ssh_keys": ["ssh-rsa CCC..."],
+            "enabled": false
+        }
+    },
+    "LOCAL_ROLE_SECURITY_POLICY": {
+        "administrator": {
+            "max_login_attempts": 5
+        },
+        "operator": {
+            "max_login_attempts": 3
+        }
+    }
+}
+```
+
+**DEVICE_METADATA Table (localhost entry):**
+* `local_user_management` (string, optional): `enabled` or `disabled`. Controls whether the declarative local user management feature is active. Defaults to `disabled` if not specified.
+
+**LOCAL_USER Table (only active when feature is enabled):**
+* `role` (string, required): `administrator` or `operator`.
+* `password_hash` (string, optional): The hashed password. Defaults to `"!"` (no password).
+* `ssh_keys` (optional): List of SSH public keys for the user. No limits are imposed on the number of SSH keys.
+* `enabled` (boolean, optional): Whether the user account is enabled. Defaults to `true` if not specified. When `false`, user shell is set to `/usr/sbin/nologin`.
+* **Validation:** Each user must have either a valid password hash (not starting with `"!"`) or at least one SSH key.
+
+**LOCAL_ROLE_SECURITY_POLICY Table (only active when feature is enabled):**
+* **`max_login_attempts`** (integer, optional): Number of failed login attempts before accounts with this role are locked.
+
+**Notes:**
+* Session timeouts are managed by the system's default timeout policy and are not configurable per user.
+* **Password Hash Generation:** Password hashes can be generated using `mkpasswd` command-line utility or programmatically using libraries like `passlib` in Python for NETCONF/RESTCONF implementations. **Note:** Modern Debian 12 systems use **Yescrypt** as the default hash function.
+* **Integration with Password Hardening:** For CLI-created passwords (using `--password-prompt`), the system will validate that passwords meet any configured password hardening requirements before hashing and storing. When a pre-computed hash is provided (using `--password-hash`), it is the administrator's responsibility to ensure the original password met hardening requirements, as the hash cannot be reverse-validated.
+* **Password Security and Logging:** **IMPORTANT:** All password handling must ensure passwords are never logged in plaintext:
+  - CLI commands must not echo passwords to terminal or log them in bash history
+  - Syslog and auth logs must never contain plaintext passwords
+  - Memory containing plaintext passwords must be cleared immediately after hashing
+* **CONFIG_DB Security:** To prevent unauthorized access to password hashes:
+  - `/etc/sonic/config_db.json` file permissions changed from `644` to `640` (root:redis)
+  - Password hashes only visible in show commands when run with `sudo` privileges
+  - `show user` displays usernames only, `sudo show user` displays hashes
+  - `show runningconfiguration` redacts password_hash fields unless run with `sudo`
+* The CLI provides both `--password-hash` (for pre-hashed passwords) and `--password-prompt` (for interactive secure password entry) options for improved security and usability.
+* **CONFIG_DB as Source of Truth:** On startup, `userd` performs a consistency check and removes any users that were added directly to Linux (bypassing CONFIG_DB) to ensure CONFIG_DB remains the authoritative source for user management.
+* **User Account Status:** When `enabled` is set to `false`, `userd` will change the user's shell to `/usr/sbin/nologin` to completely disable the account. When `enabled` is `true` (default), the shell is set to `/bin/bash`.
+* **Complete Account Disabling:** Disabled accounts cannot login via any method (password, SSH keys, console) providing complete security when accounts need to be temporarily deactivated.
+* **Role-Based Security Policies:** Security policies are applied per role (administrator/operator) rather than per user for simplified management, consistent security posture, and scalability.
+* **Account Disabling Use Cases:** The `enabled: false` feature supports real operational needs including temporary leave, security incidents, onboarding/offboarding workflows, and compliance requirements.
+* **Administrator Availability:** At least one administrator user must remain enabled at all times to prevent complete loss of administrative access. This constraint is enforced by the Yang model.
+* **Admin User Requirement:** The `admin` user is NOT included in `init_cfg.json` to avoid overwriting existing configurations. Users must follow the proper enablement sequence: (1) run `config user import-existing` to import existing users including admin (preserving any password changes), (2) enable the feature with `sudo config user feature enabled`. The feature **cannot** be enabled unless at least one administrator user exists in CONFIG_DB. This ensures existing user configurations are never lost during feature enablement.
+* **TACACS+/RADIUS Compatibility:** This local user management feature operates independently of and is fully compatible with existing TACACS+ and RADIUS authentication systems. Local users are managed through standard Linux PAM authentication stack, which can coexist with remote authentication methods. The authentication order and fallback behavior are controlled by existing AAA configuration in CONFIG_DB.
+
+### 8.2 Yang Model Enhancements
+```
+module sonic-user {
+    yang-version 1.1;
+    namespace "http://sonicproject.com/sonic-user";
+    prefix "sonic-user";
+
+    import sonic-ext {
+        prefix "sonic-ext";
+    }
+
+    // Common typedef for user roles
+    typedef user-role {
+        type enumeration {
+            enum "administrator" {
+                description "Grants administrative privileges (e.g., member of sudo, docker groups).";
+            }
+            enum "operator" {
+                description "Grants operator-level (read-only or limited) privileges.";
+            }
+        }
+        description "User role that determines group memberships, privileges, and applicable security policies.";
+    }
+
+    // Top-level container for the User feature
+    container sonic-user {
+        description "Top-level container for local user management configuration";
+
+        list LOCAL_USER_LIST {
+            key "username";
+            description "List of declaratively managed local users.";
+
+            must "count(../LOCAL_USER_LIST[role='administrator' and (not(enabled) or enabled='true')]) >= 1" {
+                error-message "At least one administrator user must remain enabled.";
+            }
+
+            leaf username {
+                type string {
+                    pattern '[a-z_][a-z0-9_-]*[$]?' {
+                        error-message "Invalid username. Must start with a lowercase letter or underscore, followed by lowercase letters, numbers, underscores, or hyphens.";
+                    }
+                    must ". != 'root'" {
+                        error-message "Username cannot be 'root'.";
+                    };
+                    length 1..32;
+                }
+                description "The username for the local account.";
+            }
+
+            leaf role {
+                type user-role;
+                mandatory true;
+                description "The role assigned to the user, which determines their group memberships and privileges.";
+            }
+
+            leaf password_hash {
+                type string;
+                default "!";
+                description "The hashed password string for the user, as found in /etc/shadow. Defaults to '!' (no password). Password hashes can be generated using 'mkpasswd' utility or programmatically using libraries like 'passlib'. To disable an account, use the 'enabled' attribute instead of prepending '!' to the password hash.";
+            }
+
+            leaf-list ssh_keys {
+                type string;
+                description "A list of full public SSH key strings.";
+            }
+
+            leaf enabled {
+                type boolean;
+                default true;
+                description "Whether the user account is enabled. When false, the user shell is set to /usr/sbin/nologin to completely disable the account.";
+            }
+
+            must "not(starts-with(password_hash, '!')) or count(ssh_keys) > 0" {
+                error-message "User must have either a valid password hash (not starting with '!') or at least one SSH key.";
+            }
+        }
+
+        list LOCAL_ROLE_SECURITY_POLICY_LIST {
+            key "role";
+            description "Global security policies applied to users based on their role.";
+
+            leaf role {
+                type user-role;
+                description "The role for which this security policy applies.";
+            }
+
+            leaf max_login_attempts {
+                type uint32 {
+                    range "1..1000";
+                }
+                description "Maximum number of failed login attempts before accounts with this role are locked. If not set, system defaults apply.";
+            }
+        }
+    }
+}
+```
+
+### 8.3 CLI Enhancements
+The CLI is enhanced with user management and global security policy commands.
+
+**Feature Control Commands:**
+```
+sudo config user feature enabled
+sudo config user feature disabled
+```
+* Enable or disable the declarative local user management feature
+* **Important:** Before enabling, users must first run `config user import-existing` to import existing users into CONFIG_DB
+* When disabling, the feature stops managing users but doesn't remove existing Linux users
+
+**Migration Commands:**
+```
+config user import-existing [--dry-run] [--uid-range <min>-<max>]
+```
+* Import existing Linux users into CONFIG_DB LOCAL_USER table
+* `--dry-run`: Show what would be imported without making changes
+* `--uid-range`: Specify UID range to import (default: 1000-60000)
+* Automatically detects user roles based on group membership:
+  - Users in `sudo`, `docker`, `redis`, or `admin` groups → `administrator` role
+  - Other users → `operator` role
+* Preserves existing password hashes from `/etc/shadow`
+* Skips system users (UID < 1000) and users already in CONFIG_DB
+
+**User Add Command:**
+```
+config user add <username> --role <role>
+                          [--password-hash <hash> | --password-prompt]
+                          [--ssh-key <key>]
+                          [--disabled]
+```
+* Multiple `--ssh-key` flags can be provided to build a list.
+* Use `--password-hash` to provide a pre-hashed password directly.
+* Use `--password-prompt` to enter the password securely through an interactive prompt (password will be hashed automatically).
+* Use `--disabled` to create the account in disabled state. Accounts are enabled by default.
+* If neither password option is provided, the user account will be created with password login disabled (password set to `!`).
+
+**User Delete Command:**
+```
+config user del <username>
+```
+
+**User Modify Commands:**
+
+**Basic Modify:**
+```
+config user modify <username> [--password-hash <hash> | --password-prompt | --remove-password]
+                              [--enabled | --disabled]
+```
+
+**SSH Key Management:**
+```
+config user modify <username> --add-ssh-key <key>
+config user modify <username> --remove-ssh-key <key_identifier>
+config user modify <username> --clear-ssh-keys
+```
+
+**Options:**
+* `--password-hash <hash>`: Set a pre-hashed password directly
+* `--password-prompt`: Enter password securely through interactive prompt (hashed automatically)
+* `--remove-password`: Remove password authentication (sets password_hash to "!")
+* `--enabled` / `--disabled`: Change account status
+* `--add-ssh-key <key>`: Add a new SSH public key to the user
+* `--remove-ssh-key <key_identifier>`: Remove specific SSH key (by key comment/identifier or partial match)
+* `--clear-ssh-keys`: Remove all SSH keys from the user
+
+**Examples:**
+```bash
+# Change password interactively
+config user modify alice --password-prompt
+
+# Remove password authentication (SSH keys only)
+config user modify alice --remove-password
+
+# Add an SSH key
+config user modify alice --add-ssh-key "ssh-rsa AAAAB3... alice@workstation"
+
+# Remove specific SSH key by comment
+config user modify alice --remove-ssh-key "alice@workstation"
+
+# Remove specific SSH key by partial key match
+config user modify alice --remove-ssh-key "AAAAB3NzaC1yc2E"
+
+# Clear all SSH keys
+config user modify alice --clear-ssh-keys
+
+# Disable user account
+config user modify alice --disabled
+```
+
+**Validation:**
+* Users must retain at least one authentication method (password or SSH key)
+* Removing the last authentication method will fail with an error
+* The `--remove-password` and `--clear-ssh-keys` commands cannot be used together unless adding a new authentication method in the same command
+
+**Security Policy Commands:**
+```
+config user security-policy <role> --max-login-attempts <count>
+```
+* Configure global login attempt limits for a specific role (`administrator` or `operator`).
+
+```
+show user security-policy [<role>]
+```
+* Display current security policies for all roles or a specific role.
+
+**Password Prompt Implementation:**
+When `--password-prompt` is used, the CLI will:
+1. Display a secure password prompt (e.g., "Enter password for user <username>:")
+2. Hide password input (no echo to terminal)
+3. Prompt for password confirmation (e.g., "Confirm password:")
+4. Validate that both entries match
+5. Hash the password using the same algorithm as `mkpasswd` (e.g., Yescrypt on Debian 12)
+6. Store only the hashed password in CONFIG_DB
+7. Clear the plaintext password from memory immediately after hashing
+
+**Security Requirements:**
+- Passwords must never appear in bash history, syslog, auth logs, or any log files
+- All password handling must use secure memory practices
+
+**Show Commands:**
+```
+show user [<username>]                    # Shows usernames and roles only
+sudo show user [<username>]               # Shows usernames, roles, and password hashes
+show user security-policy [<role>]        # Shows security policies
+```
+* **Security Note:** Password hashes are only displayed when commands are run with `sudo` privileges to prevent unauthorized access to hash values.
+
+## 9. Testing Requirements
+### New System Test cases
+* **Feature Enablement/Disablement:**
+  * Verify the feature is disabled by default and `userd` daemon is not running.
+  * Enable the feature and verify `userd` daemon starts and imports existing users.
+  * Create local users manually, enable the feature, and verify they are imported into CONFIG_DB.
+  * Disable the feature and verify `userd` daemon stops but existing Linux users remain unchanged.
+  * Re-enable the feature and verify it resumes management of previously configured users.
+
+* **Migration and Import Testing:**
+  * Create test users with `useradd` and `usermod` commands with different group memberships.
+  * Run `config user import-existing --dry-run` and verify correct role assignment preview.
+  * Run `config user import-existing` and verify users are correctly imported into CONFIG_DB.
+  * Verify administrator users (in sudo/admin groups) get `administrator` role.
+  * Verify other users get `operator` role.
+  * Test UID range filtering with `--uid-range` parameter.
+  * Verify system users (UID < 1000) are not imported.
+  * Test import of SSH keys from `~/.ssh/authorized_keys` files.
+  * Verify users already in CONFIG_DB are skipped during import.
+
+* **Global Login Attempts Policy:**
+  * Configure `administrator` role with `max_login_attempts` of 5 and `operator` role with 3.
+  * Create users with both roles and verify that login attempt limits are enforced according to their role.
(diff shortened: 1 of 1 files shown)


doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,344 @@
...
+2.  **Core User Files:** `/etc/passwd`, `/etc/shadow`, `/etc/group` for basic user identity.
+3.  **PAM Configuration (`/etc/security/faillock.conf`):** To manage global failed login attempt policies per role via `pam_faillock`.
+
+## 6. High-Level Design
+
+### 6.1 `userd` Daemon
+The `userd` daemon's logic will be expanded to manage security configurations idempotently.
+
+**Startup Consistency Check:**
+* **System Reconciliation:** On startup, `userd` will perform a consistency check to ensure that all local users in the Linux system (`/etc/passwd`, `/etc/shadow`, `/etc/group`) match the definitions in CONFIG_DB.
```
bluecmd, line 69: What will the home directory be for users defined via CONFIG_DB? What will the shell be set to?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,344 @@
...
+This feature introduces a new dedicated daemon, **`userd`**, which manages the full lifecycle of local users based on definitions in `CONFIG_DB`. It simplifies management by providing a user-friendly CLI, mapping abstract roles (`administrator`, `operator`) to specific Linux groups, and enforcing security policies. This provides a solution for managing secure, persistent local user accounts.
+
+## 4. Requirements
+### 4.1 Functional Requirements
+1.  The system must allow an administrator to define a local user account declaratively.
+2.  The user definition must support:
+    * Username and a pre-hashed password.
+    * A role, limited to either **`administrator`** or **`operator`**.
+    * Authorized SSH keys (statically defined).
+3.  The system will **auto-generate** a unique UID for each new user.
```
bluecmd: Will adding a removing the same user end up using the same UID (e.g. some kind of hash?)? Will the same user have the same UID across multiple SONiC installations, or will it be sequential?

manoharan-nexthop: > Will adding a removing the same user end up using the same UID (e.g. some kind of hash?)

Behavior will be same as current Linux behavior. When the user is removed and re-added, it will get a new user ID.

> Will the same user have the same UID across multiple SONiC installations, or will it be sequential?

It will be sequential based on the Local device prior assignments.

doc/User_Management/User_Management.md
manoharan-nexthop: the directory will be `/home/<username>/` and the shell will be set to `/bin/bash`.

thanks for bringing it up, will update these information in the document.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+    * `administrator`: members of `sudo`, `docker`, `redis` and `admin` groups.
+    * `operator`: members of a standard, non-privileged and would belong to `users` group.
+4.  The system will **auto-generate** a unique UID for each new user.
+5.  There should be atleast one user configured in the role of `administrator`
+6.  **Default Admin User:**
+    * The **`admin`** user (or the user specified during compilation) must be included in the `init_cfg.json` with default settings.
+    * This ensures the admin user is present in CONFIG_DB by default and can be managed like any other user.
+    * The admin user can be modified or disabled or deleted through standard CONFIG_DB operations.
+7.  **Security Policy Requirements:**
+    * **Login Attempts:** The system must support configuring a global maximum number of failed login attempts per role before accounts are temporarily locked.
```
ymd-arista, line 46: Can you explain a bit more about what does `locked` mean here?

Does it mean that we move the `enabled` attribute for the user from True -> False. If this is the case, is this the only reason i.e to enable locking and unlocking why we keep disabled users in the DB and not delete them?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+|:------------:|:--------------------------------------------------------------------------|
+| **`userd`** | The new User Daemon proposed in this design.                              |
+| **PAM** | Pluggable Authentication Modules                                          |
+
+
+## 3. Overview
+This feature introduces a new dedicated daemon, **`userd`**, which manages the full lifecycle of local users based on definitions in `CONFIG_DB`. It simplifies management by providing a user-friendly CLI, mapping abstract roles (`administrator`, `operator`) to specific Linux groups, and enforcing security policies. This provides a solution for managing secure, persistent local user accounts.
+
+## 4. Requirements
+### 4.1 Functional Requirements
```
madhupalu, line 29: What language used to write UserD ? >> If it is python, move it to c++ or go to address memory concerns.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+* **Default Admin User:**
+  * Verify the `admin` user exists by default in `init_cfg.json` and is present in CONFIG_DB after system initialization.
+  * Verify the `admin` user can be modified, disabled, or deleted through standard CONFIG_DB operations.
+  * Test that the `admin` user has the configured username/password from build-time environment variables and administrator role when first created from init config.
+
+* **Startup Consistency Check:**
+  * Create a user directly in Linux using `useradd` (bypassing CONFIG_DB).
+  * Restart the `userd` daemon and verify the manually created user is automatically removed.
+  * Verify that users defined in CONFIG_DB are preserved and properly configured.
+  * Ensure system users (root, daemon, etc.) are not affected by the cleanup process.
```
madhupalu, line 516: Did you enforce any limits to the number of SSH keys stored?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+3.  The system must map roles to underlying Linux groups:
+    * `administrator`: members of `sudo`, `docker`, `redis` and `admin` groups.
+    * `operator`: members of a standard, non-privileged and would belong to `users` group.
+4.  The system will **auto-generate** a unique UID for each new user.
+5.  There should be atleast one user configured in the role of `administrator`
+6.  **Default Admin User:**
+    * The **`admin`** user (or the user specified during compilation) must be included in the `init_cfg.json` with default settings.
+    * This ensures the admin user is present in CONFIG_DB by default and can be managed like any other user.
+    * The admin user can be modified or disabled or deleted through standard CONFIG_DB operations.
+7.  **Security Policy Requirements:**
```
madhupalu, line 45: Do you support any user statistics about security policy violation ex: login attempts. Advice to have user statistics stored for telemetry needs.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+
+## 6. High-Level Design
+
+### 6.1 `userd` Daemon
+The `userd` daemon's logic will be expanded to manage security configurations idempotently.
+
+**Startup Consistency Check:**
+* **System Reconciliation:** On startup, `userd` will perform a consistency check to ensure that all local users in the Linux system (`/etc/passwd`, `/etc/shadow`, `/etc/group`) match the definitions in CONFIG_DB.
+* **Cleanup of Unmanaged Users:** Any users found in the Linux system that are not defined in CONFIG_DB (except for system users like `root`, `daemon`, etc.) will be automatically removed to maintain consistency.
+* **CONFIG_DB as Source of Truth:** This ensures that CONFIG_DB remains the single source of truth for user management and prevents configuration drift.
```
madhupalu, line 71: How does this new feature affect existing tacacs+/ radius user management?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+
+**`userd`'s Points of Interaction:**
+1.  **CONFIG_DB:** Single source of truth for user configuration and global security policies.
+2.  **Core User Files:** `/etc/passwd`, `/etc/shadow`, `/etc/group` for basic user identity.
+3.  **PAM Configuration (`/etc/security/faillock.conf`):** To manage global failed login attempt policies per role via `pam_faillock`.
+
+## 6. High-Level Design
+
+### 6.1 `userd` Daemon
+The `userd` daemon's logic will be expanded to manage security configurations idempotently.
```
madhupalu, line 66: What language used to write UserD ? >> If it is python, move it to c++ or go to address memory concerns.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+7.  **Security Policy Requirements:**
+    * **Login Attempts:** The system must support configuring a global maximum number of failed login attempts per role before accounts are temporarily locked.
+8.  User accounts and their configurations must persist across system reboots and upgrades.
+9.  **System Consistency:** On startup, the system must perform a consistency check to ensure Linux users match CONFIG_DB definitions and automatically remove any users that were added directly to Linux (bypassing CONFIG_DB).
+
+## 5. Architecture Design
+The architecture centers on the new `userd` daemon. This daemon will now interact with several core Linux subsystems to enforce the configured security policies.
+
+**`userd`'s Points of Interaction:**
+1.  **CONFIG_DB:** Single source of truth for user configuration and global security policies.
```
madhupalu, line 59: Is this feature controlled by the feature flag? >> right now it is controlled by init.cfg

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+2.  **Core User Files:** `/etc/passwd`, `/etc/shadow`, `/etc/group` for basic user identity.
+3.  **PAM Configuration (`/etc/security/faillock.conf`):** To manage global failed login attempt policies per role via `pam_faillock`.
```
Yarden-Z: Why is this per role and not per UID?

doc/User_Management/User_Management.md
Yarden-Z: Why is this required? How is it different from hostfgd in this case?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+
+## 6. High-Level Design
+
+### 6.1 `userd` Daemon
+The `userd` daemon's logic will be expanded to manage security configurations idempotently.
+
+**Startup Consistency Check:**
+* **System Reconciliation:** On startup, `userd` will perform a consistency check to ensure that all local users in the Linux system (`/etc/passwd`, `/etc/shadow`, `/etc/group`) match the definitions in CONFIG_DB.
+* **Cleanup of Unmanaged Users:** Any users found in the Linux system that are not defined in CONFIG_DB (except for system users like `root`, `daemon`, etc.) will be automatically removed to maintain consistency.
```
Yarden-Z: How will cleanup occur?
Will there be separate service going over active and configured users in the systetm and deleting those not managed by the config db?
Also - why are we enforcing user deletion when users are not defined?
In addition - which users are deleted? Only those with /bin/bash? Or also users with nologin?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+* **CONFIG_DB as Source of Truth:** This ensures that CONFIG_DB remains the single source of truth for user management and prevents configuration drift.
+
+**Default Admin User:**
+* **Init Config Integration:** The `admin` user (or the user specified during compilation) is included in `init_cfg.json` with default settings, ensuring it's always present in CONFIG_DB.
+* **Standard Management:** The admin user is managed through the same CONFIG_DB interface as other users, with no special exception handling required.
+* **Default Configuration:** The admin user in `init_cfg.json` will use the username and password hash from the `USERNAME` and `PASSWORD` environment variables during build, with fallback defaults if not specified.
+
+NOTE: If there are no `administrator` role users defined in `CONFIG_DB`, `userd` will create an `admin` user with default settings if the `admin` user does not exist, otherwise will ensure the user account is enabled.
```
Yarden-Z: Why not populate config_db with a default user at all times?
And if defined otherwise in init_fcg.json - populate it with different values?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+
+**Configuration File Hierarchy:**
+* **`init_cfg.json`:** Contains base system defaults including the default admin user. This file is generated at build-time from a Jinja2 template and is always present. It ensures every SONiC system has essential user management configuration.
+* **`golden_config_db.json`:** Optional site-specific configuration file that can override defaults when used with `config load_minigraph --override_config`. This allows deployments to customize user configurations without modifying the base system defaults.
+* **Configuration Precedence:** When both files are present, `golden_config_db.json` takes precedence over `init_cfg.json` during minigraph reload operations with override enabled.
+
+**User Account Management:**
+* **Account Status Control:** `userd` will monitor the `enabled` attribute for each user. When `enabled` is `false`, it will prepend `!` to the password hash in `/etc/shadow` to disable password-based login. When `enabled` is `true`, it will restore the original password hash.
```
Yarden-Z: What is the goal of this functionality?
If we want to disable password-based login, we can have different knobs in ssh-server and other entry points rather than altering the /etc/shadow file.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+* **Password Hash Preservation:** When disabling a user, `userd` will store the original password hash and restore it when the user is re-enabled, ensuring seamless account management.
+
+**New Logic for Security Policies:**
+
+* **To enforce Login Attempts:**
+    * `userd` will manage the PAM configuration file at `/etc/security/faillock.conf`.
+    * For global role-based limits (e.g., administrators with limit of 5), it will ensure appropriate configuration in the global section.
+    * `userd` will configure PAM to apply different limits based on user group membership (administrator vs operator roles).
```
Yarden-Z: Why are we enforcing different policies dependent on user role?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+The `USER` table schema is defined for individual user accounts, and a new `USER_SECURITY_POLICY` table is added for global role-based security policies.
+
+**Schema:**
+```json
+// Example for CONFIG_DB
+{
+    "USER": {
+        "admin": {
+            "role": "administrator",
+            "password_hash": "$6$salt$hash_of_YourPaSsWoRd",
```
Yarden-Z, line 170: How are we protecting the hashed password in the DB so that it is not user readable?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+        }
+    }
+}
+```
+
+**USER Table:**
+* `role` (string, required): `administrator` or `operator`.
+* `password_hash` (string, required): The hashed password.
+* `ssh_keys` (optional): List of SSH public keys for the user.
+* `enabled` (boolean, optional): Whether the user account is enabled. Defaults to `true` if not specified.
```
Yarden-Z: Who can change the role of a user? Only and admin, right?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+
+**USER_SECURITY_POLICY Table:**
+* **`max_login_attempts`** (integer, optional): Number of failed login attempts before accounts with this role are locked.
+
+**Notes:**
+* Session timeouts are managed by the system's default timeout policy and are not configurable per user.
+* Password hashes can be generated using `mkpasswd` command-line utility or programmatically using libraries like `passlib` in Python for NETCONF/RESTCONF implementations.
+* The CLI provides both `--password-hash` (for pre-hashed passwords) and `--password-prompt` (for interactive secure password entry) options for improved security and usability.
```
Yarden-Z, line 224: We need to ascertain that any password entries are obfuscated in the logs to prevent disclosure.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+* **CONFIG_DB as Source of Truth:** On startup, `userd` performs a consistency check and removes any users that were added directly to Linux (bypassing CONFIG_DB) to ensure CONFIG_DB remains the authoritative source for user management.
```
Yarden-Z, line 225: Is there an option that this service removes the last remaining admin user on top of the system?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+* **User Account Status:** When `enabled` is set to `false`, `userd` will prepend `!` to the password hash in `/etc/shadow` to disable password-based login while preserving SSH key access. When `enabled` is `true` (default), the original password hash is restored.
```
Yarden-Z: Why don't we change the login from /bin/bash to nologin - should be less destructive and more informative

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+```
+* Display current security policies for all roles or a specific role.
+
+**Password Prompt Implementation:**
+When `--password-prompt` is used, the CLI will:
+1. Display a secure password prompt (e.g., "Enter password for user <username>:")
+2. Hide password input (no echo to terminal)
+3. Prompt for password confirmation (e.g., "Confirm password:")
+4. Validate that both entries match
+5. Hash the password using the same algorithm as `mkpasswd` (e.g., SHA-512)
```
Yarden-Z: Debian 12 base hash function is not sha512 but rather Yescrypt

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+6. Store only the hashed password in CONFIG_DB
+7. Clear the plaintext password from memory immediately after hashing
```
Yarden-Z, line 446: Not only memory - also logs (syslog, auth log, potentially others

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+* `password_hash` (string, required): The hashed password.
+* `ssh_keys` (optional): List of SSH public keys for the user.
+* `enabled` (boolean, optional): Whether the user account is enabled. Defaults to `true` if not specified.
+
+**USER_SECURITY_POLICY Table:**
+* **`max_login_attempts`** (integer, optional): Number of failed login attempts before accounts with this role are locked.
+
+**Notes:**
+* Session timeouts are managed by the system's default timeout policy and are not configurable per user.
+* Password hashes can be generated using `mkpasswd` command-line utility or programmatically using libraries like `passlib` in Python for NETCONF/RESTCONF implementations.
```
qiluo-msft: >Python [](http://example.com/codeflow?start=127&length=6)

Could you use Rust as implementation programming language for this feature? some benefit:
1. reduce overall memory usage
2. high performance library, for example, hash related library.
 #Closed

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+    * Username and a pre-hashed password.
+    * A role, limited to either **`administrator`** or **`operator`**.
+    * Authorized SSH keys (statically defined).
+3.  The system must map roles to underlying Linux groups:
+    * `administrator`: members of `sudo`, `docker`, `redis` and `admin` groups.
+    * `operator`: members of a standard, non-privileged and would belong to `users` group.
+4.  The system will **auto-generate** a unique UID for each new user.
+5.  There should be atleast one user configured in the role of `administrator`
+6.  **Default Admin User:**
+    * The **`admin`** user (or the user specified during compilation) must be included in the `init_cfg.json` with default settings.
```
qiluo-msft: >init_cfg.json [](http://example.com/codeflow?start=95&length=13)

Could you keep init_cfg.json unchanged for this design? some general suggestions:
1. the image default credential is already in /etc/shadow (admin user).
2. the "Declarative Local User Management" feature should be disabled by default
3. If operator enable this feature, this feature will automatically move/copy username and their hashes in /etc/shadow into ConfigDB.
4. if operator disable this feature, this feature will write ConfigDB back into /etc/passwd and /etc/shadow.
5. Can we limit that there is only 1 admin user called "admin"?

doc/User_Management/User_Management.md
qiluo-msft: >pam_faillock [](http://example.com/codeflow?start=120&length=12)

What is the difference between this file vs /etc/ssh/sshd_config MaxAuthTries field?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+
+## 8. Configuration and Management
+### 8.1 Config DB Enhancements
+The `USER` table schema is defined for individual user accounts, and a new `USER_SECURITY_POLICY` table is added for global role-based security policies.
+
+**Schema:**
+```json
+// Example for CONFIG_DB
+{
+    "USER": {
```
qiluo-msft: >USER [](http://example.com/codeflow?start=5&length=4)

USER -> LOCAL_USER

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+            "enabled": true
+        },
+        "showuser": {
+            "role": "operator",
+            "password_hash": "hashed_password",
+            "ssh_keys": ["ssh-rsa CCC..."],
+            "enabled": false
+        }
+    },
+    "USER_SECURITY_POLICY": {
```
qiluo-msft: >USER_SECURITY_POLICY [](http://example.com/codeflow?start=5&length=20)

USER_SECURITY_POLICY -> LOCAL_ROLE_SECURITY_POLICY

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,347 @@
...
+* If neither password option is provided, the user account will be created with password login disabled (password set to `!`).
+
+**User Delete Command:**
+```
+config user del <username>
+```
+
+**User Modify Command:**
+```
+config user modify <username> [--password-hash <hash> | --password-prompt]
```
qiluo-msft: >config user modify [](http://example.com/codeflow?start=0&length=18)

Currently the commands should be "passwd" or "chpasswd". If you invent new command lines, it will be easy to mix use.

It is possible to align with original commands?

aidan-gallagher:
Hi @manoharan-nexthop, thanks for putting this together - great to see local login being discussed as that's a feature my organisation need.

Just a heads-up that I’ve done some work for local-login / ssh-login implementation as well.

**I’ve pushed up my work here in case it’s helpful:  https://github.com/aidan-gallagher/SONiC/pull/1**

It’s a working implementation, though I hadn’t fully planned to submit it as a PR yet. Feel free to take a look or reference anything if it’s useful. Happy to help out if you like.

doc/User_Management/User_Management.md
manoharan-nexthop: this has been updated. thanks.

doc/User_Management/User_Management.md
manoharan-nexthop: Added a note that, will use C++ for this daemon.

doc/User_Management/User_Management.md
manoharan-nexthop: thanks for the review.

The intention was to allow more than one administrator. The reason to have `init_cfg.json` was that, if passwords were changed and we do a config reload with default configuration go back to the original state, we need the original password. Thats the reason to write in init_cfg.json.

* Added a feature knob and have that as disabled by default, so that no process will get spawned and would work as it is today.
* Added the mightration support if there are existing users that are to be transfered to CONFIG_DB.

doc/User_Management/User_Management.md
manoharan-nexthop: Have added a future enhancements for this. Currently we have to rely on linux logging and we can expand this in the next iteration.

doc/User_Management/User_Management.md
manoharan-nexthop: Its a temorary lockout that Linux does. If there are consecutive login attempts, the account is temporarily locked for 15 minutes before one could attempt again.

The config `enabled`, is to manually control if the user need to be disabled.

doc/User_Management/User_Management.md
manoharan-nexthop: Yes, have added a note for the same. thanks.

doc/User_Management/User_Management.md
manoharan-nexthop: We didnt see a need for per UID.. Incase in future, if we do see, we can improvise to have per UID and have this as default.

doc/User_Management/User_Management.md
manoharan-nexthop: The reason for not using existing hostcfgd is that, with python we could make mistakes easily where it throws an exception and exits.. If that happens, we could even endup in a state where we could not login. Having it separate daemon reduces the of other things in hostcfgd affecting this feature.

doc/User_Management/User_Management.md
manoharan-nexthop: Have updated the document on this..

doc/User_Management/User_Management.md
manoharan-nexthop: The cleanup will be done by userd on its startup. If the administrator wishes to directly manage the users in Linux, the feature could just be disabled. Else it becomes difficult to manage on overlap, modifications as we would miss that CONFIG_DB not having the true information anymore.

The users >1000 will only be managed by userd.

doc/User_Management/User_Management.md
manoharan-nexthop: It should work together as it does today (for the users that are locally created in Linux).

doc/User_Management/User_Management.md
manoharan-nexthop: The intention was, suppose an operator will not be available for a month, for security we could disable and re-enable when the operator becomes available. Its a value add by providing an easy control.

doc/User_Management/User_Management.md
manoharan-nexthop: For an `administrator` we would need a stricter control than the `operator` since `administrator` can disrupt the system. Hence the policy is based on role.

doc/User_Management/User_Management.md
manoharan-nexthop: The hashed password is non reversible, so that should give the protection already.

doc/User_Management/User_Management.md
manoharan-nexthop: yes, only admin can change the role.

doc/User_Management/User_Management.md
manoharan-nexthop: true, they will not be logged not even in console, hence the reason to make it interactive.

doc/User_Management/User_Management.md
manoharan-nexthop: We need atleast one administrator and enforced in the yang validation.

doc/User_Management/User_Management.md
manoharan-nexthop: since this is an optional feature, when it is disabled, we didnt want anything to be affected. This is based on the click CLI and we wanted to update the config_db.

If we are using the feature, we have everything directly here.. If not, we just use what we have in Linux.

doc/User_Management/User_Management.md
manoharan-nexthop: true, it will be logged as well.

doc/User_Management/User_Management.md
manoharan-nexthop: Have updated to use nologin. Thanks.

manoharan-nexthop:
> Hi @manoharan-nexthop, thanks for putting this together - great to see local login being discussed as that's a feature my organisation need.
> 
> Just a heads-up that I’ve done some work for local-login / ssh-login implementation as well.
> 
> **I’ve pushed up my work here in case it’s helpful: [aidan-gallagher#1](https://github.com/aidan-gallagher/SONiC/pull/1)**
> 
> It’s a working implementation, though I hadn’t fully planned to submit it as a PR yet. Feel free to take a look or reference anything if it’s useful. Happy to help out if you like.

Thank you @aidan-gallagher , I will go through your change set as well.. would be happy to collaborate.

doc/User_Management/User_Management.md
Yarden-Z: They might be logged via auth log, syslog or bash history.
I think this should be reviewed and obfuscated

manoharan-nexthop: Have added an explicit note for the same, thanks.

doc/User_Management/User_Management.md
bluecmd: If an unauthorized user can read out hashes they can crack the password offline, bypassing any form of local brute-force protection.

doc/User_Management/User_Management.md
bluecmd: Will the old home directories of deleted users be recursively deleted or will they remain?

doc/User_Management/User_Management.md
Yarden-Z: I think as a whole - the use case for this is per UID.
If we want the faillock - it should be per account distinguishable by UID (to prevent brute forcing)

doc/User_Management/User_Management.md
Yarden-Z: Is this a real use-case?
I understand a no-login for a user, but user accounts should be accessible at all times.
This will potentially require that user1 (which could potentially be an administrator) disable user 2, and when user2 wants to get back to the switch - user1 will be required to re-enable it.
Is this is a real scenario?

doc/User_Management/User_Management.md
Yarden-Z: I think we should have a unified approach for all users.
Since this is a basic brute forcing prevention functionality.
I Understand that the impact of an admin user being compromised is greater than an operator - but both achieve a foothold.
One is effectively root and the other requires additional privilege escalation to acquire root level access.

doc/User_Management/User_Management.md
Yarden-Z: And basically bypassing the faillock functionality

aidan-gallagher: In my [implementation of this feature](https://github.com/aidan-gallagher/SONiC/pull/1) I had it so the password hashes were only visible if the commands were invoked with `sudo`.

For example
```
$ show local-login
+------------+
| Username   |
+============+
| admin      |
+------------+
| aidan      |
+------------+
```

```
$ sudo show local-login
+------------+---------------------------------------------------------------------------+
| Username   | Password Hash                                                             |
+============+===========================================================================+
| admin      | $y$j9T$NtunjoczrFjJILXbKVQ/0.$YDM7E6dfbD/OlKUlo4qoRBs2kCd6.OusrkYMQNs38g3 |
+------------+---------------------------------------------------------------------------+
| aidan      | $y$j9T$WHVZxD7z2gvHHLNfFZuXH0$8frMVuUfFH.WXAp0ldKhN/YUPctCgf3aB1UkyfF5/j5 |
+------------+---------------------------------------------------------------------------+
```

This same idea was extended to `show runningconfiguration all`. 
`show runningconfiguration all` would not show sensitive information such as encrypted passwords. `sudo show runningconfiguration all` would show all config including hashed passwords.

File permissions of `/etc/sonic/config_db.json` were modified so they were only readable by root to prevent normal users bypassing the show commands.

aidan-gallagher:
> > Hi @manoharan-nexthop, thanks for putting this together - great to see local login being discussed as that's a feature my organisation need.
> > Just a heads-up that I’ve done some work for local-login / ssh-login implementation as well.
> > **I’ve pushed up my work here in case it’s helpful: [aidan-gallagher#1](https://github.com/aidan-gallagher/SONiC/pull/1)**
> > It’s a working implementation, though I hadn’t fully planned to submit it as a PR yet. Feel free to take a look or reference anything if it’s useful. Happy to help out if you like.
> 
> Thank you @aidan-gallagher , I will go through your change set as well.. would be happy to collaborate.

I have uploaded a built VS image from my change set in case you want to play about with it:
https://pub-b38eb23466d442f88f85da770f4feea7.r2.dev/sonic-vs.img.gz

vitaliibylinka:
@manoharan-nexthop - Do you have some draft implementation or just HLD changes? If you have any specific details or examples, please share them.

doc/User_Management/User_Management.md
manoharan-nexthop: I have updated the section to add safeguard against operators learning the password hash. Thanks.

doc/User_Management/User_Management.md
manoharan-nexthop: Yes, the administrator has to manually delete to avoid loosing the data.

doc/User_Management/User_Management.md
manoharan-nexthop: By default the account is enabled. This is only used when the one wants to disable. If they dont want to disable, one could just not change the configuration for that user.

doc/User_Management/User_Management.md
manoharan-nexthop: Would it be okay, if I set the same default value so that if there is a need for different there is a flexibility to choose for that.

doc/User_Management/User_Management.md
manoharan-nexthop: Have updated the section to use C++ instead of Python. Thanks.

doc/User_Management/User_Management.md
manoharan-nexthop: Done.

doc/User_Management/User_Management.md
manoharan-nexthop: I have updated the wordings. Say oper1 user fails configured times, only that user is locked out not others.. (added the clarification for same)

doc/User_Management/User_Management.md
manoharan-nexthop: Thanks, have reworded this.

manoharan-nexthop:
> @manoharan-nexthop - Do you have some draft implementation or just HLD changes? If you have any specific details or examples, please share them.

@vitaliibylinka , yes I have the draft changes, have to include the changes for the comments and updates to the document. Will share it by tomorrow.

doc/User_Management/User_Management.md
aidan-gallagher: What happens if
1) user bob is created with home directory /home/bob.
2) user bob is deleted and /home/bob remains.
3) user bob is created again. What is bob's home directory? 


You need to be careful because I found an edge case when implementing this. `useradd  --create-home` will succeed with warning but the system fails because bob has a different UID than the old bob who owns /home/bob. Therefore the new user gets permission denied error when trying to write to their home directory.

```
admin@sonic:~$ sudo useradd --create-home bob
useradd: warning: the home directory /home/bob already exists.
useradd: Not copying any file from skel directory into it.
admin@sonic:~$ echo $?
0
```

manoharan-nexthop: we can add a note so that user are aware of this.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,484 @@
...
+* **Standard Management:** Once imported, the admin user is managed through the same CONFIG_DB interface as other users.
+
+**Configuration File Hierarchy:**
+* **`init_cfg.json`:** Contains the feature flag `LOCAL_USER_MANAGEMENT` set to `disabled` by default AND includes the default admin user in LOCAL_USER table. This ensures the admin user is always present in CONFIG_DB and can be managed consistently.
+* **`golden_config_db.json`:** Optional site-specific configuration file that can enable the feature and provide initial user configurations when used with `config load_minigraph --override_config`.
+* **Configuration Precedence:** When both files are present, `golden_config_db.json` takes precedence over `init_cfg.json` during minigraph reload operations with override enabled.
+
+**User Account Management:**
+* **Account Status Control:** `userd` will monitor the `enabled` attribute for each user. When `enabled` is `false`, it will change the user's shell to `/usr/sbin/nologin` to completely disable the account. When `enabled` is `true`, it will set the shell to `/bin/bash`.
```
aidan-gallagher, line 100: This won't completely disable the account.

* This won't kick the user out if they are currently locked in.
* It will still allow the disabled user's existing cron tasks to run.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,484 @@
...
+    }
+}
+```
+
+**FEATURE Table (for LOCAL_USER_MANAGEMENT):**
+* `state` (string, required): `enabled` or `disabled`. Controls whether the declarative local user management feature is active. Defaults to `disabled`.
+
+**LOCAL_USER Table (only active when feature is enabled):**
+* `role` (string, required): `administrator` or `operator`.
+* `password_hash` (string, required): The hashed password.
```
aidan-gallagher: Why is the password_hash required?

It is valid that users may only want to configure their ssh_key to login and not a password.

manoharan-nexthop: Changed the validation to have atleast one of ssh key or password.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,484 @@
...
+* **Administrator Availability:** At least one administrator user must remain enabled at all times to prevent complete loss of administrative access. This constraint is enforced by the Yang model.
+* **Default Admin User:** The `admin` user (or the user specified during compilation) is always included in `init_cfg.json` with default settings, ensuring it's consistently present in CONFIG_DB regardless of feature enablement status. When the feature is disabled, the admin user exists in CONFIG_DB but is not actively managed by `userd`. When enabled, `userd` takes over management of the existing admin user.
+* **TACACS+/RADIUS Compatibility:** This local user management feature operates independently of and is fully compatible with existing TACACS+ and RADIUS authentication systems. Local users are managed through standard Linux PAM authentication stack, which can coexist with remote authentication methods. The authentication order and fallback behavior are controlled by existing AAA configuration in CONFIG_DB.
```
aidan-gallagher, line 232: You need to be careful that this feature doesn't delete tacas/radius users.

In my [implementation](https://github.com/aidan-gallagher/sonic-buildimage/pull/1/files#diff-b7da562b77c93aea4e412573999095e366397fceb1bdd78cd240e0b0bb0b0fa1R46) I had the `is_remote_user`  function that looked for remote_user in the gecos field.

I believe this works for both tacas and radius as they add that field when creating users. See below files:
* [sonic-buildimage/src/radius/nss/libnss-radius/nss_radius_common.c](https://github.com/sonic-net/sonic-buildimage/blob/master/src/radius/nss/libnss-radius/nss_radius_common.c#L161) 
* [sonic-buildimage/src/tacacs/nss/patch/0001-Modify-user-map-profile.patch](https://github.com/sonic-net/sonic-buildimage/blob/master/src/tacacs/nss/patch/0001-Modify-user-map-profile.patch#L575)

I still had to determine is ldap does this. Or make the necessary changes to LDAP do it does.

manoharan-nexthop: The user management / cleanup is done based on the /etc/shadow file and the tacacs/radius created users wont exist in this file and these features would work seemless.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,484 @@
...
+* If neither password option is provided, the user account will be created with password login disabled (password set to `!`).
+
+**User Delete Command:**
+```
+config user del <username>
+```
+
+**User Modify Command:**
+```
+config user modify <username> [--password-hash <hash> | --password-prompt]
```
aidan-gallagher: Is it preferred to have both `config user add` and `config user modify` instead not just the one command `config user set` which covers both?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,484 @@
...
+```
+* Multiple `--ssh-key` flags can be provided to build a list.
+* Use `--password-hash` to provide a pre-hashed password directly.
+* Use `--password-prompt` to enter the password securely through an interactive prompt (password will be hashed automatically).
+* Use `--disabled` to create the account in disabled state. Accounts are enabled by default.
+* If neither password option is provided, the user account will be created with password login disabled (password set to `!`).
+
+**User Delete Command:**
+```
+config user del <username>
```
aidan-gallagher, line 371: Do we want more granular control to delete an individual ssh key or just the password?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,484 @@
...
+7. Clear the plaintext password from memory immediately after hashing
+
+**Security Requirements:**
+- Passwords must never appear in bash history, syslog, auth logs, or any log files
+- All password handling must use secure memory practices
+
+**Show Commands:**
+```
+show user [<username>]                    # Shows usernames and roles only
+sudo show user [<username>]               # Shows usernames, roles, and password hashes
```
aidan-gallagher, line 455: Which show command shows the public ssh keys - `show user` or `sudo show user`?

doc/User_Management/User_Management.md
aidan-gallagher: As far as I understand it the yang validation is checked when `config apply-patch` is used but not when other CLI commands are used i.e `config user del <username>`.

The CLI command will also have to enforce this rule.

Users could still bypass both the CLI and apply-patch enforcement by writing directly to the configuration database. We could either accept that possibility or add checks in the script which reads from the database and applies the changes.

doc/User_Management/User_Management.md
aidan-gallagher: This feature will only be hashing the password when a user is changing a password. I don't think users will be changing their password often enough to make memory or performance an issue.

doc/User_Management/User_Management.md
aidan-gallagher: I think the is proposed approach of adding new commands in sonic-utilities (`config user ...`) is preferred over modifying existing linux commands (passwd, chpasswd, useradd, userdel, etc) and making them write to configdb.

doc/User_Management/User_Management.md
aidan-gallagher: **Use hostcfgd without implementing logic in Python**

Hostcfgd doesn't have to contain any of the business logic - it can simple initiate another process which contains the logic for creating accounts, setting passwords, etc.

In my [implementation](https://github.com/aidan-gallagher/sonic-host-services/pull/1/files#diff-f46602bac3380a0d28ddfcbe698d95d87c7bc57e504fac08dd6906bdb31ec388R1735) hostcfgd simply starts the local-login.service which then has the business logic.

I implemented the business logic (local-login.py) in python however any other programming language could be used.

**Using Python**
I don't think implementing the feature in python makes it more likely to cause a problem. The feature simply needs to call Linux commands (adduser, chpasswd, etc) and write to a file. I think a high level scripting language is ideal for this rather than using C++. Mistakes in the code will cause problems whether it's implemented in Python or C++, and I think mistakes are more likely in C++ code.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,484 @@
...
+**LOCAL_USER Table (only active when feature is enabled):**
+* `role` (string, required): `administrator` or `operator`.
+* `password_hash` (string, required): The hashed password.
+* `ssh_keys` (optional): List of SSH public keys for the user. No limits are imposed on the number of SSH keys.
+* `enabled` (boolean, optional): Whether the user account is enabled. Defaults to `true` if not specified. When `false`, user shell is set to `/usr/sbin/nologin`.
+
+**LOCAL_ROLE_SECURITY_POLICY Table (only active when feature is enabled):**
+* **`max_login_attempts`** (integer, optional): Number of failed login attempts before accounts with this role are locked.
+
+**Notes:**
```
aidan-gallagher, line 211: I think it would be worth having some notes on how this interacts with the password hardening feature.

I have a section on ['Integration With Password Hardening' ](https://github.com/aidan-gallagher/SONiC/pull/1/files#diff-1b7884df6498dc944c50ed82dd942ccbe9c312af7b56f04e6ab2adbbc8d449c0)

The gist of the problem is just encrypting a password and writing it the configdb bypasses any password hardening requirements essentially making the password hardnening feature pointless.

doc/User_Management/User_Management.md
manoharan-nexthop: The intention is to avoid new logins when the user is away.

doc/User_Management/User_Management.md
manoharan-nexthop: For the CLI created password, we will ensure it matches the requirement in password hardening. When a hash is provided, it will be left to the administrator creating it. Added a note for the same.

doc/User_Management/User_Management.md
manoharan-nexthop: The current proposal similar semantics that we have in other commands.

doc/User_Management/User_Management.md
manoharan-nexthop: Modified the same. thanks.

doc/User_Management/User_Management.md
manoharan-nexthop: It will be shown only for `sudo show user`.

doc/User_Management/User_Management.md
aidan-gallagher: Are you sure the tacas users won't exist in `/etc/shadow`?

When I log into a SONiC switch with tacas user `aidan` I see the following 
```
aidan@sonic:~$ sudo cat /etc/shadow
root:*:20279:0:99999:7:::
daemon:*:20279:0:99999:7:::
...
admin:$y$j9T$6BE1U6d10j3e7fZsww4nX.$DGJaNI5p.kIqm43VzpwwNJsulAUqolwWgPNaB4hX1w7
...
aidan:!:20285:0:99999:7:::
```

```
aidan@sonic:~$ cat /etc/passwd
root:x:0:0:root:/root:/bin/bash
daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
...
admin:x:1000:1000:Default admin user,,,:/home/admin:/bin/bash
...
aidan:x:1005:1000:remote_user_su:/home/aidan:/bin/bash
```

I think you will have to also read `/etc/passwd` to look at the users UID as you need to ensure you don't delete system & service accounts with UID < 1000.

Once you have an image ready I can test this out with a device using tacas.

manoharan-nexthop: mid of this week, I should have the image and the PRs out.. (will post in this PR once ready).

doc/User_Management/User_Management.md
manoharan-nexthop: There are no limits on the ssh keys.

doc/User_Management/User_Management.md
anders-nexthop: @manoharan-nexthop can you please address:

"which users are deleted? Only those with /bin/bash? Or also users with nologin?"

doc/User_Management/User_Management.md
anders-nexthop: @manoharan-nexthop please address this comment

doc/User_Management/User_Management.md
anders-nexthop: @manoharan-nexthop please address this comment

anders-nexthop:
> @manoharan-nexthop - Do you have some draft implementation or just HLD changes? If you have any specific details or examples, please share them.

@vitaliibylinka PRs with implementation details have been posted.

doc/User_Management/User_Management.md
manoharan-nexthop: The document is updated.

doc/User_Management/User_Management.md
manoharan-nexthop: Only the local users that are managed by the SONiC (userd) will be removed. The users are added to a group `local_mgd` and those users belonging to this group alone are removed. This ensures other users (like the RADIUS or TACACS users are not removed)

doc/User_Management/User_Management.md
manoharan-nexthop: Had been updated. Missed to respond before.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,533 @@
...
+        },
+        "operator": {
+            "max_login_attempts": 3
+        }
+    }
+}
+```
+
+**FEATURE Table (for LOCAL_USER_MANAGEMENT):**
+* `state` (string, required): `enabled` or `disabled`. Controls whether the declarative local user management feature is active. Defaults to `disabled`.
```
spandan-nexthop: Did we intentionally not use boolean for the feature (`"enabled": false`) to accommodate for any other "state" that we plan to have in future?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,533 @@
...
+config feature state LOCAL_USER_MANAGEMENT enabled
+config feature state LOCAL_USER_MANAGEMENT disabled
+```
+* Enable or disable the declarative local user management feature
+* When enabling, existing local users are imported into CONFIG_DB
+* When disabling, the feature stops managing users but doesn't remove existing Linux users
+
+**Migration Commands:**
+```
+config user import-existing [--dry-run] [--uid-range <min>-<max>]
```
spandan-nexthop, line 345: Since there is no command to add just one UID, is it safe to assume that if min=max, then a single UID will get imported?

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,533 @@
...
+config user modify alice --clear-ssh-keys
+
+# Disable user account
+config user modify alice --disabled
+```
+
+**Validation:**
+* Users must retain at least one authentication method (password or SSH key)
+* Removing the last authentication method will fail with an error
+* The `--remove-password` and `--clear-ssh-keys` commands cannot be used together unless adding a new authentication method in the same command
```
spandan-nexthop, line 425: Is this allowed? Should we add a comment about it?
```
config user modify alice --remove-password
config user modify alice --clear-ssh-keys
```

spandan-nexthop (commented):
LGTM!
Just added some comments/questions.

arlakshm:
@qiluo-msft can you please help signoff on this feature?

doc/User_Management/User_Management.md
manoharan-nexthop: No, we cant remove both ssh and password. Line 424 mentions the same. `* Removing the last authentication method will fail with an error`

doc/User_Management/User_Management.md
manoharan-nexthop: thats right.

doc/User_Management/User_Management.md
manoharan-nexthop: Generally, I see both patterns being used. In other features as well this `enabled` and `disabled` gets used. Hence did the same here.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,535 @@
+# Declarative Local User Management
+
+## Revision
+
+| Rev | Date           | Author      | Change Description                                                                                                                                   |
+|:---:|:--------------:|:-----------:|:----------------------------------------------------------------------------------------------------------------------------------------------------:|
+| 1.0 | June 16, 2025  | Manoharan Sundaramoorthy   | Initial HLD                                                                                                                                          |
+
+
```
venkatmahalingam, line 9: Currently the local users that are created are not persistent across upgrades. This HLD proposes to manage the users across upgrades by storing the configuration in CONFIG_DB and a daemon to manage them in Linux.

-> For this requirement, is it not better to migrate the users from one partition to another partition during upgrade operation instead of exposing the users & password keys in the CONFIG_DB, filtering them in the tech bundle for security reasons?

manoharan-nexthop: There were few things that we had considered for this:
1. SSH keys - we would have to create the user directories and the authorized_keys to set this up. It would be possible with scripts but would be cumbersome.
2. We would not know which were local users and which were from RADIUS / TACACS.
3. We wont be able to port another device (incase device has to be replaced).

BYGX-wcr:
Password hashes are also considered sensitive data since they can be used for reverse-engineering to get the password. Normally, /etc/shadow is only readable to root-level user. Making this change in SONiC essentially expose the /etc/shadow file content to general users and breaks the security policy of Linux.

doc/User_Management/User_Management.md
```diff
@@ -0,0 +1,535 @@
...
+* **Integration with Password Hardening:** For CLI-created passwords (using `--password-prompt`), the system will validate that passwords meet any configured password hardening requirements before hashing and storing. When a pre-computed hash is provided (using `--password-hash`), it is the administrator's responsibility to ensure the original password met hardening requirements, as the hash cannot be reverse-validated.
+* **Password Security and Logging:** **IMPORTANT:** All password handling must ensure passwords are never logged in plaintext:
+  - CLI commands must not echo passwords to terminal or log them in bash history
+  - Syslog and auth logs must never contain plaintext passwords
+  - Memory containing plaintext passwords must be cleared immediately after hashing
+* **CONFIG_DB Security:** To prevent unauthorized access to password hashes:
+  - `/etc/sonic/config_db.json` file permissions changed from `644` to `640` (root:redis)
+  - Password hashes only visible in show commands when run with `sudo` privileges
+  - `show user` displays usernames only, `sudo show user` displays hashes
+  - `show runningconfiguration` redacts password_hash fields unless run with `sudo`
```
BYGX-wcr, line 223: What if the user just writes a script/process to connect to the running CONFIG_DB and read content? Right now, there is no table-level protection in redis.

aidan-gallagher:
> Password hashes are also considered sensitive data since they can be used for reverse-engineering to get the password. Normally, /etc/shadow is only readable to root-level user. Making this change in SONiC essentially expose the /etc/shadow file content to general users and breaks the security policy of Linux.

I believe the CLI show command will only show password hashes when the user is root. 
See `can_view_passwords` in https://github.com/sonic-net/sonic-utilities/pull/4053/files#diff-3311722931fe6ac91c84f68dc5b09d27dbc6624d78e59165e1f0c54564a5b947R29.


I believe the file `/etc/sonic/config_db.json` is currently world readable
```
-rw-r--r-- 1 root root 17K Sep 23 09:07 /etc/sonic/config_db.json
```
And it will need to be changed to this
```
-rw--r---- 1 root redis 731 Sep 16 14:25 /etc/sonic/config_db.json
```
I don't see the code to do that in this PR? I would expect something like the following code at the end of generate_config() (
https://github.com/sonic-net/sonic-buildimage/blob/master/files/image_config/config-setup/config-setup#L255)
```
sudo chmod 640 ${DEST_FILE}
sudo chown root:redis ${DEST_FILE}
```

doc/User_Management/User_Management.md
aidan-gallagher: I think currently a user only has to be part of the `docker` group to view the redis database.

```
admin@sonic:~$ sudo su - newusername 
$ groups
newusername docker
$ redis-cli
127.0.0.1:6379> KEYS *
 1) "GEARBOX_TABLE_KEY_SET"
 2) "PORT_TABLE:Ethernet96"
 3) "PORT_TABLE:Ethernet80"
 4) "PORT_TABLE:Ethernet88"
 5) "PORT_TABLE:Ethernet16"
...
```

Can/should this be changed so only users who are part of the redis group can view the db?

rookie-who:
Hi @manoharan-nexthop — the 202605 branch cut is at end of May. This PR has had no activity since Dec 2025. Are you still planning to land this for 202605? If so, please address @aidan-gallagher's review comments. Otherwise we may need to defer to the next release.

anders-nexthop:
> Hi @manoharan-nexthop — the 202605 branch cut is at end of May. This PR has had no activity since Dec 2025. Are you still planning to land this for 202605? If so, please address @aidan-gallagher's review comments. Otherwise we may need to defer to the next release.
> 
> CC: @vaibhavhd @radha-danda (branch managers)

@rookie-who there is recent activity on all of the linked PRs, including LGTM responses on all three from @aidan-gallagher. These PRs still require sign-off from a reviewer with write-access before they can be merged.

rookie-who:
Hi @qiluo-msft @maipbui — could you please review and sign off on this HLD from the AAA and user management perspective? The 202605 branch cut is at end of May.

securely1g:
@qiluo-msft @maipbui Could you review and sign off on this HLD from the AAA and user management perspective? Thanks!

rookie-who:

PR #2021, opened 2025-06-24
Add design for migrating systemd-managed docker containers to Kubernetes with resource control

FengPan-Frank:
Add design for migrating systemd-managed docker containers to Kubernetes with resource control

|Description | Repo  | PR link| State |
|sidecar container sample (telemetry) | sonic-buildimage |https://github.com/sonic-net/sonic-buildimage/pull/23936|GitHub issue/pull request detail|
|docker container change sample (telemetry) | sonic-buildimage |https://github.com/sonic-net/sonic-buildimage/pull/23756|GitHub issue/pull request detail|
|watchdog container sample (telemetry) | sonic-buildimage |https://github.com/sonic-net/sonic-buildimage/pull/23724|GitHub issue/pull request detail|


doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,255 @@
...
+
+if [[ -z "$NAME" || -z "$ACTION" ]]; then
+  echo "Usage: $0 <container-name> {start|stop|restart|status}"
+  exit 1
+fi
+
+case "$ACTION" in
+  start)
+    echo "[INFO] Scaling $NAME deployment to 1"
+    kubectl scale deployment "$NAME" --replicas=1 -n "$NAMESPACE"
```
make1980: we use daemonset instead of deployment so this command won't work.

for stop we can add a taint to the node and then delete the pod:

kubectl taint nodes key=value:NoSchedule

kubectl delete pod [pod-name] -n <SONiC-device-name>

for start we'll need to remove the taint - in our current setup k8s will automatically bring up the pod if the pod was deleted.

for restart we just need to delete the pod.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,352 @@
...
+Disable any native auto-restart logic (e.g., systemctl disable, docker rm -f && docker rm, etc.). But this may break some existing feature like CriticalProcessHealthChecker, featured, systemHealth, etc. Thus this option is not preferable.
+
+### Mirror the Real State into the Config Flag
+
+This means once we enable feature via KubeSonic, we can keep updating the FEATURE table to reflect Kubernetes’ real state, by this existing service/script which depends on FEATURE table will not be broken.
+
+#### FEATURE table update via kubernetes
+
+- When the K8s BMP Deployment is enabled → set `FEATURE|bmp state = enabled`
+- When the K8s Deployment is rolled back → set `FEATURE|bmp state = disabled`
```
make1980: this can only happen the first time when k8s deployment happens - after that rollback only means k8s pod is using a previous version but it's still running so we can't set feature table state as disabled. #Closed

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,352 @@
...
+
+We can implement this as a Kubernetes-based CronJob as below:
+
+```yaml
+apiVersion: batch/v1
+kind: CronJob
```
make1980: we don't support cronjob today - only daemonset is used. i don't think we have to use cronjob though - implementing a script as a sidecar container which runs periodically inside the daemonset will achieve the same, right? #Closed

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,352 @@
...
+    spec:
+      template:
+        spec:
+          containers:
+          - name: redis-updater
+            image: redis:latest
+            command: ["/bin/sh", "-c"]
+            args:
+              - |
+                replicas=$(kubectl get deploy bmp -n sonic -o jsonpath='{.spec.replicas}')
```
make1980: we don't use deployment/replica etc - i think the script can be as simple as this:

1. in v0 of the container, it always sets the feature state as enabled and restore all the systemd script files. and in v0 there is no k8s managed bmp container defined.

2. in v1 of this sidecar container, it sets the feature state as enabled and update systemd script to interact with k8s.

we can choose how to support disabling bmp(or any other feature) through the config db feature table.

it seems that taint can be used for that:

kubectl taint nodes <node-name> key=value:NoSchedule

so when a feature is disabled we add a taint to the node and delete the pod.

make1980 (changes requested):

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,352 @@
...
+                    "set_owner": "local",
+                    "state": "enabled",
+                    "support_syslog_rate_limit": "true"
+                },
+```
+  
+- `"set_owner": "local"` → NDM owns the feature. The `state` field in `FEATURE|*` is authoritative and **must not be modified externally**.  
+- `"set_owner": "kube"` → Kubernetes owns the feature. The `state` field in `FEATURE|*` should reflect the current deployment state in Kubernetes.
+
+The sync logic described above **must only be enabled when `set_owner` is set to `"kube"`**. MM should be explicitly notified of this ownership transition to avoid inconsistencies or race conditions.
```
make1980: an alternative approach is that we don't use NCM to control the switch between local mode and kube mode.

k8s is already a goal state management system, very similar to config DB based goal state management. so introducing additional goal state like set_owner is unnecessary.

we can just rely on the following facts to decide whether we're in local mode or k8s mode.

v0 pod or no pod running - local mode
v1 and v1+ pod running - k8s mode

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,352 @@
...
+ACTION="$2"
+NAMESPACE="default"
+
+if [[ -z "$NAME" || -z "$ACTION" ]]; then
+  echo "Usage: $0 <container-name> {start|stop|restart|status}"
+  exit 1
+fi
+
+case "$ACTION" in
+  start)
```
make1980: please refer to my other comment - this won't work because we use daemonset today. #Closed

make1980 (changes requested):

doc/bmp/k8s_migration_design.md
FengPan-Frank: I think we need some path to notify NDM, since after k8s takes control of feature state, we don't want NDM to overwrite the FEATURE table after k8s rewrite the table state.

make1980: NDM doesn't control local vs kube mode today - because we're not using it.

NDM can still control feature enable/disable config but that should apply to both local and kube modes. w/o local/kube mode the design will be much more simplified.

FengPan-Frank: I mean we could use this mode to control the ownership of feature enable/disable config transition.

if local mode, NDM own source of truth, and it should overwrite and alert if any inconsistency exists in FEATURE table.
if kube mode, k8s start owning source of truth, NDM should skip FEATURE table updating/monitoring for this feature.

make1980: I don't know if we stop supporting feature table how many existing cases will be broken.

Regardless I don't see a need of introducing the local/kube mode - what's the benefit of that? Feature table anyway doesn't work in kube mode so what do we gain by that?

FengPan-Frank: AFAIK, we have
1. cli like "show feature XXX", "config feature XXX", some processes like warm-reboot may use the cli as well.
2. external service like CriticalProcessHealthChecker
3. many mgmt test like systemhealth_checker, container_checker

Actually I'm not sure if it's acceptable that we break feature table, thus try to keep it compatible in the design, we can discuss in meeting for this in detail.

doc/bmp/k8s_migration_design.md
FengPan-Frank: updated into daemonset for all parts.

doc/bmp/k8s_migration_design.md
make1980: so we'll have to support feature table - then we just modify GCU to start/stop k8s service which should be enough.

I still don't see a need of the local/kube mode table.

FengPan-Frank: Talked with Yun offline. I got currently there're many config change is via GCU, meanwhile modify NDM golden config for this change's consistency, then we should not need local/kube mode indeed.

doc/bmp/k8s_migration_design.md
FengPan-Frank: updated.

doc/bmp/k8s_migration_design.md
FengPan-Frank: updated.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
+
+# Migrating Image-Managed Docker Containers to Kubernetes with Resource Control
+
+## Background
+
+In current SONiC architecture, many containers are image-managed, which means it's packed into build image and managed by NDM Golden config. And commonly deployed and managed using `systemd` and monitored using tools like `monit`. But after KubeSonic comes into picture, this deolpyment lacks advanced orchestration and native resource management features offered by Kubernetes.
```
make1980: and enabled/disabled by feature table setting, which is controlled by NDM golden config.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+The FEATURE controls lifecycle.
+- v0, current version iteration in SONiC
+- v1, KubeSonic rollout feature container, but keeps fully backward compatible
+
+
+| Version | Container Installed By | Container service launched                                          | systemd Handling         | Kubernetes Presence |
+|---------|------------------------|---------------------------------------------------------------------|--------------------------|----------------------|
+| v0      | SONiCImage             | NDM FEATURE table (featured monitor)                                | Native systemd file used | None                 |
+| v1      | KubeSonic              | Kubernetes but NDM FEATURE table control real daemon running or not | stub service file        | DaemonSet            |
+
```
make1980: this title needs to be updated - it's more about "Making Feature Table and SystemD service Backward Compatible"

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+
```
make1980: needs to be updated

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+
```
make1980: this all needs to be removed, rigth?

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+```
+
+#### Feature|state Ownership and Versioning
+
+The FEATURE controls lifecycle.
+- v0, current version iteration in SONiC
+- v1, KubeSonic rollout feature container, but keeps fully backward compatible
+
+
+| Version | Container Installed By | Container service launched                                          | systemd Handling         | Kubernetes Presence |
```
make1980: nit: container service launched -> container service running or not

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+  "set_owner": "kube",
+  "support_syslog_rate_limit": "true"
+}
+```
+
+#### Feature|state Ownership and Versioning
+
+The FEATURE controls lifecycle.
+- v0, current version iteration in SONiC
+- v1, KubeSonic rollout feature container, but keeps fully backward compatible
```
make1980: we should call out that in v1 pod there is no telemetry/bmp container image, it's only a sidecar container that's applying the v0 systemd script to the host.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+
+#### Possible scenario for KubeSonic rollout
+
+##### FEATURE state is enabled in NDM golden config, like telemetry
+
+
+```mermaid
+sequenceDiagram
+    autonumber
+    participant Sidecar
```
make1980: note that when a feature is disabled and we deploy the v2 container(e.g. bmp) for the first time, bmp will be running initially till sidecar found that feature is disabled and execute systemctl stop.

so we'll need to make the sidecar container an initContainer in k8s: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/

make1980: how do you achieve step 6? telemetry.sh in docker image doesn't do anything like that. /usr/bin/telemetry.sh on the host doesn't check feature table either.

I think step 6 is achieved by sidecar container running "systemctl start" after systemd script is replaced.

btw - if you look at /usr/bin/telemetry.sh there are a lot of preparation work for telemetry as well like mounting the necessary directories to telemetry docker. how do we plan to address that? 

maybe this can all be handled in the initContainer - in daemonset spec we mount all the root folders to container, like /usr/share/sonic/device/ and the initContainer is going to get the sku/platform and create symbol link for telemetry to use.

the goal here is to avoid any change to telemetry code so that we can keep everything the same between the build-in container and kubesonic deployed container.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+    NDM->>CONFIG_DB: set FEATURE|bmp enablement, which will start stub bmp.service by featured.service
+    bmp->>bmp: launch bmp daemon as usual
+
+```
+
+##### After KubeSonic rollout, use FEATURE state to disable container, like feature switch-off in some livesite issue, etc
+
+```mermaid
+sequenceDiagram
+    autonumber
```
make1980: I think you really want to change NDM to ConfigUpdater or GCU.

NDM is the service generating golden config on the control plane. it doesn't talk to device or config DB directly.

ConfigUpdater is a control plane service that applies golden config to devices following SDP/safety check.

GCU is the config binary running on the device that's really changing config DB. 

so putting NDM here is confusing...

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+    participant K8s API
+    participant CONFIG_DB
+    participant systemd
+    participant NDM
+    participant telemetry
+
+    NDM->>CONFIG_DB: set FEATURE|telemetry disable, which will stop stub telemetry.service by featured.service
+    alt Pod Running
+        systemd->>telemetry: kill docker container
+        telemetry->>telemetry: container restarted by k8s automatically
```
make1980: to disable k8s container the sequence won't be this complicated.

when feature table has telemetry as disabled, hostcfgd will notice that and execute "systemctl stop telemetry" which will add a taint to the k8s node representing the device and trigger k8s to remove the pod from the node so that telemetry service is not running.

so there isn't too much involvement from telemetry or sidecar here.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+          Sidecar->>systemd: no-op, pending later service start
+        else FEATURE|telemetry is enabled
+          Sidecar->>systemd: start telemetry service
+        end
+    else KubeSonic not rollback
+        Sidecar->>systemd: no-op
+    end
+```
+
+Here need a preStop in sidecar to restore systemd script before rollback
```
make1980: this is too complicated - instead of doing this we should have v0 pod which only has sidecar container and systemd v0 version(so only run the OS build-in telemetry).

when rollback happens from v1 to v0, we first run v0 pod, which has sidecar container and systemd v0 version, but without telemetry container.

sidecar applies systemd v0 version to the system and run "systemctl start telemetry" which will start the OS build-in telemetry

there is no need of having preStop to stop the service with this design. the issues with preStop are: 1. additional complexity doing the same thing that v0 is supposed to do already. 2. it runs only once so in case it fails there is no retry logic to apply v0 systemd scripts.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+#### container startup script update
+
+Since FEATURE table still control container start/stop via featured.service, after container is rollout-ed via kubernetes, it need read from FEATURE table and determine whether it will launch daaemon or not.
+
+```bash
+#!/usr/bin/env bash
+
+STATE=$(redis-cli -n 4 HGET "FEATURE|telemetry" state)
+
+if [ "$STATE" == "enabled" ]; then
```
make1980: this is not needed - we don't have this today and won't need this because hostcfgd already handles it.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+##### Rationale
+Here we leverage a sidecar container inside the `DaemonSet` to perform periodic sync logic.
+
+- The sidecar container:
+  - Detects whether the container is running (via `kubectl`, `docker`, or other APIs).
+  - Patch or restore systemd service files.
+
+##### Sidecar Design
+- Runs a simple shell script patch_systemd.sh that loops with `sleep`.
+- Ensures FEATURE state is in sync with container state.
```
make1980: this needs to be changed right? 

the only thing that sidecar container needs to do periodically is to compare the systemd script in its container image with the systemd script on host and apply its own version when they're different.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+- For `v0`, restores systemd service files.
+- For `v1`, read from FEATURE table and optionally updates systemd stub.
```
make1980: it doesn't need to care about feature table - just apply the systemd script blindly will also work.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+  template:
+    metadata:
+      labels:
+        app: telemetry
+    spec:
+      containers:
+      - name: telemetry
+        image: ksdatatest.azurecr.io/docker-sonic-telemetry:latest
+        command: ["/usr/local/bin/supervisord"]
+      - name: telemetry-feature-sidecar
```
make1980: as mentioned we also need a initContainer

make1980: sidecar container can still be here since it needs to run all the time - we need to have a way of sharing code between initContainer and sidecar container if needed

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+    kubectl get pods -n "$NAMESPACE" \
+      -l "$LABEL_SELECTOR" \
+      --field-selector spec.nodeName="$NODE_NAME" \
+      -o jsonpath='{.items[0].metadata.name}' 2>/dev/null
+}
+
+kill_container() {
+    POD_NAME=$(get_pod_name)
+    if [[ -n "$POD_NAME" ]]; then
+        kubectl exec "$POD_NAME" -n "$NAMESPACE" -- pkill -f telemetry
```
make1980: this won't work because k8s will restart the container.

the right way of killing a container in k8s is to use a taint on the node and update node selector so that node with the taint won't run the pod.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+          exec:
+            command: ["/usr/bin/pgrep", "telemetry"]
+          initialDelaySeconds: 30
+          periodSeconds: 15
+```
+
+
+## Monitoring and Alerting
+
+Currently SONiC uses monit check memory for specific container, like below
```
make1980: I think we should get rid of all these with kubesonic and rely on cgroup to restart telemetry when memory usage exceeds the threshold.

The only thing we need to make backward compatible is the processstat table - we need to make sure the cpu/memory usage collected there for telemetry will still work when kubesonic manages the container.

not sure how it works but please take a look at that part.

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+        except IOError:
+            syslog.syslog(syslog.LOG_ERR, ERROR_CONTAINER_MEMORY_USAGE_NOT_FOUND.format(container_id))
+            sys.exit(INTERNAL_ERROR)
+
+    syslog.syslog(syslog.LOG_ERR, ERROR_CGROUP_MEMORY_USAGE_NOT_FOUND.format(docker_memory_usage_file_path, container_id))
+    sys.exit(INTERNAL_ERROR)
+```
+
+
+During transition period if `monit` must be retained temporarily, we need to rewrite /usr/bin/memory_check to use Kubernetes data (e.g., via `kubectl top`) instead of reading Docker or CGroup files like above.
```
make1980: a big piece that's missing is the handling of /usr/bin/telemetry.sh - please schedule some time for us to discuss on this.

doc/bmp/k8s_migration_design.md
FengPan-Frank: Because here the telemetry is the container rollout-ed via KubeSonic, we should be able to update telemetry.sh code.

Thus I thought we could update https://github.com/sonic-net/sonic-buildimage/blob/master/dockers/docker-sonic-telemetry/telemetry.sh, which makes Step 6 read from configDB, if FEATURE table is disabled, it does not launch telemetry daemon; if FEATURE is enabled, it launches telemetry as normal. By this, FEATURE table update will restart docker via featured monitoring.

doc/bmp/k8s_migration_design.md
FengPan-Frank: what I want to show in this scenario is after KubeSonic rollout, we still support NDM FEATURE table disable service, thus it's NDM triggers FEATURE table update, like we want to switch-off the feature in some live site issue.

doc/bmp/k8s_migration_design.md
FengPan-Frank: here my thought is after we patched stub systemd script, start/stop/restart will NOT be supported really since k8s will manage the service liveness instead of systemd, thus we only need to kill so that restart telemetry container (telemetry startup script will take care if it's idle or working based on FEATURE table).

doc/bmp/k8s_migration_design.md
FengPan-Frank: pls refer #### container startup script update

This is the startup script for launching telemetry daemon within container, which is to replace current telemetry.sh, will discuss this offline.

doc/bmp/k8s_migration_design.md
make1980: I don't quite understand - we don't know how many cases are depending on feature table today so we'll have to make it work. If we just kill telemetry and feature table marks telemetry as disabled, wouldn't this break any existing workflow?

doc/bmp/k8s_migration_design.md
make1980: I understand that - but it's going to be confusing to phrase this way for most of the people.

doc/bmp/k8s_migration_design.md
make1980: this script doesn't read feature table today. i think it would be better if we can handle everything outside of this script and have just very few places that we need to read feature table.

make1980: btw - there is nowhere to run https://github.com/sonic-net/sonic-buildimage/blob/master/dockers/docker-sonic-telemetry/telemetry.sh in kubesonic world except init container. so i guess we're talking about the same thing.

doc/bmp/k8s_migration_design.md
isabelmsft: nit: deolpyment -> deployment

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+
+The FEATURE controls lifecycle.
+- v0, current version iteration in SONiC
+- v1, KubeSonic rollout feature container, but keeps fully backward compatible
+
+
+| Version | Container Installed By | Container service launched                                          | systemd Handling         | Kubernetes Presence |
+|---------|------------------------|---------------------------------------------------------------------|--------------------------|----------------------|
+| v0      | SONiCImage             | NDM FEATURE table (featured monitor)                                | Native systemd file used | None                 |
+| v1      | KubeSonic              | Kubernetes but NDM FEATURE table control real daemon running or not | stub service file        | DaemonSet            |
```
isabelmsft: What is meant by "real daemon"? Which key in FEATURE table indicates whether real daemon is running or not (is it `set_owner` or `state`?)?

isabelmsft: Is there an example of native systemd service file vs stub service file?

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+       systemctl daemon-reload
+       systemctl start telemetry.service
+       "]
+```
+
+---
+
+#### container startup script update
+
+Since FEATURE table still control container start/stop via featured.service, after container is rollout-ed via kubernetes, it need read from FEATURE table and determine whether it will launch daaemon or not.
```
isabelmsft: nit: daaemon -> daemon

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+
+### v0 Behavior
+- Container is fully managed by `systemd` (e.g., `telemetry.service`).
+- FEATURE table controls startup (`state: Enabled/Disabled`).
+- Actual container starts or stops via `systemctl` by featured.service
+
+### v1+ Behavior with Kubernetes DaemonSet
+- The container is deployed via Kubernetes DaemonSet.
+- The systemd service is retained as a **stub**, to avoid breaking automation or tools that query it.
+- For start/stop/restart, it will just kill container simply so that kubernetes will relaunch it automatically.
```
isabelmsft: Does this mean we lose functionality of `systemctl stop` if kubernetes will relaunch automatically?

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+## Monit configuration for telemetry container
+###############################################################################
+check program container_memory_telemetry with path "/usr/bin/memory_checker telemetry 419430400"
+    if status == 3 for 10 times within 20 cycles then exec "/usr/bin/restart_service telemetry" repeat every 2 cycles
+
+```
+
+memory_checker implementation
+```
+def get_memory_usage(container_id):
```
isabelmsft: `monit` currently checks for standard container names in output of `docker ps`. I understand this part of monit will no longer be used, but want to take note of the consideration that other parts of code may similarly have dependencies on standard container name. To maintain backwards compatibility after migrating to KubeSonic, we may need to do a check through sonic code base for other dependencies on standard container_name. For example: https://github.com/sonic-net/sonic-host-services/blob/4fed73d76abd64967157f586a272f04bc2abcbb0/host_modules/reboot.py#L84

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+```json
+"telemetry": {
+  "auto_restart": "enabled",
+  "state": "enabled",
+  "delayed": "False",
+  "check_up_status": "False",
+  "has_per_asic_scope": "False",
+  "has_global_scope": "True",
+  "high_mem_alert": "disabled",
+  "set_owner": "kube",
```
isabelmsft: Do we ignore this field, and we only check the "state": "enabled" field?

doc/bmp/k8s_migration_design.md
```diff
@@ -0,0 +1,547 @@
...
+    participant Sidecar
+    participant K8s API
+    participant CONFIG_DB
+    participant systemd
+    participant NDM
+    participant telemetry
+
+    Sidecar->>K8s API: Query telemetry pod status
+    alt Pod Running
+        Sidecar->>CONFIG_DB: Query FEATURE|telemetry state enabled
```
isabelmsft: If the pod is running and telemetry state is enabled, why do we stop running the telemetry service? Unless the goal is to transition from kube-managed telemetry to SONiC local image managed telemetry?

doc/bmp/k8s_migration_design.md
make1980: that's a good point but how do we find all the places that container name is used? for this particular place it seems to be used only for pmon.

isabelmsft: I'd start by doing a grep through the repo for: 
* `docker ps` in various forms (like https://github.com/sonic-net/sonic-buildimage/blob/82bf05019c0f68c64732a00394f3237772a742dd/files/image_config/monit/memory_checker#L95) 
* `container_name`
* docker python library containers.list call (such as https://github.com/sonic-net/sonic-buildimage/blob/82bf05019c0f68c64732a00394f3237772a742dd/files/image_config/monit/memory_checker#L95)

doc/bmp/k8s_migration_design.md
FengPan-Frank: fixed

FengPan-Frank: fixed

doc/bmp/k8s_migration_design.md
FengPan-Frank: Yes, in this design "set_owner" field is not used.

doc/bmp/k8s_migration_design.md
FengPan-Frank: offline discussed

doc/bmp/k8s_migration_design.md
FengPan-Frank: updated

doc/bmp/k8s_migration_design.md
FengPan-Frank: offline discussed

doc/bmp/k8s_migration_design.md
FengPan-Frank: updated, I mean the real agent inside of container which supervisord controls

FengPan-Frank: updated

doc/bmp/k8s_migration_design.md
FengPan-Frank: Right, unless we tint the node but that will require all container in dedicated daemon set. Call this out as known limitation.

doc/bmp/k8s_migration_design.md
FengPan-Frank: updated

doc/bmp/k8s_migration_design.md
FengPan-Frank: prodockerstatd uses psutil for getting those information, thus we should be good for getting rid of this.
refer https://github.com/sonic-net/sonic-host-services/blob/d409f270edc6184d286676872b224802d0353891/scripts/procdockerstatsd#L144

doc/bmp/k8s_migration_design.md
FengPan-Frank: Added an upgrade flow diagram and call out the container name changes, maybe we can get others' comments if there's other places out of sonic depend on container name like CPHC?

doc/bmp/k8s_migration_design.md
FengPan-Frank: offline discussed

doc/bmp/k8s_migration_design.md
FengPan-Frank: offline discussed

doc/bmp/k8s_migration_design.md
FengPan-Frank: updated

doc/bmp/k8s_migration_design.md
FengPan-Frank: here stop is for patching systemd script.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+break any existing feature like CriticalProcessHealthChecker, featured, systemHealth, etc.
+
+### 4.1 Keep FEATURE table as the source of truth of feature enablement
+
+This means even after we enable feature via KubeSonic, we will still keep FEATURE table and NDM Golden as its owner, KubeSonic will only deploy container and manage container start/stop which coordinate with FEATURE table.
+
+#### FEATURE Table Snapshot
+```json
+"telemetry": {
+  "auto_restart": "enabled",
```
make1980, line 53: we should call out that some of the systemd related parameters won't be supported because we believe they're not used. like auto_restart/delayed/check_up_status/set_owner

make1980: we can do some rudimentary research and conclude that these are not used in automation and in case we find them being used we'll fix the places to use kubectl instead.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+| v0      | SONiCImage             | NDM FEATURE table control it via featured service monitor           | Native systemd file used | None                 |
+| v1      | KubeSonic              | Not launched as it's only a short life version to patch systemd script  | stub service file        | DaemonSet            |
+| v2+     | KubeSonic              | Service launched by Kubernetes but NDM FEATURE table controls supervisorctl managed daemon running or not | stub service file        | DaemonSet            |
+
+
+
+
+### 4.2 container startup script update
+
+Since FEATURE table still control container start/stop via featured.service, after container is rollout-ed via kubernetes, it keeps reading from FEATURE table and determines whether it will launch daaemon or enter idle.
```
make1980: typo: daemon

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+```
+COPY docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+RUN chmod +x /usr/local/bin/docker-entrypoint.sh
+ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
+```
+
+### 4.3 container watchdog
+
+Since container will be installed via kubernetes, but FEATURE table still controls its enable or disable, watchdog will need to coordinate with this meanwhile. 
+watchdog needs to read FEATURE, if FEATURE is enabled, watchdog execute its normal probe and return healthcheck status; if FEATURE is diabled, watchdog will skip probe and return OK directly.
```
make1980: execute -> executes

make1980: typo: disabled.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+
+fn main() {
+    let listener = TcpListener::bind("127.0.0.1:50069")
+        .expect("Failed to bind to 127.0.0.1:50069");
+
+    if !is_telemetry_enabled() {
+        let response = "HTTP/1.1 200 OK\r\nContent-Type: application/json\r\nContent-Length: 29\r\n\r\n{\"message\":\"telemetry disabled\"}";
+        stream.write_all(response.as_bytes()).ok();
+    } else {
+        // normal health check
```
make1980: we should describe what's monitored by the watchdog - maybe just a simple physical path probe to telemetry will be a good starting point.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+
+#### v0 Behavior
+- Container is fully managed by `systemd` (e.g., `telemetry.service`).
+- FEATURE table controls startup (`state: Enabled/Disabled`).
+- Actual container starts or stops via `systemctl` by featured.service
+
+#### v1+ Behavior with Kubernetes DaemonSet
+- The container is deployed via Kubernetes DaemonSet.
+- The systemd service is retained as a **stub**, to avoid breaking automation or tools that query it.
+- For start/stop/restart, it will just kill container simply so that kubernetes will relaunch it automatically. Here is some limitation: since kubernetes takes over the container thus systemd stop will not STOP container really, unless we tint the node but that requires each container should use dedicated daemon set which is not preferrable as well.
```
make1980: tint -> taint

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+    kubectl get pods -n "$NAMESPACE" \
+      -l "$LABEL_SELECTOR" \
+      --field-selector spec.nodeName="$NODE_NAME" \
+      -o jsonpath='{.items[0].metadata.name}' 2>/dev/null
+}
+
+kill_container() {
+    POD_NAME=$(get_pod_name)
+    if [[ -n "$POD_NAME" ]]; then
+        kubectl exec "$POD_NAME" -n "$NAMESPACE" -- pkill -f telemetry
```
make1980: have you tried kubectl exec? this may not work in our setup - we don't have a kubemaster to device channel today. so maybe you should use docker command to find the container with the pod name in it and restart it through docker instead

this is again some complexity because we do not put telemetry into its own daemonset/pod.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+### 4.5 Patch systemd script via DaemonSet Sidecar
+
+To patch systemd script aligned with actual runtime state, here we leverage a sidecar container inside the `DaemonSet` to perform periodic sync logic.
+
+- The sidecar container:
+  - Detects whether the container is running (via `kubectl`, `docker`, or other APIs).
+  - Patch systemd service files.
+
+
+patch_systemd.sh
```
make1980: i'd prefer to make the sidecar container written in golang so that we don't need a base image for it.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+          mountPath: /scripts
+      volumes:
+      - name: scripts
+        configMap:
+          name: feature-sync-script
+```
+
+
+### 4.6 Desired state driven for the files patching  (like systemd patch)
+Since postupgrade and Kubernetes rollout may change the same files, which may lead to uncertain result. we propose to use desired state driven for files patching, so that any apply order should coverged into the same desired state.
```
make1980: typo: coverged -> converge

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+The desired state script will based on top of tree from specific code branch.
```
make1980: will based on - will be based on

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+
+```mermaid
+sequenceDiagram
```
make1980, line 316: one thing we should probably clarify in this diagram is that we don't intend to change postupgrade script to use a desired state file(or the same desired state file that we use in the sidecar). if any postupgrade script is doing a diff based update it will still be that way but sidecar will overwrite it.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,579 @@
...
+    NDM->>CONFIG_DB: set FEATURE|telemetry disable, which will stop stub telemetry.service by featured.service
+    systemd->>telemetry: kill docker container
+    telemetry->>telemetry: container restarted by k8s automatically
+    telemetry->>CONFIG_DB: read FEATURE|telemetry as disabled
+    telemetry->>telemetry: launch container as idle
+    telemetry-watchdog->>CONFIG_DB: read FEATURE|telemetry as disabled
+    telemetry-watchdog->>k8s: skip detect and return OK   
+```
+
+### 5.4 After KubeSonic rollout, rollback the container. (v2+ -> v2)
```
make1980: featured seems to use "systemctl show telemetry --property UnitFileState" to get the status of a service - in our current design how do we make sure it will reflect whether telemetry is running or not?

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: fixed

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: These are NDM config, not all are systemd related parameter, it is monitored via https://github.com/sonic-net/sonic-host-services/blob/master/scripts/featured, and for k8s rollout patched script we need to care about multi-asic https://github.com/sonic-net/sonic-buildimage/tree/master/files/build_templates/per_namespace, other config parameter could still be monitored by featured and trigger corresponding systemd command.

doc/kubernetes/k8s_migration_design.md
make1980: this doesn't seem to work as expected today so we might be fine. the logic is: 1. check show --property result, if it's enabled don't do anything and set state table as enabled. 2. try enable and start if no error, set state table as enabled.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,582 @@
...
+                time.sleep(0.1)  # retry after short delay
+            else:
+                break
+        except IOError:
+            syslog.syslog(syslog.LOG_ERR, ERROR_CONTAINER_MEMORY_USAGE_NOT_FOUND.format(container_id))
+            sys.exit(INTERNAL_ERROR)
+
+    syslog.syslog(syslog.LOG_ERR, ERROR_CGROUP_MEMORY_USAGE_NOT_FOUND.format(docker_memory_usage_file_path, container_id))
+    sys.exit(INTERNAL_ERROR)
+```
```
make1980: we should probably just say after k8s we simplify the restart logic to be OOM based only - I'm not sure why we can tolerate memory usage exceeding the limit for several times as what's implemented right now.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,582 @@
...
+      labels:
+        app: telemetry
+    spec:
+      containers:
+      - name: telemetry
+        image: ksdatatest.azurecr.io/docker-sonic-telemetry:latest
+        command: ["/usr/local/bin/supervisord"]
+        resources:
+          requests:
+            memory: "100Mi"
```
make1980: based on monit config the memory limit is really 400M, right?

make1980 (changes requested):

doc/kubernetes/k8s_migration_design.md
make1980: do we still support auto_restart with systemd stub? maybe delayed is still supported through the stub? not sure about that part...

same for check_up_status. we don't use set_owner anymore with this design, right?

make1980 (dismissed):

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,582 @@
...
+  STATE=$(redis-cli -n 4 HGET "FEATURE|telemetry" state)
+  if [ "$STATE" == "enabled" ]; then
+    echo "Telemetry is enabled. Starting service..."
+    break
+  else
+    echo "Telemetry is disabled. Entering idle..."
+    sleep 1000
+  fi
+done
+exec /usr/local/bin/supervisord
```
losha228: what if the feature is enabled and is running by k8s pod, then someone disables the feature, I suppose the systemd service may be stopped, but what's the behavior of the container ? will it exit ? We should clear the behavior

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,582 @@
...
+
+#### v0 Behavior
+- Container is fully managed by `systemd` (e.g., `telemetry.service`).
+- FEATURE table controls startup (`state: Enabled/Disabled`).
+- Actual container starts or stops via `systemctl` by featured.service
+
+#### v1+ Behavior with Kubernetes DaemonSet
+- The container is deployed via Kubernetes DaemonSet.
+- The systemd service is retained as a **stub**, to avoid breaking automation or tools that query it.
+- For start/stop/restart, it will just kill container simply so that kubernetes will relaunch it automatically. Here is some limitation: since kubernetes takes over the container thus systemd stop will not STOP container really, unless we tint the node but that requires each container should use dedicated daemon set which is not preferrable as well.
```
losha228: for start action, we may return immediately, because kubelet will always start it if it's not running.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,582 @@
...
+- For status, it will return runtime state via kubectl.
```
losha228: how do you get the status for the container by kubectl ?  If you want to do it, you may need to get the corresponding pod and extract the container status from the pod, there is not a clear mapping between pod and the feature container, so we may return the status based on feature status.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,582 @@
...
+NAMESPACE="sonic"
+LABEL_SELECTOR="app=telemetry"
+NODE_NAME=$(hostname)
+
+usage() {
+    echo "Usage: $0 {start|stop|restart|status}"
+    exit 1
+}
+
+get_pod_name() {
```
losha228: it assumes there is only one pod on the node, but it may not be true.

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,582 @@
...
+    kubectl get pods -n "$NAMESPACE" \
+      -l "$LABEL_SELECTOR" \
+      --field-selector spec.nodeName="$NODE_NAME" \
+      -o jsonpath='{.items[0].metadata.name}' 2>/dev/null
+}
+
+kill_container() {
+    POD_NAME=$(get_pod_name)
+    if [[ -n "$POD_NAME" ]]; then
+        kubectl exec "$POD_NAME" -n "$NAMESPACE" -- pkill -f telemetry
```
losha228: in the pod, the service is still managed by systemd inside the container, so I think you may call supervisorctl to restart it

losha228: please note, we can deploy multiple containers in a pod, we usually use the feature name as the container name, you will need to consider to use '-c <container-name>' in kubectl exec

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,582 @@
...
+    telemetry->>telemetry: supervisord launch telemetry daemons as usual
+    telemetry-watchdog->>K8s: normal detect and health check response
+    alt telemetry-watchdog detect failed
+        K8s->>telemetry: rollback to v2
+    else  telemetry-watchdog detect successful
+        K8s->>telemetry: no-op
+    end
+```
+
+### 5.4 After KubeSonic rollout, rollback the container to imaged based version. (v2 -> v0)
```
losha228: what's the scenario for v2 -> v0 ?  In k8s rollout, if v2 is failed, it will rollout to v1 instead of v0, right ?

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: added real detection code for gnmi service listen 50051

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: fixed

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: updated per https://github.com/FengPan-Frank/sonic-buildimage/blob/bcfb0fa0934336714e87c3366847f82f556de5a9/dockers/docker-sonic-telemetry/base_image_files/monit_telemetry

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: I'm not sure previous intention for retry, updated this part as well.

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: updated.

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: we'll have no change in current featured monitoring workflow.

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: systemd will kill and restart container, after restart container will read from FEATURE table and know it's disabled then enters idle status.

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: For guarantee the correctness we may simply kill/restart container, so that latest FEATURE table could be read.

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: Can we add a label like "feature: telemetry" which has the same name as FEATURE table now?

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: we can update during implementation.

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: v1 only has systemd stub files, here I mean rollback to original SONiC image based version.

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: Can we add a label like "feature: telemetry" which has the same name as FEATURE table?

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,589 @@
...
+- `systemctl stop <service>`
+- `systemctl status <service>`
+
+To prevent breaking these expectations during the migration, a `systemd` service stub can be provided.
+
+
+#### v0 Behavior
+- Container is fully managed by `systemd` (e.g., `telemetry.service`).
+- FEATURE table controls startup (`state: Enabled/Disabled`).
+- Actual container starts or stops via `systemctl` by featured.service
```
ganglyu: featured has used other systemctl commands like `mask` and `unmask`, how to support them?

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,589 @@
...
+
+    syslog.syslog(syslog.LOG_ERR, ERROR_CGROUP_MEMORY_USAGE_NOT_FOUND.format(docker_memory_usage_file_path, container_id))
+    sys.exit(INTERNAL_ERROR)
+```
+
+
+As section "Enforce CPU and memory resource constraints natively" mentioned, this monit functionality could be covered by Kubernetes, thus for monit we need to get rid of if from the KubeSonic rollouted container. However, during transition period if there's any case decalres that `monit` must be retained temporarily, we can also rewrite /usr/bin/memory_check to use Kubernetes data (e.g., via `kubectl top`) instead of reading Docker or CGroup files like above.
+
+---
+
```
ganglyu, line 515: How to support container_checker?

doc/kubernetes/k8s_migration_design.md
```diff
@@ -0,0 +1,589 @@
...
+#### v0 Behavior
+- Container is fully managed by `systemd` (e.g., `telemetry.service`).
+- FEATURE table controls startup (`state: Enabled/Disabled`).
+- Actual container starts or stops via `systemctl` by featured.service
+
+#### v1+ Behavior with Kubernetes DaemonSet
+- The container is deployed via Kubernetes DaemonSet.
+- The systemd service is retained as a **stub**, to avoid breaking automation or tools that query it.
+- For start/stop/restart, it will just kill container simply so that kubernetes will relaunch it automatically. Here is some limitation: since kubernetes takes over the container thus systemd stop will not STOP container really, unless we taint the node but that requires each container should use dedicated daemon set which is not preferrable as well.
```
ganglyu: We don't really support the container, if the container has memory issue, this might not be enough.

doc/kubernetes/k8s_migration_design.md
make1980: @FengPan-Frank - let's take a look at the featured implementation. It seems that telemetry systemd service doesn't even support mask/unmask/enable/disable stuff but for the containers that do support we should probably just treat them all as restart.

FengPan-Frank: image

Verified on this, mask/unmask will flap symlink to /dev/null thus we should be fine with this command

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: do you mean OOM issue? OOM should be handled via below section 6

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: image

we can use label to map feature and container status.

doc/kubernetes/k8s_migration_design.md
FengPan-Frank: updated container_check into doc as well.

make1980 (dismissed):

make1980 (approved):

PR #2032, opened 2025-07-04
Liquid Cooling leakage detection in SONiC

yuazhe:
This is my HLD for the liquid cooling leakage detection, it provides a basic liquid cooling object to reflect the hardware as well as a  mechanism to monitor the leakage sensors of it.

PR List:

| [[sonic-buildimage] [Add support for liquid cooling inside hardware checker](https://github.com/sonic-net/sonic-buildimage/pull/24124) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-buildimage] [Add a function to get container name by asic ID](https://github.com/sonic-net/sonic-buildimage/pull/24125)] | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-utilities] [Add multi ASIC support for syslog rate limit feature](https://github.com/sonic-net/sonic-utilities/pull/4068)] | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-platform-common] [Add new liquid cooling base object](https://github.com/sonic-net/sonic-platform-common/pull/603)] | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-platform-daemons] [Add suport for liquid cooling leakage detection](https://github.com/sonic-net/sonic-utilities/pull/4068)] | GitHub issue/pull request detail | GitHub pull request check contexts |
| [[sonic-mgmt] [Add test plan and tests for liquid cooling leakage detection](https://github.com/sonic-net/sonic-utilities/pull/4068)] | GitHub issue/pull request detail | GitHub pull request check contexts |


lc_hld.md
```diff
@@ -0,0 +1,126 @@
...
+    "" string return to get the sensor's name
+    def get_name():
+
+    "" boolean returen to indicate whether there is a leak detected
+    def is_leak():
+    
+```
+
+## 5. Thermal control daemon
+A new object `LiquidCoolingUpdater` will be added to the Thermal Control daemon that dedicated monitoring the liquid device status.
```
judyjoseph: Have a question on whether we can define this sensor in the "sonic-sensormond" daemon which is like a common daemon handling all sensors in system? Is this daemon used in your platform?

yuazhe: Leakage detection is part of the liquid cooling system which may have additional functionalities or sensors in the future. And since liquid cooling is in parallel with the air cooling system, it is more appropriate to include it within the thermal control daemon.

volodymyrsamotiy:
@judyjoseph , could you please help to review?

bingwang-ms:
@judyjoseph Can you help review this HLD? Thanks

doc/bmc/leakage_detection_hld.md
```diff
@@ -0,0 +1,126 @@
...
+    def is_leak():
+    
+```
+
+## 5. Thermal control daemon
+A new object `LiquidCoolingUpdater` will be added to the Thermal Control daemon that dedicated monitoring the liquid device status.
+
+During initialization, a separate thread will be launched to periodically call the `get_leak_sensor_status` API. The reason for separating this thread from the main thermal control daemon process is that the main process has a period of 1 minute, which is reasonable for normal thermal device updates but not suitable for leak events. Leak events are critical and need to be reported as soon as possible. However, these events are also abnormal and rare, so we don't need to involve them in the main thread and increase the main thread period to 1 second, which would be overkill and lower performance.
+
+A new configuration will be added to pmon_daemon_control.json, to indicate whether the system has liquid cooling system, if not, the object and thread will not be created in the initialization of thermalctld at all to avoid performance overheading.
```
judyjoseph: Can you share details of the attribute that is planned to be introduced in pmon_daemon_control.json

yuazhe: thanks, I have added more details here

yuazhe:
> LGTM, minor comments

thanks, add more details on your comment

judyjoseph (approved):
LGTM

judyjoseph:
@lguohan could you review too, anything which you would like w.r.to cooling leakage detection. Any other related working going on in cooling. thanks

judyjoseph:
@yuazhe Can you plan to have new sonic-mgmt testcase also for liquid monitoring thread that is introduced in thermalctld ?

yuazhe:
> @yuazhe Can you plan to have new sonic-mgmt testcase also for liquid monitoring thread that is introduced in thermalctld ?

Thanks for the suggestion, I believe several factors had been tested in the UT for example when exception raised, so what are you intended to cover in the mgmt test?

PR #2034, opened 2025-07-07
HLD of SONiC FIPS POST support

ysmanman:
swss-common change https://github.com/sonic-net/sonic-swss-common/pull/1071 [merged]
sairedis change  https://github.com/sonic-net/sonic-sairedis/pull/1650
swss change https://github.com/sonic-net/sonic-swss/pull/3836

Reference links:
OCP SAI PR https://github.com/opencomputeproject/SAI/pull/2167
SAI header update in master https://github.com/sonic-net/sonic-sairedis/pull/1628


doc/fips/SONiC-FIPS-POST.md
```diff
@@ -0,0 +1,89 @@
...
+
+### State DB
+
+The following table is added to State DB to track POST status.
+```
+POST_STATUS_TABLE
+
+;Store POST status
+
+key    = POST_STATUS_TABLE|macsec
```
senthil-nexthop: This entry in the STATE_DB provides the switch-level POST status, however the Macsec SAI API supports per-port POST APIs, besides switch-level APIs. Should this be recorded in the STATE_DB? Something like POST_STATUS_PORT_TABLE|macsec|< port-name >?

doc/fips/SONiC-FIPS-POST.md
```diff
@@ -0,0 +1,89 @@
...
+
+Since SAI supports POST completion callback, a callback or notification function is provided when enabling POST in SAI switch or MACSec creation. Once being invoked, the callback function sets POST status in State DB accordingly based on the POST status returned by SAI.
+
+### Handling SAI POST failure
+
+If SAI POST fails,  MACSecOrch reads POST status of all MACSec ports and finds out which port has failed in POST.  MACSecOrch then adds the details of the failure in syslog.
+
+### Enforcing FIPS compliance
+
+In order to be compliant to FIPS, SONiC should process MACSec configuration only after POST passes. This is achieved by enhancing MACSecMgr, running in MACSec container, to check POST status published in State DB before processing any MACSec configuration, as shown in the flow chart below:
```
senthil-nexthop: If the POST status is in-progress, how does the MacSecMgr process the port configuration when POST has not finished?

doc/fips/SONiC-FIPS-POST.md
```diff
@@ -0,0 +1,89 @@
...
+Orchagent enables POST when creating SAI switch regardless of whether the MACSec feature is supported or enabled. This can avoid triggering POST after SAI switch creation, e.g., MACSec feature is enabled later. The below flow chart depicts the process.
+
+![](images/fips-post-switch-init.png)
+
+Orchagent enables POST when creating SAI switch. After SAI switch is created, Orchagent queries SAI POST capability. If POST is supported in switch init, Orchagent sets POST status to in-progress and waits for POST completion callback. Otherwise, if POST is supported only in MACSec init, Orchagent sets POST status to not-started, which informs MACSecOrch of performing POST in its initialization.
+
+If POST is not supported in either switch or MACSec init, then SAI does not support POST. The possible scenarios are:
+- The platform does not support MACSec. Or
+- The platform does support MACSec. But the deployment does not require FIPS compliance. Or
+- The platform does support MACSec, and the vendor incrementally supports FIPS when they can.
```
senthil-nexthop: Can you clarify what does incremental support mean? What is a partial FIPS support?

doc/fips/SONiC-FIPS-POST.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Handling SAI POST failure
+
+If SAI POST fails,  MACSecOrch reads POST status of all MACSec ports and finds out which port has failed in POST.  MACSecOrch then adds the details of the failure in syslog.
+
+### Enforcing FIPS compliance
+
+In order to be compliant to FIPS, SONiC should process MACSec configuration only after POST passes. This is achieved by enhancing MACSecMgr, running in MACSec container, to check POST status published in State DB before processing any MACSec configuration, as shown in the flow chart below:
+
+![](images/fips-post-compliance.png)
```
senthil-nexthop: The flowchart seems to have an infinite loop if POST status is "failed" or if Macsec is NOT configured.

doc/fips/SONiC-FIPS-POST.md
```diff
@@ -0,0 +1,89 @@
...
+
+The following figure depicts the data flow and SONiC components in the design. Orchagent is responsible for triggering POST via SAI calls and publishing POST status in State DB.   MACSec container, precisely MACSecMgr, is enhanced to be POST aware and only process MACSec configuration after POST has passed. 
+
+![](images/fips-post-overview.png)
+
+### State DB
+
+The following table is added to State DB to track POST status.
+```
+POST_STATUS_TABLE
```
wumiaont: Better to distinguish this is a FIPS TABLE. I would recommend FIPS_POST_STATUS. Also please be aware there's another FIPS table in STATE DB already FIPS_STATS. Should you consider combine your table with the existing one? Such as a separate key under FIPS_STATS?

doc/fips/SONiC-FIPS-POST.md
```diff
@@ -0,0 +1,89 @@
...
+         "not-started"    ; POST was not enabled in switch init.
+         "pass"           ; POST passed.
+         "fail"           ; POST failed. 
+         "unsupported"    ; POST is not supported by SAI.
+
+```
+
+### Enabling POST in SAI switch init
+
+Orchagent enables POST when creating SAI switch regardless of whether the MACSec feature is supported or enabled. This can avoid triggering POST after SAI switch creation, e.g., MACSec feature is enabled later. The below flow chart depicts the process.
```
wumiaont: FIPS can be enabled or disabled on the chassis. It's disabled by default. Question would be should the POST be done only when FIPS is enabled? Or you will do FIPS POST no matter FIPS is enabled or not? I am thinking about a scenario where some vendors do need macsec but does not need FIPS and their chip does not have POST support yet. Need to make sure in this scenario macsec can still be used when FIPS is disabled on chassis.

doc/fips/SONiC-FIPS-POST.md
```diff
@@ -0,0 +1,89 @@
...
+
+![](images/fips-post-switch-init.png)
+
+Orchagent enables POST when creating SAI switch. After SAI switch is created, Orchagent queries SAI POST capability. If POST is supported in switch init, Orchagent sets POST status to in-progress and waits for POST completion callback. Otherwise, if POST is supported only in MACSec init, Orchagent sets POST status to not-started, which informs MACSecOrch of performing POST in its initialization.
+
+If POST is not supported in either switch or MACSec init, then SAI does not support POST. The possible scenarios are:
```
wumiaont: I would take if FIPS is not enabled then no POST is required. If FIPS is enabled POST should be mandatary. If no POST support or POST is not a success, then all secure association APIs should be disabled.

doc/fips/SONiC-FIPS-POST.md
```diff
@@ -0,0 +1,89 @@
...
+
+The design must meet the following requirements:
+- In order to accommodate different forwarding ASIC architecture or SAI implementation, the design should support enabling POST at either switch level (during switch init) or at MACSec engine level (during MACSec engine init).
+- SONiC MACSec configuration must be processed only after POST passes. 
+- POST failure must not affect the operation of non-MACSec ports.
+- Explicit visibility must be provided if POST fails, for example, in syslog. The syslog message must include the details of the failure. For example, SAI object Id of ports that fail POST and the corresponding MACSec engine.
+
+## Deisgn details
+
+The following figure depicts the data flow and SONiC components in the design. Orchagent is responsible for triggering POST via SAI calls and publishing POST status in State DB.   MACSec container, precisely MACSecMgr, is enhanced to be POST aware and only process MACSec configuration after POST has passed.
```
wumiaont: New SAI contains POST for macsec and ipsec notification APIs. You will not need IPSEC for now. But all notifications need to add handles for IPSEC as well in sonic-sairedis when handling notification. Otherwise it will not compile.

You can mention IPSEC is out of scope with this HLD or you can take IPSEC into the picture as well. Such as using the same POST_STATUS_TABLE? or different tables? Or same table but different keys to distinguish IPSEC and MACSEC?Some chassis could support both IPSEC and MACSEC in the future. Need to leave room for IPSEC with this designing.

doc/fips/SONiC-FIPS-POST.md
```diff
@@ -0,0 +1,89 @@
...
+
+```
+
+### Enabling POST in SAI switch init
+
+Orchagent enables POST when creating SAI switch regardless of whether the MACSec feature is supported or enabled. This can avoid triggering POST after SAI switch creation, e.g., MACSec feature is enabled later. The below flow chart depicts the process.
+
+![](images/fips-post-switch-init.png)
+
+Orchagent enables POST when creating SAI switch. After SAI switch is created, Orchagent queries SAI POST capability. If POST is supported in switch init, Orchagent sets POST status to in-progress and waits for POST completion callback. Otherwise, if POST is supported only in MACSec init, Orchagent sets POST status to not-started, which informs MACSecOrch of performing POST in its initialization.
```
wumiaont: Will this cause race condition for modifying DB? At least need to have a way to avoid race condition happen. From the ASIC support for FIPS, the POST will be initiated during switch creation. It will trigger callback notification if notification API is passed. It's needed to make sure notification handling where orchagent will update the POST result to STATE db is always after you set in-progress here. I have a concern here of that if you set in-progress after creating switch will it be late?

doc/fips/SONiC-FIPS-POST.md
```diff
@@ -0,0 +1,89 @@
...
+
+This document describes SONiC design for Federal Information Processing Standards (FIPS) 140-3 standard compliance. Especially, the focus of the document is to trigger Pre-Operational Self-Test (POST) in SONiC and also ensure SONiC’s behavior is compliant to FIPS standard after POST.
+
+## Design requirements
+
+The design must meet the following requirements:
+- In order to accommodate different forwarding ASIC architecture or SAI implementation, the design should support enabling POST at either switch level (during switch init) or at MACSec engine level (during MACSec engine init).
+- SONiC MACSec configuration must be processed only after POST passes. 
+- POST failure must not affect the operation of non-MACSec ports.
+- Explicit visibility must be provided if POST fails, for example, in syslog. The syslog message must include the details of the failure. For example, SAI object Id of ports that fail POST and the corresponding MACSec engine.
```
wumiaont: The POST test is happening at switch creating or macsec creating time. I would say if POST failed, there might be no ports added to this macsec engine yet (From sonic macsec config you create macsec profile first and then add profile to the ports). How can the syslog shows which ports are failing POST?

ysmanman: When POST failed,  we can use SAI API to examine the POST status of HW macsec ports, not front panel ports, and syslog these ports accordingly.  Please see https://github.com/opencomputeproject/SAI/pull/2167 for details.  This syslog is mainly for providing information for ASIC vendors when POST fails.

doc/fips/SONiC-FIPS-POST.md
ysmanman: Can you elaborate more on the following? Did you mean the sairedis bump-up to point to SAI with POST support?

```But all notifications need to add handles for IPSEC as well in sonic-sairedis when handling notification. Otherwise it will not compile.```

File: doc/fips/SONiC-FIPS-POST.md (hunk 5 above)
Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: Thanks for the suggestion. Yes, it makes sense to move post status into existing FIPS_STATS table.

File: doc/fips/SONiC-FIPS-POST.md (hunk 1 above)
Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: @senthil-nexthop  the proposed POST status is a aggregated POST status, which is used by MACSecMgr to check if POST passes or not before it starts to process MACSec config. MACSecMgr does not really care about the POST status of individual ports of MACSec engin. It just needs to know if POST passed on all ports or not.

File: doc/fips/SONiC-FIPS-POST.md (hunk 6 above)
Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: @wumiaont With current SAI POST API, switch level POST needs to be triggered when creating SAI switch. There is no way to check if POST is supported or not before creating SAI switch because the query API also requires to pass in SAI switch object.  For vendors that needs MACSec but their chips does not have POST support, then enabling POST in switch or MACSec creation is just no-op, right? Further, the capability query API should also indicate POST is not supported by SAI. In this case, MACSec can still be used. Please refer to  the flowchart `fips-post-swith-init.png` for the details.

File: doc/fips/SONiC-FIPS-POST.md (hunk 9 above)
Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: @wumiaont yes, this is a valid point. I think we can set `in-progress` state *before* switch creation to avoid this race condition.  And after switch creation, we can update the state accordingly:
- If POST is supported in switch creation, wait for notification callback to update the state;
- If POST is *only* supported in MACSec creation, set the state to `not-started` to info MACSecOrch to performs POST;
- If POST is not supported at all, update the state to `unsupported`.

File: doc/fips/SONiC-FIPS-POST.md (hunk 7 above)
Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: @wumiaont can you clarify more on `FIPS is not enabled` ?  So SONiC already has a knob for FIPS?

File: doc/fips/SONiC-FIPS-POST.md (hunk 3 above)
Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: @senthil-nexthop  By `incremental support`, I meant vendor may add FIPS support in their SAI implementation later.

File: doc/fips/SONiC-FIPS-POST.md (hunk 2 above)
Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: @senthil-nexthop  MACSecMgr does not process any SONiC MACSec config until POST passed or is not supported.  When POST is in-progress,  OA is still waiting for the POST result from SAI, and thus no MACSec config should be processed because POST result is still unknown.

File: doc/fips/SONiC-FIPS-POST.md (hunk 4 above)
Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: @senthil-nexthop Yes, this is just to showing that MACSecMgr should not process any SONiC MACSec config if POST failed.

**Comment by abdosi** [comment]
@ysmanman : can you review the latest comments

@judyjoseph : will also check

File: doc/fips/SONiC-FIPS-POST.md (hunk 1 above)
Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: @senthil-nexthop the proposed POST status is a aggregated POST status, which is used by MACSecMgr to check if POST passes or not before it starts to process MACSec config. MACSecMgr does not really care about the POST status of individual ports of MACSec engin. It just needs to know if POST passed on all ports or not.

File: doc/fips/SONiC-FIPS-POST.md (hunk 7 above)
Comment (wumiaont) [reply doc/fips/SONiC-FIPS-POST.md]: > @wumiaont can you clarify more on `FIPS is not enabled` ? So SONiC already has a knob for FIPS?

On Sonic there's a flag to set the FIPS state of the device. By default FIPS is disabled on device. This can be done using the Sonic CLI "sudo sonic_installer set-fips --enable-fips (--disable-fips)". A reboot is required after set FIPS enable or disable to take effect. This flag is written in uboot

When FIPS is not enabled, crypto algorithms are using openssl provided ones. When FIPS is enabled, crypto will use MSFT Symcrypt which is undergo FIPS 140-3 certification (control plane).

So should consider only take POST test when FIPS is enabled.

File: doc/fips/SONiC-FIPS-POST.md (hunk 6 above)
Comment (wumiaont) [reply doc/fips/SONiC-FIPS-POST.md]: There's a flag on Sonic to indicate FIPS is enabled | disabled on device. Should combine this flag to do POST when FIPS is enabled on device.

File: doc/fips/SONiC-FIPS-POST.md (hunk 8 above)
Comment (wumiaont) [reply doc/fips/SONiC-FIPS-POST.md]: New SAI defined FIPS notifications including MACSEC and IPSEC port level and switch level POST status. When new SAI header file are included into Sonic (sairedis), it's needed to define the notification handlings. You need to add support for IPSEC notifications even if you are not going to support IPSEC at the moment. Otherwise you will not be able to compile sairedis. 

IPSEC handling is pretty similar to MACSEC. Just wondering if your HLD can take IPSEC also into the picture or you can just ignore the IPSEC at this moment. But when designing the database key for FIPS should leave room for future IPSEC support.

File: doc/fips/SONiC-FIPS-POST.md (hunk 6 above)
Comment (vikram-nexthop) [reply doc/fips/SONiC-FIPS-POST.md]: shoudn't the 'POST_STATUS_TABLE|macsec' status reflect the combined FIPS POST status of data-plane (SAI Switch) and the control-plane (crypto library)? Will the Orchagent enable both the POSTs?

Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: In order  to follow up on the HLD review,  we need to agree on how to notify OA that SAI POST should be enabled in switch or macsec init.  @wumiaont what's exactly the flag you suggested? Was it published in configDb before OA starts?

Comment (ysmanman) [reply doc/fips/SONiC-FIPS-POST.md]: @vikram-nexthop OA should only enable SAI POST.  It should be more cleaner to have a dedicated state table for MACSEC so that we can add both SAI and also, if needed, control-plane POST status in the table.  For example, the following table entry can be used to track SAI POST status:
```
FIPS_MACSEC_POST_TABLE|sai
```

doc/fips/SONiC-FIPS-POST.md
ysmanman: @wumiaont  Compiling sairedis does not require actual support of IPSEC POST notification. Please see https://github.com/sonic-net/sonic-sairedis/pull/1650 for details.  The HLD focus  on MACSec and IPSec is out of hte scope. Having separate state table for IPSec and MACSec may be better since it allows different sec type to have different set of POS status.  Also, separate state table can avoid potential multi-writer issue in the future.

doc/fips/SONiC-FIPS-POST.md
wumiaont: The flag on sonic is set/read using cli "sudo sonic-install get-fips" "sudo sonic-install set-fips --enable-fips" and follow a reboot. After boot up the FIPS state is saved on STATE_DB "FIPS_STATS|state". You can query this table to get the fips state. Be aware this table does not have multi-asic support. If you are usning multi-asic need to add support to set this table on namespace DB also. Related code is in sonic-host-service/scripts/procdockerstatsd.

ysmanman: I updated the HLD based on our discussion in https://github.com/sonic-net/sonic-swss/pull/3836.  So Orchagent will trigger SAI POST if FIPS is enabled in either /proc/cmdline or /etc/fips/fips_enable .

doc/fips/SONiC-FIPS-POST.md
ysmanman: @wumiaont updated HLD to enable SAI POST only when FIPS is enabled.

ysmanman:
> @ysmanman : can you review the latest comments
> 
> @judyjoseph : will also check

All open comments are addressed.

abdosi:
@ysmanman : please add SAI PR link here also that point header change to support this.

abdosi:
@eswaran-nexthop  and @senthil-nexthop please approve it so that we can get it merged.

ysmanman:
> @ysmanman : please add SAI PR link here also that point header change to support this.

Added in PR description.

abdosi:
Another set of supported PR:
wpa-supplicant changes: https://github.com/sonic-net/sonic-wpa-supplicant/pull/99
show command: https://github.com/sonic-net/sonic-buildimage/pull/24067

abdosi:
please update/extend the HLD for above changes for wpa-supplicant.

abdosi (approved):

PR #2039, opened 2025-07-15
HLD for SAI Validation and PySAI

wuzy0318:
SAI Validation is a Python-based framework for testing and developing SAI C APIs. It fills two key gaps:
1. Single-Device Testing: Enables thorough API validation on one DUT without needing complex topologies or RPC servers.
2. All-In-One Toolkit: Combines features like sairedis.rec recording/replay and additional diagnostic tools into one package, streamlining both issue investigation and API development.


doc/SAI_validation/SAI_Validation_HLD.md
```diff
@@ -0,0 +1,98 @@
...
+#### 6.2 PySAI
+PySAI provides a concise set of Python APIs, which helps SAI Validaiton avoid invoking complicated SAI API in C
+<div align="center">
+  <img width="600" alt="pysai_code" src="https://github.com/user-attachments/assets/d30b22cd-3168-4d85-8664-33e7ef28c490" />
+  <br><em>Figure 3. SAI Code VS PySAI Code</em>
+</div>
+
+
+##### 6.2.1 PySAI Code-gen
+1. Parsing: parse.py uses libclang to extract enums, structs, and function signatures from SAI headers. These are mapped to Pydantic models as intermediate representation.
```
rck-innovium, line 78: This is a very promising proposal and holds a lot of potential. Please consider presenting this in the OCP SAI community forum as well.
1) Can you provide some details about the design choice of say not directly converting from SAI C headers to Python API say by using ctypesgen and then build higher layer Python classes on top of Ctypes APIs?
2) Do you plan to use the intermediate artefacts like the Pydantic models for some other purpose in the future?

doc/SAI_validation/SAI_Validation_HLD.md
```diff
@@ -0,0 +1,98 @@
...
+  <img width="600" alt="sai_validation_comparison" src="https://github.com/user-attachments/assets/0153f7e4-6781-41e1-bf81-d01d5df70e04" />
+  <br><em>Figure 1. Comparison of PTF, SAI Challenger and SAI Validation</em>
+</div>
+
+### 5. Requirement / Goals
+
+1. The testbed should be self-contained, using only a single switch without dependence on external devices like a fanout switch or server.
+2. Call SAI APIs in the same way as syncd, to reflect the real-world performance of SAI API invocations in SONiC.
+3. Provide interactive mode for step-by-step debugging and incremental developing.
+4. Support sairedis.rec recording and replay for the test process, with automatic filtering and isolation of entries based on relevance.
```
rck-innovium, line 38: As discussed in the community meeting, please explain how a SONiC sairedis.rec log can be replayed to PySAI. Please explain how the Virtual OIDs are converted to Real OIDs.

wuzy0318: Our approach to converting Virtual Object IDs (VIDs) to Real Object IDs (RIDs) mirrors the strategy used by sonic-sairedis. During replay of a sairedis.rec log in PySAI, we maintain a dedicated mapping between VIDs and RIDs:

- For each object creation operation, we associate the VID referenced in the sairedis.rec with the actual RID returned by [libsai.so](http://libsai.so/) after the operation is completed.
- For set and update operations, we look up the corresponding RID for each VID in our mapping, and pass the request using the RID to [libsai.so](http://libsai.so/) for accurate execution.

doc/SAI_validation/SAI_Validation_HLD.md
wuzy0318: Thanks for your suggestions! Here are the answers
1. Our code generator produces direct Python bindings to all SAI C APIs using ctypes, enabling low-level access to the interface. However, to improve usability for developers, especially when working interactively or managing dependencies in object-oriented environments such as IPython, we also provide higher-level SAI object abstractions. These classes wrap the core functionality—including object creation/removal, bulk operations, attribute management, and statistics—making it simpler and more intuitive to work with SAI directly from Python.

2. We plan to leverage the intermediate artifacts, especially the Pydantic models, for future use. Pydantic schemas make it explicit which flags and attribute values are valid, supporting better validation and code clarity when handling SAI structures and comments. The current OCP-SAI code generation primarily uses Perl scripts, which can be difficult to extend and maintain. By introducing YAML files that formally specify SAI models, we can decouple the generation logic for various tools (Thrift, SWIG, serialization, etc.), streamline development, and facilitate new integrations—such as generating boilerplate for sonic-sairedis.

FPOOOOOOO-ali:
1、Has there been any evaluation or measurement of the performance overhead when converting C to Python?
2、Regarding the two testing frameworks you compared, OCP SAI (PTF) already has a significant number of test cases. Can these test cases be migrated to SAIvalidation with a one-click approach?
3、Why not choose the existing OCP SAI thriftv2 framework and optimize it instead? Are there any additional considerations behind this decision?

rck-innovium:
As suggested during the community review, this proposal need not have any dependency on SONiC. So, instead of SONiC, can we place this in [SAI](https://github.com/opencomputeproject/SAI/). This could get help increase the adoption of this framework beyond SONiC users.

PR #2050, opened 2025-07-31
SONiC MACsec FIPS 140-3 HLD

rajshekhar-nexthop:


doc/fips/SONiC-MACsec-FIPS-140-3.md
```diff
@@ -0,0 +1,446 @@
...
+- Secure memory handling using madvise(), mlock(), and explicit\_bzero()  
+- Authentication of all client-server communications  
+- Secure handling of process termination and zeroization
+
+#### 5.2. FIPS POST Requirements:
+
+- MACsec configuration must be blocked until POST validation passes.  
+- Use FIPS POST results provided by OpenSSL/SymCrypt for KMS and MACsec control-plane POST.  
+- Use SAI FIPS POST results from the switch for data-plane POST.  
+- Continuous monitoring support.
```
rajshekhar-nexthop, line 66: Expand

rajshekhar-nexthop: Ignore for now

doc/fips/SONiC-MACsec-FIPS-140-3.md
```diff
@@ -0,0 +1,446 @@
...
+
+##### 7.1.5.  SAI API
+
+**No impact**
+
+#### 7.2. FIPS POST
+
+The KMS library will fetch FIPS POST results from OpenSSL/SymCrypt during initialization to ensure the required FIPS compliance. The library will be usable/enabled only if the validation succeeds. A failure will force the library to move into a disabled state. In this state, any subsequent API calls will fail.
+
+The POST status can be published in the *‘POST\_STATUS\_TABLE’* table, referenced in SONiC FIPS POST support HLD.
```
rajshekhar-nexthop, line 302: Separate key for control plane status? Just for observability/CLI not for decision making

doc/fips/SONiC-MACsec-FIPS-140-3.md
```diff
@@ -0,0 +1,446 @@
...
+  ┌───────────────────────────────────────────────┐  
+  │          Publish POST STATUS.                 │  
+  └───────────────────────────────────────────────┘
+```
+
+### 8\. Testing Requirements/Design
+
+Testing ensures FIPS POST and KMS integration maintains existing SONiC functionality while providing secure key management and FIPS POST compliance.
+
+#### 8.1. Unit Test Cases
```
rajshekhar-nexthop, line 371: Remove test cases

deepak-singhal0408:
@kenneth-arista @mlok-nokia Please help review/forward to your team for review..

mlok-nokia:
@wumiaont  Please take a look at this

doc/fips/SONiC-MACsec-FIPS-140-3.md
```diff
@@ -0,0 +1,357 @@
...
+    "auth_key": "35FC8F2C81BCA28A95845A4D2A1EE6EF",  // unchanged
+    "lowest_acceptable_pn": "1",                       // unchanged
+    "ssci": "0",                                       // unchanged
+    "salt": "000000000000000000000000"                 // unchanged
+}
+```
+
+**Key Changes**:
+
+- Encrypted SAK will be 60 bytes long and will contain IV(12) | Ciphertext (32) | Tag (16)
```
wumiaont, line 213: Can you explain a little bit more here why you want the SAK to be IV|Ciphertext|tag not just encrypt of plaintext SAK?Also how would the client get the IV and tag before encrypting the keys from wpa_supplicant? 

Also do you need to encrypt the auth keys and salt in DB?

rajshekhar-nexthop: Hi @wumiaont ,
Thanks for the review. Please find below answers inline

First question:
AES-GCM decryption requires all three components:
IV: Must be unique per encryption, prevents pattern leakage
Tag: Detects tampering/corruption during decryption
Ciphertext: The encrypted data
Without IV → can't decrypt. Without tag → can't verify integrity.

Second question:
Client doesn't need to get them - KMS generates automatically:
kms_encrypt() generates random IV + ciphertext + tag, returns complete blob
kms_decrypt() extracts IV/tag from blob automatically
wpa_supplicant just stores/retrieves the blob, never handles IV/tag directly

Third question:
Auth keys need to be encrypted in DB. Salt not a secret ( just random value), hence no need to encrypt it. Are you asking these in context of IPSec?

wumiaont: Thanks. Another question is that is this project been raised by you because there's some issue you experienced during the FIPS 140-3 certification?

rajshekhar-nexthop: Yes, that is correct.

doc/fips/SONiC-MACsec-FIPS-140-3.md
```diff
@@ -0,0 +1,357 @@
...
+  │    │  └───────────────────┘    └───────────────────────┘ │   │  
+  │    └─────────────────────────────────────────────────────┘   │  
+  └──────────────────────────────────────────────────────────────┘
+```
+
+### 7\. High-Level Design
+
+#### 7.1. Key Management Service
+
+KMS is a Key Management Service library that provides secure key distribution for SONiC MACsec implementations. The library uses a client-server architecture where syncd acts as the KMS server hosting KEKs, while wpa\_supplicant processes act as clients that receive KEKs for SAK encryption. KMS server generates KEK on the first client request and uses the same KEK for subsequent requests. The library implements multiple layers of memory protection including mmap allocation, mlock to prevent swapping, and madvise to exclude sensitive data from core dumps. All sensitive memory is explicitly zeroed using explicit\_bzero before deallocation. Communication occurs through UNIX domain sockets with restricted permissions and message integrity validation. Encrypt and decrypt functionality is implemented using OpenSSL APIs backed by the SymCrypt cryptographic provider.
```
wumiaont, line 143: Should you also provide a commands (sonic cli) to zeroize sensitive memory holding key when user is required to do so? FIPS 140-3 requires zeroization to ensure that sensitive information within a cryptographic module is destroyed when the security of the module is compromised. Of course a reboot should take care of it as new key will be generated. Wondering if providing a zeroization method is necessary.

doc/fips/SONiC-MACsec-FIPS-140-3.md
```diff
@@ -0,0 +1,357 @@
...
+}
+```
+
+**Modified Schema (with KMS)**:
+
+```json
+// APP_DB MACsec SA entry (with KMS):
+"MACSEC_EGRESS_SA_TABLE:Ethernet1:5254008f4f1c0001:1": {
+    "sak": "60_byte_hex_string",
+    "auth_key": "35FC8F2C81BCA28A95845A4D2A1EE6EF",    // unchanged
```
wumiaont, line 195: Does auth_key need encryption here? If yes, please modify this part.

doc/fips/SONiC-MACsec-FIPS-140-3.md
```diff
@@ -0,0 +1,357 @@
...
+| FIPS | Federal Information Processing Standards |
+| RNG | Random Number Generator |
+| DRBG | Deterministic Random Bit Generator |
+| POST | Pre-Operational Self-Test  |
+
+### 4\. Overview
+
+The SONiC MACsec FIPS HLD defines a comprehensive security framework that enables FIPS 140-3 compliance for MACsec deployments in SONiC network devices. The primary goal is to establish a validated cryptographic infrastructure that meets federal security standards through secure key management, cryptographic module validation, and operational compliance monitoring. 
+
+KMS is a generic library which provides a secure key management infrastructure for SONiC network devices. The primary goal is to protect sensitive cryptographic material (example MACsec SAKs) by centralizing key operations within the library using client server architecture. KMS is required to achieve FIPS compliance in SONiC FIPS mode by ensuring cryptographic keys are never stored in plaintext outside of the secure boundary.
```
null0link, line 43: Will "KMS" be certified for FIPS 140-3? Even if yes, cryptographic keys (CSPs) can be outputted from the cryptographic module in a plaintext form for Security Levels 1 and 2. Reference FIPS 140-3 IG 9.5.A.
Which FIPS comilance this HLD is trying to achieve?
I do not think that MACsec keys, due to their limited lifetime (unless they are PSKs), will be stored outside of the RAM. If keys are stored in processes memory areas only, then the Operational Environment requirements are enough for FIPS compliance, and I think that encryption of the ephemeral keys stored in RAM is redundant.

doc/fips/SONiC-MACsec-FIPS-140-3.md
```diff
@@ -0,0 +1,357 @@
...
+- All cryptographic operations use AES-256-GCM for encryption/decryption  
+- All random number generation use CTR\_DRBG/HASH\_DRBG/HMAC\_DRBG with security strength 256bits  
+- Secure memory handling using madvise(), mlock(), and explicit\_bzero()  
+- Authentication of all client-server communications  
+- Secure handling of process termination and zeroization
+
+#### 5.2. FIPS POST Requirements:
+
+- MACsec configuration must be blocked until POST validation passes.  
+- Use FIPS POST results provided by OpenSSL/SymCrypt for KMS and MACsec control-plane POST.
```
null0link, line 64: I think that deriving MACsec keys by using FIPS-certified KBKDF implementation ( NIST SP 800-108rev1 ) is an important control plane requirement for FIPS inside approach compliance.

doc/fips/SONiC-MACsec-FIPS-140-3.md
```diff
@@ -0,0 +1,357 @@
...
+
+KMS is a generic library which provides a secure key management infrastructure for SONiC network devices. The primary goal is to protect sensitive cryptographic material (example MACsec SAKs) by centralizing key operations within the library using client server architecture. KMS is required to achieve FIPS compliance in SONiC FIPS mode by ensuring cryptographic keys are never stored in plaintext outside of the secure boundary.
+
+POST compliance ensures that cryptographic modules in SONiC meets FIPS 140-3 requirements before processing any MACsec operations. The primary goal is to validate the integrity and functionality of cryptographic hardware and software components during system initialization, preventing the use of potentially compromised or non-compliant crypto modules. 
+
+### 5\. Requirements
+
+#### 5.1. KMS Requirements:
+
+- Secure encryption/decryption of key material using KEK
```
null0link, line 51: IIUC KMS will not generate MACsec keys, it will accept them as an input from other processes in plaintext form?

null0link: From the flow in `7.1.3. Sequence Diagrams for KMS integration` it looks like the KMS will only deliver a master key which other applications will use to encrypt their secrets. Encrypted secrets will land in the database. Then KMS is needed again to decrypt the carried secret. So overall it is about storing SAKs in database in encrypted form AFAIU.

doc/fips/SONiC-MACsec-FIPS-140-3.md
rajshekhar-nexthop: Yes, auth_key also need encryption. Thanks for catching this. Will update the HLD

doc/fips/SONiC-MACsec-FIPS-140-3.md
eswaran-nexthop: Copy-pasting commentary from the FIPS consultant Nexthop used

```
FIPS 140-3 IG 9.5.A allows plaintext CSPs/SSPs to be entered or output at Security Levels 1 and 2 for software and hybrid software modules  provided that the CSPs remain within the Operational Environment (OE) and that OE protections (process isolation, memory management etc.) are enforced. From that perspective, if Redis is strictly local  non persistent  and access controlled plaintext storage of ephemeral SAKs could be justified under Level 1 compliance.

My recommendation that SAKs should be encrypted before entering Redis came from a different guidance point   IG 9.6.A (Acceptable Storage of SSPs). This guidance explains that CSPs stored outside of the module boundary should be protected using Approved methods. When CSPs are protected in this way, they are treated as protected CSPs which basically means:

- Redis can be cleanly excluded from the cryptographic boundary, since it only ever holds ciphertext.
- Zeroization requirements (AS09.28/29) no longer apply to those Redis-resident values, as they are not unprotected secrets.
- Our narrative is simplified since we no longer need to rely on proving Redis is properly configured and operating as part of the OE.
- The design is future proofed if the module ever targets Level 2+ validation 
```

IOW, @null0link's comment is correct. We do not technically need this for FIPS 140-3 level1 certification. We can discuss in the community if it's worth doing this as a more conservative measure.

judyjoseph: Another viewpoint which I have -- seeing the initial comment of do we need "KMS" certification for FIPS 140-3 --  is, Can we use an alternative channel other than redis to pass the SAK key for macsec_orch to consume? We could avoid storing SAK in redis itself

doc/fips/SONiC-MACsec-FIPS-140-3.md
rajshekhar-nexthop: Yes, thats correct @null0link

doc/fips/SONiC-MACsec-FIPS-140-3.md
rajshekhar-nexthop: Hi @judyjoseph ,
Yes, alternative mechanisms to avoid Redis have been evaluated, including the use of the Linux kernel keyring for storing SAKs and direct Unix domain sockets between processes. While these approaches provide stronger isolation, they introduce additional code churn and reduce portability. To maintain a generic, reusable framework for broader cryptographic use cases, we determined that a KMS-based approach aligns better with our long-term design goals. Given that FIPS 140-3 Level 1 permits storage of SSPs in plaintext within the operating environment, we have deferred adoption of these alternatives or KMS for now, with a plan to revisit them later when we target higher assurance levels.

doc/fips/SONiC-MACsec-FIPS-140-3.md
rajshekhar-nexthop: Hi @null0link ,
Thank you for the inputs. We are indeed evaluating the integration of a FIPS-certified KBKDF implementation (for example, through the OpenSSL FIPS provider or SymCrypt in FIPS mode) directly into the control plane that manages MACsec.

abdosi:
It seems this HLD is not needed for FIPS compliance. More clarity next week on this.

rajshekhar-nexthop:
Closing this PR as FIPS 140-3 Level 1 permits storage of SSPs in plaintext within the operating environment. We have deferred adoption of KMS for now, with a plan to revisit it later when we target higher assurance levels.

PR #2060, opened 2025-08-16
RIB FIB HLD

eddieruan-alibaba:
This HLD describes why we need to introduce RIB/FIB in SONiC and how we approach it. 

Phrase 1 target: routes download 

Repo | PR title | State
-- | -- | --
SONiC | [[doc/ribfib]: Add RIB/FIB LLD document #2275](https://github.com/sonic-net/SONiC/pull/2275) | GitHub issue/pull request detail
SONiC | [[doc/ribfib]: Add sonic-fib Low Level Design Document #2274](https://github.com/sonic-net/SONiC/pull/2274) | GitHub issue/pull request detail
sonic-buildimage | [Add sonic-fib library to build libnexthopgroup deb for RIB/FIB#26423](https://github.com/sonic-net/sonic-buildimage/pull/26423)  | GitHub issue/pull request detail
sonic-buildimage | [dplane_fpm_sonic: build and manage dplane NHGs from route events #29200](https://github.com/sonic-net/sonic-buildimage/pull/29200)  | GitHub issue/pull request detail
sonic-swss | [[fpmsyncd]: add nhgmgr to support RIB/FIB design #4395](https://github.com/sonic-net/sonic-swss/pull/4395) | GitHub issue/pull request detail
SONiC | [Add RIB/RIB nhg mgr LLD #2270](https://github.com/sonic-net/SONiC/pull/2270) | GitHub issue/pull request detail
sonic-swss | [[fpmsyncd]: add libnexthopgroup in Makefile.am to support rib/fib #4394](https://github.com/sonic-net/sonic-swss/pull/4394) | GitHub issue/pull request detail
sonic-swss | [[fpmsyncd]: Add RIB/FIB processing support #4419](https://github.com/sonic-net/sonic-swss/pull/4419) | GitHub issue/pull request detail
sonic-swss | [[fpmsyncd]: Add Support for SRv6 processing based on RIB/FIB #4420](https://github.com/sonic-net/sonic-swss/pull/4420) | GitHub issue/pull request detail
sonic-swss | [[fpmsyncd]: Support fpmsyncd to use APP_STATE DB #4421](https://github.com/sonic-net/sonic-swss/pull/4421) | GitHub issue/pull request detail


eddieruan-alibaba:
https://github.com/eddieruan-alibaba/SONiC/blob/ribfib/doc/ribfib/ribfib.md

StormLiangMS:
Could we add a knob to enable this feature selectively, based on demand? The reason I ask is that the NHG process between kernel and zebra has not been stable—we’ve had it disabled since 202405 to address a production issue (see https://github.com/sonic-net/sonic-buildimage/pull/23459
). I can imagine this new design will introduce additional complexity, so until it matures and is qualified in production, it would be helpful to have a knob to control it.

selva-nexthop (commented):
I have one design question regarding where the NHG block is being maintained.  Currently, this HLD proposes to keep the NHG block in fpmsyncd and persist it in Redis for warm restart reconciliation. While this works for the FRR-based stack, it introduces a few long-term architectural risks:

Tight Coupling to FRR:
I am not sure about other deployments of SONiC using different RIB entities.  But modularization-wise, just want to understand the rationale of the placement of NHG blcok.  
Maintaining NHG mapping in fpmsyncd makes the design dependent on FRR as the RIB. If in future deployments another routing entity (or alternate northbound programming mechanism) is used instead of FRR, the cache management logic would need to be duplicated in other agents

Placement of State Ownership:
Conceptually, NHG reconciliation feels closer to orchagent, since orchagent is the consumer responsible for applying NHG state into ASIC DB. Having the reconciliation state live in fpmsyncd risks splitting ownership of critical NHG state between two daemons.

Redis as Central Persistency:
Persisting the cache in Redis is fine for warm restart, but the ownership boundary still matters. If orchagent owns NHG lifecycle, it would be more consistent for it to also own the NHG cache, while Redis just acts as the persistence layer.

eddieruan-alibaba:
> I have one design question regarding where the NHG block is being maintained. Currently, this HLD proposes to keep the NHG block in fpmsyncd and persist it in Redis for warm restart reconciliation. While this works for the FRR-based stack, it introduces a few long-term architectural risks:
> 
> Tight Coupling to FRR: I am not sure about other deployments of SONiC using different RIB entities. But modularization-wise, just want to understand the rationale of the placement of NHG blcok. Maintaining NHG mapping in fpmsyncd makes the design dependent on FRR as the RIB. If in future deployments another routing entity (or alternate northbound programming mechanism) is used instead of FRR, the cache management logic would need to be duplicated in other agents

EDDIE: 
FPM is Forwarding Plane Manager. It is  a component of FRR that provides a mechanism to export routing updates (like route additions, modifications, or deletions) from FRR’s internal Routing Information Base (RIB) to an external forwarding plane.

Once we use FPM, it is tight coupled with FRR. 

Currently, I don't see any valid use cases to replace FRR to somethingelse. I would not worry about this tight couple to FRR, since it is current arch design. Any change in this arch design would need to be discussed in TSC. 



> 
> Placement of State Ownership: Conceptually, NHG reconciliation feels closer to orchagent, since orchagent is the consumer responsible for applying NHG state into ASIC DB. Having the reconciliation state live in fpmsyncd risks splitting ownership of critical NHG state between two daemons.

EDDIE: 
We map zebra NHG information to various SONiC APP DB objects here., SONiC NHG OBJ is just one of them.  Currently, this step is handled in zebra, now we want to move fpmsyncd to provide FIB functionality. 



> 
> Redis as Central Persistency: Persisting the cache in Redis is fine for warm restart, but the ownership boundary still matters. If orchagent owns NHG lifecycle, it would be more consistent for it to also own the NHG cache, while Redis just acts as the persistence layer.

EDDIE: We discussed it in WG before. Current logic, during warmreboot, fpmsyncd would check if routes information are changed or not and skip routes updating to avoid unnecessary hardware update. We want to keep this design philosophy as well. 

Redis is an expensive access. We want to keep only SONIC OBJs in there. FIB would break RIB information into corresponding SONiC OBJs.

eddieruan-alibaba:
> Could we add a knob to enable this feature selectively, based on demand? The reason I ask is that the NHG process between kernel and zebra has not been stable—we’ve had it disabled since 202405 to address a production issue (see [sonic-net/sonic-buildimage#23459](https://github.com/sonic-net/sonic-buildimage/pull/23459) ). I can imagine this new design will introduce additional complexity, so until it matures and is qualified in production, it would be helpful to have a knob to control it.

That is a valid ask. We don't plan to remove current code path. If NHG usage is disabled, we will use current code path. But if NHG usage is enabled, we will go with RIB/FIB path.

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,386 @@
...
+![image](images/zebra_nhg_chain.png)
+
+Once the resolve-through and resolve-via information is available, the SONiC forwarding chain can be represented as shown in the following graph.
+
+![image](images/SONiC_nhg_chain.png)
+
+Note that we use dotted lines to represent dependencies because we plan to use ID lookups to identify dependent NHGs rather than pointers. This approach makes the code more robust.
+
+## Resolved NHG and Unresolved NHG
+Unresolved NHG is also known as original NHG. FRR team plans to keep these NHG to provide information coming from protocol clients. In most cases, we use resolved NHG in SONiC. But in SRv6 VPN case, we use unresolved NHG due to SONiC expects tunnel NH to be provided, a.k.a unresolved NHG here.
```
pbrisset: can you add clear definition of resolved and unresolved NHG? Terminology.

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,386 @@
...
+Similar to the above example, except we use v6 address as next hop. Assume vrf foo 100.0.0.0/24 via 1::1 with vpn SID A and vrf bar 100.0.0.0/24 via 1::1 with vpn SID B. 1::1/128 via final nexthop 2::2, 3::3 and 4::4
+
+![image](images/Srv6_vpn.png)
+
+Similar to the previous example, the green section represents the logical configuration, while the grey section shows the RIB chain. In the SRv6 VPN case, the unresolved NHG 200 is used to create a PIC NHG 400 with 1::1 as the nexthop. A PIC CONTEXT object is also created for VPN SID A. The unresolved NHG 400 then references PIC NHG 400 and creates an additional PIC CONTEXT object for VPN SID B.
+
+# Route Congernce Handling
+A key objective of introducing the RIB/FIB model is to enhance route convergence. The design principle is to perform rapid updates on affected NHGs by removing failed paths based on existing NHG information. This mechanism mitigates traffic loss and provides sufficient time for routing protocols to complete reconvergence.
+
+## NHT Trigger
```
pbrisset, line 617: can an example of local failure e.g., interface failure, be added and show how that work?

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,386 @@
...
+                                             via fc08::2, Ethernet4, weight 1, 3d02h57m
+```
+
+The Zebra chain can be represented as the following graph. NHG 239 and NHG 244 (highlighted in yellow) are the final next-hops for fc06::2 and fc08::2. NHG 243 (in light green) is used by the routes 2064:100::1d/128 and 2064:200::1e/128. The four NHGs shown in light brown are dependents of NHG 243, while the NHGs in light blue are dependents of the light brown NHGs.
+
+Note that the current NHG structure only contains top-level next-hops and final resolved next-hops. For instance, NHG 275 does not include intermediate “via” path information. After discussions with the FRR team, they proposed adding “resolve through” and “resolve via” mechanisms to better maintain these dependencies.
+
+![image](images/zebra_nhg_chain.png)
+
+Once the resolve-through and resolve-via information is available, the SONiC forwarding chain can be represented as shown in the following graph.
```
pbrisset, line 561: can you add clear definition of resolved-through  and -via? Terminology is missing

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,386 @@
...
+Currently, Zebra leverages struct zebra_dplane_ctx to transfer information from the main thread to the FNC thread. The FNC thread then converts Zebra information into SONiC object representations and forwards them to fpmsyncd.
+
+![image](images/process_model.png)
+
+Going forward, NHG-related events will be updated so that struct zebra_dplane_ctx is passed directly to fpmsyncd, while the handling of other Zebra object events remains unchanged. Upon receiving such a message, fpmsyncd will invoke functions from the NHG module to translate struct zebra_dplane_ctx into the corresponding SONiC objects. Processing for all other events will continue to follow the existing workflow.
+
+In general, the FIB is responsible for handling both NHG and route events. However, since SONiC’s slow path relies on the Linux kernel and does not process route information in fpmsyncd, the SONiC FIB primarily handles NHG events. The code in fpmsyncd responsible for processing Zebra NHG events is referred to as the NHG block.
+
+## NHG Block
+![image](images/NHG_block.png)
```
pbrisset: I believe there are few things missing here:
1- a clear explanation about the difference of programming information from zebra to kernel versus SONiC. 
2- a section describing the assumptions / expectations of what FRR is providing (old approach vs new approach)
3- Clear API definition about the extension coming from FRR
4- a section describing why RIB/FIB design address warm reboot requirements and how.
5- fpmsyncd managing a new NHG-ID - why is that required? why are NHG-ID provided by zebra cannot be used?

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,386 @@
...
+* Step 3: For NHG 265, which does not involve 2064:100::1d, the walk can stop at this node.
+* Step 4: From NHG 260, continue the walk to NHG 264. Since NHG 264 has two paths and one of them involves 2064:100::1d, an update is triggered in APPDB by rewriting its NHG with the remaining valid paths from the forward walk.
+* Further steps: The walk then proceeds from NHG 260 onward, and the process continues recursively.
+
+PIC core cases would be handled in this walk.
+
+### Backwalk Step 2
+Fpmsyncd uses 2064:100::1d to trigger a lookup in NEXTHOP to SONIG NHG ID hash table. This lookup returns a list of SONiC NHGs which contains 2064:100::1d as its nexthop.
+
+![image](images/backwalk_2.png)
```
pbrisset, line 687: seems like this picture defer from previous one with a new NHG-ID A. This is confusing..

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,386 @@
...
+via fc06::2, Ethernet12 (vrf default), weight 1
+via fc08::2, Ethernet4 (vrf default), weight 1
+Client list: bgp(fd 71)
+```
+The following information would be sent to SONiC via dplane
+
+* Nexthop address : used in walk spec or for SONiC NHG table walk, a.k.a PIC edge case.
+* Its current resolved NHG ID : this NHG id is used to trigger backwalk update , a.k.a PIC core case.
+
+## Backwalk infra
```
pbrisset, line 648: I don't understand why a backwalk function is required. As soon a NHG-ID is updated, nothing else is required. Isn't the whole purpose of this?

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,386 @@
...
+
+The introduction of PIC (Prefix Independent Convergence) further complicates matters, requiring forwarding chains that may differ from routing chains. The FRR community reverted PR #16879 due to concerns that adding this complexity to Zebra would make it harder to serve different data planes.
+
+Given these challenges, the FRR community recommends redefining Zebra to function solely as the RIB, delegating FIB management to each data plane. This allows forwarding chains to be derived and optimized according to the specific requirements of each data plane implementation.
+
+## FIB's location
+The FIB functionality in SONiC can be introduced at two possible points: before APPDB, within fpmsyncd, or after APPDB, within orchagent. Following discussions in the Routing Working Group, the current consensus is to implement FIB functionality in fpmsyncd.
```
pbrisset: I'm not convince about this. I'm not sure if FIB should be part of fpmsyncd. Question: do we need to change APP_DB object and orchagent to process in a much better way NHG-ID coming from FRR since the structure is better than before. (not flatten anymore). I would have expect changes due to that.

pbrisset:
Regarding the comment about FRR and FIB manager being tightly coupled, I must agree with it. FIB manager should also work for controllers taping on SONiC. 

Perhaps, Requirements should be clearly stated at the beginning of this document.

doc/ribfib/ribfib.md
eddieruan-alibaba: Use resolved and original instead and added them in Definition section

| Resolved NHG | NHG resolved by Zebra, a.k.a containing nexthops with final outgoing interfaces and nexthop addresses  |
| Original NHG | NHG information received by Zebra from protocol clients  |

eddieruan-alibaba:
> Regarding the comment about FRR and FIB manager being tightly coupled, I must agree with it. FIB manager should also work for controllers taping on SONiC.
> 
> Perhaps, Requirements should be clearly stated at the beginning of this document.

I added a section at the very beginning of this document to clarify these discussions are not in the scope of this document. 

https://github.com/eddieruan-alibaba/SONiC/blob/ribfib/doc/ribfib/ribfib.md#topics-not-in-the-scope

doc/ribfib/ribfib.md
eddieruan-alibaba: It is required.

In the backwalk step 1 example,  if 2064:100::1d is withdrawn by IGP ( eBGP in this case), you will get an NHT event on 2064:100::1d  with NHG at 243. The event for updating 243 is not enough. We need to backwalk from 243 to other depending NHG and updating involved NHG. For example, we need to update 265, but no need to update 260.

doc/ribfib/ribfib.md
eddieruan-alibaba: ## Example of backwalk 

This section is used to give example on how convergence gets handled.

doc/ribfib/ribfib.md
selva-nexthop: Explaining the problem in detail and how backwalk is solving it would be helpful.

doc/ribfib/ribfib.md
eddieruan-alibaba: Updated

doc/ribfib/ribfib.md
eddieruan-alibaba: https://github.com/eddieruan-alibaba/SONiC/blob/ribfib/doc/ribfib/ribfib.md#fibs-location

The main idea is to determine prefix' SONiC NHG ID before updating to APPDB.  This way, we could avoid too many unnecessary routes updates to APPDB.

doc/ribfib/ribfib.md
eddieruan-alibaba: https://github.com/eddieruan-alibaba/SONiC/blob/ribfib/doc/ribfib/ribfib.md#example-of-backwalk

Here is an example to explain how backwalk is used. It is a very common method in normal FIB handling.

doc/ribfib/ribfib.md
eddieruan-alibaba: 1. Please take a look https://github.com/eddieruan-alibaba/SONiC/blob/ribfib/doc/ribfib/ribfib.md#handling-srv6-vpn-forwarding-chain-different-from-linux
2. Please take a look https://github.com/eddieruan-alibaba/SONiC/blob/ribfib/doc/ribfib/ribfib.md#fib-high-level-design I added a section to explain the differences in term of data structures in two approaches.
3. I will let Mark to handle them for FRR design.
4. and 5 are discussed in WG many times. I attached some meeting minutes links in https://github.com/eddieruan-alibaba/SONiC/blob/ribfib/doc/ribfib/ribfib.md#enable-nhg-id-handling-for-improving-route-convergence

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,489 @@
...
+#### Why we can't use FRR to handle NHG ID persistence
+We discussed this with the FRR team. But FRR team believes that NHG ID persistence is a data plane requirement and therefore it should not be managed by FRR. They recommended leveraging the Linux kernel as a mechanism to maintain NHG ID persistence. However, our proof of concept try revealed that the kernel lacks the necessary information to reliably restore NHG IDs in all scenarios. Therefore, we can't use FRR to handle NHG ID persistence.
+
+#### Use fpmsyncd to manage NHG ID persistence
+The current conclusions from working group discussions are
+* We can't retreat the same NHG ID table from Linux kernel, which leads Zebra can't recover zebra NHG ID mapping table during warm reboot.
+* FRR team doesn't want to add addtional complexity to handle this persistency requirement since it is a data plane requirement.
+* We don't want to keep NHG ID assignment in orchagent for avoiding unnecessary route events trigger redundant hardware programming.
+
+With these conclusions in mind, we conclude to manage the persistence of this zebra NHG ID to SONiC NHG ID mapping table within fpmsyncd.
```
selva-nexthop, line 151: is this in-memory table or persisted in redis?

eddieruan-alibaba: In memory table to avoid extra DB overhead.

selva-nexthop: the SONiC NHG ID will be used to update the app_db right?  if it is in-memory, how this ID will be persisted after BGP container restart?

selva-nexthop: or is it like SONiC NHG ID will be idempotent based on any hashing mechanism?

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,489 @@
...
+
+This topic has been discussed multiple times in routing working group. Here are a list of meeting minutes related to this topic.
+
+* 03/07 https://lists.sonicfoundation.dev/g/sonic-wg-routing/wiki/36235 NHD_ID handling approach during warm reboot.
+* 03/14 https://lists.sonicfoundation.dev/g/sonic-wg-routing/wiki/36288 NHD_ID handling approach during warm reboot.
+* 01/16 https://lists.sonicfoundation.dev/g/sonic-wg-routing/wiki/38418 NHG ID persistence discussion for warm reboot.
+* 02/13 https://lists.sonicfoundation.dev/g/sonic-wg-routing/wiki/38640 Fengsheng went over the proposal to use kernel to recover NHG ID  during warm reboot
```
selva-nexthop, line 94: I watched this recording.  The limitation mentioned in this discussion is recursive dependency.  Like zebra has multi-level NHG recursive structure and when it is installed into kernel, it might resolve and install it to single-level NHG recursiveness.  But when I see the zebra code, it looks like zebra itself resolving the multi-level NHG recursiveness to single-level NHG recursiveness and uses this for DP programming. 

```
File: sonic-frr/frr/zebra/zebra_nhg.c

/* Recursively construct a grp array of fully resolved IDs.
 *
 * This function allows us to account for groups within groups,
 * by converting them into a flat array of IDs.
 *
 * nh_grp is modified at every level of recursion to append
 * to it the next unique, fully resolved ID from the entire tree.
 *
 *
 * Note:
 * I'm pretty sure we only allow ONE level of group within group currently.
 * But making this recursive just in case that ever changes.
 */
static uint16_t zebra_nhg_nhe2grp_internal(struct nh_grp *grp, uint16_t curr_index,
					   struct nhg_hash_entry *nhe,
					   struct nhg_hash_entry *original, int max_num)
{
	struct nhg_connected *rb_node_dep = NULL;
	struct nhg_hash_entry *depend = NULL;
	struct nexthop *nexthop;
	uint16_t i = curr_index;

	frr_each(nhg_connected_tree, &nhe->nhg_depends, rb_node_dep) {
		bool duplicate = false;

		if (i >= max_num)
			goto done;

		depend = rb_node_dep->nhe;

		/*
		 * If its recursive, use its resolved nhe in the group
		 */
		if (CHECK_FLAG(depend->flags, NEXTHOP_GROUP_RECURSIVE)) {
			depend = zebra_nhg_resolve(depend);
			if (!depend) {
				flog_err(
					EC_ZEBRA_NHG_FIB_UPDATE,
					"Failed to recursively resolve Nexthop Hash Entry in the group id=%pNG",
					nhe);
				continue;
			}
		}

		if (!zebra_nhg_depends_is_empty(depend)) {
			/* This is a group within a group */
			i = zebra_nhg_nhe2grp_internal(grp, i, depend, nhe,
						       max_num);
		} else {
			bool found;
```

doc/ribfib/ribfib.md
eddieruan-alibaba: SONiC only supports device warm reboot, not container level restart.  For container level restart, a lot of additional efforts are requirement.

doc/ribfib/ribfib.md
eddieruan-alibaba: Zebra only keeps first level and last level. That's why we want to add resolve through and resolve via to maintain the middle level's recursive. 

Currently, zebra only send final NH list to dplane. That's why we want to carry all needed information from zebra to fpm, let SONiC handle them properly.

selva-nexthop: So, we will be making this change in zebra for this purpose?

rookie-who:
Hi @eddieruan-alibaba — the 202605 branch cut is at end of May. @selva-nexthop left review comments in March. Please address them.

eddieruan-alibaba:
> Hi @eddieruan-alibaba — the 202605 branch cut is at end of May. @selva-nexthop left review comments in March. Please address them.
> 
> CC: @vaibhavhd @radha-danda (branch managers)

The review comments have been addressed. Currently, I am waiting all Code PRs to be merged

doc/ribfib/ribfib.md
eddieruan-alibaba: Resolve through and resolved via changes would be done by Cisco team in FRR. Currently, we will use depends and dependents first until those changes are in FRR.

rookie-who:

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,489 @@
...
+* DPLANE_OP_NH_UPDATE
+* DPLANE_OP_NH_DELETE
+
+When any of these events are received, the SONiC Zebra NHG table will be updated accordingly within this block.
+
+* This table is indexed by the Zebra NHG ID and acts as a slave table to Zebra’s NHG hash table.
+* Each entry in the SONiC Zebra NHG table must correspond to an entry in Zebra’s NHG hash table.
+* The entry stores the struct zebra_dplane_ctx as the Zebra context, which is the yellow and green parts in the following diagram.
+* In addition, it maintains SONiC-specific metadata, such as the SONiC NHG ID, which is the blue part in the following diagram.
+![image](images/zebra_nhg_table_entry.png)
```
venkit-nexthop, line 415: What is the "uint32_t id" in the yellow part of the diagram?

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,489 @@
...
+
+This design allows the SONiC Zebra NHG table to serve as a mapping table between Zebra NHG IDs and SONiC NHG IDs.
+
+#### NEXTHOP KEY to Zebra NHG ID mapping
+This mapping table is used during warm reboot to mapping new zebra NHG ID to existing SONiC NHG ID.
```
venkit-nexthop, line 420: Can you add details on what constitutes a nexthop key?

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,489 @@
...
+
+### SONiC NHG table
+In certain scenarios, it is necessary to create a new SONIC Next Hop Group (NHG) upon receiving a zebra NHG event. In such cases, the corresponding SONiC NHG will be maintained in the SONiC NHG table.
+
+For the SRv6 VPN use case, NHG messages from Zebra carry not only forwarding information but also additional context, such as the VPN SID or a policy SID list. To handle this, SONiC will create a gateway NHG that contains the forwarding information only. This gateway NHG is also referred to as the PIC NHG. The accompanying context information will be used to construct a PIC Context object in SONiC.
```
venkit-nexthop: Can we stick with a single term..either "gateway" or "pic" NHG and use that consistently in the document. It would also be good to use that consistently in the code.

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,489 @@
...
+* In addition, it maintains SONiC-specific metadata, such as the SONiC NHG ID, which is the blue part in the following diagram.
+![image](images/zebra_nhg_table_entry.png)
+
+This design allows the SONiC Zebra NHG table to serve as a mapping table between Zebra NHG IDs and SONiC NHG IDs.
+
+#### NEXTHOP KEY to Zebra NHG ID mapping
+This mapping table is used during warm reboot to mapping new zebra NHG ID to existing SONiC NHG ID.
+
+### SONiC NHG table
+In certain scenarios, it is necessary to create a new SONIC Next Hop Group (NHG) upon receiving a zebra NHG event. In such cases, the corresponding SONiC NHG will be maintained in the SONiC NHG table.
```
venkit-nexthop, line 423: The description is not very clear to me. Which scenarios require the creation of SONiC NHG? I thought that would get created all the time? Can you add more explanation here?

doc/ribfib/ribfib.md
venkit-nexthop: Also can you list what exactly is part of "gateway"/"pic" NHG. 
In the HLD, it would be good to have lesser code and have more logical content. Or you can remove the rest and only highlight the part of the code that is pertinent (with explanations about what it is doing or what the field means).

doc/ribfib/ribfib.md
eddieruan-alibaba: It is zebra NHG id.  It is just a logic view.

doc/ribfib/ribfib.md
eddieruan-alibaba: I just want to use generate term in HLD. The key would be defined in LLD, based on the features we supports. 

Logically, it is the same set of fields used in zebra. But since we don't support some MPLS features, some fields would not be carried over. So it would be a trimmed version.

doc/ribfib/ribfib.md
eddieruan-alibaba: Good point, I will remove PIC NHG term, only use gateway nhg.

doc/ribfib/ribfib.md
eddieruan-alibaba: PE3# show ip route vrf Vrf1 192.100.0.1 next
Routing entry for 192.100.0.1/32
  Known via "bgp", distance 20, metric 0, vrf Vrf1, best
  Last update 00:16:37 ago
  Nexthop Group ID: 246
  Received Nexthop Group ID: 244
  q 2064:100::1d(vrf default) (recursive), label 16, weight 1
  q   fc06::2, via Ethernet12(vrf default), label 16, weight 1
  q   fc08::2, via Ethernet4(vrf default), label 16, weight 1
  q 2064:200::1e(vrf default) (recursive), label 32, weight 1
  q   fc06::2, via Ethernet12(vrf default), label 32, weight 1
  q   fc08::2, via Ethernet4(vrf default), label 32, weight 1

PE3# show nexthop rib 244 
ID: 244 (zebra)
     RefCnt: 10     Flags: 0x403
     Uptime: 00:16:46
     VRF: default(No AFI)
     Nexthop Count: 2
     Valid, Installed
     Depends: (238) (245)
           via 2064:100::1d (vrf default) inactive, label 16, seg6 fd00:201:201:1::, weight 1
           via 2064:200::1e (vrf default) inactive, label 32, seg6 fd00:202:202:2::, weight 1


For this NHG, it has 2064:100::1d and 2064:200::1e as nexthops, with corresponding VPN contexts. 

This zebra NHG would create two SONIC object, SONiC NHG and SONiC PIC CONTEXT

127.0.0.1:6378> hgetall "NEXTHOP_GROUP_TABLE:4"
1) "nexthop"
2) "2064:100::1d,2064:200::1e"
3) "ifname"
4) "unknown,unknown"
5) "weight"
6) "1,1"
7) "seg_src"
8) "2064:300::1f,2064:300::1f"

127.0.0.1:6378> hgetall "PIC_CONTEXT_TABLE:3"
 1) "nexthop"
 2) "2064:200::1e,2064:100::1d"
 3) "ifname"
 4) "unknown,unknown"
 5) "weight"
 6) "1,1"
 7) "vpn_sid"
 8) "fd00:202:202:2::,fd00:201:201:1::"
 9) "seg_src"
10) "2064:300::1f,2064:300::1f"


SONiC NHG is sharable, while SONiC PIC Context is not since it has VPN context. 


Since SONiC NHG is sharable, it stores in SONiC NHG table.

doc/ribfib/ribfib.md
eddieruan-alibaba: Removed PIC NHG and only use gateway NHG in https://github.com/eddieruan-alibaba/SONiC/commit/06b4f30ad8da3d92f91f012b041d972e58bf308c

doc/ribfib/ribfib.md
```diff
@@ -0,0 +1,731 @@
...
+
+This results in three separate hardware NHGs. Now, if 2::2/128 is removed, only NHG ID 270 needs to be updated via removing failed path to immediately stop traffic loss and buy time to safely reprogram only the affected routes. This aligns with the goal of prefix-independent convergence (PIC).
+
+Therefore, during warm reboot, orchagent would be responsible to map the new NHG ID to the old virtual OID.
+
+#### Why we can't use FRR to handle NHG ID persistence
+We discussed this with the FRR team. But FRR team believes that NHG ID persistence is a data plane requirement and therefore it should not be managed by FRR. They recommended leveraging the Linux kernel as a mechanism to maintain NHG ID persistence. However, our proof of concept try revealed that the kernel lacks the necessary information to reliably restore NHG IDs in all scenarios. Therefore, we can't use FRR to handle NHG ID persistence.
+
+#### Use fpmsyncd to manage NHG ID persistence
+The current conclusions from working group discussions are
```
selva-nexthop, line 146: **Suggestion:** use the content hash as the NHG identity

Dplane NHG table and hash-based identity already computes a Merkle hash over exactly the right structure — role, flags, sorted child hashes, and the resolving prefix for L-B — and then discards it: "the Merkle hash never leaves the plugin." A counter-allocated ID goes on the wire instead, and because that ID is arbitrary it has to be remembered.

Proposal: widen the hash and use it as the name. Keep the u32 dplane id as a session-scoped wire handle (`RTA_NH_ID` is 32 bits), but carry a full-width digest on `RTM_NEWNHGFIB` and use it as the `NEXTHOP_GROUP_TABLE` key.

Using the document's own example, with a canonical text encoding hashed with SHA-256 truncated to 128 bits:

```
L-A  for 2064:100::1d/128     1f1cd12d7a4a56816c2ebd7b49e01d7a   (zebra 243)
L-A  for 1::1/128             fde4ede720686b122f6ee96d820b88a3   (zebra 260)
L-A  for 4::4/128             e66f219cf8d88ccdfc19e705a3709a88   (zebra 270)

```
Three distinct names for the three groups that must stay distinct, even though all three flatten to the same two nexthops.

At warm reboot, zebra hands out completely different NHG IDs. The plugin re-derives the same tree, gets `1f1cd12d`… again, and fpmsyncd writes the identical key and fields into the preserved APPL_DB. Orchagent sees a no-op. Nothing was looked up and nothing was restored.

This would remove NEXTHOP KEY to Zebra NHG ID mapping, most of SONIC NHG ID Manager, the collision re-probe path, and the FRR Phase-2 task "Preserve zebra NHG ID in some deterministic cases." Everything else in the design — PIC core/edge, backwalk, received-vs-resolved, the SRv6 gateway-NHG + PIC_CONTEXT split, route-event derivation — is unchanged.

The trade-off: an opaque ID can be migrated if the key encoding changes in a later release; a content-derived name cannot. That's manageable if the encoding covers only forwarding-affecting fields, uses names rather than kernel indices (ifname, not ifindex), and is versioned per object class — but it should be a stated rule.

eddieruan-alibaba: Thanks Selva. 

Your suggestion is to use hash value, may 128 bits as the key to identify the NHG. This hash value would not be changed during warm reboot. 

So we don't need to use in-persistent zebra NHG ID, or dplane NHG ID. In fpmsyncd, we will map this 128 bit hash id to SONIC NHG id which would be used in APP DB. 

I am not worry about encoding method change. That would mess up a lot of stuffs. I would rather treat it as major upgrade, which needs cold reboot. 

How about debug-ability? In my mind, comparing id would be easier than comparing hash values, although we could use LLM to help on syslog analysis.

selva-nexthop: Thanks @eddieruan-alibaba  — and I'm glad you see the encoding question that way. Treating a change as a major upgrade needing cold reboot is the right call.

One clarification though, because I think this is the crux:
`In fpmsyncd, we will map this 128 bit hash id to SONIC NHG id which would be used in APP DB.`

In this context, if there's still a separately allocated SONiC NHG ID behind the hash, that ID is arbitrary, so it still has to be remembered across warm reboot. The ID manager, the persistence table and the restore path all survive — only the table's key changes. The idea is that the hash is the SONiC NHG ID, i.e. the APP_DB key itself:

```
NEXTHOP_GROUP_TABLE:1f1cd12d7a4a56816c2ebd7b49e01d7a
ROUTE_TABLE:2064:100::1d/128  →  nexthop_group = 1f1cd12d...
```
fpmsyncd recomputes the key after restart and writes the same row, orchagent sees no change, and there's nothing to persist because APPL_DB is the mapping.

On debuggability — a real concern, but I'd see it mostly improving. Today "NHG 243" in yesterday's log means nothing after a restart, and you walk two mapping tables to get from zebra to SAI. A content hash is the same name in zebra logs, fpmsyncd logs, APP_DB, and on every other box in the fabric — so cross-restart and cross-device correlation become a grep instead of impossible.

eddieruan-alibaba: Interesting idea. One more question, how do you guarantee that 128 bit hash id would be unique for each NHG? 

Current zebra hash doesn't guarantee it. It registers zebra_nhg_hash_equal() as hash_cmp function to compare contents when hash collision happens.

selva-nexthop: Two cases to separate:

```
- Same content → same digest. That's dedup, not collision.
- Different content → same digest. That's the real question.
```
We don't guarantee it. We make it negligible and detect it.  zebra_nhg_hash_key() is a 32-bit jhash used as a bucket index, so collisions are routine by design — hence zebra_nhg_hash_equal() in the hot path. At 1M NHGs a 32-bit space collides with certainty. SHA-256 truncated to 128 bits collides with probability ~1e-27, well below an undetected memory or ASIC-table error on the same box. 128 bits is my suggestion; 96 if APPL_DB key size becomes a concern.

Content length doesn't matter. A cryptographic hash distributes uniformly regardless of input size, and the Merkle form folds child digests rather than child content, so the hashed string stays bounded however deep the tree.  Idempotency and uniqueness are independent. Same input always gives the same digest at any width — guaranteed by construction. Width only buys uniqueness.

We keep the compare, which is what makes the move seamless. Dplane NHG table and hash-based identity already stores full content per object and verifies every hash hit. That stays as written. Only the compare's role changes — in zebra it resolves routine collisions, here it's an assertion that should never fire:

```
- content identical → reuse, bump refcount (unchanged)
- content differs → critical log and a locally suffixed name, never a silent merge
```
The delta is the digest width, plus putting it on the wire instead of keeping it plugin-internal. onNextHopGroupFullMsg and NHGMgr are unaffected, and NhgOrch already keys on a string.

eddieruan-alibaba: I am confused. You suggested to put in APPDB via 

NEXTHOP_GROUP_TABLE:1f1cd12d7a4a56816c2ebd7b49e01d7a

for different content cases, they are two different NHGs. We can't put them in APP DB like that, right?

selva-nexthop: You're right, and I should have shown this — two different NHGs never share a key.

The plugin computes the digest, then checks its table. Three outcomes:

```
- digest not present → create, key = digest
- digest present, content identical → same NHG, reuse and bump refcount
- digest present, content differs → real collision (~1e-27). Second object gets a suffixed key, plus a critical log.  So, the chance for this case is almost NO.

```
So in the collision case APP_DB looks like this:
```
NEXTHOP_GROUP_TABLE:1f1cd12d7a4a56816c2ebd7b49e01d7a
NEXTHOP_GROUP_TABLE:1f1cd12d7a4a56816c2ebd7b49e01d7a#1
```
Two keys, two hardware entries. Never merged. That one suffixed object loses idempotency — it reprograms on the next warm reboot — which seems fine at 1e-27.

On the map lookup, take `2064:100::1d/128` via `fc06::2/Eth12` and `fc08::2/Eth4`.

Today, zebra gives NHG 243 and fpmsyncd allocates SONiC NHG 400. After warm reboot zebra gives 511 instead, so to reach 400 again you walk two maps:
`nexthop key --[NEXTHOP KEY table]--> 243 --[SONiC zebra NHG table]--> 400
`
Both maps plus the allocator state have to be persisted and restored for that to work.
With the digest:
```
canonical content --sha256--> 1f1cd12d...
APP_DB:  NEXTHOP_GROUP_TABLE:1f1cd12d...
         ROUTE_TABLE:2064:100::1d/128 -> nexthop_group = 1f1cd12d...
```
After warm reboot zebra gives 511, the content is the same, so the digest is `1f1cd12d...` again and fpmsyncd writes the identical row. Orchagent sees no change.

No lookup, because there's nothing to look up — APPL_DB's own key is the mapping. That's what removes both maps and the ID manager.

Note this only works if the digest is the APP_DB key directly. If we map digest → an allocated SONiC NHG ID, that ID is still arbitrary and still has to be persisted, so we'd keep the table and the manager.

eddieruan-alibaba: Discussed with Selva offline. Current hash value would not be able to guarantee unqiueness to identify NHG. Although the probability for collision is much lower, I would prefer not to take this risk architecturally. 

Current agreements are
1. Keep current id approach
2. We will invest a way to be able to generate id uniquely based on contents in the future. 

So we will resolve this conversation for now.

pbrisset:
HI, the HLD mentioned contradictory information. From my understanding, is it what is proposed here?

In the proposed architecture, Zebra maintains the RIB and sends route events to dplane_fpm_sonic. That component builds the forwarding graph from the routing information. This plugin resides on the FRRouting side. The fpmsyncd FIB block, that resides on the SONiC side, converts the graph into SONiC next-hop groups, gateway next-hop groups, and PIC_CONTEXT entries, and writes them to APP_DB. Finally, orchagent reads APP_DB and programs the ASIC.

Can you confirm?

PR #2061, opened 2025-08-18
[thermalctld] Add sonic common infrastructure to use optics temperature data from LC for cooling algorithm in SUP

judyjoseph:
In current sonic T2 chassis implementation (especially in VoQ chassis) the optical temperature data fetched by the xcvrd in the Linecards are not used in the Supervisor card for cooling.

The change here proposes two different approaches to push data from LC --> SUP  

1. push the raw optics temperature data from LC --> SUP and let SUP use the data in cooling algorithm 
2. process the optics temperature in the LC and push the calculated values to SUP for use 

**Option 2 is preferred approach** 
                (i)  as it reduces the number updates to SUP
                (ii) faster to process locally and push result

### DIFF
--- a/doc/pmon/pmon-chassis-optics-thermal-enhancement.md
+++ b/doc/pmon/pmon-chassis-optics-thermal-enhancement.md
@@ -7,8 +7,7 @@
 * [About this Manual](#about-this-manual)
 * [Scope](#scope)
   * [1. Modular Chassis Thermalctld architecture](#1-modular-chassis-thermalctld-architecture)
-  * [2. Transceiver thermal data in Linecard](#2-transceiver-thermal-data-in-linecard)
-    * [2.1 Retrieve optics temperature data in Linecard](#21-retrieve-optics-temperature-data-in-linecard)
+  * [2. Transceiver thermal data in Linecard](#2-transceiver-thermal-data-in-linecard)    
   * [3. Approaches to send transceiver thermal data from Linecard to Supervisor](#3-approaches-to-send-transceiver-thermal-data-from-linecard-to-supervisor)
     * [3.1 Send raw optics thermal data from Linecard to Supervisor](#31-send-raw-optics-thermal-data-from-linecard-to-supervisor)
     * [3.2 Process optics thermal data in Linecard send result to Supervisor](#32-process-optics-thermal-data-in-linecard-send-result-to-supervisor)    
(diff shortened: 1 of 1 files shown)


judyjoseph:
Adding all chassis sku teams to take a look and comment on this proposal 
@Marc-nokia @mlok-nokia @kenneth-arista @bmridul

doc/pmon/pmon-chassis-design.md
```diff
@@ -415,6 +416,57 @@ Thermal 4        32           68         0             N/A            N/A
...
+ key                                   = TEMPERATURE_INFO_<card-index>|THERMAL_ALGO_RESULT; 
+ ; field                               = value
+ thermal_result_attribute_count        = int 
+ thermal_result_attribute_1            = str;
+ thermal_result_attribute_2            = str;
+ thermal_result_attribute_3            = str;
+ ... 
+ ```
+In the Linecard there will be the following changes in thermalctld
+      (i) thermalctld/main daemon --> thermal_manager.run_policy() implemented by each vendor/platform which will collect all the thermal data stored in STATE_DB locally and generate THERMAL_ALGO_RESULT, store it in STATE_DB
```
snider-nokia: For "Option 2" each vendor will have proprietary fields/attributes at LC in the context of their thermal algorithm.  Will those LC local STATE_DB fields be named generically (something like THERMAL_RESULT_ATTRIBUTE_1, THERMAL_RESULT_ATTRIBUTE_2, THERMAL_RESULT_ATTRIBUTE_3, etc. or will some type of JSON setup be used in order for each vendor to define their own Redis attribute field names?   If the answer is the former then it may be adequate to just say there is some arbitrary max of these attribute fields (say 9 for example) and then any of those that aren't used by a given vendor are set to "N/A" (or similar) by vendor specific code.  All of these generic attribute fields could then be promulgated by common thermalctld code from local STATE_DB to CHASSIS_STATE_DB.  Only issue with this solution is that these fields must then be visualized in a common way at SUP;  they would need to be displayed by some tool (if visualization is desired) alongside their generic field/attribute name (that is THERMAL_RESULT_ATTRIBUTE_1, THERMAL_RESULT_ATTRIBUTE_2, etc.) without any vendor specific field context.

judyjoseph: Thanks @snider-nokia for suggestion. I did think about that too, we could use platform_json itself. We can have a new section to define the thermal_algo_result_attributes in this file

REF : https://github.com/sonic-net/sonic-buildimage/blob/5cd0207b150897cb16ea284535681c7a0597d660/device/nokia/x86_64-nokia_ixr7250e_sup-r0/Nokia-IXR7250E-SUP-10/platform.json#L25

```
  "thermal_algo_result_attributes": {
	     "thermal_result_attribute_1",
	     "thermal_result_attribute_2",
	     "thermal_result_attribute_3"
	},
```

These are the set of attributes which vedor can define, calculated in LC and pushed to local STATE_DB and then to CHASSIS_STATE_DB in SUP. Further it will be used in SUP for cooling algorithm

bmridul:
@spilkey-cisco Pls review.

doc/pmon/pmon-chassis-design.md
```diff
@@ -415,6 +416,57 @@ Thermal 4        32           68         0             N/A            N/A
...
+
+The optics temperature infromation is retrieved by the xcvrd/dom_mgr in the respective Linecards and stored in the TRANSCEIVER_DOM_SENSOR|Ethernet<> table in STATE_DB. The temperature thresholds are stored in TRANSCEIVER_DOM_THRESHOLD|Ethernet<> table in STATE_DB. This temperature information should be used in the cooling/fan algorithm.
+
+In the chassis architectures (voq and packet) the cooling/fan algorithm is run in the context of thermalctld in the Supervisor. There are two options for thermalctld in Linecard to pass the optics temperature information to Supervisor.
+
+**Option 1**
+
+This case where thermalctld in each Linecard/module **don't** process thermal data locally, instead just send each optics Temperature_info(along with the other thermals) to Supervisor CHASSIS_STATE_DB. The Supervisor/thermalctld use this directly as input to cooling algorithm.
+ 
+Following schema could be used to store the TEMPERATURE_INFO in the CHASSIS_STATE_DB. The <Sensor-Name> here will be interface name.
```
spilkey-cisco: Does this schema align 1-to-1 with the existing temperature entries in CHASSIS_STATE_DB?

doc/pmon/pmon-chassis-design.md
```diff
@@ -415,6 +416,57 @@ Thermal 4        32           68         0             N/A            N/A
...
+ timestamp                             = str;
+ ```
+
+In the Linecard, the thermalctld/ThermalMonitor thread will read the optics temperature information from the tables TRANSCEIVER_DOM_SENSOR|Ethernet<>, TRANSCEIVER_DOM_THRESHOLD|Ethernet<> in the STATE_DB of either linux host in case of single ASIC platforms or from the STATE_DB in respective namespaces for multi-asic platforms. This data will be pushed to the CHASSIS_STATE_DB in the CHASSIS_DB in the supervisor along with other thermal data.
+
+In the Supervisor, the thermalctld/main daemon --> thermal_manager.run_policy() implemented by each vendor/platform which will collect all the thermal data pushed in CHASSIS_STATE_DB including optics and drive the cooling algotithm.
+
+**Option 2**
+
+This case where thermalctld in each Linecard/module processes thermal data locally using the vendor/platform specific thermal algorithm and sends the thermal algorithm result to Supervisor CHASSIS_STATE_DB.
```
spilkey-cisco: Vendor implementations of thermal_manager.py will just read optics data directly from `TRANSCEIVER_DOM_SENSOR|Ethernet<>, TRANSCEIVER_DOM_THRESHOLD|Ethernet<>`, correct?

doc/pmon/pmon-chassis-design.md
```diff
@@ -415,6 +416,57 @@ Thermal 4        32           68         0             N/A            N/A
...
+ 
+Following schema could be used to store the TEMPERATURE_INFO per Linecard module. This will contain vendor/platform specific attributes (eg: temperature of hotest sensor on linecard) to be used in the cooling algorithm in the supervisor. 
+ 
+ #### CHASSIS_STATE_DB Schema for Temperature_Info
+ ```
+ key                                   = TEMPERATURE_INFO_<card-index>|THERMAL_ALGO_RESULT; 
+ ; field                               = value
+ thermal_result_attribute_count        = int
```
spilkey-cisco: Is this field necessary? The fields in this entry appear entirely up to the vendor's discretion, and it's simple to get the number of entries without an explicit field defining it. Just curious if there's some use case I'm missing where this would be helpful to be stored in redis.

doc/pmon/pmon-chassis-design.md
judyjoseph: yes

doc/pmon/pmon-chassis-design.md
judyjoseph: It is not required, I moved the vendor specific attributes to be defined to platform.json file

doc/pmon/pmon-chassis-design.md
judyjoseph: yes

doc/pmon/pmon-chassis-design.md
spilkey-cisco: As long as the STATE_DB entry name `THERMAL_ALGO_RESULT` is well defined, do its individual entry names need to be defined in platform.json? thermalctld can `hgetall` or similar on `THERMAL_ALGO_RESULT` to get all the contained entries without knowing anything about the contents. Vendor implementation pushes directly into STATE_DB `THERMAL_ALGO_RESULT`, vendor can decide how/where/if to store entry names in a config file or similar. Or would this somehow be used for sonic-mgmt tests to verify contents? I don't fully understand the need to map vendor entry names to generic names for display.

PR #2067, opened 2025-09-04
Linux VRF Route Lookup Fallback Control HLD

sudharr86:
This document provides the high-level design for Linux VRF Route Lookup Fallback Control. By default, `vrfmgrd` blocks legacy Linux RPDB fall-through in applicable non-default VRFs, aligning the kernel route-miss result with ASICs that terminate a VRF miss. An explicit global compatibility setting can temporarily restore legacy kernel fallback. The feature changes Linux kernel routing only; SAI and ASIC programming remain unchanged.

### DIFF
--- a/doc/vrf/no-default-vrf-route-fallback-hld.md
+++ b/doc/vrf/no-default-vrf-route-fallback-hld.md
@@ -192,8 +192,128 @@ In prior SONiC releases, Linux kernel allowed VRF lookup fallback if a route was
 - Knob enabled: fallback restored (legacy only).
 - New deployments should not rely on kernel fallback feature.
 
-## 10–14. (Remain as in previous version, with test cases and open items updated to reflect new default and knob.)
+## 10. Warmboot and Fastboot Design Impact
 
+### 10.1. Warmboot Impact
+
+**No negative impact on warmboot functionality:**
+- Configuration persistence through CONFIG_DB
+- Route reconciliation during vrfmgrd restart
+- No additional stalls or IO operations in boot path
+- No changes to critical boot sequence
+
+### 10.2. Fastboot Impact
+
+**No impact on fastboot performance:**
+- Configuration loaded during normal service startup
+- No additional dependencies or processing overhead
+- Feature can be delayed without affecting critical services
+
+### 10.3. Performance Analysis
+
+- **Boot Time**: No measurable impact on boot time
+- **CPU Usage**: Minimal CPU overhead during configuration changes only
+- **IO Operations**: Limited to route add/delete operations
+- **Critical Path**: Not in critical boot path
+
+## 11. Memory Consumption
+
+### 11.1. Memory Analysis
+
+**Static Memory:**
+- Configuration storage in CONFIG_DB: ~100 bytes per VRF
+- Additional code in vrfmgrd: ~5KB
+
+**Dynamic Memory:**
+- No growing memory consumption during operation
+- Kernel route entries: ~64 bytes per unreachable route
+- No memory leaks or unbounded growth
+
+**Total Impact:**
+- Negligible memory footprint
+- No memory consumption when feature is disabled
+- Scales linearly with number of configured VRFs
+
+## 12. Restrictions/Limitations
+
+### 12.1. Design Limitations
+
+- **L-1**: Unreachable routes affect only software-forwarded traffic
+- **L-2**: Feature requires Linux kernel support for unreachable routes
+- **L-3**: Does not modify BGP or other routing protocol behavior
+- **L-4**: Limited to VRF route lookup control, not general routing policy
+
+### 12.2. Platform Dependencies
+
+- **No platform-specific dependencies**
+- **Standard Linux iproute2 functionality required**
+- **Compatible with all SONiC-supported platforms**
+
+### 12.3. Scale Limitations
+
+- **Limited by kernel route table capacity**
+- **Recommended maximum: 1000 VRFs (platform dependent)**
+
+## 13. Testing Requirements/Design
+
+### 13.1. Unit Test Cases
+
+#### 13.1.1 CLI Tests
+- **UT-CLI-1**: Validate CLI command syntax and parameter validation
+- **UT-CLI-2**: Test CONFIG_DB updates for valid configurations
+- **UT-CLI-3**: Test error handling for invalid VRF names
+- **UT-CLI-4**: Test show command output formatting
+
+#### 13.1.2 CONFIG_DB Tests
+- **UT-DB-1**: Schema validation for global fallback entries
+- **UT-DB-2**: Schema validation for per-VRF fallback entries
+- **UT-DB-3**: Invalid configuration handling
+- **UT-DB-4**: Configuration persistence testing
+
+#### 13.1.3 VRF Manager Tests
+- **UT-MGR-1**: Configuration change monitoring
+- **UT-MGR-2**: Route addition/removal logic
+- **UT-MGR-3**: IPv4 and IPv6 handling independently
+- **UT-MGR-4**: Error handling for route operations
+
+### 13.2. System Test Cases
+
+#### 13.2.1 Functional Tests
+- **ST-FUNC-1**: Verify route fallback disabled with unreachable route present
+- **ST-FUNC-2**: Verify per-VRF override of global setting
+- **ST-FUNC-3**: Verify IPv4 and IPv6 independent operation
+- **ST-FUNC-4**: Test with multiple VRFs and mixed configurations
+
+#### 13.2.2 Integration Tests  
+- **ST-INT-1**: End-to-end CLI to kernel route verification
+- **ST-INT-2**: Configuration persistence across reboots
+- **ST-INT-3**: Config reload testing
+- **ST-INT-4**: Warm reboot with feature enabled
+
+#### 13.2.3 Upgrade/Downgrade Tests
+- **ST-UPG-1**: Seamless upgrade from pre-feature version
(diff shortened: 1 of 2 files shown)


doc/vrf/no-default-vrf-route-fallback-hld.md
```diff
@@ -0,0 +1,411 @@
...
+- Remove: `ip route del unreachable default vrf <VRF_NAME>`
+
+#### 7.5.2. CLI Implementation
+
+**Configuration Commands:**
+- `config vrf route-fallback [ipv4|ipv6] [enable|disable]`
+- `config vrf route-fallback [ipv4|ipv6] [enable|disable] -v <VRF_NAME>`
+
+**Show Commands:**
+- `show vrf-route-fallback`
```
selva-nexthop: can be "show vrf route-fallback" which will be consistent to config commands.

prhoskot: 'show vrf' command is implemented as a flat CLI. If we want to include subcommands to this, then we should make it as a group. But if we introduce group, one more keyword is added for 'show vrf' which will break existing commands/test scripts. So 'show vrf-route-fallback' is implemented as a separate command itself.

doc/vrf/no-default-vrf-route-fallback-hld.md
```diff
@@ -0,0 +1,411 @@
...
+### 7.7. Fastboot Impact
+
+- No additional dependencies for fastboot
+- Configuration loaded during normal startup sequence
+
+## 8. SAI API
+
+**No SAI API changes are required for this feature.**
+
+The unreachable routes are maintained only in the Linux kernel and are explicitly not programmed into the ASIC hardware. This design choice ensures:
```
rck-innovium: This may not be true. Consider the below config:
```
"VRF_ROUTE_FALLBACK_GLOBAL|ipv4": {
    "status": "enabled"
},
"VRF|VrfRed": {
    "fallback": "enabled",
},
"VRF|VrfGreen": {
    "fallback": "disabled",
}
```
In this case, in the ASIC route tables, **VrfGreen|0.0.0.0/0** route needs to be installed to avoid fallback VRF lookup for packets in VrfGreen.

rck-innovium: Can you please explain what is the relation of the proposed feature with SAI_SWITCH_ATTR_DEFAULT_VIRTUAL_ROUTER_ID

doc/vrf/no-default-vrf-route-fallback-hld.md
selva-nexthop: ok

doc/vrf/no-default-vrf-route-fallback-hld.md
```diff
@@ -0,0 +1,411 @@
...
+
+## 1. Revision
+
+| Rev | Date       | Author      | Description                                |
+|-----|------------|-------------|--------------------------------------------|
+| 1.0 | <Date>     | <Your Name> | Initial draft for VRF route fallback control |
+
+## 2. Scope
+
+This high-level design document covers the implementation of a configurable mechanism to disable VRF (Virtual Routing and Forwarding) route fallback behavior in SONiC. The scope includes:
```
ashutosh-agrawal: Please include some explicit wording to clarify that the HLD covers only the Kernel routes and software forwarding.

doc/vrf/no-default-vrf-route-fallback-hld.md
```diff
@@ -0,0 +1,334 @@
...
+| ASIC | Application-Specific Integrated Circuit |
+| BGP | Border Gateway Protocol |
+| FRR | Free Range Routing |
+
+## 4. Overview
+
+In prior SONiC releases, Linux kernel allowed VRF lookup fallback if a route was missing in a non-default VRF, which could cause divergence from ASIC behavior and security/operational risks.
+
+**This design ensures:**
+- By default, fallback is disabled both in kernel and ASIC by installing an unreachable default route in every new non-default VRF.
```
ashutosh-agrawal: I thought that unreachable route installed by the proposed implementation doesn't get propogated to ASIC DB, so this design doesn't do anything specific to disable fallback in the ASIC.

doc/vrf/no-default-vrf-route-fallback-hld.md
```diff
@@ -0,0 +1,334 @@
...
+
+- On creation of any non-default VRF, vrfmgrd adds:
+  - `ip route add unreachable default metric 4278198272 vrf <VRF_NAME>`
+  - (For both IPv4 and IPv6.)
+- This disables fallback in the kernel.
+- **This unreachable default route will not be added to APP_DB and will not be propagated to lower layers.**
+
+### 7.2. Global Fallback Knob
+
+- New global configuration (in CONFIG_DB and CLI): `KERNEL_VRF_FALLBACK`
```
ashutosh-agrawal: IIRC, in the HLD review meeting, we agreed to implement only the CONFIG_DB mechanism and not the CLI.

doc/vrf/no-default-vrf-route-fallback-hld.md
```diff
@@ -0,0 +1,334 @@
...
+- See section 7.4 for updated config and show commands.
+
+### 9.2. CONFIG_DB Schema
+
+- See section 7.3.
+
+### 9.3. Backward Compatibility
+
+- Default (knob unset): fallback disabled, matches ASIC.
+- Knob enabled: fallback restored (legacy only).
```
ashutosh-agrawal: Please change to "fallback restored in the kernel forwarding path only"

doc/vrf/no-default-vrf-route-fallback-hld.md
```diff
@@ -0,0 +1,334 @@
...
+In prior SONiC releases, Linux kernel allowed VRF lookup fallback if a route was missing in a non-default VRF, which could cause divergence from ASIC behavior and security/operational risks.
+
+**This design ensures:**
+- By default, fallback is disabled both in kernel and ASIC by installing an unreachable default route in every new non-default VRF.
+- A temporary global knob (`kernel-vrf-fallback`) can be set to restore the old fallback behavior. When set, the unreachable default route is removed from all non-default VRFs, allowing the kernel to fall back as before.
+- The knob is intended only for backward compatibility and will be deprecated.
+
+> **Note:**  
+> The unreachable default route added to non-default VRFs exists only in the Linux kernel routing table.  
+> **It will not be added to APP_DB and will not be propagated to lower layers such as ASIC_DB, SAI, or programmed into hardware.**
```
rck-innovium: This causes the kernel forwarding and h/w forwarding behavior to be different. i.e., fallback can be disabled in the kernel but not in the h.w. Isn't this a concern?

gord1306:
I've submitted a similar PR before. Here's just a link for reference: https://github.com/sonic-net/sonic-swss/pull/2943

doc/vrf/no-default-vrf-route-fallback-hld.md
sudharr86: I beleive there is no relation.

doc/vrf/no-default-vrf-route-fallback-hld.md
sudharr86: @ashutosh-agrawal the unreachable route is NOT propagated to ASIC DB. The feature makes the kernel behavior consistent with ASIC. will re-word accordingly.

doc/vrf/no-default-vrf-route-fallback-hld.md
sudharr86: agreed.

doc/vrf/no-default-vrf-route-fallback-hld.md
sudharr86: in ASIC there is no fallback today. it makes the kernel and HW behavior same.

doc/vrf/no-default-vrf-route-fallback-hld.md
sudharr86: agreed.

sudharr86: Addressed in the current HLD. Section 2 limits the change to Linux VRF route lookup for host-local and software-forwarded traffic, and Sections 4 and 8 explicitly state that ASIC, SAI, and hardware forwarding behavior are unchanged.

doc/vrf/no-default-vrf-route-fallback-hld.md
sudharr86: Addressed in the current HLD. Section 4 and Section 7.4 state that fpmsyncd filters the RTN_UNREACHABLE notifications before APP_DB mutation, so the managed sentinels do not enter APP_DB, ASIC_DB, SAI, or the hardware FIB.

doc/vrf/no-default-vrf-route-fallback-hld.md
sudharr86: Addressed in the current HLD. Sections 2, 5.3 EX-3, 7.3, and 9 explicitly define CONFIG_DB-only management and state that no feature-specific CLI, show command, or management API is introduced.

doc/vrf/no-default-vrf-route-fallback-hld.md
sudharr86: Addressed in the current HLD. Sections 2 and 7.1 now state that status=enabled restores legacy RPDB fall-through only in the Linux kernel route-lookup path; APP_DB, ASIC_DB, SAI, and hardware behavior remain unchanged.

PR #2080, opened 2025-10-06
HLD - Device Local Diagnosis: Vendor Rules and Daemon Designs

gregoryboudreau:
This PR includes the HLDs for the common vendor rules schema and the daemon design for the "Device Local Diagnsosis" topic.

gNOI Healthz [[HLD](https://github.com/sonic-net/SONiC/blob/master/doc/mgmt/gnmi/gnoi_healthz_hld.md)]
Openconfig
Faults [[PR](https://github.com/openconfig/public/pull/1346/)]
Healthz [[Readme](https://github.com/openconfig/gnoi/tree/main/healthz)]


chengzhi11:
Deployment form factor — a data point from a production user

Hi Gregory and the DLD team,

We operate a production device-local self-healing agent on a large SONiC-based fabric, with an architecture very similar to DLD (STATE_DB alarm subscription → local remediation → controller escalation). Our agent runs as a Docker container depending only on the database container, and crash isolation has been acceptable in practice — sibling container crashes have never taken down our agent or the database in production.

Given that, we'd like to suggest reconsidering — or at least explicitly documenting the trade-off of — the host-systemd deployment choice:

Upgrade & maintenance: containerized delivery locks DLDD's runtime and its pinned Pydantic / pydantic-core dependency chain inside its own image. As a host service, every dependency bump touches the host image's shared Python package set (the versions-py3 snapshot / sonic_debian_extension wheel install path), increasing coupling with other host packages and making out-of-band upgrades harder.

Consistency with the SONiC lifecycle: containerized services get standard image distribution, the FEATURE/featured enable-disable flow, and config reload sequencing for free. The HLD currently has to define bespoke systemd units, timers, and host-side wiring to achieve the same outcomes.

Crash isolation is a real but manageable concern: our production data point is that depending only on the database container is sufficient — especially since the HLD already designs DLDD to degrade gracefully when gNMI/gNOI are unavailable.

We'd be happy to share more production details (resource footprint, restart semantics, upgrade workflow) if useful for the decision. Looking forward to the discussion — and congratulations on getting DLD into the 202611 release.

gregoryboudreau:
> Deployment form factor — a data point from a production user
> 
> Hi Gregory and the DLD team,
> 
> We operate a production device-local self-healing agent on a large SONiC-based fabric, with an architecture very similar to DLD (STATE_DB alarm subscription → local remediation → controller escalation). Our agent runs as a Docker container depending only on the database container, and crash isolation has been acceptable in practice — sibling container crashes have never taken down our agent or the database in production.
> 
> Given that, we'd like to suggest reconsidering — or at least explicitly documenting the trade-off of — the host-systemd deployment choice:
> 
> Upgrade & maintenance: containerized delivery locks DLDD's runtime and its pinned Pydantic / pydantic-core dependency chain inside its own image. As a host service, every dependency bump touches the host image's shared Python package set (the versions-py3 snapshot / sonic_debian_extension wheel install path), increasing coupling with other host packages and making out-of-band upgrades harder.
> 
> Consistency with the SONiC lifecycle: containerized services get standard image distribution, the FEATURE/featured enable-disable flow, and config reload sequencing for free. The HLD currently has to define bespoke systemd units, timers, and host-side wiring to achieve the same outcomes.
> 
> Crash isolation is a real but manageable concern: our production data point is that depending only on the database container is sufficient — especially since the HLD already designs DLDD to degrade gracefully when gNMI/gNOI are unavailable.
> 
> We'd be happy to share more production details (resource footprint, restart semantics, upgrade workflow) if useful for the decision. Looking forward to the discussion — and congratulations on getting DLD into the 202611 release.

Thanks for sharing, it's a useful data point. I considered a dedicated container, but DLDD needs to support vendor hooks that may access host platform APIs, device files, I2C, sysfs, or other vendor custom paths (think maybe even vendor ASIC debug shells or the like). Those facilities are not necessarily available inside a container, and exposing them would require platform-specific mounts, SDK dependencies, broad privileges, or host-execution proxies.
For that reason, I am strongly inclined towards the host service approach. I do agree that host Python dependency coupling is a valid concern, so I will look into possibility of further isolating DLDD-specific dependencies and document the deployment trade-off more clearly in the HLD.

PR #2083, opened 2025-10-08
Update and rename SONiC-SAI-POST.md

vikram-nexthop:
- include Control Plane POST.
- CLI to fetch the POST status.

Implementation PRs:
https://github.com/sonic-net/sonic-buildimage/pull/24067
https://github.com/sonic-net/sonic-swss/pull/3907
https://github.com/sonic-net/sonic-wpa-supplicant/pull/99

### DIFF
--- a/doc/fips/SONiC-MACSec-POST.md
+++ b/doc/fips/SONiC-MACSec-POST.md
@@ -105,7 +105,7 @@ POST is performed by MACSecOrch when POST is supported only in MACSec init. The 
 
 ### Enabling Control Plane POST in MACSecMgr init
 
-MACSecMgr performs POST status query during process init and publishes the status in the _FIPS_MACSEC_POST_TABLE|crypto_ table. Any w. A failure will result in MACSecMgr not processing any configuration beyond this.
+MACSecMgr performs POST status query during process init and publishes the status in the _FIPS_MACSEC_POST_TABLE|crypto_ table. A failure will result in MACSecMgr not processing any configuration beyond this.
 
 ![](images/fips-cp-post-macsec-init.png)
 


vikram-nexthop:
implementation PRs:
https://github.com/sonic-net/sonic-buildimage/pull/24067
https://github.com/sonic-net/sonic-swss/pull/3907
https://github.com/sonic-net/sonic-wpa-supplicant/pull/99

doc/fips/SONiC-MACSec-POST.md
```diff
@@ -94,8 +96,49 @@ MACSec level POST failure
...
+  --dump-file                   store show output to a file
+  --post-status                 show macsec FIPS POST(Pre-Operational Self-Test) status
+  -d, --display [all|frontend]  Show internal interfaces  [default: all]
+  -n, --namespace []            Namespace name or all
+  -?, -h, --help                Show this message and exit.
+```
+Sample output
+```
+admin@sonic:~$ show macsec --post-status
+POST Status:
```
ysmanman: On multi-ASIC, is POST status displayed for every ASIC?

vikram-nexthop: without the '-n' option, this should display the POST status from the default namespace.

ysmanman: On mutli-ASIC,  SAI POST status is published for each ASIC namespace, but not host namespace. What does  SAI POST status mean in the output when without `-n` option?

vikram-nexthop: Sorry, please ignore my previous comment. Without the '-n' option, the command should display the POST status (both SAI and Crypto) from all the namespaces. My understanding from [SONiC_multi_asic_hld.md](https://github.com/sonic-net/SONiC/tree/master/doc/multi_asic/SONiC_multi_asic_hld.md)
is that there will be a macsec service per ASIC NS.

judyjoseph:
@vikram-nexthop Can you share more details on "Control Plane POST" ? I understand the SAI POST to check the ASIC/SecY -- but no so clear on the requirements on control plane POST

vikram-nexthop:
@judyjoseph Control plane POST validates the cryptographic module used by wpa_supplicant for MKA operations, which is separate from the SAI/ASIC crypto engines, and requires independent FIPS validation too.
Please see my clarification in PR #[3907](https://github.com/sonic-net/sonic-swss/pull/3907#issuecomment-3460257699).

ysmanman (approved):

doc/fips/SONiC-MACSec-POST.md
vikram-nexthop: Sample CLI output section updated with multi asic post status example.

doc/fips/SONiC-MACSec-POST.md
```diff
@@ -94,8 +97,93 @@ MACSec level POST failure
...
 
+### Enabling POST in SAI MACSec init
+
+POST is performed by MACSecOrch when POST is supported only in MACSec init. The following flow chart demonstrates the process.
+
+![](images/fips-post-macsec-init.png)
+
+### Enabling Control Plane POST in MACSecMgr init
+
+MACSecMgr performs POST status query during process init and publishes the status in the _FIPS_MACSEC_POST_TABLE|crypto_ table. Any w. A failure will result in MACSecMgr not processing any configuration beyond this.
```
lihuay: some typo? "Any w. A failure"

doc/fips/SONiC-MACSec-POST.md
```diff
@@ -24,19 +26,22 @@ This document describes SONiC design for Federal Information Processing Standard
 
 The design must meet the following requirements:
 - In order to accommodate different forwarding ASIC architecture or SAI implementation, the design should support enabling POST at either switch level (during switch init) or at MACSec engine level (during MACSec engine init).
+- Should enable control plane POST at the MACSec level.
```
lihuay, line 29: Is this a must for FIPS compliance?

vikram-nexthop: I can't say if it is a must(couldn't find a direct reference to control plane POST validation in the official FIPS documentation), but I do consider it necessary for SONiC MACsec FIPS compliance due to the following:

1. These libraries (FIPS-certified OpenSSL FIPS Provider, SymCrypt) are not part of the SONiC repository but are critical to the security model - they handle authentication, data plane key generation, etc. Independent validation ensures their integrity and expected operational status (FIPS mode).
2. Provides initialization-time detection of configuration errors, self-test failures, library corruption, etc enabling graceful service disabling rather than runtime failures.
3. A one-time validation that reduces the attack surface by ensuring cryptographic libraries are in a known-good state.

This could prove redundant if the entire system (SONiC + 3rd party crypto libraries) is under a common FIPS boundary.

doc/fips/SONiC-MACSec-POST.md
vikram-nexthop: thanks for noticing this.

doc/fips/SONiC-MACSec-POST.md
senthil-nexthop: To satisfy FIPS 140-3 requirements, all cryptographic module algorithms must undergo a self-test before being used by applications to ensure integrity. The goal of this PR is to implement this self-test during wpa_supplicant initialization.

However, after reviewing the implementation, it appears that the SymCrypt library already executes the necessary self-test and integrity checks when the library itself is loaded. If this is the case, duplicating this functionality in wpa_supplicant may be redundant. We will close this PR.

Reference to self-tests executed in SymCrypt: https://github.com/microsoft/SymCrypt/blob/b39181fbfb3e54e1b471f0d10864d0e7077626b8/modules/posix/common/module.c#L32

vikram-nexthop:
just to add. wpa_supplicant code changes only checks if the crypto provider POST succeeded, which services like macsecmgr can use to detect FIPS error early (on startup) and move into an error state. 

This doesn't seem to be a necessary requirement, hence closing this PR.

PR #2091, opened 2025-10-27
[DASH] Move route rule table priority to key

theasianpianist:
The SAI DASH inbound routing header has the routing entry priority as part of the entry key. To prevent having multiple DASH_ROUTE_RULE_TABLE entries with the same priority, deprecate the priority field and store the priority in the key to match the SAI definition.

sonic-dash-api change: https://github.com/sonic-net/sonic-dash-api/pull/62


KrisNey-MSFT:
hi @theasianpianist - do we need to wait for this one to be fixed (https://github.com/sonic-net/sonic-buildimage/issues/23590) before merging this current PR?

prsunny:
This is an API contract change. @mzms, please review

KrisNey-MSFT:
pinged @Michal mzms to review

KrisNey-MSFT:
per @mzms this is non-blocking for March 2026

prsunny:
@theasianpianist , is there a corresponding dash-api change PR? if so, can you link it here?

theasianpianist:
> @theasianpianist , is there a corresponding dash-api change PR? if so, can you link it here?

not yet, had originally planned to merge this first. will open the dash-api change PR and link it here today

PR #2096, opened 2025-11-06
LLDP Custom TLV HLD

cshivashgit:
This document provides a high-level design for LLDP Custom TLV feature in SONiC. The feature allows either users to configure custom TLVs that users want to be advertised or another feature in SONiC which needs to use custom TLVs for communicating with the neighbouring devices.


ashutosh-agrawal:
@saiarcot895 @venkatmahalingam Appreciate your review of this HLD.

doc/lldp_custom_tlv/lldp-custom-tlv-hld.md
```diff
@@ -0,0 +1,746 @@
...
+    }
+  }
+}
+```
+**Note** : Global and port-level configurations are mutually exclusive but can coexist; interface-level configurations do not override global settings.
+
+
+### State DB Schema
+
+#### LLDP_RX_CUSTOM_TLV Table
```
shak5009, line 273: To keep things consistent with the current codebase's naming conventions, should we rename LLDP_RX_CUSTOM_TLV Table to something like LLDP_CUSTOM_TLV_ENTRY_TABLE?

I do understand that STATE DB is likely the better placement for both LLDP_ENTRY_TABLE and LLDP_CUSTOM_TLV_ENTRY_TABLE. Since the existing standard LLDP entries reside in APPL_DB under LLDP_ENTRY_TABLE, would it make sense to explicitly place this new table in APPL_DB as well to ensures they stay in sync when neighbors update or time out.

The proposal (if it makes sense) to put these tables to STATE_DB can possibly come in a different PR?

doc/lldp_custom_tlv/lldp-custom-tlv-hld.md
```diff
@@ -0,0 +1,746 @@
...
+      "custom_tlv_name": "vendor1-tlv"
+    },
+    "Ethernet8": {
+      "admin_status": "enabled",
+      "custom_tlv_name": "vendor2-tlv"
+    }
+  }
+}
+```
+**Note** : Global and port-level configurations are mutually exclusive but can coexist; interface-level configurations do not override global settings.
```
shak5009, line 268: This feels counter-intuitive to the generic LLDP configuration behavior in SONiC, where per-interface configurations strictly override global settings.

To maintain a consistent configuration paradigm, I suggest that custom_tlv_name follow the standard override behavior. If a TLV is set globally (e.g., TLV-G) and we want a specific interface to send an additional TLV (TLV-A), the user should explicitly configure that interface with [TLV-G, TLV-A]. This keeps the inheritance model predictable and aligns with how network operators expect port-level overrides to function.

shak5009 (commented):
Thanks for putting this PR together. I left a couple of inline comments regarding the data modeling (synchronization between APPL_DB and STATE_DB) and the configuration inheritance behavior for the custom TLVs. Let me know your thoughts!

doc/lldp_custom_tlv/lldp-custom-tlv-hld.md
```diff
@@ -0,0 +1,747 @@
...
+LLDPD integrates with the LLDP Manager through lldpcli interface:
+   - LLDP Manager communicates with lldpd through lldpcli commands
+   - Immediate application of custom TLV changes globally or to the respective interfaces
+   - Include configured custom TLVs in transmitted LLDP packets
+
+
+#### LLDP Sync Daemon (lldp_syncd)
+   - Retreive the operational data from lldpd using the lldpcli commands
+   - Extract custom TLVs from the operational data and update State DB (LLDP_RX_CUSTOM_TLV table) for feature daemon consumption
+   - Existing operational data sync happens every 10 seconds, while the LLDP custom TLV information will be synced every 1 second to ensure the feature daemon gets the updates quickly
```
lucas-li-micas, line 385: Will a 1-second polling interval be too frequent?

PR #2133, opened 2025-11-21
Fix spelling in Debug Framework Design Spec

lihuay:
Keep original capitalization: PseudoCode

Addressed feedback from @qiluo-msft on PR #2120


doc/debug_framework_design_spec.md
```diff
@@ -0,0 +1,394 @@
+# Debug Framework in SONiC
```
qiluo-msft, line 1: The file is moved to doc/debug-framework/debug_framework_design_spec.md.

It appears as new file in this PR.

PR #2147, opened 2025-12-03
HLD : OpenConfig support for Static Route 

rajakushwah:
### Summary
This PR adds comprehensive OpenConfig YANG model support for Static Route configuration in SONiC, enabling management through REST API and gNMI interfaces using industry-standard OpenConfig models.

### Supported Features
- Multi-protocol support (IPv4/IPv6)
- Multiple next-hop types (IP, interface, Null0)
- BFD configuration support
- REST API and gNMI interfaces


rajakushwah:
Creating a new PR

PR #2159, opened 2025-12-16
Feature/ PFCWD Hardware Recovery HLD

pinky-nexthop:
## Summary
This PR introduces a comprehensive High Level Design document for PFC Watchdog Hardware Recovery feature that provides **better accuracy in storm detection and restoration** compared to software-based approaches.

## Key Features
- **Hardware-based PFC recovery** mechanism using SAI attributes with **improved accuracy and reduced latency**
- **Architecture refactoring** with new PfcWdHwOrch orchestrator for clean separation of hardware/software logic
- **New CLI command**: `show pfcwd status` with hardware-specific information including actual timer values and granularity
- **Flow diagrams** illustrating hardware vs software recovery decision flows and detailed workflows
- **Comprehensive comparison** between software and hardware recovery mechanisms

## Hardware Recovery Advantages
- **Better accuracy in storm detection**: Hardware operates at line rate without software polling delays
- **Better accuracy in storm restoration**: More precise timing without dependency on software polling intervals
- **Reduced latency**: Immediate hardware detection and recovery actions
- **Lower CPU overhead**: Eliminates continuous software polling, reducing system load

## Document Structure
- Complete SONiC HLD format following established conventions
- Placed in `doc/pfcwd/` directory.
- References existing PFC Watchdog Design and Test Plan documents
- Includes SAI API specifications and comprehensive testing requirements

## Technical Highlights
- Hardware timer granularity constraints and actual vs configured values
- Runtime platform capability detection for automatic hardware/software selection
- Backward compatibility with existing software implementation
- Error handling and validation for hardware constraints

## Review Focus Areas
- Architecture design and orchestrator refactoring approach
- SAI attribute usage and hardware integration strategy
- CLI design and user experience for hardware-specific information
- Testing strategy and validation approach for accuracy improvements

## Related Pull Requests

| Index | Description | PR Number | Pre-requisites |
|-------|-------------|-----------|----------------|
| 1 | Add PFC Watchdog hardware state table definition | [sonic-swss-common#1169](https://github.com/sonic-net/sonic-swss-common/pull/1169/) | None |
| 2 | Update th5 parameter to support hw pfcwd | [sonic-buildimage#26471](https://github.com/sonic-net/sonic-buildimage/pull/26471) | None |
| 3 | New cli "show pfcwd status" support | [sonic-utilities#4384](https://github.com/sonic-net/sonic-utilities/pull/4384) | None |
| 4 | CLI UT for PFCWD status CLI | [sonic-utilities#4408](https://github.com/sonic-net/sonic-utilities/pull/4408/) | 3 |
| 5 | Refactor PFCWD | [sonic-swss#4358](https://github.com/sonic-net/sonic-swss/pull/4358/) | None |
| 6 | Skeleton for PFCWD Hw Recovery | [sonic-swss#4403](https://github.com/sonic-net/sonic-swss/pull/4403/) | 1, 5 |
| 7 | Implementation for Hw based PFCWD | [sonic-swss#4412](https://github.com/sonic-net/sonic-swss/pull/4412) | 1, 5, 6 |
| 8 | UT for flexcounters and orchagent changes | [sonic-swss#4418](https://github.com/sonic-net/sonic-swss/pull/4418) | 1, 5, 6, 7 |
| 9 | Add hardware-aware PFC watchdog test support | [sonic-mgmt#23430](https://github.com/sonic-net/sonic-mgmt/pull/23430) | None |


doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,359 @@
...
+
+```shell
+admin@sonic:~$ show pfcwd status
+PORT        RECOVERY TYPE    STATUS     HW DETECTION TIME    DETECTION GRANULARITY    HW RESTORATION TIME    RESTORATION GRANULARITY
+----------  -------------    --------   -------------------  ---------------------    ---------------------  -----------------------
+Ethernet0   hardware         success    300                  100ms                    500                    100ms
+Ethernet4   software         success    N/A                  N/A                      N/A                    N/A
+Ethernet8   hardware         success    200                  50ms                     400                    100ms
+Ethernet12  hardware         failed     N/A                  100ms                    N/A                    100ms
+```
```
kperumalbfn, line 392: In a switch, all ports will be either hardware or software based PFCWD. Can we have different per interface? Please clarify in the above example.

kperumalbfn, line 392: How are these timers determined for hardware-based model?

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,359 @@
...
+## 8. Manifest
+
+Not applicable - this is a core SONiC feature enhancement.
+
+## 9. CLI-YANG Model Enhancements
+
+No YANG model changes are required for this feature. The new `show pfcwd status` command displays operational state data, not configuration data.
+
+**Rationale**: The new command displays runtime operational data (actual hardware values, granularity constraints, recovery type) rather than user configuration. This operational data is derived from:
+- Platform hardware capabilities (granularity values)
```
kperumalbfn, line 540: Any SAI capability query to get the hardware granularity? If so, please include SAI attributes

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Yes, all ports will be either s/w or h/w based Pfcwd. Let me break this table into 2 for clarity. The table was intended to show the fields but I agree this is confusing. Thanks.

pinky-nexthop: Added a section `Timer determination for hardware-based model` to explain the behavior. 

First we query the range of the timers and current granularity of the port and then adjust the granularity and value to be programmed.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Added the relevant attributes to the section

pinky-nexthop (commented):
Thanks for the review, @kperumalbfn.
Updated the doc as per comments.

kperumalbfn:
Thanks @pinky-nexthop Could you also add update to STATE_DB with PFCWD detection and restoration time per port/queue. This can be used for both S/w and H/w based PFCWD.

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,387 @@
...
+Hardware recovery might have some constraints as listed below:
+
+**Timer granularity**
+There may be a granularity associated with the timers, so it is possible that the value configured in h/w is different from the value configured through the cli.
+For example: if the hardware granularity is 100 ms and configured detection time through cli is 250 ms, then hardware might program either 200/300 ms.
+
+**Upper limit on the timer value**
+Some platforms like broadcom only allow the timer value to be within the range <1-15> multiplied with the current granularity. So, it is possible that the value in the cli config cannot be programmed in the hardware.
+
+#### 5.1 New CLI command
```
vmittal-msft: @pinky-nexthop Instead of adding new CLI, Shall we enhance existing one and add new columns ? Since this about configuration of PFCWD.

Image

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,387 @@
...
+We propose to add a new CLI command `show pfcwd status` to display hardware-specific information including recovery type, programming status, hardware detection time, hardware restoration time, detection time granularity, and restoration time granularity. The STATUS column indicates whether the pfcwd configuration was successfully programmed (success/failed). Configuration can fail due to unsupported timer value ranges or hardware constraints.
```
vmittal-msft: @pinky-nexthop  Can you please share if we are going to keep o/p same for "show pfcwd stats" command ? Any enhacement we need for this ?

Image

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,387 @@
...
+**Timer granularity**
+There may be a granularity associated with the timers, so it is possible that the value configured in h/w is different from the value configured through the cli.
+For example: if the hardware granularity is 100 ms and configured detection time through cli is 250 ms, then hardware might program either 200/300 ms.
+
+**Upper limit on the timer value**
+Some platforms like broadcom only allow the timer value to be within the range <1-15> multiplied with the current granularity. So, it is possible that the value in the cli config cannot be programmed in the hardware.
+
+#### 5.1 New CLI command
+We propose to add a new CLI command `show pfcwd status` to display hardware-specific information including recovery type, programming status, hardware detection time, hardware restoration time, detection time granularity, and restoration time granularity. The STATUS column indicates whether the pfcwd configuration was successfully programmed (success/failed). Configuration can fail due to unsupported timer value ranges or hardware constraints.
+
```
vmittal-msft, line 308: @pinky-nexthop  please list any changes/enhacement coming to this list of config commands due to HW based PFCWD -

Image

vmittal-msft: 1. As we expect change in interval range programming, we may need update there.
2. pfc_stat_history : Will there be any impact to this feature due to move to HW based detection ?

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,387 @@
...
+#### Implementation Details
+
+`PfcWdBaseOrch` is the new base class that will have the common logic of parsing the CLI and creating the tasks. This implementation exists in `PfcWdOrch` today and will be moved to this new class. `PfcWdOrch` will be removed and renamed to `PfcWdBaseOrch` without the template parameters.
+
+`PfcWdHwOrch` is the new class that will handle hardware recovery mechanisms. There will be no handlers defined in hardware-based recovery unlike software recovery which defines drop and forward handlers. To update the counters periodically, `PfcWdHwOrch` itself will provide a function to read the required SAI attributes and get updated counters.
+
+Runtime selection in `orchdaemon.cpp`:
+
+```c
+// Check hardware capability
```
vmittal-msft: @pinky-nexthop  In addition, we may need chip type check else it may enable for existing SKUs where HW base support is there but we are still on SW model.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,387 @@
...
+- Configured values may differ from actual hardware-programmed values
+- Upper limits may apply based on platform (e.g., Broadcom: 1-15 × granularity)
+
+**Hardware recovery advantages:**
+- **Better accuracy in storm detection**: Hardware-based detection operates at line rate without software polling delays, providing more precise detection timing
+- **Better accuracy in storm restoration**: Hardware-based restoration timing is more accurate as it doesn't depend on software polling intervals or system load
+- **Reduced latency**: Hardware detection and recovery actions happen immediately without waiting for software polling cycles
+- **Lower CPU overhead**: Eliminates continuous software polling, reducing CPU utilization and system load
+
+#### 4.1.1 Hardware vs Software Recovery Flow
```
vmittal-msft, line 129: @pinky-nexthop  there are some flex counters for PFCWD stats which give us more information regarding trigger/restoration of event e.g. queue pause, pfc frames, xon off/on etc.. Will those still be available if we move to HW based model. Reason being those are usually helpful in case SW PFCWD misbehaves and helpful in debugging. If not, Are we going to add some syslogs to print state of the HW when event happens/restores ?

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,387 @@
...
+        HW->>HW: Apply Recovery Action
+    end
+
+    Note over HW: Storm Subsides
+    HW->>HW: Restoration Timer Expires
+    HW->>HW: Restore Normal Operation
+    HW->>SAI: Generate Restoration Event
+    SAI->>Orch: Storm Restored Event
+    Orch->>DB: Update Restoration Counters
+```
```
kperumalbfn, line 214: When PFCWD gets detected and restored, could we add that state in the STATE_DB? This could be applicable for both s/w and h/w based method

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,387 @@
...
+
+This design document is an extension of the existing PFC watchdog functionality described in the [PFC Watchdog Design](https://github.com/sonic-net/SONiC/wiki/PFC-Watchdog-Design). Newer network chips are capable of PFC storm detection and mitigation in the hardware without software intervention. Current implementation of software based PFC watchdog detection depends on constant polling of the queue's pause state, mitigation requires installation of ACLs to take drop action and restoration requires constant polling of PFC frames received on the queue.
+
+This document provides an overview of the programming requirements to enable hardware based recovery for PFC watchdog.
+
+## 2. Requirements
+
+The hardware-based PFC watchdog recovery feature shall provide:
+
+1. **Hardware Detection**: Automatic PFC deadlock detection in hardware without software polling
```
vmittal-msft, line 86: @pinky-nexthop  by HW detection, i believe it's a combination of HW and SDK. If you have information, Can you please share what part is covered by HW and SDK separately ?

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: > @pinky-nexthop Can you please share if we are going to keep o/p same for "show pfcwd stats" command ? Any enhacement we need for this ?
> 
> Image

There is no change in this cli.

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: Thanks for bringing this up. We thought about using the same config cli, however, chose to keep it as a different "status" cli  because of the following:
The key distinction is that the new CLI is not displaying configuration - it's displaying actual operational state and hardware-programmed values, which may differ from the configured values.
In software-based recovery, the configured detection/restoration times are always honored exactly as specified by the user. However, with hardware-based recovery, there are important differences:
Hardware Granularity Constraints: The hardware may have timer granularity limitations (e.g., 100ms increments). If a user configures 350ms, the hardware might actually program 400ms (ceiling) or 300ms (floor) based on the platform's capabilities. It is important to display both - the configured as well as operational value

Hardware Register Limitations: Some values may not be directly programmable in hardware registers and require adjustment to fit within supported ranges. In this case, the operational state should display failure to program. For s/w based recovery, failure was never an option.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Great suggestion! I will add it in the next revision. Thanks.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Sure, I can add this in the document

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Thanks for bringing the interval range programming. 
We have two options, either provide the ranges per sku in the cli. That would involve detecting the platform logic within the cli. The other option is to keep the cli as is with the current range and handle the programming based on the logic given in section 5.1. Any failures are reflected in the `show pfcwd status` cli. Keeping the cli as is, also prevents any breakage in the existing scripts used by consumers. 

There will be no change in pfc_stat_history as the counters work similarly as s/w based recovery.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: All the flex counters for pfcwd stats will still be available in the h/w mode. 
Yes, the implementation will ensure proper syslogging of events for better visibility/debuggability.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Sure, will add the relevant details.

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,527 @@
...
+
+| Phase | Software Recovery | Hardware Recovery |
+| ----- | ----- | ----- |
+| Deadlock detection requirement | Polling period for lua script configured in S/W | Enable h/w deadlock detection and recovery. Configure detection timer. Register callback for h/w events |
+| Detection mechanism | Lua script polls the queue to check if queue received PFC frames continuously for detection time | H/w detects the deadlock and publishes an event to the s/w to take action, if required. |
+| PFC pause frame handling | Software disables PFC by clearing the PFC mask bit for the affected queue (ignores incoming PFC pause frames) | Hardware sets the registers that ignores PFC pause frames for the affected queue during storm mitigation |
+| Mitigation | Orchagent programs ACLS to take drop action or programs a zero-buffer profile | H/w takes the drop/forward action |
+| Recovery requirement | Polling period for lua script configured in S/W | Configure recovery timer attribute |
+| Recovery | Lua script polls if the queue remains unpaused for restoration-time amount of time to recover | H/w recovers the queue and publishes an event to the s/w to take action, if required. |
```
rlhui: when h/w is set to ignore pfc pause frames, is it still able to check that it has received any pfc frames during that period?  so counters will still reflect that?

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,527 @@
...
+admin@sonic:~$ show pfcwd status
+PORT        RECOVERY TYPE    PROGRAMMING_STATUS    HW DETECTION TIME    DETECTION GRANULARITY    HW RESTORATION TIME    RESTORATION GRANULARITY
+----------  -------------    ------------------    -------------------  ---------------------    ---------------------  -------------------------
+Ethernet0   hardware         success               300                  100ms                    500                    100ms
+Ethernet4   hardware         failed                2000                 N/A                      N/A                    N/A
+Ethernet12  hardware         success               400                  100ms                    800                    100ms
+```
+
+**Key Observations:**
+- **Ethernet0**: Configuration (350ms/550ms) was adjusted to (300ms/500ms) by hardware due to 100ms granularity. The `show pfcwd config` shows what the user requested (350/550), but `show pfcwd status` reveals the actual hardware-programmed values (300/500).
```
lguohan: do you know who adjust the configuration? is that we program the SAI, SAI automatically round the value? Then, how do we know the real value being programmed in the hardware. @eswaran-nexthop ? Usually, that is not expectation of the SAI attribute. If user program a value, it should stay on that value. @kperumalbfn  as well.

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,527 @@
...
+### 6.1 SAI Attributes
+
+**Configuration Attributes:**
+
+| SAI Attribute | Description |
+| ------------- | ----------- |
+| `SAI_QUEUE_ATTR_ENABLE_PFC_DLDR` | Enable PFC deadlock detection and recovery |
+| `SAI_SWITCH_ATTR_QUEUE_PFC_DEADLOCK_NOTIFY` | Register callback for PFC deadlock events |
+| `SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL_RANGE` | Query supported detection timer range per port |
+| `SAI_PORT_ATTR_PFC_TC_DLD_TIMER_INTERVAL` | Set the granularity for the detection timer per port |
```
lguohan: this needs to be added to the SAI header?

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,527 @@
...
+      ├── PfcWdHwOrch (new hardware-based class - no handlers)
+      └── PfcWdSwOrch<DropHandler, ForwardHandler> (existing template class)
+```
+
+#### Implementation Details
+
+`PfcWdBaseOrch` is the new base class that will have the common logic of parsing the CLI and creating the tasks. This implementation exists in `PfcWdOrch` today and will be moved to this new class. `PfcWdOrch` will be removed and renamed to `PfcWdBaseOrch` without the template parameters.
+
+`PfcWdHwOrch` is the new class that will handle hardware recovery mechanisms. There will be no handlers defined in hardware-based recovery unlike software recovery which defines drop and forward handlers. To update the counters periodically, `PfcWdHwOrch` itself will provide a function to read the required SAI attributes and get updated counters.
+
```
rlhui, line 248: how frequent is this reading? is this part same (read SAI attributes) or any different than the SW-based approach?

doc/pfcwd/pfc_hardware_recovery_hld.md
rlhui: but it's better (much easier) when user only needs to do the same cli to check for pfcwd status.  Users in general (esp during suspected storm or packet drops situation) will not know what sku is doing hw or sw pfcwd.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,527 @@
...
+admin@sonic:~$ show pfcwd config
+PORT        ACTION    DETECTION TIME    RESTORATION TIME
+----------  --------  ----------------  ------------------
+Ethernet0   drop      350               550
+Ethernet4   drop      2000              2000
+Ethernet12  drop      400               800
+```
+
+Now, let's use the new `show pfcwd status` command to see the actual hardware-programmed values and programming status:
+
```
rlhui, line 362: "show pfcwd status" - "status" is a general term.  Is this going to support both sw-based and hw-based pfcwd approach?

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,527 @@
...
+- `SAI_QUEUE_STAT_BYTES` - Queue byte counters
+- `SAI_QUEUE_STAT_DROPPED_PACKETS` - Queue dropped packet counters
+- `SAI_QUEUE_STAT_DROPPED_BYTES` - Queue dropped byte counters
+
+## 7. Configuration and Management
+
+The hardware-based PFC watchdog uses the same configuration interface as the existing software implementation:
+
+- Configuration is done through CONFIG_DB PFC_WD table
+- Runtime selection between hardware and software is automatic based on platform capabilities
```
rlhui, line 451: if user want to stay with "sw" before switching to hw-based, even platform supports it, how to do so?

doc/pfcwd/pfc_hardware_recovery_hld.md
eswaran-nexthop: This adjustment is done in orchagent. SAI should program the value we give it. @pinky-nexthop for confirmation

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,527 @@
...
+    "last_restoration_time": <timestamp>
+    "storm_duration_ms": <value>
+    "detection_time_configured": <value_in_ms>
+    "detection_time_programmed": <value_in_ms>  // N/A for software mode or if programming failed
+    "detection_time_granularity": <value_in_ms>  // N/A for software mode or if programming failed
+    "restoration_time_configured": <value_in_ms>
+    "restoration_time_programmed": <value_in_ms>  // N/A for software mode or if programming failed
+    "restoration_time_granularity": <value_in_ms>  // N/A for software mode or if programming failed
+    "action": "drop" | "forward"
+```
```
rlhui, line 489: can this be supported with telemetry as well?  Please add test case for that.

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: Yes, that's right. 
Section 5 has the details as well

Timer determination for hardware-based model For hardware-based recovery, the actual timer values programmed in hardware are determined through the following process:

User Configuration: User configures desired timer values through CLI (e.g., detection_time=250ms)
Platform Capability Query: Software queries SAI capability attributes to determine hardware granularity constraints (see section 6.1 for SAI capability attributes)
Value Adjustment: Software adjusts the configured value to the nearest supported hardware value based on granularity
Hardware Programming: Adjusted value is programmed via SAI attributes (SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL / SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL)
Verification: Software reads back the actual programmed value from hardware to display in show pfcwd status

anilpannala:
Here is the recording of HLD review with community - https://zoom.us/rec/share/3Ue-srcURF5g3-UE2RPq7BeWIqNKJ-6B5HzehB8Tu25cnOm26qhdPJGns5ON7Q48.wo5M_LO4Ndaf7FEE

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: Yes, 
There is one more switch equivalent that is also part of saiswitchextensions.h which should move to saiswitch.h
SAI_SWITCH_ATTR_PFC_TC_DLD_TIMER_INTERVAL

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Thanks for bringing this up. This part is same as sw-based approach. 
I will add a note in the doc.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: We discussed this part in the meeting. 
For software based mode, the cli will just return `NA for software recovery mode. `
I will update the document with this info.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: The code will have a list of such platforms to override h/w based recovery. It is mentioned under  `SKU-Based Override Mechanism`

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: Yes, hardware still counts PFC frames even when configured to ignore them during storm mitigation.

The "ignore PFC pause frames" setting only affects the flow control action (queue pausing), not the frame reception and counting. The hardware:

- Continues to receive PFC frames at the MAC layer
- Continues to increment SAI_PORT_STAT_PFC_*_RX_PKTS counters
- Ignores the XOFF signal at the queue scheduler level (doesn't pause transmission)

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Sure.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
kperumalbfn: Could you add the list of past storm/restoration events? This will be useful to check the history of watchdog events

doc/pfcwd/pfc_hardware_recovery_hld.md
kperumalbfn: @pinky-nexthop let us add the SAI capability attributes to query the valid range and the granularity(could be port or switch level). This will be queried during orchagent init, so the range/granularity will be available during the config validation phase.

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: Updating the cli section such that there is no failed programming by using the min/max detection timer range. 
But still maintaining the `show pfcwd status` command to display if there was a round-off to the actual configured value based on granularity.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Updating the cli section such that there is no failed programming by using the min/max detection timer range.
But still maintaining the show pfcwd status command to display if there was a round-off to the actual configured value based on granularity.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Hey @kperumalbfn,
Do you mean to maintain a state table which stores the past events upto some configured limit (100 events) or so? 
```
PFC_WD_EVENT_HISTORY|Ethernet0|3|1
    "event_type": "storm_detected"
    "timestamp": "2026-02-05T10:00:00Z"
    
PFC_WD_EVENT_HISTORY|Ethernet0|3|2
    "event_type": "storm_restored"
    "timestamp": "2026-02-05T10:00:03Z"
    "storm_duration_ms": 3000```

File: doc/pfcwd/pfc_hardware_recovery_hld.md (hunk 11 above)
Comment (pinky-nexthop) [reply doc/pfcwd/pfc_hardware_recovery_hld.md]: Update the SAI attributes and also orchagent init section to clarify how it will be used.

File: doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md (hunk 8 above)
Comment (kperumalbfn) [reply doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md:214]: @pinky-nexthop could we have the list in one table entry per Interface|<Queue>? We could sort them based on the timestamp for the sequence

File: doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md [hunk 17]
```diff
@@ -0,0 +1,641 @@
...
+**Rationale**: The new command displays runtime operational data (actual hardware values, granularity constraints, recovery type) rather than user configuration. This operational data is derived from:
+- Platform hardware capabilities (granularity values)
+- Runtime hardware programming (actual timer values)
+- Platform detection logic (recovery type selection)
+
+The existing PFC watchdog configuration YANG model remains unchanged as all user-configurable parameters (detection_time, restoration_time, action, etc.) are already supported.
+
+## 10. Warmboot and Fastboot Design Impact
+
+There are no impacts to warmboot or fastboot. The hardware PFC watchdog configuration will be restored during the warmboot process through the standard CONFIG_DB restoration mechanism.
```
yunang-c, line 548: If the switch is capable of hardware recovery and is currently running an software watchdog, will the transition from software to hardware PFC watchdog via warmboot upgrade be supported?

If so, how is the transition handled? Suppose the software watchdog has already put some queues in the recovery mode before warmboot, would the hardware watchdog orch clean up these queues' states?

cc @akarshgupta25

rsudarsh-eridu (commented):
@pinky-nexthop  Thanks for adding this feature. The HLD references both SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL (port-level) and SAI_SWITCH_ATTR_PFC_TC_DLD_TIMER_INTERVAL (switch-level). Could you clarify Is per-TC (per-queue) granularity supported, or is the timer uniform across all PFC-enabled TCs on a port, if only the switch-level attribute is used, does that mean all ports share the same detection/restoration timer?

Thanks for confirming syslogs will be added for storm detection /restoration events. Could the HLD be updated with details on the log format, severity level, and whether any rate limiting is planned to avoid log flooding during repeated storm oscillation?

Could a test case for fallback to SW mode when the SAI capability query returns false be added to Section 12, with the actual test living in sonic-swss alongside the existing PFCWD orchagent tests? Thank you.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: @yunang-c 
No, the transition won't happen until the code for that chipset enables it. For example: in this PR https://github.com/sonic-net/sonic-swss/pull/4412/changes only the switches that return true for isHwPfcWdSupportedSku() will enable h/w based recovery. 
Once the sku/platform has tested the h/w recovery code, they can add to this list and use the new code.

pinky-nexthop:
> @pinky-nexthop Thanks for adding this feature. The HLD references both SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL (port-level) and SAI_SWITCH_ATTR_PFC_TC_DLD_TIMER_INTERVAL (switch-level). Could you clarify Is per-TC (per-queue) granularity supported, or is the timer uniform across all PFC-enabled TCs on a port, if only the switch-level attribute is used, does that mean all ports share the same detection/restoration timer?
> 
> Thanks for confirming syslogs will be added for storm detection /restoration events. Could the HLD be updated with details on the log format, severity level, and whether any rate limiting is planned to avoid log flooding during repeated storm oscillation?
> 
> Could a test case for fallback to SW mode when the SAI capability query returns false be added to Section 12, with the actual test living in sonic-swss alongside the existing PFCWD orchagent tests? Thank you.

@rsudarsh-eridu 
The timer is uniform across all PFC-enabled TCs on a port since the cli is applied at the port level. 
I added the switch-level attribute for completeness and is mostly legacy. Most newer chips support per port/per tc detection/timer intervals. 

Sure, we will add a test case in orchagent tests. I will also add the log format to the document.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,641 @@
+# PFC Watchdog Hardware Recovery
+
```
eddyk-nvidia, line 2: I personally find a bit misleading the use of "recovery" here and in the document when the feature is about Detection & Recovery. SAI API also clearly use DLD and DLR notation

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,641 @@
...
+## Revision
+
+| Rev | Date       | Author        | Change Description |
+|-----|------------|---------------|--------------------|
+| 0.1 | 12/13/2025 | Pinky Agrawal | Initial version    |
+| 0.2 | 02/05/2026 | Pinky Agrawal | Addressed review comments: Added configuration validation using SAI_SWITCH_ATTR_PFC_TC_DLD_INTERVAL_RANGE, updated SAI attributes section with extension attributes, clarified counter polling mechanism, updated show pfcwd status for software mode, added telemetry test cases |
+
+## Scope
+
+This document describes the design for hardware-based PFC (Priority Flow Control) watchdog recovery mechanism in SONiC. It covers the extension of existing software-based PFC watchdog functionality to support hardware-accelerated detection and recovery.
```
eddyk-nvidia: Even using the term of "HW-based" I would mention that precise partitioning of PFC WD processing is vendor-specific solution: it can be done fully in HW (ASIC), in ASIC FW, in the vendor SAI or even using any mix of them (e.g. detection in HW or FW and recovery in vendor SAI)

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,641 @@
...
+### 4.1 Hardware Recovery Mechanism
+
+Hardware recovery mechanism requires the application to program certain SAI attributes to initiate hardware based PFC deadlock detection and recovery. The below table highlights the key differences between hardware and software recovery.
+Each phase of hardware detection and recovery happens based on the programming of a corresponding SAI attribute.
+
+| Phase | Software Recovery | Hardware Recovery |
+| ----- | ----- | ----- |
+| Deadlock detection requirement | Polling period for lua script configured in S/W | Enable h/w deadlock detection and recovery. Configure detection timer. Register callback for h/w events |
+| Detection mechanism | Lua script polls the queue to check if queue received PFC frames continuously for detection time | H/w detects the deadlock and publishes an event to the s/w to take action, if required. |
+| PFC pause frame handling | Software disables PFC by clearing the PFC mask bit for the affected queue (ignores incoming PFC pause frames) | Hardware sets the registers that ignores PFC pause frames for the affected queue during storm mitigation |
```
eddyk-nvidia: HW handling can be vendor-specific which can include also disabling of PFC handling for affected queues

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,641 @@
...
+| Mitigation | Orchagent programs ACLS to take drop action or programs a zero-buffer profile | H/w takes the drop/forward action |
+| Recovery requirement | Polling period for lua script configured in S/W | Configure recovery timer attribute |
+| Recovery | Lua script polls if the queue remains unpaused for restoration-time amount of time to recover | H/w recovers the queue and publishes an event to the s/w to take action, if required. |
```
eddyk-nvidia: Recovery implementation is vendor-specific: can be HW, FW or even SAI SW

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,641 @@
...
+**Hardware recovery advantages:**
+- **Better accuracy in storm detection**: Hardware-based detection operates at line rate without software polling delays, providing more precise detection timing
+- **Better accuracy in storm restoration**: Hardware-based restoration timing is more accurate as it doesn't depend on software polling intervals or system load
+- **Reduced latency**: Hardware detection and recovery actions happen immediately without waiting for software polling cycles
+- **Lower CPU overhead**: Eliminates continuous software polling, reducing CPU utilization and system load
+
+#### 4.1.1 Hardware vs Software Recovery Flow
+
+The following diagram illustrates the decision flow for selecting hardware vs software recovery:
+
```
eddyk-nvidia, line 132: Is app_managed_recovery set by the vendor SAI to indicate whether SWSS should trigger it by sending SAI_QUEUE_ATTR_PFC_DLR_INIT (when app_managed_recovery == true) or not (when app_managed_recovery == false) ? If so, SAI_QUEUE_ATTR_PFC_DLR_INIT should not be used as a platform indication of support of PFC WD HW mode. Use something else for this purpose (e.g. DLD interval which must be set in any case for HW mode)

By the way, why indeed do we need the app to manage it ? What is exactly the point in control by app ? In some cases the app (SWSS) can decide not to do recovery ? Why not to immediately start handling recovery in SAI without loop through SWSS and sending the explicit trigger ?

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,641 @@
...
+    HW->>SAI: Generate Restoration Event
+    SAI->>Orch: Storm Restored Event
+    Orch->>DB: Update Restoration Counters
+    Orch->>StateDB: Update status=storm_restored,<br/>last_restoration_time,<br/>storm_duration_ms
+```
+
+#### 4.1.3 SDK and Hardware Responsibilities
+
+In hardware-based PFC watchdog recovery, the responsibilities are distributed between the SDK layer and the hardware (ASIC).
+
```
eddyk-nvidia: This information is vendor-specific. Particular partitioning might vary between different vendors

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,641 @@
...
+For such cases, the orchagent can maintain a list of such SKUs and skip creating `PfcWdHwOrch` for those SKUs as given in the code below.
+
+Runtime selection in `orchdaemon.cpp`:
+
+```c
+// Check hardware capability
+bool pfcHwRecoverySupported = gSwitchOrch->checkPfcHwRecoverySupport();
+
+// Check if SKU is in the override list to force software-based PFCWD
+std::vector<std::string> swPfcwdSkuList = getSwPfcwdSkuOverrideList();
```
eddyk-nvidia: I don't think it is a good direction - to use the SKU information. 

What about making a decision on which PFC WD OrcAgent to use based on 2 criterias:

1. Platform capabilities (from STATE_DB). By the way, I see PFC_WD_HW_CAPABILITIES table was mentioned in the doc. new one ? 
2. New global flag that can be introduced in PFC_WD|GLOBAL "mode" that can be set to "SW" or "HW". This flag can be configured by user and can override the platform capability. In this case even if the platform supports the PFC WD HW mode the user can override it setting the "SW" to PFC_WD|GLOBAL

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,641 @@
...
+std::string currentSku = gSwitchOrch->querySwitchSku();
+bool skuForcesSwPfcwd = std::find(swPfcwdSkuList.begin(), swPfcwdSkuList.end(), currentSku) != swPfcwdSkuList.end();
+
+// Check hardware capability
+bool pfcHwRecoverySupported = gSwitchOrch->checkPfcHwRecoverySupport();
```
eddyk-nvidia: checkPfcHwRecoverySupport() - which SAI attribute will be queried ? See my commend above about not using SAI_QUEUE_ATTR_PFC_DLR_INIT  for that . Also about some override that I proposed to support via CONFIG DB instead of dealing with SKU list

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,641 @@
...
+**Purpose of this command:**
+- Display the actual hardware-programmed timer values (which may differ from configured values due to granularity constraints) in h/w recovery mode. This command is N/A for s/w recovery mode.
+- Since there is no SAI attribute to query the current granularity setting, users need visibility into what values were actually programmed in hardware 
+
+**Timer determination for hardware-based model**
+For hardware-based recovery, the actual timer values programmed in hardware are determined through the following process:
+1. **User Configuration**: User configures desired timer values through CLI (e.g., detection_time=250ms)
+2. **Range Validation**: Software validates the configured value against the queried hardware range (`SAI_SWITCH_ATTR_PFC_TC_DLD_INTERVAL_RANGE`)
+3. **Granularity Selection**: Software determines the appropriate granularity (1ms, 10ms, or 100ms) based on the configured value
+4. **Value Adjustment**: Software adjusts the configured value to the nearest multiple of the selected granularity (e.g., 250ms with 100ms granularity → 200ms or 300ms)
```
eddyk-nvidia: I think this calculation can be vendor-specific

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,641 @@
...
+**Hardware Constraints:**
+
+Even with validation, hardware may have additional constraints:
+
+**Timer granularity**
+There may be a granularity associated with the timers, so it is possible that the value programmed in hardware is different from the value configured through the CLI.
+For example: if the hardware granularity is 100 ms and configured detection time through CLI is 250 ms, then hardware might program either 200ms or 300ms.
+
+Since there is no SAI attribute to query the current granularity setting today, the `show pfcwd status` command is needed to display the actual hardware-programmed values and verify if they match the configured values.
+
```
eddyk-nvidia, line 295: There might be additional HW constrains. For example, PFC WD DLDR SAI API allows the fully flexibility in configuration of detection/restoration intervals per port. So, each port can be configured with the unique timers. The vendor HW/FW might not support such flexibility. For example, it can support limited variation - might be a single set (same timer values for all ports) or limited set (e.g. 2 that can provide different configuration for uplink and downlink ports)

Does it make sense to support SAI_SWITCH_ATTR_PFC_TC_DLD_INTERVAL and SAI_SWITCH_ATTR_PFC_TC_DLR_INTERVAL for global configuration which can be overriden (if supported) by SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL / SAI_PORT_ATTR_PFC_TC_DLR_INTERVAL ? Similar configuration is used for QoS with few hierarchy - switch/port/tunnel - which can override each other for more specific configuration)

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,641 @@
...
+Ethernet12  drop      400               800
+```
+
+Now, let's use the new `show pfcwd status` command to see the actual hardware-programmed values:
+
+```shell
+admin@sonic:~$ show pfcwd status
+PORT        RECOVERY TYPE    HW DETECTION TIME    DETECTION GRANULARITY    HW RESTORATION TIME    RESTORATION GRANULARITY
+----------  -------------    -------------------  ---------------------    ---------------------  -------------------------
+Ethernet0   hardware         300                  100ms                    500                    100ms
```
eddyk-nvidia: "Recovery type" per port can confuse the user who can think that the detect/recovery type can be different per port. It is not the case, right ? Either of modes but one at a time can be activated on the device

doc/pfcwd/pfc_hardware_recovery_hld.md
eddyk-nvidia: I think this granularity information in CLI command might be confusing to a user. Still the real value configured to HW might be very vendor-specific and can't be calculated from the configured time and granularity

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,641 @@
...
+| `SAI_PORT_ATTR_PFC_TC_DLD_TIMER_INTERVAL` (Extension) | sai_map_list_t | CREATE_AND_SET | Port PFC Deadlock Detection timer granularity of all PFC priorities |
+
+**Switch Attributes (SAI_OBJECT_TYPE_SWITCH):**
+
+| SAI Attribute | Type | Flags | Description |
+| ------------- | ---- | ----- | ----------- |
+| `SAI_SWITCH_ATTR_QUEUE_PFC_DEADLOCK_NOTIFY` | sai_pointer_t | CREATE_AND_SET | Register callback for PFC deadlock events |
+| `SAI_SWITCH_ATTR_PFC_DLR_PACKET_ACTION` | sai_packet_action_t | CREATE_AND_SET | Configure drop/forward action during recovery |
+| `SAI_SWITCH_ATTR_PFC_TC_DLD_INTERVAL_RANGE` | sai_u32_range_t | READ_ONLY | Query supported detection timer range at switch level (capability query) |
+| `SAI_SWITCH_ATTR_PFC_TC_DLD_TIMER_INTERVAL` (Extension) | sai_map_list_t | CREATE_AND_SET | PFC Deadlock Detection timer granularity in milliseconds (1ms, 10ms, or 100ms) |
```
eddyk-nvidia: what about SAI_SWITCH_ATTR_PFC_TC_DLD_INTERVAL and SAI_SWITCH_ATTR_PFC_TC_DLR_INTERVAL ?

Please see my proposal on using the switch level intervals and when needed override with PORT based intervas

Let's see how to use all of them. I know that MSFT uses today 1 set of detection/recovery timers for all ports

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,641 @@
...
+
+The following STATE_DB table is used to track PFCWD detection and restoration events for both hardware and software-based methods. This table serves as the data source for the `show pfcwd status` command and telemetry reporting.
+
+**PFC_WD_STATE Table:**
+
+This table tracks the current state and configuration of PFC watchdog on each port/queue.
+
+```
+PFC_WD_STATE|<port_name>|<queue_index>
+    "recovery_type": "hardware" | "software"
```
eddyk-nvidia: check if you want this data per port. It is actually global

doc/pfcwd/pfc_hardware_recovery_hld.md
```diff
@@ -0,0 +1,641 @@
+# PFC Watchdog Hardware Recovery
```
eddyk-nvidia: @pinky-nexthop - please check my comments below. Thanks

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Changed title to hardware-based detection and recovery for clarity.

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: Added a note in the scope section to clarify, that pfcwd can be implemented in a vendor specific way and this document is presented from a sonic-perspective using SAI as the interface.

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: Under the scope section, added a note to clarify the use of word hardware throughout the document. IT could mean any of the vendor-specific implementation as mentioned by you.

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: Under the scope section, added a note to clarify the use of word hardware throughout the document. IT could mean any of the vendor-specific implementation as mentioned by you.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: No, app_managed_recovery is not set by vendor SAI. It has to be set by swss during the callback triggered as part of detection. 
One example where this could be useful:  there may be a requirement where a platform might want to forward certain kind of traffic for drop action, so it could install some custom acl before setting `SAI_QUEUE_ATTR_PFC_DLR_INIT = true.`

SAI_QUEUE_ATTR_PFC_DLR_INIT=TRUE is required to start recovery, else the queue remains in stormed state indefinitely.

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: Removing this section as each vendor might have different implementation.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: The sku based determination is short term until all platforms have verified this code works for their skus. 
there may be a requirement for systems to keep using s/w based recovery until this code is fully verified.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: broadcom uses SAI_QUEUE_ATTR_PFC_DLR_INIT. This may be different for different platforms. I am not sure.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Sure, the design is flexible and doesn't impose the use of SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL. 
We will provide a base implementation using SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL and other vendors can build on top as per their requirements.

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: Please find updated cli output in the document. This can also be viewed in the PR https://github.com/sonic-net/sonic-utilities/pull/4384/changes

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: yes, we had a SAI OCP review on the broadcom extension about exposing granularity config through a SAI attribute. Broadcom agreed to hide this detail behind the SAI_PORT_ATTR_PFC_TC_DLD_INTERVAL programming since NOS doesn't need to be aware of this vendor specific detail. 
Removing all granularity related items from the document.

doc/pfcwd/pfc_hardware_recovery_hld.md
pinky-nexthop: I added SAI_SWITCH_ATTR_PFC_TC_DLD_INTERVAL and SAI_SWITCH_ATTR_PFC_TC_DLR_INTERVAL in the SAI attribute list.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
eddyk-nvidia: so, in this case it is not clear who and how sets "app_managed_recovery". I thought some vendor can start the recovery process without triggering by SWSS and in this case the vendor SAI starts recovery immediately without waiting for SAI_QUEUE_ATTR_PFC_DLR_INIT=TRUE. Please clarify the meaning of app_managed_recovery=false

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
eddyk-nvidia: why from the beginning not to consider the flag in PFC_WD|GLOBAL ? It doesn't require any special processing. Just can override the platform capability. If it is not set in CONFIG DB the reported HW capability is used. Otherwise, it can override the reported HW capability. I still think that the use of SKU information in generic SWSS code is not a good direction

@kperumalbfn  - what do you think ?

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
eddyk-nvidia: As I mentioned: there might 2 different types of handling by vendors:

1. SAI reports the detection of PFC storm --> SWSS initiates the recovery by setting SAI_QUEUE_ATTR_PFC_DLR_INIT
2. SAI reports the detection of PFC storm and starts the recovery without waiting for trigger from SWSS.

I thought it is the meaning of "app_managed_recovery". If needed we can introduce a new SAI attribute to know the platform preference

See for example SAI_SWITCH_ATTR_HOSTIF_OPER_STATUS_UPDATE_MODE. It can be read/write or write only

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
eddyk-nvidia: @pinky-nexthop - could you please address my last question about meaning of "app_managed_recovery" ?

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
eddyk-nvidia: @pinky-nexthop , @kperumalbfn  - could you please address my question about use of SKU data in SONIC ?

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
eddyk-nvidia: @pinky-nexthop  - please address my comment. How to handle #2 option from my list ?

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,668 @@
...
+## 2. Requirements
+
+The hardware-based PFC watchdog recovery feature shall provide:
+
+1. **Hardware Detection**: Automatic PFC deadlock detection in hardware without software polling
+2. **Hardware Recovery**: Automatic recovery actions (drop/forward) performed by hardware
+3. **Backward Compatibility**: Existing software-based PFC watchdog functionality remains unchanged
+4. **Runtime Selection**: Automatic selection between hardware and software implementations based on platform capabilities
+5. **Enhanced CLI**: New `show pfcwd status` command to display recovery type, programming status, actual hardware timer values
```
vmittal-msft, line 90: @pinky-nexthop  Did we decide to keep on existing CLI ?

rookie-who:
Hi @pinky-nexthop — the 202605 branch cut is at end of May. Please address @vmittal-msft's review feedback. This looks close to ready.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: I used this approach based on the comment https://github.com/sonic-net/SONiC/pull/2159#discussion_r2743100301 from @vmittal-msft

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: Yes, that's correct — app_managed_recovery=false means SAI/SDK starts and completes recovery on its own, without waiting for DLR_INIT=true from SWSS, and emits a SAI_QUEUE_PFC_DEADLOCK_EVENT_TYPE_RECOVERED notification when done (per the spec comment on sai_queue_deadlock_notification_data_t.app_managed_recovery in inc/saiqueue.h).

The field is set by the host (SWSS) as a return value from the DETECTED notification callback — so it's a per-event decision made by SWSS, not reported by vendor SAI.

Concrete vendor mapping: the Broadcom SDK supports both modes via the pfc_deadlock_seq_control SOC property. With pfc_deadlock_seq_control=0 (SDK default), the SDK auto-runs recovery and the host return value is ignored. With pfc_deadlock_seq_control=1, the SDK honors the host's return — recovery start/stop is then driven by SWSS via SAI_QUEUE_ATTR_PFC_DLR_INIT=true/false.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: You are right on the 2 different types of handling. Worth noting app_managed_recovery is the host → vendor return value, not a vendor capability — SWSS returns it on the callback. On Broadcom the SDK gate is pfc_deadlock_seq_control: with =1 (current SONiC default) the SDK honors the host's return so both methods are reachable; with =0 the SDK forces method 2 regardless of host return.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
prsunny: @vmittal-msft , this should be based on capability check and/or config knobs.

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
pinky-nexthop: We had this discussion https://github.com/sonic-net/SONiC/pull/2159#discussion_r2742990718 and concluded that the cli will only allow valid intervals of the h/w recovery based on each platform, so that failure doesn't happen while programming because of invalid range. However, if a platform doesn't support programming of certain SAI attributes or fails for some reason, then there is no way to know if the config made it successfully to the h/w. So, technically the user doesn't have to look at this new CLI as it will be successful most of the time. 
This CLI also provides information about which mode (sw/hw) is on and what are the intervals for restoration/detection and the actual values that got programmed to the h/w [there can be some granularity based round-off, so this cli can provide the actual h/w value vs configured value]. 
Let me know if you think this is not required for the feature, we can certainly remove it and rely on syslog for any SAI related errors. 

```
admin@gold409:~$ show pfcwd status
PFC Watchdog Mode: Hardware
Supported hardware detection interval range: 100-1500 ms
Supported hardware restoration interval range: 100-1000 ms

     PORT      STATUS    HW DETECTION TIME (ms)    HW RESTORATION TIME (ms)
---------  ----------  ------------------------  --------------------------
Ethernet0  configured                       400                         400
Ethernet8  configured                       800                         800
 ```

File: doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md (hunk 17 above)
Comment (yunang-c) [reply doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md:548]: Thanks for the response. I want to follow up on a specific warmboot upgrade scenario: 
1. The switch is running PFC software WD (either on a old image with no hardware WD support, or is not part of `isHwPfcWdSupportedSku` ), and detects deadlocks on some queues.
2. While these queues are still in recovery, a warmboot upgrade happens.
3. Assume the new image determines the switch should run hardware recovery and starts the PFC hardware WD orch, how does it handle these queues?

The software WD orch currently reads PFC_WD_TABLE in APP DB to re-initiate the recovery of such queues after warmboot. But in the new design, it seems orchagent does not create the software WD orch when it determines the switch should run hardware recovery. Will the hardware WD orch take care of these queues?

**Comment by rookie-who** [comment]

File: doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md (hunk 17 above)
Comment (pinky-nexthop) [reply doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md:548]: Thanks for raising this point, very valid.
However, a sw→hw switchover cannot happen across a warmboot, so the problematic case you've described (queues mid-recovery under software WD, then coming up under hardware WD) doesn't actually arise.

On Broadcom chips, whether the switch runs hardware-based PFC recovery is determined by the SOC properties file, which is only consumed during a cold boot / full SDK reinitialization. Warmboot deliberately preserves the existing hardware/SDK state and does not re-read these properties, so the recovery mode (sw vs. hw) is effectively fixed across a warmboot — it cannot flip from software to hardware as part of the upgrade.

Comment (yunang-c) [reply doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md:548]: If the switch is on an older release image (that doesn't have this feature) pre-warmboot and we upgrade it to a new image with this PFC WD feature during warmboot, it seems a software->hardware transition will happen? Since the new image will read the hardware capability and enable hardware recovery.

Or if the preboot image has the SKU whitelisted for software recovery but the postboot image removed it from the whitelist, will it also trigger a transition?

Comment (pinky-nexthop) [reply doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md:548]: > If the switch is on an older release image (that doesn't have this feature) pre-warmboot and we upgrade it to a new image with this PFC WD feature during warmboot, it seems a software->hardware transition will happen? Since the new image will read the hardware capability and enable hardware recovery.

The way new image will support this capability will be by changing the SOC property on the newer image as we ll as the new orchagent code. Now, the SOC property doesn't get applied during a warm reboot (at least for broadcom), so transition will not happen. 

> Or if the preboot image has the SKU whitelisted for software recovery but the postboot image removed it from the whitelist, will it also trigger a transition?

We discussed whitelisting with @kperumalbfn and agreed that we should not hardcode skus in orchagent. All the capability decision should happen using SAI capability queries only. I will update the document to reflect that we have decided to use this method for deciding which path to take - sw vs hw in orchagent.
Hope this answers your question.

File: doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md (hunk 32 above)
Comment (vmittal-msft) [reply doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md:90]: @pinky-nexthop Is it possible to merge this information with "show pfcwd stats" & "show pfcwd config" CLI ?

File: doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md (hunk 14 above)
Comment (vmittal-msft) [reply doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md:362]: @pinky-nexthop  i would prefer merging this new information in existing commands.

**Review by vmittal-msft — APPROVED** [review]

**Comment by vmittal-msft** [comment]
@saksarav-nokia @ysmanman @arista-nwolfe please help review soon as we are planning to enable HW based PFCWD for upcoming SKUs and potentially for chassis as well.

**Review by saksarav-nokia — APPROVED** [review]

**Comment by pinky-nexthop** [comment]
@prsunny @vmittal-msft 
Could you please help merge this to master?

**Comment by vmittal-msft** [comment]
@pinky-nexthop  i am waiting for @kperumalbfn to approve as well.

**Comment by vmittal-msft** [comment]
@kperumalbfn @eddyk-nvidia please help approve.

**Comment by pinky-nexthop** [comment]
@rlhui @vmittal-msft kindly help merge

**Comment by pinky-nexthop** [comment]
Gentle reminder to merge @rlhui @prsunny @yxieca

File: doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md [hunk 33]
```diff
@@ -0,0 +1,594 @@
...
+
+`PfcWdBaseOrch` is the new base class that will have the common logic of parsing the CLI and creating the tasks. This implementation exists in `PfcWdOrch` today and will be moved to this new class. `PfcWdOrch` will be removed and renamed to `PfcWdBaseOrch` without the template parameters.
+
+`PfcWdHwOrch` is the new class that will handle hardware recovery mechanisms. There will be no handlers defined in hardware-based recovery unlike software recovery which defines drop and forward handlers. To update the counters periodically, `PfcWdHwOrch` will use the same counter polling mechanism as software-based recovery. The counter querying remains the same as the software-based approach, using SAI queue statistics APIs (`SAI_QUEUE_STAT_PACKETS`, `SAI_QUEUE_STAT_DROPPED_PACKETS`, etc.) to read hardware counters and update COUNTERS_DB.
+
+#### PfcWdHwOrch vs PfcWdSwOrch Selection
+
+`checkPfcHwRecoverySupport()` is called at init to decide whether to instantiate `PfcWdHwOrch` or fall back to `PfcWdSwOrch`. It returns true when the platform's SAI implementation advertises support for hardware-based PFC deadlock detection and recovery through a SAI capability query.
+
+## 5. CLI Changes
```
richard-nexthop, line 253: Please add nhcli configuration and operational commands.

richard-nexthop (commented):
Removed comment

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
```diff
@@ -0,0 +1,594 @@
...
+**Hardware Constraints:**
+
+Even with validation, hardware may have additional constraints:
+
+**Timer granularity**
+The actual value programmed in hardware may differ from the value configured through the CLI due to vendor-specific granularity constraints and timer implementation details.
+
+Since there is no SAI attribute to query the current granularity setting today, the `show pfcwd status` command is needed to display the actual hardware-programmed values and verify if they match the configured values.
+
+### 5.2 New CLI command
```
richard-nexthop, line 296: Please add nhcli configuration and nhcli operational commands that are required for 202605.1.

richard-nexthop (commented):
Removed comment

doc/pfcwd/pfc_watchdog_hardware_detection_recovery.md
richard-nexthop: Will work with Pinky offline to gather the nhcli components.

prsunny:
@kperumalbfn , would you approve?

kperumalbfn (approved):

PR #2173, opened 2026-01-07
Update Media-based-Port-settings.md to support custom SerDes attributes

longhuan-cisco:
sonic-swss side change: https://github.com/sonic-net/sonic-swss/pull/3764 (merged)
sonic-platform-daemons side change: https://github.com/sonic-net/sonic-platform-daemons/pull/643 (under review)

Add custom SerDes attributes ([SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION](https://github.com/opencomputeproject/SAI/pull/2156)) support in media_settings.json

### DIFF
--- a/doc/media-settings/Media-based-Port-settings.md
+++ b/doc/media-settings/Media-based-Port-settings.md
@@ -144,7 +144,9 @@
 
 Custom SerDes attributes are vendor-specific SerDes settings that are not covered by standard `SAI_PORT_SERDES_*` attributes, especially those not suitable for public standardization in SAI (for example, proprietary ones). `SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION` was introduced to carry these settings as a single JSON string based attribute.
 
-In media_settings.json, define custom attributes using keys prefixed with `CUSTOM:` (for example, `CUSTOM:XYZ`). xcvrd's media_settings_parser strips the prefix, and aggregates them into a single JSON string field `custom_serdes_attrs` in APPL_DB `PORT_TABLE`. The JSON format is a list of attribute objects, each with a per-lane `value` array sized for the logical port, and portsorch maps that field to `SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION`. portsorch doesn't interpret the contents; it simply passes it through to SAI.
+In media_settings.json, define custom attributes under the top-level `CUSTOM_MEDIA_SETTINGS` section using keys prefixed with `CUSTOM:` (for example, `CUSTOM:XYZ`). `CUSTOM_MEDIA_SETTINGS` is keyed by physical port selectors and supports the same selector syntax as `GLOBAL_MEDIA_SETTINGS` (ranges, lists, list-of-ranges, and single-port keys). xcvrd's media_settings_parser strips the prefix, and aggregates them into a single JSON string field `custom_serdes_attrs` in APPL_DB `PORT_TABLE`. The JSON format is a list of attribute objects, each with a per-lane `value` array sized for the logical port, and portsorch maps that field to `SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION`. portsorch doesn't interpret the contents; it simply passes it through to SAI.
+
+Within each port selector, the media/vendor key matching, lane speed pattern matching, and `Default` fallback behavior are identical to traditional `GLOBAL_MEDIA_SETTINGS` and `PORT_MEDIA_SETTINGS`.
 
 ### Value of `SAI_PORT_SERDES_ATTR_CUSTOM_COLLECTION` is a JSON string in the following format
 
@@ -171,7 +173,7 @@ Note: The above JSON is formatted with spaces and newlines for readability to he
 
 ```
 {
-    "GLOBAL_MEDIA_SETTINGS": {
+    "CUSTOM_MEDIA_SETTINGS": {
         "1-32": {
             "Default": {
                 "CUSTOM:XYZ": {
@@ -221,7 +223,13 @@ Example APPL_DB PORT_TABLE entry based on the above media_settings.json:
                     "lane1": "0xc",
                     "lane2": "0xd",
                     "lane3": "0xa"
-                },
+                }
+            }
+        }
+    },
+    "CUSTOM_MEDIA_SETTINGS": {
+        "1-32": {
+            "Default": {
                 "CUSTOM:XYZ": {
                     "lane0": 10,
                     "lane1": 11,


longhuan-cisco:
@prgeor Please review, thanks

doc/media-settings/Media-based-Port-settings.md
```diff
@@ -140,6 +140,113 @@
...
+
+Example APPL_DB PORT_TABLE entry based on the above media_settings.json:
+
+```
+# sonic-db-cli APPL_DB hgetall "PORT_TABLE:EthernetXXX"
+{
+    ...
+    'preemphasis': '0x112233,0x112233,0x112244,0x443322',
+    'idriver': '0xb,0xc,0xd,0xa',
+    'custom_serdes_attrs': '{"attributes":[{"XYZ":{"value":[10,11,12,13]}}]}'
```
prgeor: @longhuan-cisco if you keep the custom attributes as follows, I don't need a NEW parser for custom attribute for xcvrd. For eg

```
 'preemphasis': '0x112233,0x112233,0x112244,0x443322',
  'idriver': '0xb,0xc,0xd,0xa',
  'custom_attribute1': '{"attribute_XYZ": '0x10, 0x12, 0x29, 0x30'}',
  'custom_attribute2': '{"attribute_ABC": '0x11, 0x32, 0x56, 0x20'}',
  'custom_attribute3': '{"attribute_PQR": '0x90, 0x00, 0x99, 0x77'}',

```

doc/media-settings/Media-based-Port-settings.md
```diff
@@ -140,6 +140,113 @@
...
+    'custom_serdes_attrs': '{"attributes":[{"XYZ":{"value":[10,11,12,13]}},{"ABC":{"value":["mode_a","mode_b","mode_c","mode_d"]}}]}'
+    ...
+}
+```
+
+### Example media_settings.json with mixed traditional and custom attributes
+
+```
+{
+    "GLOBAL_MEDIA_SETTINGS": {
```
prgeor, line 212: @longhuan-cisco can you introduce "CUSTOM_MEDIA_SETTINGS" so that we can have a separate parser for custom attributes. I know this will mandate to do the port range and per port attribute parsing again for custom attribute but the decision to introduce custom attributes deviates from general sonic requirements so ok to introduce this new section. This will help to contain the changes to this feature only.

longhuan-cisco: @prgeor Thanks for the review and for raising the isolation / blast-radius concern — I agree we should keep the risk to existing platforms/vendors as low as possible, especially for those that don’t define any `CUSTOM:XXX` attributes in `media_settings.json`.

I considered the suggestion of introducing a dedicated `CUSTOM_MEDIA_SETTINGS` section. My main concern is that we’d either need to (a) replicate the existing parsing/matching behaviors (port ranges, `DEFAULT_KEY`, vendor+PN matching, regex key matching, `media_lane_speed_key`, etc.), which could become harder to maintain over time, or (b) do a larger refactor to share that logic, which may increase risk more than necessary for this feature.

As a small, low-impact compromise, I propose adding an early-return guard inside `handle_custom_serdes_attrs()`:

- If the resolved `media_dict` has **no** `CUSTOM:` keys, return immediately (no `deepcopy`, no conversion logic), so the legacy behavior stays unchanged for platforms/vendors that don’t opt in via `CUSTOM:` keys.
- If `CUSTOM:` keys exist, proceed exactly as before for the opt-in case.

I updated https://github.com/sonic-net/sonic-platform-daemons/pull/643 as the same, please refer:
https://github.com/longhuan-cisco/sonic-platform-daemons/blob/27dfe58d4668d678b2ad850b076194ca17891b76/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py#L340-L358
```
def handle_custom_serdes_attrs(media_dict, lane_count, subport_num):
    """
    Handle custom SerDes attributes in the media_dict and convert them to JSON.

    Args:
        media_dict: dictionary containing SerDes settings for all lanes of the port
        lane_count: number of lanes for this subport
        subport_num: subport number (1-based), 0 for non-breakout case

    Returns:
        media_dict: updated media_dict with custom SerDes attributes converted to JSON,
                    or the original media_dict unchanged if no custom attributes present
    """
    # Keep the custom-SerDes-attributes feature fully opt-in.
    # If there are no "CUSTOM:" keys in the resolved media_dict, do not run any of the
    # new conversion logic (and avoid deepcopy). This minimizes blast radius for
    # existing platforms/vendors that don't define custom SerDes attributes.
    if not any(isinstance(k, str) and k.startswith(CUSTOM_SERDES_ATTR_PREFIX) for k in media_dict):
        return media_dict
```

If you’d still prefer an even stricter opt-in boundary, one option would be to do a **one-time scan of `media_settings.json`** (e.g., look for the substring `"CUSTOM:"`) and only enable/invoke the custom handling when that marker exists in the file at all.

longhuan-cisco: @prgeor Let me know if [this](https://github.com/longhuan-cisco/sonic-platform-daemons/blob/27dfe58d4668d678b2ad850b076194ca17891b76/sonic-xcvrd/xcvrd/xcvrd_utilities/media_settings_parser.py#L340-L358) early-return guard addresses the isolation requirement. I’m happy to adjust further if you’d prefer an even stricter opt-in mechanism (e.g., a one-time scan of `media_settings.json` to detect `"CUSTOM:"` before enabling the custom handling), or if you still feel a dedicated `CUSTOM_MEDIA_SETTINGS` section is the better direction.

prgeor: @longhuan-cisco there is hardly 24 lines to duplicate which is already happening (between the parsing of GLOBAL_MEDIA_SETTINGS_KEY and PORT_MEDIA_SETTINGS_KEY) but you can use oops principal if you desire to minimize the duplication.

longhuan-cisco: @prgeor 

Sure, I've implemented the dedicated CUSTOM_MEDIA_SETTINGS section as requested at https://github.com/sonic-net/sonic-platform-daemons/pull/643. Please let me know if it looks ok. Thanks.

One thing to flag from the dedicated-CUSTOM_MEDIA_SETTINGS approach: the dedicated section adds a new selector/lookup path (is_port_selected,get_custom_media_settings_value) and an extra merge path in notify_media_setting. The earlier approach only added a single post‑processing helper (handle_custom_serdes_attrs) and reused the existing GLOBAL/PORT selection path, so it touched less of the parser.
  - old approach (no dedicated section) [code reference](https://github.com/longhuan-cisco/sonic-platform-daemons/compare/fd863c109637890f0dd2bfa06bcd5f01db2b82ac...27dfe58d4668d678b2ad850b076194ca17891b76)
      - Adds handle_custom_serdes_attrs()
      - Single hook in notify_media_setting()
      - Reuses existing GLOBAL/PORT lookup path; no new selector logic
  - Current approach (dedicated CUSTOM_MEDIA_SETTINGS) [code reference](https://github.com/longhuan-cisco/sonic-platform-daemons/compare/40ccd149514634a96076b8f1a85aed30e0846329...3858b95a395e859ffd5efc7905a47c4f786d1c46)
      - Adds new path: get_custom_media_settings_value()
      - Adds selector parser: is_port_selected()
      - Adds get_custom_serdes_attrs()
      - Extra control flow in notify_media_setting() to merge custom JSON
      - Extracts get_media_settings_for_key() to reuse matching logic

I’m totally good sticking with the dedicated section if separation is the priority, but wanted to make the added maintenance surface and more touching of the existing code explicit now that we have code to compare.

prgeor (approved):

PR #2182, opened 2026-01-20
hamgrd design and implementation doc

yue-fred-gao:
Adding hamgrd design and implementation doc


BYGX-wcr:
@yue-fred-gao, there is an existing HLD for HAmgrd but it does not contain all the details here. Let's merge or connect them in some way.

BYGX-wcr:
@yue-fred-gao , please fix the DCO check

yue-fred-gao:
> @yue-fred-gao , please fix the DCO check

Fixed. Thanks!

PR #2152, opened 2025-12-09
CPO support in SONiC

KroosMicas:
Describe how SONiC supports CPO devices.

PR #2183, opened 2026-01-20
HLD for banking support for CMIS

bobby-nexthop:
HLD undergoing some minor changes and in draft mode.

### DIFF
--- a/doc/layer1/cmis_banking_support.md
+++ b/doc/layer1/cmis_banking_support.md
@@ -1,0 +1,924 @@
+# I2C Transceiver Page Banking Support HLD
+
+## Table of Content
+
+- [1. Revision](#1-revision)
+- [2. Scope](#2-scope)
+- [3. Definitions/Abbreviations](#3-definitionsabbreviations)
+- [4. Overview](#4-overview)
+- [5. Requirements](#5-requirements)
+- [6. Architecture Design](#6-architecture-design)
+- [7. High-Level Design](#7-high-level-design)
+  - [7.1 Repositories Changed](#71-repositories-changed)
+  - [7.2 Kernel Driver Changes (optoe.c)](#72-kernel-driver-changes-optoec)
+  - [7.3 Memory Map Abstraction Changes (cmis.py)](#73-memory-map-abstraction-changes-cmispy)
+  - [7.4 Field Definition Changes (xcvr_field.py)](#74-field-definition-changes-xcvr_fieldpy)
+  - [7.5 CLI Utility Changes (sfputil)](#75-cli-utility-changes-sfputil)
+  - [7.6 Linear Offset Calculation](#76-linear-offset-calculation)
+  - [7.7 Backwards Compatibility](#77-backwards-compatibility)
+  - [7.8 API Design for Multi-Bank Modules](#78-api-design-for-multi-bank-modules)
+  - [7.9 DB and Schema Changes](#79-db-and-schema-changes)
+  - [7.10 SWSS and Syncd Changes](#710-swss-and-syncd-changes)
+  - [7.11 Sequence Diagram](#711-sequence-diagram)
+  - [7.12 Implementation Phases](#712-implementation-phases)
+  - [7.13 Platform Dependencies](#713-platform-dependencies)
+- [8. SAI API](#8-sai-api)
+- [9. Configuration and Management](#9-configuration-and-management)
+  - [9.1 Manifest](#91-manifest)
+  - [9.2 CLI/YANG Model Enhancements](#92-cliyang-model-enhancements)
+  - [9.3 Config DB Enhancements](#93-config-db-enhancements)
+- [10. Warmboot and Fastboot Design Impact](#10-warmboot-and-fastboot-design-impact)
+- [11. Memory Consumption](#11-memory-consumption)
+- [12. Restrictions/Limitations](#12-restrictionslimitations)
+- [13. Testing Requirements/Design](#13-testing-requirementsdesign)
+  - [13.1 Unit Test Cases](#131-unit-test-cases)
+  - [13.2 System Test Cases](#132-system-test-cases)
+- [References](#references)
+
+## 1. Revision
+
+| Rev | Date       | Author | Change Description |
+|-----|------------|--------|--------------------|
+| 0.1 | 2026-01-20 | bobby-nexthop      | Initial version    |
+
+## 2. Scope
+
+This high-level design document describes the implementation of page banking support for I2C transceivers (CMIS devices) in SONiC. Currently, the optoe kernel driver only supports Bank 0 pages. This enhancement enables access to all banks as defined in the CMIS specification, which is required for multi-lane modules with more than 8 lanes.
+
+## 3. Definitions/Abbreviations
+
+| Term | Definition |
+|------|------------|
+| CMIS | Common Management Interface Specification |
+| I2C | Inter-Integrated Circuit |
+| QSFP-DD | Quad Small Form Factor Pluggable Double Density |
+| OSFP | Octal Small Form Factor Pluggable |
+| VDM | Versatile Diagnostics Monitoring |
+| SFP-DD | Small Form Factor Pluggable Double Density |
+| optoe | Optical Transceiver Open EEPROM driver |
+
+## 4. Overview
+
+The CMIS specification defines a 3D memory addressing scheme using Bank Select (byte 0x7E) and Page Select (byte 0x7F) registers. The current SONiC implementation only supports Bank 0, limiting access to lane-specific registers for modules with more than 8 lanes.
+
+### CMIS Memory Architecture
+
+| Memory Region | Address Range | Description |
+|---------------|---------------|-------------|
+| Lower Memory  | 0x00 - 0x7F   | Fixed, always accessible (128 bytes) |
+| Upper Memory  | 0x80 - 0xFF   | Paged (128 bytes). Only certain pages support banking (see below) |
+
+**Note**: Not all upper memory pages support banking. According to the CMIS specification, only the following page ranges can be banked:
+- Pages 0x10 to 0x1F (lane-specific registers)
+- Pages 0x20 to 0x2F (VDM and lane-specific data)
+- Page 0x9F
+- Pages 0xA0 to 0xAF
+
+Other upper memory pages (e.g., 0x00-0x0F, 0x30-0x9E) are not bankable and always refer to Bank 0.
+
+### Control Registers
+
+| Register | Address | Purpose |
+|----------|---------|---------|
+| Bank Select | 0x7E (byte 126) | Selects active bank |
+| Page Select | 0x7F (byte 127) | Selects active page within bank |
+
+### Banking Use Cases
+
+- **Lane-specific registers**: Pages 10h-1Fh, 20h-2Fh and 0x30-0x3F
+  - Bank 0: Lanes 1-8
+  - Bank 1: Lanes 9-16
+  - Bank 2: Lanes 17-24, etc.
+- **VDM (Vendor Diagnostic Monitoring)**: Pages 20h-2Fh
+
+## 5. Requirements
+
+### Functional Requirements
+
+1. Support reading/writing to banks 0-7 on CMIS devices (per optoe driver limit)
+2. Maintain backward compatibility with existing Bank 0 operations
+3. Extend CLI utilities to accept bank parameter
+4. Update Python xcvr library to support bank-aware field access
+5. Auto-detect banking support by reading CMIS Banks Supported field (Page 01h, Byte 155)
+6. Use 1-byte I2C writes for transceivers without banking support; 2-byte writes only when banking is advertised
+
+### Non-Functional Requirements
+
+1. Minimize I2C overhead when switching banks
+2. No performance degradation for Bank 0 operations
+3. Support for all CMIS-compliant transceivers (QSFP-DD, OSFP, SFP-DD)
+
+## 6. Architecture Design
+
+This feature modifies the following components in the SONiC architecture:
+
+```
+┌─────────────────────────────────────────────────────────────────┐
+│                     User Space                                  │
+├─────────────────────────────────────────────────────────────────┤
+│  sfputil CLI  ──►  sonic_xcvr library  ──►  XcvrEeprom          │
+│      │                    │                      │              │
+│      │              CmisMemMap.getaddr()         │              │
+│      ▼                    ▼                      ▼              │
+│  bank parameter    bank-aware offset      read/write with       │
+│  in commands       calculation            linear offset         │
+├─────────────────────────────────────────────────────────────────┤
+│                     Kernel Space                                │
+├─────────────────────────────────────────────────────────────────┤
+│  optoe.c driver                                                 │
+│      │                                                          │
+│      ├── optoe_translate_offset() ── calculate bank from offset │
+│      │                                                          │
+│      └── optoe_eeprom_update_client() ── write bank select reg  │
+├─────────────────────────────────────────────────────────────────┤
+│                     Hardware                                    │
+├─────────────────────────────────────────────────────────────────┤
+│  I2C Bus ──► Transceiver EEPROM (Bank Select 0x7E, Page 0x7F)   │
+└─────────────────────────────────────────────────────────────────┘
+```
+
+No changes to the overall SONiC architecture are required. This feature extends existing transceiver management components.
+
+## 7. High-Level Design
+
+### 7.1 Repositories Changed
+
+| Repository | Files Modified |
+|------------|----------------|
+| sonic-linux-kernel | `drivers/misc/eeprom/optoe.c` |
+| sonic-platform-common | `sonic_xcvr/mem_maps/public/cmis.py`, `sonic_xcvr/fields/xcvr_field.py` |
+| sonic-utilities | `sfputil/main.py` |
+
+### 7.2 Kernel Driver Changes (optoe.c)
+
+**File**: `drivers/misc/eeprom/optoe.c`
+
+**Reference Implementation**: [sonic-net/sonic-linux-kernel PR #473](https://github.com/sonic-net/sonic-linux-kernel/pull/473)
+
+**Key Changes**:
+
+#### 7.2.1 New Constants
+
+```c
+#define OPTOE_BANK_SELECT_REG         0x7E
+#define OPTOE_DEFAULT_BANK_SIZE       0    /* disabled by default */
+#define OPTOE_MAX_SUPPORTED_BANK_SIZE 8
+#define OPTOE_NON_BANKED_PAGE_SIZE    16   /* pages 00h-0Fh are not banked */
+#define OPTOE_BANKED_PAGE_SIZE        240  /* pages 10h-FFh are banked */
+```
+
+#### 7.2.2 New sysfs Interface: `bank_size`
+
+A new sysfs entry `bank_size` is added to enable and configure bank support:
+
+- **Default value**: 0 (bank support disabled)
+- **Valid range**: 0-8
+- **Only applicable to**: `optoe3` device class (CMIS devices)
+
+```c
+static ssize_t show_bank_size(struct device *dev,
+            struct device_attribute *dattr, char *buf)
+{
+    /* Returns current bank_size value */
+}
+
+static ssize_t set_bank_size(struct device *dev,
+            struct device_attribute *attr,
+            const char *buf, size_t count)
+{
+    /* Only supported for CMIS devices (optoe3) */
+    if (optoe->dev_class != CMIS_ADDR)
+        return -EINVAL;
+
+    /* When enabling bank (bank_size > 0), automatically set write_max to 2
+     * to comply with CMIS 5.3 section 8.2.15 which mandates both bank and
+     * page values be updated in a single WRITE operation */
+    if (optoe->bank_size > 0 && optoe->write_max == 1)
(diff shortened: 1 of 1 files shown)


doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,922 @@
...
+Ethernet0  Ethernet8  Ethernet16  Ethernet24
+(lanes 1-8) (lanes 9-16) (lanes 17-24) (lanes 25-32)
+```
+
+#### 7.8.2 Recommended Approach: Thread-Safe SfpView Pattern
+
+The recommended design uses long-lived `SfpView` wrappers to provide thread-safe access to banked transceivers. Each SfpView is created once at chassis init and cached per interface. Each view has its own xcvr_api instance with bank context set.
+
+```
+Physical Module (32 lanes, index=1)
```
prgeor: @bobby-nexthop This is fine for a physical pluggable transceiver. For CPO systems that have separate ELSP and OE memory map support, can you show how the banking will be handled where ELSP and OE may support different number of banks? I am asking in this HLD scope:-
https://github.com/sonic-net/SONiC/pull/2152/

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,922 @@
...
+
+```python
+class CmisApi:
+    NUM_CHANNELS = 8
+
+    def __init__(self, xcvr_eeprom):
+        self.xcvr_eeprom = xcvr_eeprom
+        self._bank = 0  # Default bank
+
+    def set_bank(self, bank):
```
prgeor: @bobby-nexthop In the case of CPO , which bank will it set for? ELSFP or OE?

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,922 @@
...
+**Reference Implementation**: [sonic-net/sonic-linux-kernel PR #473](https://github.com/sonic-net/sonic-linux-kernel/pull/473)
+
+**Key Changes**:
+
+#### 7.2.1 New Constants
+
+```c
+#define OPTOE_BANK_SELECT_REG         0x7E
+#define OPTOE_DEFAULT_BANK_SIZE       0    /* disabled by default */
+#define OPTOE_MAX_SUPPORTED_BANK_SIZE 8
```
micas-net, line 202: why limit to 8, does it support more than 8? I found below spec in CMIS 5.3:
Bank Index(0-255), a Page Index(00h-FFh), and a Byte Address(0-255)

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,922 @@
...
+
+To ensure backwards compatibility with existing transceivers that may not handle 2-byte I2C writes correctly, the implementation reads the **Banks Supported** field from the transceiver before enabling banking.
+
+#### 7.7.1 CMIS Banks Supported Field
+
+Per CMIS 5.x specification, the Banks Supported field is located at:
+
+| Location | Description |
+|----------|-------------|
+| Page 01h, Byte 155, Bits 2-0 | Number of banks for lane-specific registers |
```
micas-net: 155 is for module and lane controls, not bank.

doc/layer1/cmis_banking_support.md
bobby-nexthop: I have added a section for the CPO seperate mode case

doc/layer1/cmis_banking_support.md
bobby-nexthop: Same as above, I added the case for CPO seperate mode where the device is not treated as a single CMIS compliant module

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,959 @@
...
+```
+
+### 7.4 Field Definition Changes (xcvr_field.py)
+
+**File**: `src/sonic-platform-common/sonic_platform_base/sonic_xcvr/fields/xcvr_field.py`
+
+```python
+class RegField:
+    def __init__(self, name, offset, *args, bank=0, **kwargs):
+        self.bank = bank
```
KroosMicas: This means that each CmisMemMap​ needs to register the mappings of all banks. However, since each bank has the same functionality, this will lead to a large amount of redundant code.

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,959 @@
...
+```
+
+### 7.5 CLI Utility Changes (sfputil)
+
+**File**: `src/sonic-utilities/sfputil/main.py`
+
+```python
+# Add bank parameter to page dump commands
+@click.option('--bank', default=0, type=int, help='Bank number (default: 0)')
+def read_eeprom(port, page, bank):
```
KroosMicas: Do all the read_eeprom​ and write_eeprom​ methods inherited from SfpOptoeBase​ previously need to add new parameters?

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,959 @@
...
+
+To ensure backwards compatibility with existing transceivers that may not handle 2-byte I2C writes correctly, the implementation reads the **Banks Supported** field from the transceiver before enabling banking.
+
+#### 7.7.1 CMIS Banks Supported Field
+
+Per CMIS 5.x specification, the Banks Supported field is located at:
+
+| Location | Description |
+|----------|-------------|
+| Page 01h, Byte 155, Bits 2-0 | Number of banks for lane-specific registers |
```
KroosMicas: Page 01h, Byte 155, Bits 2-0 . Is this the correct address?

doc/layer1/cmis_banking_support.md
bobby-nexthop: Yes this was left over from an initial version. Thank you for pointing it out I've removed it. RegField doesn't need a bank attribute. The bank context lives in the API layer and is passed through at read time.

doc/layer1/cmis_banking_support.md
bobby-nexthop: They shouldn't need it, these methods take the offset as a parameter, the bank is already encoded in the offset by the time it reaches these methods.

doc/layer1/cmis_banking_support.md
bobby-nexthop: Screenshot 2026-01-29 at 9 10 11 AM
Thank you, updating to byte 142

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,948 @@
...
+
+##### Layer 4: XcvrEeprom (`xcvr_eeprom.py`)
+
+Add bank parameter to read/write methods:
+
+```python
+def read(self, field_name, bank=0):
+    """Read a field from the transceiver EEPROM."""
+    field = self.mem_map.get_field(field_name)
+    offset = field.get_offset(bank=bank)
```
abhi-nexthop: how will XcvrField get_offset be implemented with the bank field? Maybe add that redefined method in the HLD.

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,948 @@
...
+
+```python
+# Constants matching optoe driver
+OPTOE_PAGE_SIZE = 128
+OPTOE_NON_BANKED_PAGE_SIZE = 16   # pages 00h-0Fh
+OPTOE_BANKED_PAGE_SIZE = 240      # pages 10h-FFh
+
+# Extended implementation
+def getaddr(self, page, offset, page_size=128, bank=0):
+    """
```
abhi-nexthop: getaddr is also used by CMISMemMap to register fields. In this case, should memory mapped fields always use getaddr with bank=0? Given that get_offset takes a bank param, it makes sense. However, I think we should explicitly specify that the memory classes inside cmis.py and its derived classes should not register other banks as it is calculated elsewhere in the offset calculation

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,985 @@
...
+),
+```
+
+#### 7.7.5 Initialization Flow
+
+```
+┌─────────────────────────────────────────────────────────────────┐
+│                  Transceiver Initialization                     │
+├─────────────────────────────────────────────────────────────────┤
+│  1. Read Page 01h, Byte 155 (Banks Supported)                   │
```
DavidHuang666: Byte 142？

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,928 @@
...
+        loff_t *offset, struct i2c_client **client, uint8_t *bank)
+{
+    unsigned int page = 0;
+    /* ... existing offset translation ... */
+
+    page = (*offset >> 7) - 1;
+    *offset = OPTOE_PAGE_SIZE + (*offset & 0x7f);
+
+    /* Calculate bank from page number */
+    *bank = page < OPTOE_PAGE_SIZE ? 0 :
```
KroosMicas, line 298: I think this should be：page < OPTOE_NON_BANKED_PAGE_SIZE ? 0 : (page - OPTOE_NON_BANKED_PAGE_SIZE) / OPTOE_BANKED_PAGE_SIZE;

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,928 @@
...
+            return page * page_size + offset
+        else:
+            # Banks 1+: only pages 10h-FFh (0x10+) are banked
+            # Pages < 0x10 are never banked, even for bank > 0
+            if page < 0x10:
+                # Non-banked pages (00h-0Fh): same as bank 0
+                return page * page_size + offset
+            else:
+                # Banked pages (10h-FFh): offset by bank * OPTOE_BANKED_PAGE_SIZE pages
+                return ((self._bank * OPTOE_BANKED_PAGE_SIZE) + page) * page_size + offset
```
KroosMicas: We have a simpler algorithm that you can refer to:

Image

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,928 @@
...
+    ▼         ▼         ▼         ▼
+Ethernet0  Ethernet8  Ethernet16  Ethernet24
+(lanes 1-8) (lanes 9-16) (lanes 17-24) (lanes 25-32)
+```
+
+#### 7.8.2 Recommended Approach: SFP Object Per Logical Interface with Internal Bank
+
+The recommended design creates one SFP object per logical interface (front panel port). Each SFP object stores its bank internally and uses it automatically when creating the xcvr_api. Multiple SFP objects may share the same physical transceiver (same I2C path) but have different bank values.
+
+```
```
Junchao-Mellanox, line 628: In case of an OE contains 4 banks, we will need 4 SFP objects. I suppose this design would cause following issue:
1. 4 times I2C access than actual needs
2. Conflict in OE FW upgrading via CDB based on current implementation. 

How do you want to fix above issues?

ishidawataru: Do we also need to think how to handle AppSel?

aditya-nexthop: > In case of an OE contains 4 banks, we will need 4 SFP objects. I suppose this design would cause following issue:

Yes, for an OE with 4 banks, there would be 4 SFP objects.
We wanted to maintain the abstraction of having one SFP object per "front panel interface" which in this case is currently no wider than 1 bank.

> 4 times I2C access than actual needs

We will likely have a new/different `DomInfoUpdateTask` for CPO devices, which can intelligently, for example, read non-banked pages only from `Sfp` objects with `bank=0`, associate them with the ID of the `OE` and replicate the information to other `Sfp` objects with the same `OE` with non-zero banks. This is a good question but beyond the scope of this specific HLD which only lays the groundwork for banking support. https://github.com/sonic-net/SONiC/pull/2211 discusses what an SFP object for CPO can look like and we can use fields from it to maintain efficient I2C access.

> Conflict in OE FW upgrading via CDB based on current implementation.

This is same as what we have with addressing multiple logical ports associated with a physical port.
An upgrade on one port will upgrade the underlying physical resource. Similarly an upgrade on any of the `Sfp`s will, presently, only use CDB instance associated with bank 0 to perform upgrades. The use of CDB instance bank 1 is beyond the scope of this HLD.

> Do we also need to think how to handle AppSel?

I think that ApSels do not cross banks so we should be okay. Please let us know if there is anything specific we misunderstood or should clarify.

Junchao-Mellanox: Sounds good to me. Could you please add this to your HLD? I think it is important to record the details here.

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,928 @@
...
+
+### 7.9 CLI Utility Changes (sfputil)
+
+**File**: `src/sonic-utilities/sfputil/main.py`
+
+```python
+# Add bank parameter to page dump commands
+@click.option('--bank', default=0, type=int, help='Bank number (default: 0)')
+def read_eeprom(port, page, bank):
+    # Include bank in offset calculation
```
benle7: The implementation uses chassis.get_sfp, and the sfp object should be bank-aware.
Why do we need a new bank param here?

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,928 @@
...
+- Non-banked pages (0x00-0x0F) have the same linear offset regardless of bank
+
+**Non-Banked Fields** (same offset for all banks):
+
+| API Call / Field | Page | Byte | Linear Offset |
+|------------------|------|------|---------------|
+| `get_module_temperature()` | 0x00 | 14 | 14 |
+| `get_voltage()` | 0x00 | 16 | 16 |
+
+### 7.5 Memory Map Abstraction Changes (cmis.py)
```
tshalvi, line 430: Why is a memory-map change needed here if bank_id is already part of the SFP object and each CMIS banked page keeps the same layout (According to [port_mapping_for_cpo.md](https://github.com/sonic-net/SONiC/pull/2211))?
Since the page layout itself does not change across banks, it seems the existing memory-map definitions should still be sufficient, as long as field access uses the correct bank_id from the SFP object. Could you clarify why the memory map itself needs to become bank-aware, instead of keeping banking only at the SFP/object access layer?

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,928 @@
...
+OPTOE_PAGE_SIZE = 128
+OPTOE_NON_BANKED_PAGE_SIZE = 16   # pages 00h-0Fh
+OPTOE_BANKED_PAGE_SIZE = 240      # pages 10h-FFh
+
+class CmisFlatMemMap(XcvrMemMap):
+    def __init__(self, codes, bank=0):
+        super().__init__(codes)
+        self._bank = bank  # Immutable after construction
+
+    def getaddr(self, page, offset, page_size=128):
```
tshalvi, line 479: The proposed getaddr() change appears to embed OPTOE-specific linear offset logic directly into the generic memory-map layer. Since this HLD is changing the definition/behavior of getaddr() for all platforms, could you clarify why this should be done at the common memory-map layer rather than in the OPTOE-specific path only? As written, it seems to affect platforms that do not use OPTOE as well.

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,928 @@
...
+            return [tx_power['TxPower%dField' % i] for i in range(1, self.NUM_CHANNELS + 1)]
+        return ["N/A" for _ in range(self.NUM_CHANNELS)]
+
+    # Module-level methods work the same - bank 0 fields are unaffected by bank setting
+    def get_module_temperature(self):
+        """Temperature is on non-banked page (page 0) - bank setting doesn't affect it."""
+        return self.xcvr_eeprom.read(consts.MODULE_TEMPERATURE_FIELD)
+```
+
+#### 7.8.4 Platform Configuration (platform.json)
```
tshalvi: According to this section, the bank is defined per interface in platform.json. However, in section 7.1 of [port_mapping_for_cpo.md](https://github.com/sonic-net/SONiC/pull/2211), platform.json explicitly remains unchanged, and the bank association is described under optical_devices.json instead. Could you clarify which file is expected to be the source of truth for bank configuration? Right now the two HLDs appear inconsistent on this point.

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,928 @@
...
+    # ...
+```
+
+### 7.9 CLI Utility Changes (sfputil)
+
+**File**: `src/sonic-utilities/sfputil/main.py`
+
+```python
+# Add bank parameter to page dump commands
+@click.option('--bank', default=0, type=int, help='Bank number (default: 0)')
```
tshalvi: What is the reason for introducing a separate --bank CLI parameter for sfputil read-eeprom / write-eeprom if the selected SFP object already carries the bank information?
If banking is part of the object/platform mapping, then exposing --bank in the CLI seems to duplicate that information and may also allow accesses that bypass the intended platform mapping. Could you clarify the use case this parameter is meant to address?

doc/layer1/cmis_banking_support.md
aditya-nexthop: We should use the the scheme in https://github.com/sonic-net/SONiC/pull/2211.

Line 723 already states this:

> The bank is explicitly defined per interface in `platform.json`. This initial banking support is for single i2c device modules.For more complicated use cases like CPO, a follow up HLD will explain the platform mapping.

doc/layer1/cmis_banking_support.md
bobby-nexthop: Sfputil is to be handled by another HLD. Removed the section entirely

doc/layer1/cmis_banking_support.md
bobby-nexthop: The mem_map has always produced optoe-linear offsets (page * 128 + offset). Banking adds one term, not a new coupling. The mem_map has to carry bank because optoe (or any driver alternative) doesn't expose (bank, page, offset), it exposes a single linear file. Somthing has to turn (bank, page, byte) into a linear address; this HLD does it once at mem_map construction time so RegField.offset is a real offset and the hot path stays a single sysfs call.

doc/layer1/cmis_banking_support.md
bobby-nexthop: That's correct, in the workgroup meeting it was discussed that OPTOE and non-OPTOE drivers like Nvidia should follow the same linear address scheme. The driver already does the bank/page work, it writes BankSelect/PageSelect over I2C. SONiC just computes which byte it wants, as a linear offset into optoe's/non-optoe's eeprom sysfs file. That's the same split optoe has used for page selection always; banking extends it by one axis. 

The alternative I think you're referring to of ("driver takes (bank, page, offset) directly") either breaks atomicity (separate sysfs writables), breaks the byte-addressed-file convention that hexdump/dd/ethtool -m rely on.

doc/layer1/cmis_banking_support.md
bobby-nexthop: This is outdated before any of these were well scoped. This will be addressed in a different HLD and removed from this.

abelamit (approved):

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,899 @@
...
+**Reference Implementation**: [sonic-net/sonic-linux-kernel PR #473](https://github.com/sonic-net/sonic-linux-kernel/pull/473)
+
+**Key Changes**:
+
+#### 7.3.1 New Constants
+
+```c
+#define OPTOE_BANK_SELECT_REG         0x7E
+#define OPTOE_DEFAULT_BANK_SIZE       0    /* disabled by default */
+#define OPTOE_MAX_SUPPORTED_BANK_SIZE 8
```
prgeor, line 202: @bobby-nexthop As per CMIS spec v5.3 there are support for MAX 4 banks. Am I missing something?

Image

doc/layer1/cmis_banking_support.md
```diff
@@ -0,0 +1,899 @@
...
+| Page 01h, Byte 142, Bits 1-0 | Number of banks for lane-specific registers |
+
+**Encoding**:
+
+| Value | Banks Supported |
+|-------|-----------------|
+| 000b  | 1 bank (Bank 0 only) |
+| 001b  | 2 banks |
+| 010b  | 4 banks |
+| 011b  | 8 banks |
```
prgeor, line 514: @bobby-nexthop Which spec mentions 8 banks? `11` is Reserved as per CMIS spec v5.3

doc/layer1/cmis_banking_support.md
aditya-nexthop: Some implementations might use 0b11 to mean 8 banks.
Thoughts, @KroosMicas ?

micas-net: Yes, bailly is 0b11 @aditya-nexthop

doc/layer1/cmis_banking_support.md
micas-net: yes, it is reserved, we can treat it as 8 banks so that  some vendor can work on cmis 5.2 like on 5.3

doc/layer1/cmis_banking_support.md
KroosMicas: > Some implementations might use 0b11 to mean 8 banks. Thoughts, @KroosMicas ?

Agree with this approach. Bailly is also 0b11, so they can be shared. If CMIS has a standard representation for 8 banks in the future, we can make modifications then.

micas-net:
The HLD looks fine to me, and I am doing local code integrating (473), once the functionality be verified, will merge this HLD too!

prgeor (approved):

PR #1631, opened 2024-03-11
PoE design proposal HLD document

plvisiondevs:
Add PoE design proposal HLD document with the following content:

- Requirements
- Component interactions
- DB table definition
- CLI

In order to isolate PoE functionality and complexity from existing SONiC design, the following main changes are made:

- Add separate container for PoE control plane routine operations.
- Introduce separate SYNCD container to support external PoE Abstraction Interface.
- Add new PoE tables into database container.

Related PRs:
- https://github.com/sonic-net/sonic-swss-common/pull/894
- https://github.com/sonic-net/sonic-sairedis/pull/1404
- https://github.com/sonic-net/sonic-swss/pull/3238
- https://github.com/sonic-net/sonic-utilities/pull/4819
- https://github.com/sonic-net/sonic-buildimage/pull/19636
- https://github.com/sonic-net/sonic-buildimage/pull/21005

SAI: https://github.com/opencomputeproject/SAI/pull/1977 (merged)

PR #2200, opened 2026-02-02
PoE Power via MDI HLD proposal

saravanan-i:
Added PoE power via MDI design proposal HLD document with the following content:

Requirements
Component interactions
DB table definition
CLI

This PR is an enhancment planned on top of this HLD
https://github.com/sonic-net/SONiC/pull/1631

### DIFF
--- a/doc/poe_power_via_mdi/poe_hld.md
+++ b/doc/poe_power_via_mdi/poe_hld.md
@@ -109,7 +109,6 @@ LLDP syncd is required to be modified as following:
 
 ### High-Level Design 
 
-To be added
 
 ###### LLDP flows
 


doc/poe_power_via_mdi/poe_hld.md
```diff
@@ -0,0 +1,347 @@
...
+| MDI                      | Media Dependent Interface |
+| LLDP                     | Link Level Disconvery Protocol |
+| PD                       | Powered Devices |
+| TLV                      | Type Length Value |
+| PSE                      | Power Supplying Equipment |
+
+### Overview 
+
+SoNiC switches with PoE capability can be augmented to support Power via MDI functionality. Once the port connected to an PD reaches delivering power stage, PD can adjust it's power limit (called as Dynamic Power Limit) by exchanging LLDP packets to the switch. 
+It prevents overloading the PSE’s power budget, which can lead to port shutdowns thus ensuring stable and efficient operation of all connected devices.
```
madhupalu, line 54: Who is responsible for restoring the power budget while system under Warm reboot? >> update the HLD

doc/poe_power_via_mdi/poe_hld.md
```diff
@@ -0,0 +1,347 @@
...
+- config poe interface power-via-mdi \<ifname\> {enable | disable}
+
+Examples:
+```
+$ config poe interface power-via-mdi Ethernet0 enable
+```
+
+**TODO**: update CLI reference https://github.com/sonic-net/sonic-utilities/blob/master/doc/Command-Reference.md
+
+#### YANG model Enhancements
```
madhupalu, line 175: Do you support OpenConfig Yang model ?

doc/poe_power_via_mdi/poe_hld.md
```diff
@@ -0,0 +1,347 @@
...
+### Memory Consumption
+This sub-section covers the memory consumption analysis for the new feature: no memory consumption is expected when the feature is disabled via compilation and no growing memory consumption while feature is disabled by configuration. 
+
+### Restrictions/Limitations  
+
+This HLD assumes an Endpoint PSE (Switch). If a Midspan PSE is used between the SONiC switch and the PD, LLDP-based power negotiation will likely fail as Midspans typically do not terminate or proxy LLDP TLVs.
+
+### Testing Requirements/Design  
+
+#### Unit Test cases
```
madhupalu, line 336: What is scale of PoE/PD ports support ?

doc/poe_power_via_mdi/poe_hld.md
```diff
@@ -0,0 +1,347 @@
...
+PDs(AF/AT/BT) power via mdi requirements sent via (LLDP TLVs) can be mocked to validate the functionality
+
+#### System Test cases
+
+System testing can be done on systems with PoE power via mdi capability and the PD (with Power via MDI) support
+
+### Open/Action items
+
+- SONiC Yang models needs to be added
+- SAI PoE definition needs to be updated in the OpenComputingProject to support Dynamic Power Limit
```
madhupalu, line 347: Do you have SAI dependency? Since it depends on the SAI, How does the design handle if the hardware won't support SAI? >> DO you have SAI capability check and handle the errors gracefully?

doc/poe_power_via_mdi/poe_hld.md
```diff
@@ -0,0 +1,347 @@
...
+	
+	pse-max-power						= STRING	;Maximum power available in PSE for this port, applicable when standard is "at"/"bt"
+```
+
+#### STATE DB Enhancements  
+
+New table LLDP_MDI_POWER introduced in State DB to enable to PoE manager to communicate with the PD. Once the dynamic power limit is set in hardware, then the poe manager will write the same in STATE_DB. This DB will be subscribed by 
+lldp manager to fill the outgoing power via mdi TLVs in LLDP protocol packets.
+
+#### LLDP_MDI_POWER
```
madhupalu, line 256: How to debug the Power negotiation issues? DO you have CLI exposed ?

doc/poe_power_via_mdi/poe_hld.md
```diff
@@ -0,0 +1,347 @@
...
+### High-Level Design 
+
+
+###### LLDP flows
+
+1) lldpmgrd updates the APPL_DB with the power via MDI LLDP packet received from PD.
+2) poemgrd subscribed to related tables in APPL_DB and upon receiving the notification from APPL_DB and then processes the dynamic power limit request by propagating the same in PoE hardware.
+3) After setting the power limit in PoE hardware, poemgrd updates the outgoing TLV information in STATE_DB
+4) lldp_syncd subscribed to related table in STATE_DB and upon receiving the update notification, it appends the outgoing LLDP packet (to the PD) with  the Power via MDI TLV.
+
```
binnyjeshan, line 119: hi @saravanan-i 

One scenario that may need explicit coverage is dynamic power promotion when the PSE is already near its PoE budget limit. If a PD requests a higher power level that cannot be immediately honored, the PSE should continue advertising the currently allocated power in outgoing LLDP TLVs, effectively indicating that the promotion was not granted.

When power later becomes available (for example, due to demotion or disconnection of another PD), the PSE should be able to re-evaluate and grant the earlier request. It would be good to clarify in the HLD whether such pending promotion handling and re-evaluation logic is already considered, or if the design needs to be updated to explicitly support this behavior.

doc/poe_power_via_mdi/poe_hld.md
```diff
@@ -0,0 +1,347 @@
...
+#### PoE interaction
+
+This section describes the detail interaction of PoE componnents, DB and PoE hardware (controller) in SONiC subsytem.
+
+![poe-interaction](poe_lldp_diagram_data_flow.png)
+
+#### SWSS changes
+
+To support setting dynamic power limit in ASIC, SWSS container needs to be enhanced to set the attribute in SAI.
```
binnyjeshan, line 84: Hi @saravanan-i

Could this parameter be called dynamic power **request** instead, since its a request for '**exact power**' needs from the PD and not a limit requested by the PD? 
If my understanding isn't aligned with the goal of this attribute, please share some details that'd help.

doc/poe_power_via_mdi/poe_hld.md
```diff
@@ -0,0 +1,347 @@
...
+
+#### CLI
+
+##### Show commands
+Existing show command is updated to display the dynamic power limit set.
+
+- show poe interface \<ifname\>
+```
+
+Port         Status      En/Dis  Priority Protocol      Class  PWR Consump  PWR Limit   Voltage   Current DYN_PWR Limit
```
binnyjeshan, line 156: @saravanan-i , same as my comment above, it'd be good to rename this from "Limit" to "Request" to indicate that its a request for more/ less power from PD.

doc/poe_power_via_mdi/poe_hld.md
```diff
@@ -0,0 +1,347 @@
...
+
+The PoE Manager uses a new SAI PoE library that implements the PoE Abstraction Interface.
+
+#### Device Capabilities
+
+#### Port Capabilities
+
+| Capabilities       		 | SAI attributes                                |
+| -------------------        | --------------------------------------------- |
+| Dynamic Power Limit        | * SAI_POE_PORT_ATTR_DYNAMIC_POWER_LIMIT         |
```
DanielaMurin, line 139: During the negotiation process between the PD and the PSE, the PD advertises a PoE power request. Following a successful negotiation, the PSE allocates power from the available budget. Is this negotiated and allocated value what SAI_POE_PORT_ATTR_DYNAMIC_POWER_LIMIT is intended to represent?
If so, I would suggest changing the name to: SAI_POE_PORT_ATTR_DYNAMIC_POWER_REQUEST

binnyjeshan: @DanielaMurin , agree with your views, My comment earlier weeks ago was around similar lines. Its not a limit. Its the desired power/requested power per my understanding as well.

PR #2207, opened 2026-02-06
HLD: ELSFP Memory map and paging support

abhi-nexthop:
This HLD refactors CMIS Memory Map into page classes and defines an ELSFP Memory Map along with its respective pages. The refactor also avoids any API level changes for any classes that may consume the CMIS map

Implementation PRs:
[Refactor CMIS memory map into pages](https://github.com/sonic-net/sonic-platform-common/pull/667)
[Create ELSFP memory map classes](https://github.com/sonic-net/sonic-platform-common/pull/678)

### DIFF
--- a/doc/layer1/elsfp-and-paging-support.md
+++ b/doc/layer1/elsfp-and-paging-support.md
@@ -182,7 +182,7 @@ Existing CMIS registers are grouped into their respective pages. For example, re
               .
               .
             ]
-  class CmisCdbMessagePage(CMISPage): #0x9F
+  class CmisCdbMessagePage(CmisPage): #0x9F
     def __init__(codes, page=0x9F, bank=0):
         super(CmisCdbMessagePage, self).__init__(codes, page, bank)
         self.fields[consts.TRANS_CDB_FIELD] = [
@@ -212,12 +212,12 @@ class CmisMemMap(XcvrMemMap):
     def __init__(self, codes):
         super(CmisMemMap, self).__init__(codes)
         #CMIS pages
-        self.administrative_upper_page = CMISAdministrativeUpperPage(codes) #0x00U
-        self.advertising_page = CMISAdvertisingPage(codes) #0x01
+        self.administrative_upper_page = CmisAdministrativeUpperPage(codes) #0x00U
+        self.advertising_page = CmisAdvertisingPage(codes) #0x01
         .
         .
         .
-        self.cdb_message_page = CMISCDBMessagePage(codes) #0x9F
+        self.cdb_message_page = CmisCdbMessagePage(codes) #0x9F
         self.TRANS_CDB= RegGroupField(consts.TRANS_CDB_FIELD,
             *self.get_field_from_pages(consts.TRANS_CDB_FIELD, self.advertising_page, self.cdb_message_page)
         )
@@ -326,8 +326,8 @@ class ElsfpMemMap(CmisFlatMemMap):
     def __init__(self, codes, bank):
         super(ElsfpMemMap, self).__init__(codes, bank)
         #CMIS pages
-        self.administrative_upper_page = CMISAdministrativeUpperPage(codes, bank=bank)
-        self.advertising_page = CMISAdvertisingPage(codes, bank=bank)
+        self.administrative_upper_page = CmisAdministrativeUpperPage(codes, bank=bank)
+        self.advertising_page = CmisAdvertisingPage(codes, bank=bank)
         .
         .
         .
--- a/doc/layer1/elsfp-and-paging-support.md
+++ b/doc/layer1/elsfp-and-paging-support.md
@@ -1,0 +1,451 @@
+# Paging support in CMIS memory maps and ELSFP support #
+
+## Table of Content
+
+- [1. Revision](#1-revision)
+- [2. Scope](#2-scope)
+- [3. Definitions/Abbreviations](#3-definitionsabbreviations)
+- [4. Overview](#4-overview)
+- [5. Requirements](#5-requirements)
+  - [Functional Requirements](#functional-requirements)
+  - [Non-Functional Requirements](#non-functional-requirements)
+- [6. Architecture Design](#6-architecture-design)
+- [7. High-Level Design](#7-high-level-design)
+  - [7.1 Repositories Changed](#71-repositories-changed)
+  - [7.2 Memory Map Abstraction Changes (cmis.py)](#72-memory-map-abstraction-changes-cmispy)
+    - [7.2.1 The CmisPage class](#721-the-cmispage-class)
+    - [7.2.2 New CMIS pages](#722-new-cmis-pages)
+    - [7.2.3 The CmisMemMap class](#723-the-cmismemmap-class)
+  - [7.3 ELSFP Memory mapping](#73-elsfp-memory-mapping)
+    - [7.3.1 ELSFP constants](#731-elsfp-constants)
+    - [7.3.2 ElsfpPage classes](#732-ElsfpPage-classes)
+    - [7.3.3 ElsfpMemMap class](#733-ElsfpMemMap-class)
+    - [7.3.4 Custom Page remapping](#734-custom-page-remapping)
+- [8. SAI API](#8-sai-api)
+- [9. Configuration and Management](#9-configuration-and-management)
+  - [9.1 Manifest](#91-manifest)
+  - [9.2 CLI/YANG Model Enhancements](#92-cliyang-model-enhancements)
+  - [9.3 Config DB Enhancements](#93-config-db-enhancements)
+- [10. Warmboot and Fastboot Design Impact](#10-warmboot-and-fastboot-design-impact)
+- [11. Memory Consumption](#11-memory-consumption)
+- [12. Restrictions/Limitations](#12-restrictionslimitations)
+- [13. Testing Requirements/Design](#13-testing-requirementsdesign)
+  - [13.1 Unit Test Cases](#131-unit-test-cases)
+  - [13.2 System Test Cases](#132-system-test-cases)
+- [References](#references)
+
+### 1. Revision  
+
+| Rev | Date       | Author | Change Description |
+|-----|------------|--------|--------------------|
+| 0.1 | 2025-02-03 | abhi-nexthop      | Initial version    |
+
+
+### 2. Scope  
+
+This high-level design document describes the design for adding paging support in CMIS memory maps and ELSFP support in SONiC.
+
+### 3. Definitions/Abbreviations 
+
+| Term | Definition |
+|------|------------|
+| CMIS | Common Management Interface Specification |
+| I2C | Inter-Integrated Circuit |
+| ELSFP | External Laser Small Form Factor Pluggable |
+
+### 4. Overview 
+
+The CMIS spec provides the memory layout for certain i2c transceivers. This layout consists of a fixed lower memory, and a paged upper memory. The upper memory has certain pages that are banked, providing a 3D memory layout for the device.
+
+While SONiC does not currently provide banking support, it has been proposed in the [Banking HLD](https://github.com/sonic-net/SONiC/pull/2183) PR and this document depends on it.
+
+CMIS layout (Taken from the Banking HLD PR):
+
+| Memory Region | Address Range | Description |
+|---------------|---------------|-------------|
+| Lower Memory  | 0x00 - 0x7F   | Fixed, always accessible (128 bytes) |
+| Upper Memory  | 0x80 - 0xFF   | Paged (128 bytes). Only certain pages support banking |
+
+The ELSFP-CMIS spec is an extension of CMIS that provides memory specification for External Laser Sources. This is critical for emerging technologies such as CPO where optical engines and laser sources are used instead.
+
+ELSFP describes Upper memory CMIS pages it supports in its memory map. Existing CMIS pages used:
+
+| Page | Description |
+|------|-------------|
+| 0x00 | Administrative Information |
+| 0x01 | Advertising |
+| 0x02 | Threshold Information |
+| 0x03 | User NV RAM |
+| 0x9F | CDB Command/Response with local payload |
+| A0-AF | CDB EPL extended payload segments |
+
+
+While, SONiC registers these paged registers, it does not explicitly sort them in individual pages. As the CMIS spec grows and new specs that branch off CMIS are defined (such as ELSFP and C-CMIS) , more registers may be added and the memory map will continue to grow. Furthermore, currently, other memory classes such as the CCmisMemMap inherit the CMIS memory map and add their own registers. However, not all pages in the CMIS memory map are used by every device. For example, if an ELSFPMemoryMap object derives from the CMIS memory map, it will contain registers from pages that do not comply to the ELSFP spec. Therefore, there is a need to explicitly define pages and provide the ability to pick a subset of CMIS pages for a memory map.
+
+In addition, the ELSFP spec describes new pages that are not currently implemented in SONiC and these will need to be defined
+
+| Page | Description |
+|------|-------------|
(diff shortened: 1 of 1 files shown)


doc/layer1/elsfp-and-paging-support.md
```diff
@@ -0,0 +1,464 @@
...
+For example, consider Device 1 that controls Device 2. Device 2's advertising page is mapped onto a vendor reserved page B0.
+
+This is how a vendor may implement this:
+
+```python
+
+class Device2AdvertisingPage(CmisAdvertisingPage):
+    def __init__(codes, page=0xB0, bank=0):
+        # Device 2's advertising page which is set to 0x01 by default is mapped onto page 0xB0
+        super(Device2AdvertisingPage, self).__init__(codes, page, bank)
```
KroosMicas: By "Device," do you mean different ELSFPs?

doc/layer1/elsfp-and-paging-support.md
```diff
@@ -0,0 +1,464 @@
...
+    def __init__(self, codes, bank):
+        super(Device1MemMap, self).__init__(codes, bank)
+        .
+        .
+        .
+        # Device 1's own page is mapped to the default CMIS page
+        self.advertising_page = CmisAdvertisingPage(codes)
+        # Device 2's advertising page is mapped to the vendor reserved page 0xB0
+        self.device2_advertising_page = Device2AdvertisingPage(codes)
+        .
```
KroosMicas: Is this targeting the scenario in joint mode​ where multiple ELSFPs​ are mapped to the same CMIS?

prgeor (commented):
@Junchao-Mellanox is there any impact to caching of eeprom ?

prgeor:
@abhi-nexthop please fix the DCO checker. The commit needs to be signed-off

doc/layer1/elsfp-and-paging-support.md
abhi-nexthop: Yes this is a possible use case

doc/layer1/elsfp-and-paging-support.md
abhi-nexthop: Could be but its left more generic in the HLD for the page refactor part

doc/layer1/elsfp-and-paging-support.md
```diff
@@ -0,0 +1,464 @@
...
+|------|-------------|
+| 0x00 | Administrative Information |
+| 0x01 | Advertising |
+| 0x02 | Threshold Information |
+| 0x03 | User NV RAM |
+| 0x9F | CDB Command/Response with local payload |
+| A0-AF | CDB EPL extended payload segments |
+
+
+While, SONiC registers these paged registers, it does not explicitly sort them in individual pages. As the CMIS spec grows and new specs that branch off CMIS are defined (such as ELSFP and C-CMIS) , more registers may be added and the memory map will continue to grow. Furthermore, currently, other memory classes such as the CCmisMemMap inherit the CMIS memory map and add their own registers. However, not all pages in the CMIS memory map are used by every device. For example, if an ELSFPMemoryMap object derives from the CMIS memory map, it will contain registers from pages that do not comply to the ELSFP spec. Therefore, there is a need to explicitly define pages and provide the ability to pick a subset of CMIS pages for a memory map.
```
Natgerbi, line 85: Could you please explain the benefit of splitting it into separate memory maps instead of simply extending CmisMemMap to support ELSFP using inheritance?

doc/layer1/elsfp-and-paging-support.md
```diff
@@ -0,0 +1,464 @@
...
+
+class Device2AdvertisingPage(CmisAdvertisingPage):
+    def __init__(codes, page=0xB0, bank=0):
+        # Device 2's advertising page which is set to 0x01 by default is mapped onto page 0xB0
+        super(Device2AdvertisingPage, self).__init__(codes, page, bank)
+        .
+        .
+        .
+  
+class Device1MemMap(CmisMemMap):
```
Natgerbi, line 355: Should this part be handled within each vendor’s platform implementation?
I believe we would like to keep the creation of the API object and the MemMap object consistent with the existing create_xcvr_api function, rather than requiring a vendor-specific override for each implementation.

doc/layer1/elsfp-and-paging-support.md
```diff
@@ -0,0 +1,464 @@
...
+| Page | Description |
+|------|-------------|
+| 0x1A | ELSFP Advertisements, Flags |
+| 0x1B | ELSFP Controls and Monitors |
+
+### 5. Requirements
+
+### Functional Requirements
+
+1. Support paging in CMIS memory maps
```
Natgerbi, line 98: Could you please clarify the motivation behind this change? Why not keep the existing support? SONiC already supports different page IDs, so I don’t see the need to redesign the entire memory layout.

doc/layer1/elsfp-and-paging-support.md
```diff
@@ -0,0 +1,464 @@
...
+2. Add ELSFP memory map and pages in SONiC
+
+### Non-Functional Requirements
+
+1. Existing maps should remain functionally identical and cause no API changes
+2. Easy remapping of pages to another page
```
Natgerbi, line 104: Could you please clarify why we shouldn’t add all the non-vendor-specific fields to the existing map and use only the relevant ones from the API object?

doc/layer1/elsfp-and-paging-support.md
bgallagher-nexthop: The ELSFP memory map is not a superset of the common CMIS memory map. Pages like 10h for lane/datapath control do not exist in the ELSFP memory map for instance, since an ELSFP is a resource module with no datapath. Extending the CmisMemMap via inheritance would end up defining a memory map for the ELSFP that contains these pages erroneously.

Screenshot 2026-04-09 at 12 35 33 PM

The rationale behind factoring pages out into their own classes is that it allows for easier composition of higher-level memory maps (ELSFP/CMIS/C-CMIS) that share pages but are not strict subset or superset extensions of each other.

doc/layer1/elsfp-and-paging-support.md
abhi-nexthop: The idea is to cleanly separate the pages of the Memory map rather than having them as a massive dump. This happens to work well with the idea that each device does not support every CMIS page, for example ELSFP CMIS explicitly mentions that page 6 and 7h are not supported

doc/layer1/elsfp-and-paging-support.md
abhi-nexthop: This simply enables easy remapping in vendor specific solutions. Its a benefit of the refactor that this HLD brings along with addition of ELS memory classes

doc/layer1/elsfp-and-paging-support.md
Natgerbi: If we look at the ELSFP CMIS specification itself, it states the following:

Image

Image

I agree that there may be a few pages that are not used, but we can simply choose not to use them when calling the CmisApi functions.
As I see it, MemMap should represent only the layout and should not contain functionality. Because of that, I think it is perfectly acceptable for some pages to remain unused as a result of inheritance.

I understand that in separate mode there may be a need for two CMIS interfaces. However, I still believe the right approach would be to have two memory maps: one for the common CMIS layout, and another for ELSFP CMIS that also inherits from CMIS for the ELSFP-specific information.

Also, in joint mode we use both CMIS and ELSFP CMIS through the same interface, so this approach does not fit joint mode well. With this design, we would end up duplicating the EEPROM layout.

doc/layer1/elsfp-and-paging-support.md
Natgerbi: I think this approach may not fully align with the requirement to reuse the existing framework. It may be helpful to also consider the proposed HLD in PR #2291 for the DOM vendor-specific extension mechanism. With that approach, we can continue leveraging the existing common framework while also enabling vendor-specific functionality through the same framework.

doc/layer1/elsfp-and-paging-support.md
aditya-nexthop: > I agree that there may be a few pages that are not used, but we can simply choose not to use them when calling the CmisApi functions.

Why not do it properly that reflects the actual pages implemented? We already have classes like CCmisMemMap, CmisFlatMemMap that try to portray the correct pages and fields available.

> MemMap should represent only the layout and should not contain functionality

Please can you explain how the MemMap is containing functionality here?

> Also, in joint mode we use both CMIS and ELSFP CMIS through the same interface, so this approach does not fit joint mode well. With this design, we would end up duplicating the EEPROM layout.

I think your comment might be implementation specific. Please can you explain how the EEPROM layout is duplicated in this proposal?

doc/layer1/elsfp-and-paging-support.md
aditya-nexthop: The linked HLD only adds support for DOM vendor specific extensions and can coexist with this proposal.
It does not seem to allow for the ability to read static and identification information from the ELSFP memory map space that is not a part of DOM/VDM/CDB - things like serial number, vendor information, advertisements etc.

This is possible by having the ELSFP be its own memory map, and the vendor can set up the composite memory maps however they like. Any custom, non-standard DOM registers can still use the vendor specific extensions so that those registers are polled automatically alongside DOM.

In our view, an ELSFP is a standalone device, with its own memory map and not just a small set of vendor-specific register extensions. In joint mode, some or all pages of the ELSFP may be mapped to custom/vendor-specific pages. Our proposal allows for an easy way of accomplishing that without having to re-define all the ELSFP registers in a vendor-custom space.

doc/layer1/elsfp-and-paging-support.md
aditya-nexthop: Also, the redesign is strictly internal. The existing memory maps - CmisFlatMemMap, CmisMemMap, CCmisMemMap are functionally identical with no required changes to their users. This also allows for easier composition of strictly applicable pages into future mem maps (Elsfp as an example) and makes them more maintainable by separating out fields into their respective pages rather than them being in one monolithic CmisMemMap class.

doc/layer1/elsfp-and-paging-support.md
aditya-nexthop: Same reason there exist CmisFlatMemMap and CCmisMemMap. We tried to be true to the actually implemented pages, just like these memory maps applying to their respective module types.

doc/layer1/elsfp-and-paging-support.md
Natgerbi: My point is not that ELSFP uses all CMIS pages, but that it should still be modeled as an extension of the CMIS memory map, similar to how CCmisMemMap extends CmisMemMap.

I do not see inheriting unused pages as a problem by itself, since inheritance here is mainly about preserving the common structure, while the implementation can still use only the relevant subset.

Regarding duplication, what I meant is that in joint mode, if both CMIS and ELSFP are represented separately over the same interface, the lower/common pages could end up being defined twice. That said, as you noted, this is likely implementation-specific, and from our side I expect we will use a single memmap anyway.

Because of that, I still think it would be better to keep a consistent approach across both joint and separate modes, and model ELSFP through inheritance from the common CMIS map rather than through a selective-page definition.

abhi-nexthop:
Our Implementation PRs can currently be found here (will be moved to sonic-net once parent PR is merged upstream):
PR 1: Memory Map Refactor into pages
https://github.com/nexthop-ai/sonic-platform-common/pull/1

PR 2:  ELSFP pages and Memory Map
https://github.com/nexthop-ai/sonic-platform-common/pull/2

doc/layer1/elsfp-and-paging-support.md
aditya-nexthop: Perhaps https://github.com/sonic-net/SONiC/pull/2211 will help explain how these separate memory maps will be used.

Even though there are two separate memory maps, and their corresponding separate CmisApis for them, the `CompositeSfp` concept bundles them together to ensure that they appear to `xcvrd` as a single device. 

In joint mode, ELSFP memory pages may be mapped at different vendor pages (00L at B0h, or some other vendor-custom page) but what remains the same is the content of the page - which is based on ELSFP-CMIS.
Our design offers the flexibility of defining a standards based ELSFP memory map that can be flexibly remapped as per requirements, allowing use of the same ElsfpCmisApi to access ELSFP information, no matter where a given ELSFP page is mapped in a combined/joint-mode memory map.

I think our design offers a consistent approach for both separate mode and joint mode.

For the sake of discussion, if ELSFP memory map were to extend CmisMemMap, it would also have Page 10h and 11h. Would your proposal be to use this ElsfpMemMap for joint mode and have:
- Pages 00L, 00h, 01h, 02h, 10h, 11h reflect OE information,
- then 1Ah and 1Bh show ELSFP information and
- the vendor custom pages B0h-B3h show ELSFP information again?

This approach works just because ELSFP 1Ah and 1Bh are luckily mapped to the standard assigned pages for ELSFP, but a different vendor that maps ELSFP information at a different set of vendor custom pages will be unable to use the custom-defined non 1Ah/1Bh ELSFP register definitions without duplicating code.

Also, the ELSFP memory map used in such a way would not truly reflect an ELSFP and would instead be a mix of OE and ELSFP information.

abelamit (dismissed):

prgeor (dismissed):

doc/layer1/elsfp-and-paging-support.md
```diff
@@ -0,0 +1,475 @@
...
+              .
+            ]
+            self.fields[consts.ADVERTISING_FIELD] = [
+                        NumberRegField(consts.INACTIVE_FW_MAJOR_REV, self.getaddr(128), format="B", size=1),
+                        NumberRegField(consts.INACTIVE_FW_MINOR_REV, self.getaddr(129), format="B", size=1),
+              .
+              .
+              .
+            ]
+  class CmisCdbMessagePage(CmisPage): #0x9F
```
micas-net, line 187: Can we split this into two HLDs? The ELSFP part may have less impact, while the paging part could significantly affect the current architecture. Also, is the paging class really necessary? I hope this won‘t block the 202605 branch.

aditya-nexthop: Yes, the code PR is split into two.
Paging allows for better organization of the EEPROM in order to compose the ELSFP memory map easily.
Code PR is ready - [Private version](https://github.com/nexthop-ai/sonic-platform-common/pull/1) and I just created a [public one](https://github.com/sonic-net/sonic-platform-common/pull/667) for merge
The second one is currently in a [private fork](https://github.com/nexthop-ai/sonic-platform-common/pull/2) but will be made public once the first merges to make it easier to review.

doc/layer1/elsfp-and-paging-support.md
```diff
@@ -0,0 +1,475 @@
...
+#### 7.2.3 The CmisMemMap class
+
+The CmisMemMap class is refactored to be a container for CmisPage objects, while remaining functionally identical. A helper is added to get fields from multiple pages at once, since a single field may have registers across multiple pages.
+
+```python
+class CmisMemMap(XcvrMemMap):
+    def get_field_from_pages(self, field_name, *pages):
+        fields = []
+        for page in pages:
+            fields.extend(page.get_field_values(field_name))
```
prgeor: @abhi-nexthop based upon how frequent this function is called, can we get the fields saved in object during construction time?

aditya-nexthop: This part was out of date. Updated now.

doc/layer1/elsfp-and-paging-support.md
aditya-nexthop: The second PR is now up at https://github.com/sonic-net/sonic-platform-common/pull/678

abelamit (dismissed):

prgeor (approved):

PR #2211, opened 2026-02-06
HLD: Port Mapping for CPO Platforms

bgallagher-nexthop:
Add a HLD describing how ports/interfaces driven by multiple independent hardware devices may be supported. This is motivated by upcoming hardware that utilizes co-packaged optics (CPO), since there exists some CPO hardware where each port leverages both an optical engine device and external laser source device to drive traffic through the port.

The design attempts to remain generic to support hardware platforms that leverage an arbitrary number of I2C configurable devices, both shared and not shared, for each given port.

PRs that implement the proposal in this HLD are below. They are currently stacked on other PRs yet to merge upstream, but will be later rebased against upstream master once those dependent PRs merge.
- https://github.com/nexthop-ai/sonic-platform-common/pull/3
- https://github.com/nexthop-ai/sonic-platform-common/pull/4


prgeor (commented):
@bgallagher-nexthop can the title be updated to `Port mapping for CPO platforms` ?

bgallagher-nexthop:
> @bgallagher-nexthop can the title be updated to `Port mapping for CPO platforms` ?

Updated the PR title and title of the HLD to the suggested title. I had initially kept the naming of this HLD not scoped specifically to CPO because the design attempts to remain generic enough to support potential future hardware platforms that leverage multiple (potentially shared) hardware devices for each port.

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,464 @@
...
+| OE | Optical Engine |
+| ELSFP | External Laser Small Factor Pluggable |
+| MCU | Microcontroller Unit |
+| SerDes | Serializer/Deserializer |
+| ASIC | Application-Specific Integrated Circuit |
+| EEPROM | Electrically Erasable Programmable Read-Only Memory |
+
+### 4. Overview 
+
+This HLD proposes adding a new configuration file called `optical_devices.json` that will contain information about the hardware devices used to drive traffic through the ports of a switch, and how each device is mapped to each port/interface on the switch. Specifically, this will enable configuration of hardware where:
```
prgeor:
```suggestion
This HLD proposes adding a new configuration file called cpo.json` that will contain information about the hardware devices used to drive traffic through the ports of a switch, and how each device is mapped to each port/interface on the switch. Specifically, this will enable configuration of hardware where:
```

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,630 @@
...
+          device = optical_device_data.devices[dvc_info.device_id]
+          if device.device_type == "optical_engine":
+            oe = VendorSfp(..., bank=dvc_info.bank)
+          elif device.device_type == "external_laser_source":
+            elsfp = VendorSfp(..., bank=dvc_info.bank)
+
+        # Create a composite SFP to represent this interface
+        assert oe, "No optical engine found for this interface"
+        assert elsfp, "No ELSFP found for this interface"
+        self._sfp_list.append(CpoSfpBase(oe=oe, elsfp=elsfp))
```
KroosMicas: Where is CpoSfpBase? Do you mean CpoSfpOptoeBase ?

bgallagher-nexthop: Thanks, good catch 👍 this is supposed to be a vendor-implemented class (which would inherit from `CpoSfpOptoeBase`)

I have changed it to be `VendorCpoSfp(...)`

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,631 @@
...
+      // This allows arbitrary devices to be added easily -- for instance,
+      // if there were separate TX and RX optical engines that required different
+      // configuration in the future, these could be accommodated without any
+      // schema change by simply adding new device types of optical_engine_rx
+      // and optical_engine_tx.
+      "device_type": "optical_engine",
+      // The number of CMIS banks associated with this device.
+      "max_banks": 4,
+      // The overall number of physical lanes available in this device.
+      "lanes": "41,42,43,44,45,46,47,48,49,50,51,52,53,54,55,56,...",
```
KroosMicas, line 140: Are these the physical lanes​ of the ASIC chip? What is the purpose of the OE​ storing this information?

bgallagher-nexthop: Yeah, that is correct -- these are the ASIC lanes the optical engine is responsible for. 

It can be used to understand the entire topology of the system. For instance, one immediate usage of this info is to use it to validate our configuration files (i.e the ASIC lanes in `platform.json` for an interface should be present in the ASIC lanes of the OE called out in the `associated_devices` field, otherwise we have a misconfiguration).

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,631 @@
...
+### 8. SAI API 
+
+There are no changes to SAI API in this HLD.
+
+### 9. Configuration and management 
+The introduction of a composite SFP abstraction will require changes to `sfputil`, because composite SFPs do not implement `read_eeprom()` and `write_eeprom()` methods and `sfputil` does not yet know how to reason about composite SFPs in general. `sfputil` will need to be modified to access the underlying SFPs inside a composite SFP directly.
+
+#### 9.1 sfputil Changes
+
+A new command will be added to `sfputil` to allow users to list all the names of the underlying Sfps in a composite Sfp.
```
keboliu: Besides sfputil, any other daemons will access this new Composite SFP object? If yes, would you please give an example?

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,631 @@
...
+
+`sfputil` will require changes across most, if not all, commands to support the composite SFP abstraction. A separate HLD in the future will cover those changes in greater detail.
+
+#### 9.2 Config DB Enhancements
+
+Schema changes will be required in CONFIG_DB to store the information encoded in optical_devices.json.
+
+The `associated_devices` field will be added to the PORT table, so that the mapping of interface to devices is stored in CONFIG_DB.
+```yang
+list associated_devices {
```
keboliu: Why is this information needed to be populated in the config DB?  How will it be consumed?

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,631 @@
...
+        }
+    }
+}
+```
+
+A new OPTICAL_DEVICES table will be added to CONFIG_DB to store information about each device associated with a port (optical engine, external laser source, etc.).
+```yang
+// sonic-yang-models/yang-models/sonic-optical-devices.yang
+
+module sonic-optical-devices {
```
keboliu: How will this information be consumed? Would you please give an example?

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,631 @@
...
+Device Name    Type                   Bank  Part Number  Vendor
+-----------    ---------------------  ----  -----------  ----------------
+OE1            Optical Engine         0     012345678    ExampleVendor
+ELS1           External Laser Source  0     123456789    ExampleVendorTwo
+```
+
+In addition, all existing commands to query/manipulate transceivers will be modified to accept a new optional parameter `-d <device>` to specify the name of the underlying Sfp to query/manipulate. This parameter is not required for non-composite Sfps, but if not specified for a composite Sfp an error will be returned asking the user to specify a device.
+
+```
+$> sfputil show eeprom-hexdump -p Ethernet4
```
keboliu: I assume in a logical CPO port, a common SfpOptoeBase sfp object will also be created and associated with it. How will sfputil decide to block access to this SfpOptoeBase sfp object associated with this logical port?

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,631 @@
...
+This HLD proposes adding a new configuration file called `optical_devices.json` that will contain information about the hardware devices used to drive traffic through the ports of a switch, and how each device is mapped to each port/interface on the switch. Specifically, this will enable configuration of hardware where:
+- hardware devices are shared amongst multiple ports/interfaces.
+- multiple hardware devices are used by a single port/interface.
+
+### 5. Requirements
+
+- The HLD should provide support for configuring upcoming generations of hardware using co-packaged optics.
+- The HLD should provide a mechanism generic enough to cover future hardware with complex topologies where ports/interfaces either share hardware devices, or leverage multiple hardware devices per port/interface.
+- The HLD will not support configuration for ports/interfaces that span multiple CMIS banks (a transceiver wider than 8 lanes for instance).
+
```
eddyk-nvidia, line 90: Please add the need to support both "joint" and "separate" modes of work with CPO module as a requirement

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,631 @@
...
+We can achieve modeling of complex multi-device topologies via the **platform.json** file and a new **optical\_devices.json** file.
+
+```json
+# platform.json
+// No changes to platform.json
+{
+  "interfaces": {
+    "Ethernet0": {
+      "index": "1,1,1,1,1,1,1,1",
```
eddyk-nvidia, line 107: Based on your suggestion in another HLD this index will be a number of SFP object covering 8 lanes ?

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: That is correct, the composite SFP class for CPO will inherit from either `SfpBase` or `SfpOptoeBase` (depending on whether the switch vendor uses the optoe driver). The composite SFP class overrides the methods for direct EEPROM access (`read_eeprom()` and `write_eeprom()`) to throw exceptions when called on a composite SFP, since the composite SFP is a logical entity and direct EEPROM access semantics do not make sense. The code where we ban direct EEPROM access by default can be seen in section 7.3.1.

The recommended approach instead is to use the methods to fetch underlying devices directly and access the EEPROM directly that way. For instance, something like the below could be done:
```
def eeprom_hexdump_for_composite_sfp(composite_sfp):
  for internal_device in composite_sfp.get_internal_devices():
    # internal_device will just be a SfpOptoeBase or SfpBase derived object.
    # we can now do things like internal_device.read_eeprom(0, 8)
    # or internal_device.write_eeprom(0, 8, buffer)
```

This means `sfputil` would likely encounter an exception when attempting to call these methods during eeprom dump without any changes. Note that there is an upcoming follow-up HLD that will go into the specifics of how exactly `sfputil` will need to change for CPO hardware. That HLD will propose changes to ensure that we don't just dump a stacktrace, but provide an intuitive behaviour when dumping eeprom for these composite SFPs. The section in this HLD is mostly just a call out that `sfputil` will certainly require changes to become aware of the idea of composite SFPs.

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: `xcvrd` will also use this composite SFP object.

The composite SFP object can be set up in a way that allows it to be a drop-in, backwards compatible replacement for a typical `SfpBase`/`SfpOptoeBase` derived object (in particular, see the code snippet for CpoApi constructor in section 7.3.2).

```
class CpoApi(CmisApi):
    """
    An XcvrApi implementation for hardware with co-packaged optics (CPO) comprising of 
    an optical engine and an external laser source.
    """
    def __init__(self, optical_engine_xcvr_api, external_laser_source_xcvr_api) -> None:
        # We use the optical engine API as our "default" choice for existing CmisApi methods.
        super().__init__(optical_engine_xcvr_api.xcvr_eeprom)
        self.optical_engine_xcvr_api = optical_engine_xcvr_api
        self.external_laser_source_xcvr_api = external_laser_source_xcvr_api

    # Implement ELSFP specific methods below here
    def get_elsfp_lane_output_power_alarm(self):
        return self.external_laser_source_xcvr_api.get_elsfp_lane_output_power_alarm()

    ...
```

This is set up so that most code that assumes a 1:1 relationship between ports and transceivers (like `xcvrd`) will not be broken and instead just mostly report optical engine stats (or whatever device a vendor chooses to delegate to by default). However, we will need changes to `xcvrd` to report telemetry for CPO hardware properly, probably in some dedicated DOM info update task specific to CPO that will call the ELSFP specific methods to fetch telemetry from the ELSFP in addition to the OE. The composite SFP abstraction allows code to choose when it wants to use new methods (internal devices are directly accessible, CPO subclasses can expose OE/ELSFP specific information in new methods implemented on the subclass) while remaining somewhat backwards compatible with existing SfpBase/SfpOptoeBase derived objects.

The changes `xcvrd` will require is left out of scope of this HLD however. The intention is that another HLD will  outline what changes are required in xcvrd and how the DB schema in state DB and counters DB should change to allow OE and ELSFP information to be published to redis for CLI commands to read/write.

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: While this information is purely static and won't change (it represents physical hardware topology, so not mutable by an operator of the network switch), I have included it so that `sonic-cfggen` can read the `optical_devices.json` file and publish its contents to Config DB so that CLI commands can use the information.

For instance, there could be a subcommand of `show interfaces` (maybe `show interfaces devices` or something like that) which would enumerate the physical devices associated with a certain interface by reading this field in the PORT table (and the information in the OPTICAL_DEVICES table below if more detailed information is desired).

A rough draft example on a hypothetical system using one OE and two ELSFPs per interface:
```
$> show interfaces devices Ethernet0
Device                Associated ASIC Lanes     CMIS Bank
OE1                   1,2,3,4,5,6,7,8           0
ELS1                  1,2,3,4                   0
ELS2                  5,6,7,8                   0
```

I decided to include this information in Config DB to follow existing precedent where `sonic-cfggen` already publishes information from the `interfaces` section of `platform.json` to the PORT table in Config DB so that CLIs and daemons can read it instead of parsing the JSON file directly. Let me know if you have opinions against this, as I am not very strongly for/against this until details of show commands for CPO are more clear.

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: See the above comment for how this information may be used (i.e for CLI commands and daemons to read from + follow existing precedent for how other files like platform.json are published to Config DB for simplicity)

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: Done 👍

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: The schema and data of the "interfaces" section of `platform.json` remains completely unchanged by this HLD. It is just included for clarity in how `optical_devices.json` interacts with the data already present in that file.

The meaning of the index file remains the same as outlined in the [existing HLD](https://github.com/sonic-net/SONiC/blob/master/doc/dynamic-port-breakout/sonic-dynamic-port-breakout-HLD.md#platform-capability-design) that introduced the field:

> "index": the indexes for individual lanes. This usually matches the front panel port #, and is used by platform plugins for optical module queries.

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: Another thing I forgot to mention -- this schema can also be extended to accommodate configurable fields of CPO hardware. For instance, if an operator wants to set per-lane output power or laser bias for an ELSFP, fields can be added to this schema for CLI commands to set those fields in Config DB, which daemons will react to and program the setting into hardware.

keboliu: Under the same platform, different SKUs can be created with different logical port mappings. Is a per-platform optical_devices.json sufficient, or do we need it as per SKU?  Or can we even re-use the existing `hwsku.json` file?

bgallagher-nexthop: That makes total sense. I have updated the HLD to allow defining an optical_devices.json file in the hwsku directory, with a fallback to the platform directory if none exists in the hwsku folder at the end of section 7.1.

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,636 @@
...
+```
+
+##### 7.3.4 CPO Joint Mode
+There is a mode of operation for CPO hardware called "joint mode" where a single I2C device called an "MCU" handles all reads/writes, redirecting them to the optical engine EEPROM or the ELSFP EEPROM appropriately.
+
+<br>
+
+![](./cpo_joint_mode.png)
+<br>
+
```
Natgerbi: Based on the last meeting with Microsoft, separate mode will not be supported.

With your proposed design, are you expecting to maintain two memory maps that inherit from CmisMemMap, one for OE and one for ELS? If so, both would still reference the same lower pages, since they are exposed through the same interface.

In addition, since we are discussing joint mode, where a single interface is used for two different EEPROMs, defining a fully generic interface for the entire ELS is currently somewhat challenging. This is because the mapping may vary between switch vendors. Unlike separate mode, this is not standardized, as there is no direct access to the EEPROM.

tshalvi (commented):
Following the recent cross-team discussion, the aligned direction is to support **Joint Mode only**, and Separate Mode is currently not part of the agreed scope.
Given that, most of the document is based on Separate Mode assumptions (e.g., modeling OE and ELS as separate Sfp objects and building the platform API around this separation), which does not align with the expected implementation direction.
In Joint Mode, the system exposes a single I2C device that handles all accesses, and the software stack interacts with a unified module abstraction. OE and ELS are not modeled as separate entities in SONiC.
 
However, the current Joint Mode description would benefit from further clarification.
Specifically, it is not clear what the required changes are for supporting Joint Mode, independent of the Separate Mode design. The current description suggests reusing the OE and ELS Sfp objects and simply pointing them to the same sysfs path. This approach still relies on Separate Mode abstractions and effectively carries over their complexity into Joint Mode.
From a Joint Mode perspective, this raises a concern that the OE/ELS separation is being preserved even though it is not required. Maintaining this structure may be unnecessary, given that Joint Mode exposes a single MCU-backed interface and does not require modeling OE and ELS as separate entities in SONiC.
 
It would be helpful to clarify:
- What are the changes strictly required for Joint Mode
- Whether the OE/ELS Sfp object separation is actually needed in this mode, or if it can be avoided entirely

This clarification would help ensure that the Joint Mode design remains simple and avoids introducing abstractions that are only relevant for Separate Mode.

@prgeor FYI

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: I have added some code examples to the section on CPO joint mode (section 7.3.4) that more comprehensively outlines how the composite SFP abstraction proposed in this design can be used for joint mode.

It describes how the memory map proposal [here](https://github.com/sonic-net/SONiC/pull/2207/) can be used to define an ELSFP memory map located at an arbitrary address space by a vendor using reusable CMIS page "building blocks", and how that custom memory map can then be used as part of a composite SFP.

The primary benefit is that this approach is not strictly limited to CPO joint mode, and can be leveraged for:
- CPO separate mode, if the wider SONiC community requires support for a hardware platform utilizing separate mode
- arbitrary hardware configurations beyond CPO, if a new hardware platform is introduced that utilizes multiple independent i2c configurable/monitorable devices for each front-panel port.

bgallagher-nexthop: cc @tshalvi, as I think this addresses your below questions also. The OE and the ELS would be presented as separate entities to software regardless of whether the hardware has a separate or joint mode design. The composite SFP design allows for both "direct" access to each device, while also preserving a single unified interface for backwards compatibility with existing code that assumes a 1:1 relationship between front-panel ports and transceiver devices. This leaves the software abstraction presented to SONiC decoupled from the reality of how i2c access occurs in the underlying hardware.

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,735 @@
...
+# The vendor only needs to override _init_pages() to construct the relevant page instances at their
+# remapped locations. The RegGroupField wiring is inherited from ElsfpMemMap.
+class CustomVendorElsfpMemMap(ElsfpMemMap):
+    def _init_pages(self, codes, bank):
+        # Remapped CMIS pages
+        self.advertising_page = CmisAdvertisingPage(codes, bank=bank, page=0xB0)
+        self.thresholds_page = CmisThresholdsPage(codes, bank=bank, page=0xB1)
+        self.performance_monitoring_page = CmisVdmAdvertisingCtrlPage(codes, bank=bank, page=0xB2)
+        self.cdb_message_page = CmisCdbMessagePage(codes, bank=bank, page=0xB3)
+
```
Natgerbi: those are not banked, I think it will be better to send static values 0 instead of sending the bank id ..

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,735 @@
...
+    def get_number_of_internal_devices(self) -> int:
+        return 2
+
+    def get_internal_device(self, name: str) -> SfpBase:
+        if name == "OE":
+            return self._oe
+        elif name == "ELS":
+            return self._els
+        raise ValueError(f"No SFP found for {name}")
```
Natgerbi: how is it going to work in case of multiple ELS ?

abelamit (dismissed):

doc/platform/port_mapping_for_cpo.md
aditya-nexthop: We would have to create a `MultiElsfpCpoSfpBase` class inheriting from `CompositeSfpBase` in a new file `multi_elsfp_cpo_base.py`
Does your implementation need it? We can add it to the code changes.

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: Removed the bank parameter here

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,785 @@
...
+      "device_type": "external_laser_source",
+      // Number of individual lasers this device provides.
+      "lasers": 8,
+      "max_banks": 1,
+      // laser_to_lane_mapping provides a mapping of laser to which lane it is powering.
+      "laser_to_lane_mapping": {
+        "1": "41,42",
+        "2": "43,44",
+        "3": "45,46",
+        "4": "47,48",
```
pavannaregundi, line 156: In this example, see that max_banks is 1, which is bank-0 used by Ethernet0. However there are 8 lasers, so it feels like it can have another bank. Either max_banks or lasers looks incorrect.

bgallagher-nexthop: Hi Pavan, the max number of lanes available in a bank is 8 per CMIS, so these 8 lanes can be described by a module with a single bank is the intent behind this configuration.

ELS1 would be used by Ethernet0, Ethernet8, Ethernet16 and Ethernet24 -- all with bank 0. The laser-to-lane mapping then helps figure out which lanes of those 8 are associated with a given interface.

I might be missing something or misunderstanding, let me know.

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,785 @@
...
+    "OE1": {
+      // device_type is an indicator for what type of device we are dealing with.
+      // This allows arbitrary devices to be added easily -- for instance,
+      // if there were separate TX and RX optical engines that required different
+      // configuration in the future, these could be accommodated without any
+      // schema change by simply adding new device types of optical_engine_rx
+      // and optical_engine_tx.
+      "device_type": "optical_engine",
+      // The number of CMIS banks associated with this device.
+      "max_banks": 4,
```
micas-net, line 138: The max bank is same with bank_size, it should be also read from EEPROM, we can remove it from here.

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,785 @@
...
+      // The overall number of physical lanes available in this device.
+      "lanes": "41,42,43,44,45,46,47,48,49,50,51,52,53,54,55,56,...",
+      // If a vendor needs to encode I2C mapping information, that information
+      // can be included here. However, this is platform dependent and may be
+      // achieved via other mechanisms.
+      "i2c_path": "/sys/bus/i2c/devices/32-0050"
+    },
+    "ELS1": {
+      "device_type": "external_laser_source",
```
micas-net, line 147: The type of ELS should be read from EEPROM, we can remove it from here.

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,785 @@
...
+      // Number of individual lasers this device provides.
+      "lasers": 8,
+      "max_banks": 1,
```
micas-net, line 150: The max bank is same with bank_size, it should be also read from EEPROM, we can remove it from here.

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,785 @@
...
+        "3": "45,46",
+        "4": "47,48",
+        ...
+      }
+    },
+    "OE2": {...},
+    "ELS2": {...},
+    ...
+  },
+  "interfaces": {
```
micas-net, line 164: It is better to merge this part into platform.json, no need to split them into two files.

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,785 @@
...
+            return self._oe
+        elif name == "ELS":
+            return self._els
+        raise ValueError(f"No SFP found for {name}")
+
+    def refresh_xcvr_api(self):
+        self._xcvr_api = CpoApi(
+            optical_engine_xcvr_api=self._oe.get_xcvr_api(),
+            external_laser_source_xcvr_api=self._els.get_xcvr_api()
+        )
```
Natgerbi: Based on my last conversation with Prince, my understanding is that the proposed approach is to expose the OE and ELS APIs based on the EEPROM vendor name, and possibly also the vendor PN in the future. @prgeor , please correct me if I misunderstood.

Please review section 7.2 DOM: CPO API Wiring in https://github.com/sonic-net/SONiC/pull/2273, as I believe it provides a clearer explanation of the approach.

doc/platform/port_mapping_for_cpo.md
aditya-nexthop: The `max_banks` here is the system capability and is needed at chassis creation time to determine the number of OE `Sfp` objects to instantiate. The value read from the EEPROM is used to write to the sysfs path so that we don't have weird behavior. In the case of CPO systems these two should match.

In the case of say, a system with OSFP-XD ports, we would instantiate 2 `Sfp` objects based on max_banks, but if a 8 lane transceiver is inserted that supports/advertises only one bank, only the first Sfp object (bank=0) would be used.

doc/platform/port_mapping_for_cpo.md
aditya-nexthop: Please can you share the other types possible? This field was used to differentiate an OE from a laser source. Perhaps we should rename it to just `laser_source`?

doc/platform/port_mapping_for_cpo.md
aditya-nexthop: same response as above. The platform capability in this case, only supports ELSFPs with a single bank. If a multi-bank ELSFP is inserted, the platform only knows to use bank 0.

doc/platform/port_mapping_for_cpo.md
aditya-nexthop: we thought it better to keep it separate since it is applicable only for CPO, but open to putting it in `platform.json`.
@prgeor, thoughts?

doc/platform/port_mapping_for_cpo.md
aditya-nexthop: We should discuss this more.

Do we really need to read EEPROM vendor name or PN to figure out the OE that is fixed on the board for that platform? I agree that ELSFP identification can possibly be driven based on EEPROM because it can be removed/inserted.

In any case, we can still achieve what you're looking for in this class which is in the HLD, without making changes to the common XcvrApiFactory and introduce vendor-custom codes in there:
```
class VendorCpoJointModeSfp(CpoSfpBase):
    def refresh_xcvr_api(self):
        # Pick the appropriate code/MemMap/API combination for each underlying device.
        # Additional logic can be applied here at refresh time -- e.g. selecting a different
        # MemMap or API class based on what ELSFP module is currently inserted.
        cmis_mcu_config = XcvrApiConfig(codes_cls=CmisCodes, mem_map_cls=VendorCmisMcuMemMap, api_cls=CmisApi)
        elsfp_config = XcvrApiConfig(codes_cls=CmisCodes, mem_map_cls=CustomVendorElsfpMemMap, api_cls=ElsfpApi)

        self._oe.set_xcvr_api_config(cmis_mcu_config)
        self._els.set_xcvr_api_config(elsfp_config)

        # Each underlying Sfp builds its XcvrApi from the most recently assigned config.
        self._xcvr_api = CpoApi(
            optical_engine_xcvr_api=self._oe.get_xcvr_api(),
            external_laser_source_xcvr_api=self._els.get_xcvr_api(),
        )
```
We bypass the XcvrApiFactory altogether by using the XcvrApiConfig mechanism.

doc/platform/port_mapping_for_cpo.md
micas-net: I think the max_banks is equal to bank_size. For the SFP instance, it will be defind in the relationship of interface and oe bank id. please help figure out where it can be used, I can't find it in the HLD.

abelamit (dismissed):

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,620 @@
...
+### 7. High-Level Design 
+
+#### 7.1 Configuration Data Model
+
+We can achieve modeling of complex multi-device topologies via the **platform.json** file and a new **optical\_devices.json** file.
+
+```json
+# platform.json
+// No changes to platform.json
+{
```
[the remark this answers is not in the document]
bgallagher-nexthop: I will leave this as is, the syntax highlighting is valuable from the `json` marker and the comments are intentionally there (despite not being valid JSON) for descriptive purposes.

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,620 @@
...
+  - The ELSFP should be initialized with a memory map that is aware of the memory layout that the MCU exposes (in joint mode, the ELSFP EEPROM will likely be mapped into some part of the MCU's address space).
+
+So, in order to instantiate a CpoBase-derived object for a port of a hardware platform using CPO joint mode, the vendor defines a memory map that describes where the ELSFP memory has been remapped to using the approach outlined in [the companion HLD for ELSFP memory map layout here](https://github.com/sonic-net/SONiC/pull/2207). Each existing page class (both common CMIS pages and ELSFP-specific pages) accepts an optional `page` parameter that defaults to its spec-defined location, so vendors can remap a page simply by passing a different page number at construction time -- no page subclassing is required:
+
+```python
+class CustomVendorElsfpMemMap(CmisFlatMemMap):
+    def __init__(self, codes, bank=0):
+        self._bank = bank
+        super(CmisFlatMemMap, self).__init__(codes)
+        self.pages = []
```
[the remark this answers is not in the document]
bgallagher-nexthop: This is intentional.

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,620 @@
...
+4. **Cross-File Validation**
+   - Interface names in `optical_devices.json` match interfaces defined in `platform.json`
+   - Consistency between port counts and device counts
+   - Assigned lanes in each file are consistent across both files for each interface, and that interfaces don't have conflicting configuration for devices that are shared across multiple interfaces (overlapping lane assignment for instance)
+
+#### 13.2 System Test cases
+
+| Test Case |
+|------------------------------------------------|
+| Load proposed configuration onto a real device with co-packaged optics. Ensure that links come up successfully. |
```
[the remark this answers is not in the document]
bgallagher-nexthop: Done

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,620 @@
...
+2. **JSON Schema Validation**
+   - Required fields present (`devices`, `interfaces`)
+   - Correct data types for all fields
+   - Valid enumeration values (e.g., `device_type`)
+
+3. **Semantic Validation**
+   - All device IDs referenced in `interfaces` section exist in `devices` section
+   - Bank numbers are valid for CMIS devices (0-3)
+   - No duplicate device IDs
+   - Interface names are valid (match expected patterns)
```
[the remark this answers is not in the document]
bgallagher-nexthop: Done

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,620 @@
...
+    # List of SfpBase-derived objects representing all sfps
+    # available on the chassis
+    self._sfp_list = []
+    optical_device_data = device_info.get_optical_devices_data()
+    if optical_device_data:
+      self.construct_sfp_list(optical_device_data)
+
+  def construct_sfp_list_for_topology(self, optical_device_data):
+    """Subclasses should implement this method to create sfp objects based on topology data in optical_devices.json"""
+    raise NotImplementedError
```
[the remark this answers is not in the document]
bgallagher-nexthop: Fixed.

abelamit (dismissed):

doc/platform/port_mapping_for_cpo.md
```diff
@@ -0,0 +1,624 @@
...
+        "2x800G": [
+          "Port1/1",
+          "Port1/2"
+        ]
+      }
+    }
+  }
+}
+
+# optical_devices.json
```
prgeor: @bgallagher-nexthop can we rename this to cpo.json` file?

prgeor: @bgallagher-nexthop also where will this file reside? I think we don't need this .json file per hwsku as it describes the mapping at platform level

doc/platform/port_mapping_for_cpo.md
prgeor: agree with @aditya-nexthop .

doc/platform/port_mapping_for_cpo.md
prgeor: @bgallagher-nexthop aren't these information/mapping already available in .json file and the fact these mapping are static per CPO platform means we don't need it to be published to config DB?

micas-net (dismissed):
LGTM!

abelamit (dismissed):

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: Renamed the file to `cpo.json`.

The file contains mappings of logical interfaces to the devices and banks that are used by that logical interface (see the `interfaces` field of the JSON file), so I believe we need the ability to have per hwsku files so different breakout configurations can be modelled.

doc/platform/port_mapping_for_cpo.md
bgallagher-nexthop: That is correct. The CONFIG_DB section has been deleted from the HLD in a previous commit for this reason.

prgeor (approved):

PR #2215, opened 2026-02-10
SONiC BMC platform management and monitoring 

judyjoseph:
Platform architecture and monitoring planned for BMC running SONiC 
 - Interaction between Sonic on BMC and Switch-Host
 - Interaction with Rack-Manager
 - DB schema and various tables introduced
 - Update to the thermalctld in Switch-Host to push thermal sensor data to redisDB/BMC
 - Config/Show command changes


doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+#### 2.2.4 Platform APIs
+
+# Platform APIs Used
+
+Listing down the platform common APIs planned for sonic bmc support. 
+
+The following docs from Nvidia team which is already present define many platform API's for bmc, leak and liquidCooling
+
+* https://github.com/sonic-net/SONiC/blob/master/doc/bmc/bmc_hld.md                                                                                                       
+* https://github.com/sonic-net/SONiC/blob/master/doc/bmc/leakage_detection_hld.md
```
fraserg-arista: There are some proposed changes to the leak detection HLD to add leak severity: https://github.com/sonic-net/SONiC/pull/2217

It also contains changes to add switch host power management but that can be moved out (to the SwitchHostBase class mentioned below).

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+
+* BMC can be accessed in two ways (i) via the external management interface (ii) from the Switch Host via the internal Host-Bmc-Link
+* BMC can access Switch-Host redis DB over this internal Host-Bmc-Link.
+* BMC will manage the Switch Host to support operations like soft reboot, power up/down, get operational status.  
+* BMC will read local leak sensors, its severity and take appropriate actions based on policy.
+* BMC will get inputs from external Rack Manager on Inlet Liquid temperature, Inlet Liquid flow rate, Inlet Liquid Pressure and Rack level Leak. It takes action based on policy.  
+* BMC and Switch-Host shall each enable an independent Hw watchdog timer.
+  
+* Switch-Host can access BMC redis DB over this internal Host-Bmc-Link.
+* Switch-Host will manage its thermal sensors and automatically power down when any thermal sensor temperature exceeds the policy-defined thresholds.
```
fraserg-arista: In systems with a hybrid cooling setup (air + liquid cooling), fan speed control would need to be aware of leak status since high fan speeds might cause any leaking liquid to spray.

Since this doc makes the leak status visible to the switch host via Redis and the Host-Bmc-Link, this should maybe state that the switch host's thermalctld also needs to monitor the leak state.

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+(i) Local leak detection uses the leak status in LIQUID_COOLING_DEVICE|leakage_sensors{X}.
+    The result ( whether it is critical/minor ) will be updated in LEAK_STATUS table defined in [2.2.2.1 DB schema](#2221-db-schema)
+        
+(ii) External Rack manager alert status is updated by redfish/bmcweb in RACK_MANAGER_ALERT table defined in [2.1.2.1 DB schema](#2121-db-schema).
+    
+        
+#### 2.1.6 BMC event logging
+
+The general syslogs will be placed in /var/log/syslog where /var/log directory will be mounted on **tmpfs **. Syslogs will be sent to remote server as well.
+The Leak, Switch-Host state and interactions, Rack-manager interactions will be persistently stored on disk/eMMC in "/host/bmc.log" (Note: yet to conclude on exact location )
```
fraserg-arista: To maximise the lifetime of eMMC storage, this logging should be limited to important events, e.g. when leak state changes and not every time the leak polling loop runs.

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+##### 2.2.1.1 bmcctld on BMC
+
+The bmc controller daemon "bmcctld" is started first in BMC pmon container. The following logic is applied
+
+```
+Loop on this logic 
+ Check if the Switch-Host is reachable and up ( In case when pmon/bmcctld restarts OR BMC alone restarts )
+ if YES
+ {
+   Check if the redfish connectivity with rack_manager is established ( This should be updated in redis DB by redfish docker processes)
```
fraserg-arista: If connectivity with the rack manager is lost and the switch host is powered on, what is the expected behaviour? Should it keep running or should it be powered off after some interval?

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+     else
+          - update DB SWITCH_HOST_STATE table with the status
+ }
+ if NO
+ {
+   if it is First Boot
+   {
+     Sleep for (5min - bootup time) ( this is configurable value in config_db)
+   }
+   Check if the redfish connectivity with rack_manager is established ( This should be updated in redis DB by redfish docker processes)
```
fraserg-arista: If connectivity with the rack manager cannot be established, should the switch host remain powered off?

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+
+#### 2.2.1 BMC controller - bmcctld
+
+##### 2.2.1.1 bmcctld on BMC
+
+The bmc controller daemon "bmcctld" is started first in BMC pmon container. The following logic is applied
+
+```
+Loop on this logic 
+ Check if the Switch-Host is reachable and up ( In case when pmon/bmcctld restarts OR BMC alone restarts )
```
fraserg-arista: 1. This line implies that a BMC restart does not cause a switch host restart. If so, that should probably be called out elsewhere.
2. If the switch host is powered but does not become reachable (or later becomes unreachable), should the BMC power cycle it? i.e should be BMC act as a watchdog for the switch host? (In addition to the HW watchdog but testing reachability rather than CPU responsiveness)

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+ if YES
+ {
+   Check if the redfish connectivity with rack_manager is established ( This should be updated in redis DB by redfish docker processes)
+   Check for Local Leak status + external Rack manager leak status
+     if CRITICAL LEAK
+          - syslog
+          - [enhancement] Try to do a soft reboot of Switch Host with a timeout of ~2min ( configurable )
+          - if unsuccessfull POWER_OFF Switch_host
```
fraserg-arista: Its not clear to me what a "successful" soft reboot of the switch host would look like here since the leak would still be present. Is the intention to do a clean shutdown (rather than reboot) of the switch host and do a hard power off if that fails?

It would also make sense for the BMC to do a hard power off unconditionally after the timeout, in case the switch host shutdown gets stuck after shutting down its containers and link but before the CPU powers off. Otherwise, the CPU might be left powered with no cooling available. While the HW watchdog would reset the CPU in this case, that leaves the CPU booting back up (with high power and cooling demands) with no cooling available.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+reason                    = STR                                       ; optional reason string
+
+key                       = BMC_STATE|bmc                             ; STATE_DB on Switch-Host to store state of BMC
+; field                   = value
+reachability              = REACHABLE | UNREACHABLE
+last_change_timestamp     = STR
+```
+
+#### 2.2.2 thermalctld
+In Liquid cooled platform, thermalctld will skip the PSU, FAN, SFP thermals but have additional responsibilities to check leak sensors and apply leak policy.
```
fraserg-arista, line 420: I presume this refers to thermalctld on the BMC and not thermalctld on the switch host?

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+       +--------------------------------------+-----------------------+
+       | 1 or more (Critical) leaks           |        CRITICAL       |
+       | 2 or more leaks with any Severity    |        CRITICAL       |
+       | 1 (Minor) leak stay leaking for MAX-T|        CRITICAL       |
+       | 1 (Minor) leak detected              |        MINOR          |
+       +--------------------------------------+-----------------------+
+
+       Additional considerations:
+         - MAX-T mins defined before which a Minor leak can be considered CRITICAL need to be tested, Should it be per platform ?
+         - Need to apply debounce timers
```
fraserg-arista: Should these debounce timers be on top of the platform API or should they be internal to it?

I'd say they're an implementation detail since the right values are going to depend on sensor technology and how they're monitored at a hardware level.

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+#### 2.2.2 thermalctld
+In Liquid cooled platform, thermalctld will skip the PSU, FAN, SFP thermals but have additional responsibilities to check leak sensors and apply leak policy.
+
+```
+Loop on this logic 
+  (i) Check local leak sensors using platform API
+         -- store the result in LIQUID_COOLING_DEVICE table
+         -- Leak severity can either be CRITICAL or MINOR
+
+  (ii) Apply the thermal policy based on the number and severity of leak sensors with leak
```
fraserg-arista: "Apply the thermal policy" here implied to me that action was being taken to respond to the leak. Could it be reworded to make it clearer that this is about mapping leak sensors to overall system leak status?

e.g. "Apply the following policy to map leak sensor status and severity to an overall system leak status/severity".

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+       | Leak conditions                      | System Leak Severity  |
+       +--------------------------------------+-----------------------+
+       | 1 or more (Critical) leaks           |        CRITICAL       |
+       | 2 or more leaks with any Severity    |        CRITICAL       |
+       | 1 (Minor) leak stay leaking for MAX-T|        CRITICAL       |
+       | 1 (Minor) leak detected              |        MINOR          |
+       +--------------------------------------+-----------------------+
+
+       Additional considerations:
+         - MAX-T mins defined before which a Minor leak can be considered CRITICAL need to be tested, Should it be per platform ?
```
fraserg-arista: Per-platform would make sense since it's going to depend on placement of leak sensors and routing of liquid cooling pipes.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+
+```
+Loop on this logic 
+  (i) Check local leak sensors using platform API
+         -- store the result in LIQUID_COOLING_DEVICE table
+         -- Leak severity can either be CRITICAL or MINOR
```
fraserg-arista: Should there be handling of failed or missing sensors? e.g. if a sensor is non-responsive or reports an out of range value should it be treated as a leak?

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+In **Liquid cooled platforms** a new daemon **"bmcctld"** will be introduced in pmon container on both BMC and Switch-Host
+
+#### 2.2.1 BMC controller - bmcctld
+
+##### 2.2.1.1 bmcctld on BMC
+
+The bmc controller daemon "bmcctld" is started first in BMC pmon container. The following logic is applied
+
+```
+Loop on this logic
```
fraserg-arista: Add "Check for and act on commands from the rack manager" to the logic loop.

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+```
+Loop on this logic 
+ Check if the Switch-Host is reachable and up ( In case when pmon/bmcctld restarts OR BMC alone restarts )
+ if YES
+ {
+   Check if the redfish connectivity with rack_manager is established ( This should be updated in redis DB by redfish docker processes)
+   Check for Local Leak status + external Rack manager leak status
+     if CRITICAL LEAK
+          - syslog
+          - [enhancement] Try to do a soft reboot of Switch Host with a timeout of ~2min ( configurable )
```
fraserg-arista: To clarify: the BMC is expected to act autonomously on detecting a leak, without waiting for a command from the rack manager?

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+
+#### 2.1.3 Host-Bmc-Link
+There is ethernet link between the Switch-Host and Switch-BMC ( eg: Ethernet over USB )
+
+The Switch-Host will intialize the usb netdev dring the inital platform bringup and name it as bmc0.
+Similarly the BMC will intialize the usb netdev dring the inital platform bringup and name it as bmc-eth0
+
+IP address to be configured on the Switch-Host end and Switch-Bmc end can be defined sonic wide unique in file "files/image_config/constants/bmc_ip_address.json" as below
+```
+Switch_Host=169.254.100.1
```
fraserg-arista: Netmask /24?

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+; field                   = value
+shutdown_delay            = float                              ; Time in secs the BMC will wait after REBOOT command issued to Switch-Host. ( default = 2 min ).
+                                                               ; if this timer expires, BMC will go ahead and POWER OFF switch-host
+
+```
+
+##### 2.2.1.2 bmcctld on Switch-Host
+
+The bmcctld daemon on the Switch-Host will subscribe to STATE DB table "BMC_TO_SWITCH_HOST_CMD" for any shutdown/reboot command from BMC.
+On receipt of this event, the Switch-Host will be shutdown/reboot gracefully.
```
fraserg-arista: Should there also be a command for immediate (hard) power-down? e.g. complete loss of cooling system pressure

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+severity                  = "status"                                  ;CRITICAL/MINOR
+
+key                       = LEAK_STATUS|local                         ; local bmc leak status in STATE DB
+; field                   = value
+device_leak_status        = "status"                                  ;CRITICAL/MINOR
+```     
+
+#### 2.2.3 Hw watchdog
+Hw watchdog timers will be enabled on both Switch-Host and BMC. 
+This will do a CPU reset when OS hangs or becomes unresponsive and fails to service the watchdog
```
fraserg-arista, line 480: Should there be a hardware leak detection watchdog? i.e if there's a leak and software fails to check the leak sensors for some period, the hardware will auto-powerdown the switch host.

doc/bmc/pmon/pmon-bmc-design.md
```diff
@@ -0,0 +1,432 @@
...
+---
+
+####  SwitchHostBase
+This Class is a new BaseClass to define which will contain API's for switch BMC to control the Switch-Host
+
+| Method | Present | Action |
+|---------|---------|----------|
+| power_on_switch_host() | New | Power on Switch Host from standby/off |
+| power_off_switch_host() | New | Immediate HW power-off (bypass OS) |
+| reboot_switch_host() | New | Graceful shutdown of Switch Host|
```
fraserg-arista: Is "reboot" the right term for a graceful shutdown? `shutdown_switch_host()` maybe?

fraserg-arista (commented):
Thanks @judyjoseph, it's good to see a design for this!

We've been working on this internally and there's a lot of corner cases, so I've added some questions on how they should be handled. Also some requests for clarification to make sure I'm understanding it properly.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,492 @@
...
+* Switch-Host will manage its thermal sensors and automatically power down when any thermal sensor temperature exceeds the policy-defined thresholds.
+    
+### 1.2. BMC Platform Stack
+
+![BMC Platform Stack](images/sonic-bmc-arch.png)
+
+## 2. Detailed Architecture and workflows
+### 2.1 BMC platform
+Presence of the file bmc.json in the vendor/platform directory tells this platform has BMC. The contents of bmc.json in Switch-Host and BMC are as below.
+```
```
benle7, line 88: It seems to be different than the original bmc.json that was introduced here:
https://github.com/yuazhe/SONiC/blob/143bd9c677c82157f6687f56eb488fe5a27dcb59/doc/bmc/bmc_hld.md#3-bmc-ip-address-initialization

For example:
https://github.com/sonic-net/sonic-buildimage/blob/master/device/mellanox/x86_64-nvidia_sn6600_ld-r0/bmc.json

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,492 @@
...
+* Send a POWER_ON command to Switch-host if all clear. 
+
+![BMC platform power up](images/sonic-bmc-bootup.png)
+
+
+#### 2.1.2 BMC Rack Manager Interaction
+The new docker container "redfish" in sonicBMC will have openbmc/bmcweb service which terminates the redfish calls from the rack Manager.
+
+**Note: Redish docker to be enabled only on Liquid cooling platform.**
+
```
benle7, line 117: Don't we need the Redfish server for every platform with BMC?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,492 @@
...
+```  
+
+#### 2.1.3 Host-Bmc-Link
+There is ethernet link between the Switch-Host and Switch-BMC ( eg: Ethernet over USB )
+
+The Switch-Host will intialize the usb netdev dring the inital platform bringup and name it as bmc0.
+Similarly the BMC will intialize the usb netdev dring the inital platform bringup and name it as bmc-eth0
+
+IP address to be configured on the Switch-Host end and Switch-Bmc end can be defined sonic wide unique in file "files/image_config/constants/bmc_ip_address.json" as below
+```
```
benle7, line 228: What about the bmc.json that was introduced here:
https://github.com/yuazhe/SONiC/blob/143bd9c677c82157f6687f56eb488fe5a27dcb59/doc/bmc/bmc_hld.md#3-bmc-ip-address-initialization

https://github.com/sonic-net/sonic-buildimage/blob/master/device/mellanox/x86_64-nvidia_sn6600_ld-r0/bmc.json

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: Thank you for reviewing. I have modified a bit the original approach to define a new baseclass SwitchHostBase --> use an existing  ModuleBase class since Module was also used earlier to model sonic running devices ( eg: Linecard/DPU etc ) which could be powered up and down. 

Here in this BMC based switch too, the Switch-Host is a complete sonic module with CPU/ASIC and overlapping capabilities. Let me know your thoughts.

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: I haven't thought of a system with hybrid air + liquid cooling, it was assumed to be either air cooled or liquid cooled in deployments ?

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: We are not actually monitoring the connectivity with external rack manager. The switch Host will remain powered on if the connectivity with Rack Manager is lost . In the event of connectivity with Rack manager going away -- Switch Host will be power off only based on local switch leak criticality

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: same as above

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: Sure added this to the Requirements section. BMC is doing local leak and rack level leak detection and powering OFF the SwitchHost. 

I am not sure if we really need to check health of Switch-Host and reboot .. as it could be a case where sonic network stack is not responding -- but ASIC is still forwarding traffic. Feel it should just follow current procedure of hos it is handled in sonic pizza boxes ? Let me know thoughts

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: Agree We have a timer for this and if GNOI request don't have a response .. we go for hard power off. The initial soft power off request could just umount filesystems | shut ports.

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: yes

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: yes .. I have added similar changes already please review again

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: I have already added the leak sensor faulty check as a platform API. please review again

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: yes

doc/bmc/sonicBMC/pmon-bmc-design.md
fraserg-arista: Resolved - not needed.

doc/bmc/pmon/pmon-bmc-design.md
fraserg-arista: Resolved - can be deferred until/if there's a need for describing hybrid systems.

doc/bmc/pmon/pmon-bmc-design.md
fraserg-arista: Resolved - this is a BMC-level concern and not specific to the pmon container and can be noted elsewhere.

doc/bmc/pmon/pmon-bmc-design.md
fraserg-arista: Resolved - rack manager connectivity is not a factor in power up or leak response.

doc/bmc/pmon/pmon-bmc-design.md
fraserg-arista: Resolved - rack manager connectivity is not a factor in power up or leak response.

doc/bmc/pmon/pmon-bmc-design.md
fraserg-arista: 1. Resolved - BMC restart doesn't affect the switch host.
2. Resolved - not needed (existing watchdog behaviour should be sufficient)

doc/bmc/pmon/pmon-bmc-design.md
fraserg-arista: Resolved - soft power down has hard power down backup as described.

doc/bmc/pmon/pmon-bmc-design.md
fraserg-arista: Resolved.

doc/bmc/pmon/pmon-bmc-design.md
fraserg-arista: Resolved.

doc/bmc/pmon/pmon-bmc-design.md
fraserg-arista: Resolved.

judyjoseph:
@manapalai @fraserg-arista @oleksandrivantsiv @chinmoy-nexthop please include your team and share review comments

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,740 @@
...
+
+|| Switch Host State (Start) | Event | Action | Switch Host State (Final) 
+|--|---|---|---|---|
+|1| ONLINE  | LOCAL_LEAK_CRITICAL_EVENT | Syslog, Power OFF Switch Host | OFFLINE 
+|2| ONLINE  | RACK_MGR_CRITICAL_EVENT | Syslog this event + Syslog the thermal sensor data | ONLINE
+|3| ONLINE  | RACK_MGR SHUTDOWN command | Syslog, graceful-shutdown Switch Host | OFFLINE
+|4| ONLINE  | CHASSIS_MODULE admin_down user request | Syslog, graceful-shutdown Switch Host | OFFLINE 
+|5| ONLINE  | LOCAL_LEAK_MINOR_EVENT | Syslog, External monitoring tool take action | ONLINE
+|6| ONLINE  | RACK_MGR_MINOR_EVENT | Syslog, External monitoring tool to take action | ONLINE
+|7| OFFLINE  | POWER ON request | Power ON Switch Host, Syslog | ONLINE
```
yxieca: Minor issue: We should callout that from OFFLINE state, any above 6 events would end up with switch host stay at OFFLINE state.

yxieca (dismissed):
Approved with a minor issue.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,746 @@
...
+IP address to be configured on the Switch-Host end and Switch-Bmc end can be defined sonic wide unique in file "files/image_config/constants/bmc.json" as below
+```
+{
+    "bmc_if_name": "bmc0",
+    "bmc_if_addr": "169.254.100.2",    #Address on Switch-Host end
+    "bmc_addr": "169.254.100.1",       # Address on the BMC end
+    "bmc_net_mask": "255.255.255.252"
+}
+```
+A platform could override this ip-address/netmask by defining similar details in the file bmc.json in the vendor/platform directory.
```
adamcalabrigo, line 236: What about IPv6 configuration support for the host-bmc link?

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: updated Ying - thx

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+* BMC can access Switch-Host redis DB over this internal Host-Bmc-Link and viceversa.
+* BMC will manage the Switch Host to support operations like soft reboot, power up/down, power-cycle, get operational status.
+* BMC and Switch-Host shall enable an independent Hw watchdog timer.
+* BMC and Switch-Host can be power ON and OFF independently.
+* BMC shall remain operational (UP) during system leak events, power or voltage faults affecting the host system, provided standby power rail remains available
+* Firmware upgrade of components is done on Switch-Host or BMC based on who owns the component and what needs a reboot.
+
+Liquid cooled sku requirements
+* BMC will manage leak detection, read local leak sensors, its severity and enforces mitigation actions according to system-wide SONiC policy.
+* BMC will get inputs from external Rack Manager on Inlet Liquid temperature, Inlet Liquid flow rate, Inlet Liquid Pressure and Rack level Leak. It takes action based on policy.
```
chinmoy-nexthop, line 70: Rack Manager can send alerts for inlet liquid temperature, inlet liquid flow rate, inlet liquid pressure, and rack-level leak. PMON is expected to handle these alerts based on a defined policy. Is this understanding correct?

chinmoy-nexthop, line 70: How are these policies defined? Are they static or dynamic?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+"Switch_Host" flag is set to 1 on the switch host, "Switch_BMC" flag is set to 1 on the switch BMC.
+
+
+#### 2.1.1 BMC platform power up
+When device is powered ON, the BMC powers first, boots up the sonic BMC which starts the various cointainers   
+
+If it is Air cooled switch the Switch-Host is powered on immediately.
+
+If it is liquid cooled, the following actions are done before the Switch-Host is powered on.
+* Check local system leaks and external Leaks if any reported by Rack Manager
```
chinmoy-nexthop: Is there any Rack Manager API call that the BMC should wait for before initiating a SwitchHOST reset-out?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+* Send a POWER_ON command to Switch-host if all clear. 
+
+![BMC platform power up](images/sonic-bmc-bootup.png)
```
chinmoy-nexthop, line 110: Should the BMC notify the Rack Manager if it doesn't power on the switch host for any reason?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+
+#### 2.1.2.1 DB schema
+
+Redis DB will be used to store the command/data send from external Rack manager for the platform daemons to act upon.
+
+Rack Manager command and state
+```
+key                       = RACK_MANAGER_COMMAND|CMD_<command_id>         ; Commands from Rack Manager in STATE_DB in BMC
+; field                   = value                                         ; e.g. ComputerSystem.Reset
+command                   = POWER_ON | POWER_OFF
```
chinmoy-nexthop: Currently, only POWER_ON and POWER_OFF commands are defined. Is this intentional?

There are other commonly used operations such as force-off, graceful shutdown, and power cycle. Should we consider including these as well?

otherwise, all of them should either maps to or or off .

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+
+
+key                       = RACK_MANAGER_STATE|rack-manager               ; STATE_DB on BMC to store rechability of RackManager
+; field                   = value
+reachability              = REACHABLE | UNREACHABLE
+last_change_timestamp     = STR
+
+```
+
+Rack Manager alerts, it is triggered when there is a CRITICAL/MAJOR/MINOR event in Rack
```
chinmoy-nexthop, line 167: How many alerts can Redis persist?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+```
+key                       = RACK_MANAGER_ALERT|Inlet_liquid_temperature    ; Alert data from Rack Manager in STATE_DB
+; field                   = value
+severity                  = status                                         ;CRITICAL/MAJOR/MINOR
+timestamp                 = STR
+
+key                       = RACK_MANAGER_ALERT|Inlet_liquid_flow_rate      ; Alert data from Rack Manager in STATE_DB
```
chinmoy-nexthop, line 174: Do we also need an Alert_ID to uniquely identify each alert?  An Action will be performed on An Alert so probably it is required , please think about it

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+* The usecase for this is to identify when the Critical alert is cleared. 
+* This telemetry data will not be streamed out.
+
+```
+key                       = RACK_MANAGER_DATA|Inlet_liquid_temperature    ; Telemetry data from Rack Manager in STATE_DB
+; field                   = value
+InletTemperature          = float
+unit                      = C
+severity                  = status                                         ;CRITICAL/MAJOR/MINOR/NORMAL
+timestamp                 = STR
```
chinmoy-nexthop, line 200: Why telemetry is having severity ? is it treated as alert as well ?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+* Switch-Host has thermalctld managing its thermal sensors and control the fan/cooling as done today.
+    
+### 1.2. BMC Platform Stack
+The SONiC in BMC interoperate with the SONiC in Switch-Host as in below diagram.
+ 
+![BMC Platform Stack](images/sonic-bmc-arch.png)
+
+## 2. Detailed Architecture and workflows
+### 2.1 BMC platform
+Update the vendor/platform/platform_env.conf with the following flags,
```
rtorrn: Consider changing to < vendor >/< platform >/platform_env.conf to make this clearer.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+```
+Switch_Host=1
```
rtorrn: I suggest 
switch_host=1
liquid_cooled=1

or 
SWITCH_HOST=1
LIQUID_COOLED=1 

to be consistent with other values for platform_env.conf today.
default_mtu=9100
disaggregated_chassis=1
disaggregatedT2=1
dmasize=128M
dmasize=32M
dmasize=512M
dmasize=64M
is_ltsw_chip=1
is_ltsw_chip=1is_ltsw_chip=1
is_ltsw_chip=1SYNCD_SHM_SIZE=1g
is_ltsw_chip=1SYNCD_SHM_SIZE=512m
linecard_reboot_timeout=180
macsec_enabled=1
rx_buffer_size=9238
supervisor=1
SYNCD_SHM_SIZE=1g
SYNCD_SHM_SIZE=1G
SYNCD_SHM_SIZE=1gb
SYNCD_SHM_SIZE=256m
SYNCD_SHM_SIZE=512m
usemsi=0
usemsi=1

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+Liquid_cooled=true
+```
+```
+Switch_BMC=1
+Liquid_cooled=true
+```
+
+"Liquid_cooled" flag is set to true on a liquid cooled switch.
```
rtorrn: Maybe describe what if a system is 5% air cooled, do we still set liquid_cooled=1?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+**Note: Redish docker to be enabled only on Liquid cooling platform.**
+
+Few of the URIs which needs to be supported in BMC are below,
+
+    1. GET /redfish/v1   
+             -- Rack Manager to get switchBMC type eg: "SONiCBMC"
+    2. GET /redfish/v1/UpdateService/FirmwareInventory  
+             -- Rack Manager to get switch firmware details
+    3. POST /redfish/v1/Systems/System/Actions/ComputerSystem.Reset  
+             -- Rack Manager to power off/on Main_cpu_switch_board
```
rtorrn: change main_cpu_switch_board to switch host  to be consistent with references in this document?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+|| Switch Host State (Start) | Event | Action | Switch Host State (Final) 
+|--|---|---|---|---|
+|1| ONLINE  | LOCAL_LEAK_CRITICAL_EVENT | Syslog, Power OFF Switch Host | OFFLINE 
+|2| ONLINE  | RACK_MGR_SHUTDOWN command | Syslog, graceful-shutdown Switch Host | OFFLINE
+|3| ONLINE  | CHASSIS_MODULE_admin_down user request | Syslog, graceful-shutdown Switch Host | OFFLINE
+|4| ONLINE  | RACK_MGR_CRITICAL_EVENT | Syslog this event and hot thermal sensors, External monitoring tool to take action | ONLINE
+|5| ONLINE  | RACK_MGR_MINOR_EVENT | Syslog, External monitoring tool to take action | ONLINE
+|6| ONLINE  | LOCAL_LEAK_MINOR_EVENT | Syslog, External monitoring tool take action | ONLINE
+|7| OFFLINE  | RACK_MGR_ POWERON command | Power ON Switch Host, Syslog | ONLINE
+|8| OFFLINE  | CHASSIS_MODULE_admin_up user request | Power ON Switch Host, Syslog | ONLINE
```
rtorrn: The events like LOCAL_LEAK_CRITICAL_EVENT look intiuitive, but it took me a while to figure out that CHASSIS_MODULE_admin_down user request means what I think it means ("user initiated shut down of switch host"). Consider adding definitions for all 8 events in this section.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+```
+A platform could override this ip-address/netmask by defining similar details in the file bmc.json in the vendor/platform directory.
+
+
+#### 2.1.4 BMC-Switch Host Interaction
+The Switch-Host and BMC communicate over the Host-Bmc-Link for accessing redis DB.
+
+BMC controls the State of the Switch-Host based on various factors/events. Defining the various events, the start, final states of Switch-Host here,
+
+|| Switch Host State (Start) | Event | Action | Switch Host State (Final)
```
rtorrn: Consider being explicit if the BMC and /or the switch host wil emit the syslog. I think we intend to just have the BMC emit the syslog related to the event by default.
Or do we want the switch host to also emit some additional syslog prior to going down (if OFFLINE is the final state), or to mirror the BMC syslog?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+##### 2.2.1.1 bmcctld on BMC
+
+The bmc controller daemon "bmcctld" is started first in BMC pmon container. It acts on the commands/alerts from External rack manager and local system leaks reported by 'thermalctld'.
+
+![bmcctld interactions](images/bmcctld.png)
+
+Detailed workflow below
+
+```
+Sleep for BMC_BOOTUP_TIMEOUT (this is configurable value in config_db)
```
rtorrn: Is it reasonable to set a BMC_BOOTUP_TIMEOUT of 0? I think the answer is yes. For initial provisioning from manufacturing line, selecting the appropriate default seems important. By the time the BMC is powered on, there should be liquid flowing, right?  And otherwise the switch host subsystem will have self protection mechanisms in the case of overheat.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+
+```
+
+![bmcctld events](images/bmcctld_events.png)
+
+###### DB schema
+This section covers the various tables which this daemon creates/uses in Redis DB on BMC
+
+```
+key                       = BMC_BOOTUP_TIMEOUT|default   ; Config DB on BMC
```
rtorrn: Seems like this should be called SWITCH_HOST_POWER_ON_MIN_DELAY or something else.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+; field                   = value
+boot_delay                = float                        ; Time in secs after power on the device, switch BMC can power on the Switch-Host. ( default = 5 min ).
```
rtorrn: consider adding units second, and setting default to 0 seconds, instead of 5 minutes.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+leaking                   = STR                                       ; Yes or No to indicate leak status
+leak_sensor_status        = STR                                       ; Is Leak sensor good or faulty.
+type                      = STR                                       ; leak sensor type
+location                  = STR                                       ; leak sensor location
+severity                  = "status"                                  ;CRITICAL/MINOR
+
+key                       = LEAK_PROFILE|<sensor_type>                ; LEAK profile per leak sensor type in CONFIG_DB
+; field                   = value
+leak_type                 = STR                                       ; Leak sensor type
+max_minor_duration_sec    = float                                     ; MAX-T mins defined before which a MINOR leak can be considered CRITICAL
```
rtorrn: field name says "_sec" but the description says "mins". Suggest to use seconds.
Maybe consider -1 as infinity, or have a way to indicate no limit for a minor leak to become critical. I heard that in some cases you could go many hours or longer before any action actually needs to be taken.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+This base class is already defined in sonic-platform-common. Additional new platform APIs defined.
+
+| Method | Present | Action |
+|---------|---------|----------|
+| get_name() | Y | Get leak sensor name |
+| is_leak() | Y | Is there a leak detected? **Applies debounce logic before reporting or clearing a leak** |
+| is_leak_sensor_ok() | New | Is the leak sensor OK or faulty ? |
+| get_leak_sensor_type() | New | What type of leak sensor is this rope, flex, spot etc |
+| get_leak_sensor_location() | New | Location of leak sensor |
+| get_leak_severity() | New | Get the severity based on the criticality of the zone |
```
rtorrn: Suggest to distinguish leak sensor CLASSIFICATION from system leak severity. Replace leak severity with "leak classification". So e.g. a certain rope may be MAJOR or CRITICAL (choose one or the other, not both),if there is flooding happening.  Another rope may be MINOR, if it indicates a localized leak only.
MAJOR Liquid is flooding inside the chassis
MINOR Localized leak detection


Leak sensor is classified as minor or major/critical (choose one)
System leak severity is low or high/critical

Something like above could reduce confusion when talking about each leak sensor ("how is it classified?") vs how bad the system leak is (is it low or high)?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+There is a new field named "Switch-Host Serial Number" which will show the Switch serial number
+
+```
+show version 
+
+.....
+ASIC Count: 1
+Serial Number: <BMC serial number>
+Switch-Host Serial Number: <Switch serial number>
+Model Number: <>
```
rtorrn: Should model number be system model number ?
e.g. (Switch model number)
Model Number: < switch product name >

or
e.g. BMC related model number
Model Number: SONiC BMC or SONiC-BMC

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+show chassis module status
+        Name             Description   oper status       Serial
+------------  ----------------------  -------------  -----------
+         BMC   Board Management Card         Online           <>
+ Switch-Host      Switch Host System        Offline           <>
+
+
+```
+
+* show platform leak control_policy
```
rtorrn: consider changing control_policy to control-policy or control policy . having underscore in  CLI in SONiC is somewhat rare per https://github.com/sonic-net/sonic-utilities/blob/master/doc/Command-Reference.md

asic-sdk-health-event
auto-techsupport
device-global
advertised-routes
received-routes
eeprom-hexdump
flowcnt-route
mgmt-vrf
memory-stats
... etc ...

vs
dhcp_relay
dhcp_server
management_interface
mirror_session
warm_restart

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+
+Enhance the reboot reason to include the shutdown/powercycle commands from BMC 
+Applicable to (LC, AC)
+
+```
+show reboot-cause history 
+Name                 Cause                                                                 Time                             User    Comment
+-------------------  --------------------------------------------------------------------  -------------------------------  ------  ---------
+2026_03_18_04_38_17  reboot                                                                Wed Mar 18 04:37:21 AM UTC 2026  admin   N/A
+2026_03_18_02_06_06  shutdown from BMC                                                     Wed Mar 18 02:05:12 AM UTC 2026  admin   N/A
```
rtorrn: consider changing the cause to the following to be more explicit.

graceful shutdown initiated by BMC
and power down initiated by BMC

instead of 
shutdown from BMC
power cycle by BMC

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+Firmware upgrade of components is done on Switch-Host or BMC (based on who owns the component, eg: Leak CPLD could be owned and upgraded from BMC).
+The upgrade process could be either during sonicimage-install or using fwutil cli.
+
+Below is a sample reference,
+
+| Firmware upgrade   |      Switch-Host reboot    |     BMC reboot     | Who does FW upgrade |
+|--------------------|----------------------------|--------------------|---------------------|
+| BMC Uboot          |         No                 |         Yes        |    BMC              |
+| CPU CPLD           |         Yes                |         Yes        |    Switch-Host      |
+| Leak CPLD FW       |         No                 |         Yes        |    BMC              |
```
rtorrn: In some cases both switch-host and BMC could do FW upgrade; it might be good to give an example. e.g.

In some designs, Leak CPLD FW could be done by either side - BMC or Switch Host.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,750 @@
...
+    - Apply the System leak severity detection algorithm as below
+       
+       +--------------------------------------+-----------------------------+
+       | Leak sensor data                     | System Local Leak Severity  |
+       +--------------------------------------+-----------------------------+
+       | 1 Critical leak                      |        CRITICAL             |
+       | 2 or more leaks with any Severity    |        CRITICAL             |
+       | 1 Minor leak stay leaking for MAX-T  |        CRITICAL             |
+       | 1 Minor leak detected                |        MINOR                |
+       +--------------------------------------+-----------------------------+
```
rtorrn: Consider having a configurable policy wherein at least 1 minor leak detection results in CRITICAL system leak severity. This could make lab setups safer especially during software bringup.

rtorrn (commented):
Hi Judy, thanks for the thorough writeup.

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: I can add the following ipv6 config @adamcalabrigo .. will check if this overlaps with anything existing 

```
{
  "bmc_if_name": "bmc0",
  "bmc_if_addr": "169.254.100.2",
  "bmc_addr": "169.254.100.1",
  "bmc_net_mask": "255.255.255.252",
  "bmc_if_addr_v6": "fd00:169:254:100::2",
  "bmc_addr_v6": "fd00:169:254:100::1",
  "bmc_prefix_len_v6": 127
}

```

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: yes policy is defined in section [2.2.1 BMC controller-bmcctld]

judyjoseph: Currently it is in section "[2.2.1 BMC controller-bmcctld]" and it is static .. but planning to update to a dynamic approach where user can control the action for various events via CLI/redis-db

doc/bmc/sonicBMC/pmon-bmc-design.md
rtorrn: Also, what's the implementation idea for reboot-cause on the X86 if gnoi is the mechanism to initiate a graceful shutdown?  How does X86 Switch-Host realize that this is from BMC as opposed to from an external device initiating the graceful shutdown?

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: There could be an explicit POWER_ON message from Rack manager, if not we have a timeout after which switch-host to be power-on assuming no critical leaks

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: No, rack manager will not be informed instead it will result in a POWER_ON event failure syslog - which triggers an alert raised for external tool to handle.

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: sure will update ..

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: telemetry data comes every 60sec from Rack-manger and the severity field is to represent the status for eg: if there is a Critical alert on this parameter eg: Inlet_liquid_temperature  or it is Normal.

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: The idea was to just keep the latest state of the parameter in the DB Assume if we get multiple events bmcctld should be able to drain them without losing any intermediate events and handle and apply policy based on final state ?

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: So this is a case where it is liquid cooled switch, but there is air cooling too with fans ?

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: thanks for comment will do

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: ok so one option is to have 2 minor leaks .. which can cause a Critical System Local Leak Severity

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: sure .. I am trying to update this control policy a bit , to make it more configurable. Will update the doc

doc/bmc/sonicBMC/pmon-bmc-design.md
rtorrn: Yes, for some designs there will be 5% air cooling for e.g. the CPU or the SSD but all other hot components are liquid cooled.

doc/bmc/sonicBMC/pmon-bmc-design.md
rtorrn: I'm saying in a lab set up, we might want to  have a customized  policy (e.g. "Lab Safe Policy") where any leak results in bmc shutting down the switch host .   It's a different policy entirely compared to the data above. In table format this Lab Safe Policy would have a different table
       +--------------------------------------+-----------------------------+
       | Leak sensor data                     | System Local Leak Severity  |
       +--------------------------------------+-----------------------------+
       | 1 Critical leak                      |        CRITICAL             |
       | 2 or more leaks with any Severity    |        CRITICAL             |
       | 1 Minor leak stay leaking for MAX-T  |        CRITICAL             |
       | 1 Minor leak detected                |        CRITICAL                | <<< ONLY DIFFERENCE
       +--------------------------------------+-----------------------------+

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: that is a good suggestion. If the BMC and Switch-Host share same model number, I feel we could display same model number at both places then.

doc/bmc/sonicBMC/pmon-bmc-design.md
chinmoy-nexthop: Nice, Thanks

doc/bmc/sonicBMC/pmon-bmc-design.md
chinmoy-nexthop: Looks good. I was just wondering—what would happen if bmcctld is down or unable to keep up with the alert flow rate? Perhaps we could add a note describing the expected behavior in such scenarios. Otherwise, everything looks good.

doc/bmc/sonicBMC/pmon-bmc-design.md
adamcalabrigo: Could we instead use the typical link-local IPv6 address space? Like:
```
  "bmc_if_addr_v6": "fe80::2",
  "bmc_addr_v6": "fe80::1",
  "bmc_prefix_len_v6": 64
```

judyjoseph: fe80::1 and fe80::2 are ipv6 link local .. Is there some restriction that we need to specify the interface scope when we use them ? @adamcalabrigo

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: bmcctld will be a critical process in pmon .. so alert will be raised if supervisor restart fails multiple times

doc/bmc/sonicBMC/pmon-bmc-design.md
rtorrn: Okay, I agree that Model Number: <switch product name> should be here.

However, we should consider the following implications:
- For ZTP running on BMC, the plan is to publish "SONiC-BMC" instead of <switch product name> as the product name under client ID (61). I think that the ZTP implementation will need to deviate in the implementation -- instead of taking some information that was inferred from the EEPROM content such as the <switch product name> , ZTP will have to publish just the string "SONiC-BMC"
- If the BMC is field-replaceable, then the traceability for the BMC for RMA or replacement is slightly harder -- vendors need to track the BMC SKU for replacement only via serial number which is printed in "show version".  I think that this is acceptable and in many cases the BMC will be integrated into the system so it won't be field replaceable anyway.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,746 @@
...
+
+#### 2.1.2 BMC - External Rack Manager Interaction
+The new docker container "redfish" in sonicBMC will have openbmc/bmcweb service which terminates the redfish calls from the external Rack Manager Node.
+
+**Note: Redish docker to be enabled only on Liquid cooling platform.**
+
+Few of the URIs which needs to be supported in BMC are below,
+
+    1. GET /redfish/v1   
+             -- Rack Manager to get switchBMC type eg: "SONiCBMC"
```
manapalai, line 121: Let us define:
How the BMC identifies itself as SONiCBMC (e.g. Managers/Bmc → Model, Oem.SONiC.BmcType = "SONiCBMC").

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,746 @@
...
+    2. GET /redfish/v1/UpdateService/FirmwareInventory
```
manapalai, line 122: Missing is a clear resource model that matches your firmware matrix:
At least:
BMC OS → FW_BMC_0 (SoftwareInventory)
(Optionally) BMC bootloader / iRoT, if you want them visible
Main board CPLD → CPLD_0
Leak/CSM CPLD → CPLD_LEAK_0 (or equivalent)

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,746 @@
...
+
+**Note: Redish docker to be enabled only on Liquid cooling platform.**
+
+Few of the URIs which needs to be supported in BMC are below,
+
+    1. GET /redfish/v1   
+             -- Rack Manager to get switchBMC type eg: "SONiCBMC"
+    2. GET /redfish/v1/UpdateService/FirmwareInventory  
+             -- Rack Manager to get switch firmware details
+    3. POST /redfish/v1/Systems/System/Actions/ComputerSystem.Reset
```
manapalai, line 124: Spell out which ResetType values you support (ForceOff, ForceOn, PowerCycle, GracefulRestart) and make sure that’s consistent with Platform API (SwitchHostBase methods).

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,746 @@
...
+Few of the URIs which needs to be supported in BMC are below,
+
+    1. GET /redfish/v1   
+             -- Rack Manager to get switchBMC type eg: "SONiCBMC"
+    2. GET /redfish/v1/UpdateService/FirmwareInventory  
+             -- Rack Manager to get switch firmware details
+    3. POST /redfish/v1/Systems/System/Actions/ComputerSystem.Reset  
+             -- Rack Manager to power off/on Main_cpu_switch_board
+    4. POST /redfish/v1/Managers/Bmc/Oem/SONiC/RackManagerInterface/Actions/SONiC.SubmitAlert
+             -- Rack manager to post a critical alert to BMC
```
manapalai: These cover the use cases, but you’re missing:

A concrete OEM schema for Oem.SONiC.RackManagerInterface, including:
Allowed properties in the SubmitAlert and SubmitTelemetry payloads (e.g. fields for inlet temp, flow, pressure, rack-level leak, severity).
Datatypes/units and severity enum (CRITICAL/MINOR).

Alignment with PMON DB schema:
RACK_MANAGER_ALERT|* and RACK_MANAGER_COMMAND|* tables in STATE_DB already define fields; the Redfish OEM actions should map 1:1 to those.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,746 @@
...
+    5. POST /redfish/v1/Managers/Bmc/Oem/SONiC/RackManagerInterface/Actions/SONiC.SubmitTelemetry
+             -- Rack manager send periodic telemetry data of Inlet Liquid temperature, Inlet Liquid flow rate,
+                Inlet Liquid Pressure, Leak information
+    6. POST /redfish/v1/EventService/Subscriptions  
+             -- Rack manager to subscribe for events like Leak from switchBMC.
+             -- Leak sensor can be modelled under /redfish/v1/Chassis/BMC/ThermalSubsystem/LeakDetection/LeakDetectors/<ID>
```
manapalai, line 139: LeakDetectors resource path:
You propose to model sensors as
/redfish/v1/Chassis/BMC/ThermalSubsystem/LeakDetection/LeakDetectors/<ID>
That needs to be written down as the authoritative resource model and aligned with DMTF/OCP guidance (property names, links, etc.).

Event payload:
OriginOfCondition should point to the specific LeakDetectors/<ID> resource.
Message ID / severity should tie to a Message Registry (DMTF or OEM) with clear semantics for:
Local leak vs rack-level leak
MINOR vs CRITICAL

Mapping from Redis:
LIQUID_COOLING_DEVICE|leakage_sensors{X} and LEAK_STATUS|local (PMON design) → LeakDetectors resources and Redfish events.
RACK_MANAGER_ALERT|* (PMON) → which LeakDetectors or OEM telemetry resources/events.

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: Yes, we are planning to keep the BMC_BOOTUP_TIMEOUT = 0 in the manufacturing image. This default BMC_BOOTUP_TIMEOUT = 5 min is based on scenario's of a rack level leak and the whole rack needs to be powered down. When powering on the rack and liquid supply there needs some time before liquid flow gets to normal

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: The idea behind the get_leak_severity() call is to determine whether any condition indicates a system-level critical leak. This decision can be based on:
**(i) Zone criticality:**
The physical location of the leak sensor. For example, if a spot leak sensor is located near critical components (ASIC/CPU) and the measured value (e.g., resistance or other leak-detection metric) crosses a platform-defined threshold, it should be treated as a critical leak.
**(ii) Extent of liquid detection:**
Even if the sensor is located farther from critical components (e.g., a rope/cable leak sensor), the severity may still be considered critical if the sensor indicates a large or expanding leak. For example, multiple segments of the cable detecting liquid or leak detection along a significant portion of the sensing cable.

```
show platform leak status
Name             Leak  Leak-sensor-status leak-sensor-type leak-severity
---------      ------- ------------------- ---------------- -------------
leak_sensors1    YES         OK                 rope           MINOR
leak_sensors2    NO          FAULTY          flex-pcb     NA
...
leak_sensors3    NO          FAULTY             spot           CRITICAL


```

Here leak severity just talks about how bad this particular leak sensor is, **should we flag as critical**.  We can just take three states for every leak sensor CRITICAL/MINOR/NO-LEAK

Let me know your feedback

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: I have updated the comments .. I believe in the gnoi halt handler we can update the reboot cause.

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: Maybe we can have an option for thermalctld to take an input json file defining policy in a particular format ?

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: updated in header to explicitly tell in BMC

doc/bmc/sonicBMC/pmon-bmc-design.md
rtorrn: This approach seems reasonable and has enough customization.

manapalai:
HW watchdog: user-space watchdog service needs to defined.

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: updated this as a future enhancement in doc

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: We will add ""Product": "SONiCBMC"" in the response to call "GET /redfish/v1" from Rack Manager

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: This API will just get back the inventory list following the DMTF spec

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: updated

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: yes this will map 1:1 .. we could get this added to the redfish document .. will that be ok ? @chinmoy-nexthop f.y.i

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: this again will map 1:1 .. we could get this added to the redfish document .. Hope that is fine @chinmoy-nexthop f.y.i

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: Have explicitly added CRITICAL_SYSTEM_LEAK and MINOR_SYSTEM_LEAK to differentiate from crititcal/minor severity per leak sensor. Hope it gives some more clarity

doc/bmc/sonicBMC/pmon-bmc-design.md
rtorrn: For "show version" the plan is to put <switch model number> for "Model Number".

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -657,31 +656,27 @@ config liquidcool leak-control [system|rack_mgr] [enabled|disabled]
...
 ```
-config liquidcool system-leak-action   critical       [syslog_only|graceful_shutdown|power_off]
-config liquidcool system-leak-action   minor          [syslog_only|graceful_shutdown|power_off]
-config liquidcool rack-mgr-leak-action critical-alert [syslog_only|graceful_shutdown|power_off]
-config liquidcool rack-mgr-leak-action minor-alert    [syslog_only|graceful_shutdown|power_off]
+config liquidcool leak-action [system|rack_mgr] [critical|minor]  [syslog_only|graceful_shutdown|power_off]
 
    - syslog_only      : Log the event; no Switch-Host power action taken.
    - graceful_shutdown: Issue a graceful GNOI shutdown to Switch-Host; force power-off after SWITCH_HOST_SHUTDOWN_TIMEOUT if unresponsive.
    - power_off        : Immediately power off Switch-Host via platform API module->set_admin_state(DOWN).
```
rtorrn, line 684: please add description of "enabled"

rtorrn (commented):
Thanks for disambiguating system vs individual sensor.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,861 @@
...
+
+####  ModuleBase
+This base class is already defined in sonic-platform-common.
+Switch-Host can be modelled as a Module object and the APIs to control power on/off/cycle the Switch-Host are as below,
+
+| Method | Present | Action |
+|---------|---------|----------|
+| set_admin_state(up=True) | Y | Power ON Switch Host from standby/off |
+| set_admin_state(up=False) | Y | Power OFF Switch Host |
+| get_oper_state() | Y | Fetch the operational state of Switch-Host|
```
chander-nexthop: Can you elaborate on get_oper_state()? BMC can bring the CPU of reset, but it has no way of knowing if the software and services are running in the switch card, unless we build a s/w handshake. Did you have anything specific in mind?

Also, do we need a platform API to issue a GNOI graceful shutdown/reboot or will that be taken care by bmcctld? IIUC it seems you are delegating this to bmcctld without the need for an explicit platform api. Please confirm

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,861 @@
...
+
+| Method | Present | Action |
+|---------|---------|----------|
+| get_name() | Y | Get leak sensor name |
+| is_leak() | Y | Is there a leak detected? **Applies debounce logic before reporting or clearing a leak** |
+| is_leak_sensor_ok() | New | Is the leak sensor OK or faulty ? |
+| get_leak_sensor_type() | New | What type of leak sensor is this rope, flex_pcb, spot etc |
+| get_leak_sensor_location() | New | Location of leak sensor |
+| get_leak_severity() | New | Get the severity based on the criticality of the zone or how sever is the leak for a sensor for eg: more liquid presence |
+| get_leak_profile() | New | Returns the leak sensor profile associated with this sensor |
```
Yakiv-Huryk: Can you please elaborate regarding leak profiles? Is it platform-specific and comes from the vendor's platform api implementation? Or is it configurable?

doc/bmc/sonicBMC/pmon-bmc-design.md
chinmoy-nexthop: Noted, Thanks.

doc/bmc/sonicBMC/pmon-bmc-design.md
chinmoy-nexthop: Agree. Yes @manapalai good point.  an oem-extention  feature will be aded to sonic-redfish to handle the same.

chinmoy-nexthop (commented):
Thank you, @judyjoseph, for addressing the review comments. I have no further comments from the sonic-redfish perspective. LGTM.

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: I am adding more details on this in doc. It is basically a profile created per leak sensor type rope, flex_pcb, spot etc . IT can have various APIs to get different leak sensor specific parameters/settings/thresholds which could vary between vemndor/platforms.

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: My understanding is for it to get the power ON/OFF state .. whether it is ONLINE/OFFLINE - since that is all which BMC is managing (switch-host power state ) .. @yxieca  please comment as well. 

Regarding GNOI .. yes it will be done in bmcctld context directly talking to gnmi server/port in switch-host

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: changed

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: updated, so if there is a critical system leak .. we do a power off using platform API

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: I have added a section for hybrid cooled sku's -- but will add more details into the next update of this doc. But yes the thermalctld on Switch-Host will need to run the thermal algorithm ( as it is done in sonic today ) and control the  specific air cooled component/fan module. @fraserg-arista

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: I am adding it to furture enhancements for now @adamcalabrigo will update in next version of doc - Hope that is ok

doc/bmc/pmon/pmon-bmc-design.md
judyjoseph: This is already used in certain platforms in code .. eg: files/image_config/interfaces/interfaces.j2, so continuing same netmask approach

judyjoseph:
Some how the reviewers were removed ! 

[adamcalabrigo](https://github.com/adamcalabrigo), [chander-nexthop](https://github.com/chander-nexthop), [chinmoy-nexthop](https://github.com/chinmoy-nexthop), [oleksandrivantsiv](https://github.com/oleksandrivantsiv) and [rtorrn](https://github.com/rtorrn) can you sign off .. or add any additional features which you would like ? I can add in the Future planned section

doc/bmc/sonicBMC/pmon-bmc-design.md
adamcalabrigo: Ok with me.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -0,0 +1,879 @@
...
+There is a new field named "Switch-Host Serial Number" which will show the Switch serial number
+
+```
+show version 
+
+.....
+ASIC Count: 1
+Serial Number: <BMC serial number>
+Switch-Host Serial Number: <Switch serial number>
+Model Number: <Switch model number>
```
oleksandrivantsiv, line 719: Can you please explicitly specify which API will be used to query the serial numbers?

judyjoseph: I have added the platform API's which could be used @oleksandrivantsiv

rtorrn (approved):
Looks good! Thank you.
Last suggestion -- consider changing liquidcool to liquid-cool for the config command. Otherwise this looks good.

Yakiv-Huryk (dismissed):

oleksandrivantsiv (dismissed):

yxieca (approved):

PR #2217, opened 2026-02-10
Add serverity and power control to leak detection HLD

fraserg-arista:
Posting for design discussion.

Adds severity information and power control operations (for the liquid-cooled power domain) to the leak detection high-level design. The two domains are major and minor, representing leaks that require immediate and delayed shutdown, respectively. 

The interface attempts to be generic enough to support different hardware implementation approaches, for both sensor technology and for whether power control is handled in hardware/firmware, software, or both.


fraserg-arista:
@judyjoseph I've opened this for design discussion for adding severity to leak detection.

doc/bmc/leakage_detection_hld.md
```diff
@@ -5,30 +5,56 @@
...
+2. For platforms that don't support liquid cooling at all, there should be no performance overhead.
+3. The API should be flexible enough to support different liquid cooling monitoring and response designs.
 
 ## 3. Detection and alarm flow
 The leak alarm process is straightforward. The platform API first acquires the status of the leak detection sensors. Then, thermalctld has a thread that calling the API and in turn notifies system health monitor, who ultimately sends out a gNMI event.
 
 ![LCflow chart](https://github.com/sonic-net/SONiC/blob/73f11eb7ad058b214d745e9ef728b8319574edbe/images/bmc/leakage_detection_flow.png)
 
+### 3.1 Leak severity
+Depending on the placement of a leak sensor within the chassis, a given leak event will represent a "major" or "minor"
```
fraserg-arista, line 18: As per the design [here](https://github.com/sonic-net/SONiC/pull/2215), these should be changed to "critical" and "minor".

PR #2219, opened 2026-02-11
Support PDB flows

yuazhe:
PDB, power distributed board, will be installed on direct-current platform and replace PSU. This HLD proposes several facilities to support monitoring status of this new kind of hardware. 

PR List:

| [sonic-buildimage] [Add pdb support for system-health](https://github.com/sonic-net/sonic-buildimage/pull/26829) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [sonic-buildimage] [Mellanox] [platform API for support pdb object](https://github.com/sonic-net/sonic-buildimage/pull/26830)| GitHub issue/pull request detail | GitHub pull request check contexts |
| [sonic-utilities] [Support new pdb object and remove redundancy error message](https://github.com/sonic-net/sonic-utilities/pull/4463) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [sonic-platform-common] [Add new pdb base object and corresponding functions to chassis base object](https://github.com/sonic-net/sonic-platform-common/pull/656) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [sonic-platform-daemons] [Add support for pdb object of psud](https://github.com/sonic-net/sonic-platform-daemons/pull/796) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [sonic-mgmt] [Test plan for PDB support](https://github.com/sonic-net/sonic-mgmt/pull/23028) | GitHub issue/pull request detail | GitHub pull request check contexts |


bmridul:
Why is PSUd not able to extend/serve the same functionality?  What is the daemon doing differently than PSUd? 

Note that the following issue was fixed which provided vendor platform to provide the name for the specific PSU (instead of an index). So you easily label the PSU as PDB if u wanted to.

https://github.com/sonic-net/sonic-platform-daemons/pull/446

bmridul:
Are PDBs FRUable entities. Do we need SNMP support for them?

doc/pdb/support_pdb_flows.md
```diff
@@ -0,0 +1,372 @@
...
+
+```
+
+## 4. New platform daemon for pdb
+New PDB daemon pdbd shall be created, it can be configured to skip on psu based platform.
+It must be enabled on pdb based platform.
+```
+pmon_daemon_control.json
+
+“skip_pdbd”: true/false
```
bmridul: The default should be disabled so that the daemon does not get enabled for all platforms. 

In that spirit, the name of the configuration parameter should be

enable_pdbd

yuazhe:
> Why is PSUd not able to extend/serve the same functionality? What is the daemon doing differently than PSUd?
> 
> Note that the following issue was fixed which provided vendor platform to provide the name for the specific PSU (instead of an index). So you easily label the PSU as PDB if u wanted to.
> 
> [sonic-net/sonic-platform-daemons#446](https://github.com/sonic-net/sonic-platform-daemons/pull/446)

we can do that, but IMO pdb is a kind of different hardware compared to psu(like no fan, not replaceable, no firmware on it to control ), so I prefer to have a separate daemon dedicated for it which looks more straightforward and intuitive.

yuazhe:
> Are PDBs FRUable entities. Do we need SNMP support for them?

No it can not be replaceable. We don't need SNMP for it.

doc/pdb/support_pdb_flows.md
yuazhe: Sounds good, I'm updating it.

doc/pdb/support_pdb_flows.md
```diff
@@ -0,0 +1,372 @@
+# Support PDB flows
+
+## 1. Requirments
+PDB, power distributed board, will be installed on direct-current platform and replace PSU. SONiC needs to provide several facilities to monitor status of this new kind of hardware. 
+This HLD proposes a new SONiC platform object, PDBobject,as the platform abstraction and corresponding small updates to the chassis base object. In addition to that, a new platform daemon, PDB daemon, which similar to PSU daemon will be provided, as well as new CLI command.
+
+## 2. New sonic-platform-common PDBobject APIs
+
+### 2.1 APIs inherited from Device Base object
```
jeffyin-dell: As discussed in the community review, consider extending the existing PSU platform APIs, or explain why that's not feasible. 

We also discussed the possibility of representing each PDB unit as a different subsystem of the hardware; for a traditional PSU, these could be separate power rails. This is of course dependent on each platform vendor's HW design.

bmridul: +1

doc/pdb/support_pdb_flows.md
```diff
@@ -0,0 +1,372 @@
...
+maximum_temperature              = FLOAT                                ; maximum recorded temperature (C)
+critical_threshold               = FLOAT                                ; temperature critical high threshold (C)
+timestamp                        = STRING                               ; timestamp for the temperature reading
+```
+
+## 7. CLI Command 
+
+![CLI data flow](https://github.com/yuazhe/SONiC/blob/9e9b4c2cd6feba0706112f3865feef5498a6c759/images/pdb/CLI_data_flow.png) 
+
+### 7.1 show platform pdbstatus
```
krishgobinath: Since `pdbstatus` replaces `psustatus` functionally on this platform, do you think we can reuse the existing customer-facing PSU APIs and CLI (e.g., show platform psustatus) to keep power reporting consistent across platforms?

doc/pdb/support_pdb_flows.md
yuazhe: @jeff-yin does the pdb on your guy's platform have fan or not?

jeffyin-dell: Yes, we call them "DC PSUs" here and they have fans and are field-removable.

doc/pdb/support_pdb_flows.md
yuazhe: agreed, will adopt the existing command line to give a consistent user experience

doc/pdb/support_pdb_flows.md
yuazhe: @lguohan  @jeff-yin  @bmridul  we agreed to re-use the current psu daemon and corresponding CLI. But regarding the object, I suggest to inherit a PDB object from PSU object, it will be easier to manage PDB specific attribute, if any. And also avoid to name a PSU object to "PDB XXX", which is a little bit confuse. Let me know what you think, thanks!

jeffyin-dell: Thanks @yuazhe. You will really need to spell out the reasons and problems you're trying to solve with this new object, rather than just extending the existing PSU base class.

For instance, the PSU class can just be extended to provide a `get_type()` API that returns AC or DC:
https://github.com/sonic-net/sonic-buildimage/blob/master/platform/broadcom/sonic-platform-modules-dell/z9664f/sonic_platform/psu.py#L275

And then `psud` could be extended to read that type and add the AC/DC type into the DB.

doc/pdb/support_pdb_flows.md
```diff
@@ -0,0 +1,372 @@
...
+
+        Returns:
+            A boolean, False since PDB is generally not replaceable.
+        """
+
+```
+
+### 2.2  APIs new for PDB object
+```
+    def get_input_current(self):
```
jeffyin-dell: These are already supported in `psu_base.py` right?
https://github.com/sonic-net/sonic-platform-common/blob/master/sonic_platform_base/psu_base.py#L269

yuazhe: right, this is what I intend to do, inherit it from psu base.

doc/pdb/support_pdb_flows.md
yuazhe: Thanks @jeff-yin, the pdb hardware can be like psu or not, depends on vendor specific HW design, so I want to make a distinguish here. For example, on our platform the pdb has no eeprom/fan/led, not even a FRU, it's simply a dummy board. I'm aware of that on your platform it more likely to psu. But given there can exist such differences, I prefer to have a sperate object here for clarity and easier to management.

anilpannala:
Please find the community HLD review recording here - https://zoom.us/rec/share/WrT1V9VKlIQiEKWwBFVIBW_U_JwoHIHuYA-Km3Qs2mnwYjUjEPaq5bQqb_75sBC8.PGu7eKEV_e2580z8

doc/pdb/support_pdb_flows.md
bmridul: HI Yuanzhe

Few questions..

1. In the document you mention that status and presence is monitored every 3 seconds. If the device is not Fruable (ie. cannot be replaced in the field) why do we need to monitor its presence? What should be done if the status is NOT OK. Is the device operational in such a scenario? Does it need to be returned back to vendor?

2. If PDB is a simplified form of PSU, then you can provide the implementation for supported fields and return the others as N/A or dummy values. I have already mentioned how you can name the PDB modules appropriately in a previous comment. 

3. If PDB is really like an always present daughterboard with some sensors, then do we really need to monitor its status? You can monitor the temperature, voltage and current sensors using thermalctld and sensormond daemons, which are more suitable to on-board sensors' monitoring requirements.


thanks

yuazhe: @bmridul thanks for the time for review and comments, I appreciate your effort on this.
1. The not FRU is not stand any more, as dell comments on their platform the pdb can be replaced. The presence checking is to make sure the device is being detected at all, if hardware issue happens, the pdb is possible to be lost.
2. I understand your point, technically there is no berries prevent us not to do so. But, as pdb is essentially a different hardware, in our platform, the pdb hardware has many differences which I mentioned above. So name a psu object to pdb is a trick to obscure this hardware difference and I prefer not.
3. As I replied in point 1, this board can be unoperational, and there can be mulitple pdbs so one of them failed may not cause the entire system goes down, so we do need to monitor it. And this thing is still under power scope so I don't want to mix it with other normal sensors object.

bmridul: Responding to #2

Most PSUs from different vendors have differences - in types (AC/DC), fans, number of sensors, redundancy models etc. PSUd is the SONiC way of handling these devices. It is a superset as it has to cater to all these devices with different capabilities. If we create a new daemon for every device which is little bit different then we will have a lot of daemons doing mostly similar things. 

My recommendation is : Provide a PSUd platform implementation for your device. If your device has some new feature which is not supported by PSUd, consider extending PSUd for that.

yuazhe: correct, I have already aligned the hld to reuse psud and its CLI, I'm just arguing there should still exist a pdb base object.

bmridul: I think that might be ok. It would be useful if you can send a code PR and update the document.
Your platform implementation can provide PDB base object types to PSUd and that could work.

liat-grozovik:
@yuazhe  please add list of code PRs into the PR description and please work to get approval for them as this is part of 202605

yuazhe:
> @yuazhe please add list of code PRs into the PR description and please work to get approval for them as this is part of 202605

PRs had been added accordingly.

bmridul:
> > @yuazhe please add list of code PRs into the PR description and please work to get approval for them as this is part of 202605
> 
> PRs had been added accordingly.

@yuazhe If PDB is a type of PSU (as discussed in earlier comment), why do we need changes in Chassisd. 
You can provide the PDB objects (instances of PDB class that is derived from PSU class) and information in get_psu, get_num_psu, get_all_psu APIs from the platform. That will avoid/reduce any changes in chassis_base.py, psud, system_health etc. Display of PDB name can be supported using the PR I mentioned earlier.. - https://github.com/sonic-net/sonic-platform-daemons/pull/446/changes

Happy to discuss in chassis meeting or otherwise if needed.

yuazhe:
> > > @yuazhe please add list of code PRs into the PR description and please work to get approval for them as this is part of 202605
> > 
> > 
> > PRs had been added accordingly.
> 
> @yuazhe If PDB is a type of PSU (as discussed in earlier comment), why do we need changes in Chassisd. You can provide the PDB objects (instances of PDB class that is derived from PSU class) and information in get_psu, get_num_psu, get_all_psu APIs from the platform. That will avoid/reduce any changes in chassis_base.py, psud, system_health etc. Display of PDB name can be supported using the PR I mentioned earlier.. - https://github.com/sonic-net/sonic-platform-daemons/pull/446/changes
> 
> Happy to discuss in chassis meeting or otherwise if needed.

Thanks for the comment, as my understanding the pdb is a different hardware type compared to psu, in our platform, the pdb hardware has many differences such as no fan, not replaceable, no firmware on it to control. So I prefer to have dedicated object here and corresponding chassis functions, stated in the HLD. I want to emphasize that pdb is NOT identical with psu.

keboliu (approved):

Sourabh-Kumar7:
@vvolam @judyjoseph can you pls help review this?

rookie-who:
Hi @yuazhe — this is approved by @keboliu. Is there anything blocking the merge? 202605 branch cut is at end of May.

yuazhe:
@liat-grozovik  would you please help to merge this PR, thanks

PR #2223, opened 2026-02-21
HLD: Front Panel Port LED Policy

justin-oliver:
Add a HLD describing a new policy for front panel LEDs. This proposal defines consistent behavior for port LEDs and moves the LED state/color logic out of the platform layer and into LEDD daemon. The proposal covers breakout scenarios and how to map subports to 1 or more port LEDs.

### DIFF
--- a/doc/pmon/port-led-policy.md
+++ b/doc/pmon/port-led-policy.md
@@ -1,0 +1,699 @@
+# Port LED Policy #
+
+## Table of Content
+
+### 1. Revision
+ | Rev |     Date    |       Author                  | Change Description     |
+ |:---:|:-----------:|:-----------------------------:|------------------------|
+ | 0.1 | 02/17/2026  |  Justin Oliver, Aaron Payment | Initial version        |
+
+### 2. Scope
+This document describes an updated LED policy for front panel port LEDs.
+
+### 3. Definitions/Abbreviations
+
+| Term | Definition |
+|:-----|:-----------|
+| SFP | Small Form-factor Pluggable |
+| QSFP | Quad Small Form-factor Pluggable |
+| OSFP | Octal Small Form-factor Pluggable |
+| Front Panel Port Cage | A slot where a transceiver can be inserted |
+| LEDD | LED Daemon responsible for managing port LED states |
+
+### 4. Overview
+
+SONiC's current LED policy implementation uses the deprecated Platform API V1 and has several issues with ambiguous behavior, particularly around port breakout scenarios where a single physical port is split into multiple logical interfaces. The LED state determination is inconsistent across different SKUs and platform vendors.
+
+This design proposes a new LED policy that:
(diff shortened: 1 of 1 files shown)


doc/pmon/port-led-policy.md
```diff
@@ -0,0 +1,687 @@
...
+```
+
+**Description:**
+Displays the current LED status for interfaces. If an interface name is provided, shows only that interface. Otherwise, shows all interfaces with controllable LEDs.
+
+**Example Output:**
+```
+admin@sonic:~$ show interface led status
+
+Interface Name  Port Index  Port Name  Alias   LED Index  LED Name  LED Color
```
Blueve: This is referred to `sfp_name`. Should we use an aligned column name here?

justin-oliver: Yes that's right. Updated.

doc/pmon/port-led-policy.md
```diff
@@ -0,0 +1,699 @@
...
+- `sonic-platform-daemons/sonic-ledd/scripts/ledd`: Enhanced LED policy logic
+- `sonic-utilities/show/interfaces.py`: New port LED status command
+
+#### 7.3. LED Policy Details
+
+##### 7.3.1. LED-to-Interface Mapping
+
+A front panel port will have 0 to N LEDs associated with it. The LED policy will evenly split the logical interfaces of a front panel port across the available LEDs. If the number of LEDs exceeds the number of logical interfaces, the extra LEDs will remain off.
+
+For example, an OSFP port with 2 LEDs:
```
pfallon-nexthop, line 178: Can you please clarify what would be the behavior for an OSFP port with 2 LEDs if a QSFP Adaptor is inserted?

In this case there would only be 4 usable lanes. Based on the stated policy, I would expect the two LEDs to split the interfaces on the first 4 lanes between them. However, this means the physical lanes the LED pertains to for status will change with an adaptor, which could add some complexity and may be confusing.

pfallon-nexthop (commented):
Thanks Justin! I just had a question for the policy when it comes to adaptors.

doc/pmon/port-led-policy.md
justin-oliver: Hi Paul! The current plan is to distribute the usable lanes across the LEDs. As you pointed out, the mapping will be somewhat dynamic when adaptors are used but I hope the new CLI command makes it very clear which lanes map to which LEDs.

pfallon-nexthop: Thanks Justin! That seems reasonable to me. I think it would be helpful to add one example along these lines.

bingwang-ms:
@judyjoseph Can you help review?

justin-oliver:
I've create https://github.com/sonic-net/sonic-platform-common/pull/662 which has the platform API changes to support the new policy.

PR #1599, opened 2024-02-04
[bfd] BFD HW offload for BGP sessions

baorliu:
BGP can be configured to use BFD to monitor it's neighbor reachability. when the BFD session becomes DOWN, it indicates that BGP neighbor become not reachable, so BGP can updates the corresponding route quickly based on BFD state, the detection can be much faster than the detection using BGP keep alive approach. 
Currently, BGP can use frr/bfdd (a software BFD) to do this.
The proposed design let BGP use hardware offloaded BFD in SONiC, it provide a faster detection, less CPU load, and supports more BGP sessions to use BFD.

Repo | PR title | State
-- | -- | --
sonic-swss[master] |[BFD hardware offload for BGP sessions](https://github.com/sonic-net/sonic-swss/pull/4253) |  GitHub issue/pull request detail|

PR #2225, opened 2026-02-23
BFD syncd HLD

rajshekhar-nexthop:
Today, SONiC effectively has two mostly disconnected BFD planes:
     • Software BFD (FRR `bfdd`) – CPU‑based, integrated with routing protocols via ZAPI.
     • Hardware BFD (`bfdorch` / SAI) – ASIC‑offloaded per the BFD HW Offload HLD, providing fast (<10 ms) detection. 
Apart from the static‑route BFD pipeline, FRR routing protocols (BGP/OSPF, etc.) do not consume HW‑offloaded BFD directly: they only see software BFD state, while HW BFD lives in a separate plane and telemetry model. The proposed BFD‑Syncd design proposes a bridging component that synchronizes HW‑offloaded BFD sessions and state into FRR bfdd (running in data plane mode). This would give SONiC a single, unified BFD view while still leveraging hardware offload.

### DIFF
--- a/doc/bfd/BFD_Syncd_HLD.md
+++ b/doc/bfd/BFD_Syncd_HLD.md
@@ -1,0 +1,479 @@
+# BFD-Syncd: FRR and Hardware BFD Integration
+## High Level Design Document
+### Rev 1.0
+
+# Table of Contents
+
+  * [Revision](#revision)
+  * [About this Manual](#about-this-manual)
+  * [Scope](#scope)
+  * [Definitions/Abbreviations](#definitionsabbreviations)
+  * [1 Requirements Overview](#1-requirements-overview)
+  * [2 Architecture Design](#2-architecture-design)
+  * [3 Design Details](#3-design-details)
+  * [4 Database Schema](#4-database-schema)
+  * [5 SAI API](#5-sai-api)
+  * [6 Configuration and Management](#6-configuration-and-management)
+  * [7 Warm Restart](#7-warm-restart)
+  * [8 Restrictions and Limitations](#8-restrictions-and-limitations)
+  * [9 Implementation](#9-implementation)
+  * [10 Testing Requirements](#10-testing-requirements)
+  * [11 References](#11-references)
+
+# Revision
+
+| Rev  | Date       | Author             | Change Description                          |
+|:----:|:----------:|:------------------:|---------------------------------------------|
+| 0.1  | 2026-02-23 | Rajshekhar Biradar | Initial version                             |
+
+# About this Manual
+
+This document provides the high-level design for **bfd-syncd**, a new daemon that bridges FRR's BFD daemon (bfdd) with SONiC's hardware-offloaded BFD (BFDOrch). This enables routing protocols running in FRR to leverage hardware-accelerated BFD for fast failure detection.
+
+# Scope
+
+This document covers:
+- Integration of FRR bfdd with SONiC BFDOrch via the Distributed BFD protocol
+- Synchronization of BFD sessions between FRR and SONiC databases
+- State propagation from hardware BFD to routing protocols
+
+This document does not cover:
+- Modifications to BFDOrch or SAI BFD APIs (uses existing implementation per [BFD HW Offload HLD](https://github.com/sonic-net/SONiC/blob/master/doc/bfd/BFD%20HW%20Offload%20HLD.md))
+- FRR bfdd internals (see [FRR BFD Documentation](https://docs.frrouting.org/en/latest/bfd.html))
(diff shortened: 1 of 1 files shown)


baorliu:
This "BFD HW offload for BGP sessions" PR (raised in 2024) is already in review:
https://github.com/sonic-net/SONiC/pull/1599

rajshekhar-nexthop:
> This "BFD HW offload for BGP sessions" PR (raised in 2024) is already in review:
> #1599

Agree, #1599 is the main prior work for BFD HW offload for BGP and is already under review. Our intent with v0.2 HLD is to collaborate and extend it into a more complete system design (covering link‑local, convergence, unified/static‑route usage, partial offload, and restart behavior etc).  We would be happy to work with you and other authors to align on these efforts and make sure we end up with a single, coherent design.

sarabjeetsingh1509:
Hi Rajshekhar
As community is planning to move to "unified" mode for FRR, what is the plan to accomodate that?
right now in frrcfgd, I see there is a '[BFD_PEER](https://github.com/sonic-net/sonic-buildimage/blob/53eceeca4a2a1c2454d189b0017ec826c9608ac8/src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py#L104)' table.
In order to make this persistent, are we going ahead with the APP_DB table design (BFD_SESSION_TABLE) and creating the table with same name in CONFIG_DB? or we are taking the table format from frrcfgd.py (BFD_PEER) and changing the swss to handle the same?

rajshekhar-nexthop:
> Hi Rajshekhar As community is planning to move to "unified" mode for FRR, what is the plan to accomodate that? right now in frrcfgd, I see there is a '[BFD_PEER](https://github.com/sonic-net/sonic-buildimage/blob/53eceeca4a2a1c2454d189b0017ec826c9608ac8/src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py#L104)' table. In order to make this persistent, are we going ahead with the APP_DB table design (BFD_SESSION_TABLE) and creating the table with same name in CONFIG_DB? or we are taking the table format from frrcfgd.py (BFD_PEER) and changing the swss to handle the same?

Hi @sarabjeetsingh1509 ,
Thanks for the review. Yes, we do intend to support unified mode for FRR. Today, community SONiC doesn't have a dedicated BFD peer YANG model — BFD enablement is implicit through BGP neighbor and static route configuration. While the frr-mgmt-framework's frrcfgd does have BFD_PEER tables in CONFIG_DB, these lack a formal YANG schema backing them. We are discussing introducing a cleaner profile-based YANG model as part of the unified architecture. Also we plan to keep the two planes separate. The BFD_PEER tables in CONFIG_DB handle operator intent and persistence, while APPL_DB:BFD_SESSION_TABLE remains the hardware programming path. The flow is: CONFIG_DB → frrcfgd → FRR bfdd → bfd-syncd → APPL_DB → BFDOrch → SAI. We're not duplicating BFD_SESSION_TABLE into CONFIG_DB or changing SWSS to consume BFD_PEER directly.

We're currently in a focused side-track discussion to get this HLD into the right shape. Once the key architectural decisions are solid, we'll plan a wider community review.

doc/bfd/BFD_Syncd_HLD.md
```diff
@@ -0,0 +1,1683 @@
...
+    }
+  }
+}
+```
+
+**Note:** The `when` constraint restricts this leaf to the `bgp` feature entry only. Without it, the leaf would be syntactically valid for any feature (e.g., `FEATURE|swss`), which is meaningless.
+
+### 6.3.2 BFD_SESSION_TABLE — new link-local fields
+
+The `encap_type`, `src_mac`, and `dst_mac` fields added to `APPL_DB:BFD_SESSION_TABLE` (§4.4) extend the existing BFD session YANG model (`sonic-bfd.yang`):
```
yuezhoujk, line 1436: I have two questions:

1. My understanding is that YANG validation is performed against ConfigDB. Does it also apply to BFD_SESSION_TABLE in AppDB?
2. Is this sonic-yang file newly added in this PR?

rajshekhar-nexthop: 1. Yes , YANG validation in sonic-yang-models is CONFIG_DB-only. Text here is little misleading and needs fix.
2. sonic-bfd.yang is not present in sonic-yang-models today

rajshekhar-nexthop:
Since https://github.com/sonic-net/SONiC/pull/1599
 has been merged and this PR has overlapping content, I’m closing this PR. If needed, I will create a separate PR for enhancements.

PR #2229, opened 2026-02-24
CoPP: Add COPP_TRAP_EXCLUDE_PORTS documentation

bhouse-nexthop:
Keep HLD up to date with new enhancements.  This adds the information about the new COPP_TRAP_EXCLUDE_PORTS to maintain a list of ports to exclude from CoPP Traps.

This HLD update is implemented by:
* https://github.com/sonic-net/sonic-swss/pull/4234
* https://github.com/sonic-net/sonic-buildimage/pull/25571

Fixes https://github.com/sonic-net/sonic-buildimage/issues/25570

### DIFF
--- a/doc/copp/CoPP Config and Management.md
+++ b/doc/copp/CoPP Config and Management.md
@@ -10,6 +10,7 @@ The following are the high level requirements for CoPP/TRAP managment
 2. User shall be able to configure, modify CoPP table entries for various policing requirements
 3. System shall be backward compatible with old Config DB without any CoPP Tables
 4. A VS test must be added to verify the Config DB CoPP tables
+5. Ports shall be able to be excluded from traps and configured in Config DB.
 
 # Design Proposal
 
@@ -51,6 +52,15 @@ key = "COPP_TRAP|name"
 trap_ids      = name ; list of trap ids
 trap_group    = name ; copp group name
 ```
+```
+key = "COPP_TRAP_EXCLUDE_PORTS|GLOBAL"
+; field        = value
+ports          = port-list ; ports that are excluded
+
+; value annotations
+port-name      = 1*64VCHAR                         ; name of the port
+port-list      = port-name [ 1*( "," port-name ) ] ; list of the ports. Valid values range is platform dependent
+```
 
 ### StateDB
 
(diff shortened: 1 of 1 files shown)


prsunny:
@rminnikanti for review

prsunny:
@bhouse-nexthop , can you add a section for test plan and also link the sonic-mgmt test PR for this once available

rminnikanti:
@bhouse-nexthop , to avoid having packets trapped on the fanout, you can update the trap_action in copp_cfg.j2 from "trap" to "forward". So, that the fanout becomes only a transient node and simply forwards traffic that matches the trap.

For documentation purposes, could you also share the sonic-mgmt VXLAN tests you referenced in the GitHub issue?

bhouse-nexthop:
@rminnikanti 
> @bhouse-nexthop , to avoid having packets trapped on the fanout, you can update the trap_action in copp_cfg.j2 from "trap" to "forward". So, that the fanout becomes only a transient node and simply forwards traffic that matches the trap.
> 
> For documentation purposes, could you also share the sonic-mgmt VXLAN tests you referenced in the GitHub issue?

That's the point, you *can't* disable traps globally in some environments such as when using VXLAN between the testbeds and fanout, you still need the traps on some ports.

Lets say ports Ethernet0 and Ethernet8 are routed ports on the fanout participating in a BGP Unnumbered VXLAN-EVPN.  So you have BGP traffic on ports Ethernet0 and Ethernet8 and the fanout is acting as a VTEP (using a loopback ipv4 address). 

And now you have 8 of the fanout ports connected to the DUT front panel ports you are testing, so you then map VNIs from your testbed of 10001001, 10001002, 10001003, 10001004, 10001005, 10001006, 10001007, 10001008 to VLANs on the fanout 1001, 1002, 1003, 1004, 1005, 1006, 1007, 1008, respectively.  

Then you associate ports: Ethernet16 VLAN 1001, Ethernet24 VLAN 1002,  Ethernet32 VLAN 1003, Ethernet40 VLAN 1004, Ethernet48 VLAN 1005, Ethernet56 VLAN 1006, Ethernet64 VLAN 1007, and Ethernet72 VLAN 1008.

If you aren't trapping globally, this won't work.  We still want traps on Ethernet0 and Ethernet8, but not on 16, 24, 32, 40, 48, 56, 64, and 72.

PR #2230, opened 2026-02-25
HLD for Northbound ZMQ

inder-nexthop:
Add HLD for fpmsyncd to orchagent northbound ZMQ path

This commit introduces the High-Level Design (HLD) document for the northbound ZeroMQ (ZMQ) communication path between fpmsyncd and orchagent. 

The HLD details the following key areas:
* Architecture: Bypassing Redis pub/sub for route updates via a local
  TCP ZMQ socket while maintaining asynchronous Redis writes for state
  persistence.
* Configuration: Documentation of the new CONFIG_DB and YANG parameter
  `orch_northbound_route_zmq_enabled`.
* Lifecycle Management: Expected behavior and system interactions during
  warm restart and warm boot scenarios.
* Benchmarks: Route programming performance metrics comparing the Redis
  and ZMQ paths across the 202505 and master branches.


doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
```diff
@@ -0,0 +1,153 @@
...
+
+If the scenario described in the earlier section (i.e fpmsyncd restarts between 2(b) and 2(c)) happens around a warm boot, the new version of ASIC\_DB would not have the route, while the old version (and the actual ASIC itself) would have the route. The syncd process would then find a “diff” between the old and new version and would appropriately remove the route from the ASIC. 
+
+So in a nutshell this scenario is not an issue for Warm Boot. It is only a problem for a process crash/restart.
+
+## Download Speed Results
+
+Route programming speed with and without ZMQ for various route table sizes:
+
+| Build | Route Count | Without ZMQ | With ZMQ | Improvement |
```
eddieruan-alibaba, line 146: Can. you provide test results with Multi DB and ring buffer enabled for without ZMQ case?

| master | 500K | 337.576s | 292.593s | 1.15x |

This is less than 2K routes per second, which is very low.

Which platform and which chip do you use? 

Also can you collect the log for the following performance measurements ?

https://github.com/sonic-net/SONiC/blob/master/doc/bgp_loading_optimization/bgp-loading-optimization-hld.md#how-to-analyze-the-performance-related-logs

doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Key characteristics
+
+- **fpmsyncd** receives netlink route messages from FRR's zebra over the FPM TCP connection. It parses these messages in `RouteSync::onMsg()` and writes them into Redis APPL\_DB using `ProducerStateTable`.  
+    
+- **orchagent** reads from Redis APPL\_DB using `ConsumerStateTable`. The Redis notification mechanism (keyspace events) wakes orchagent's event loop when new route entries arrive.  
+    
+- Every route update involves two Redis round-trips: the producer writes the entry, then Redis notifies the consumer. At scale (hundreds of thousands of routes), this pub/sub overhead becomes a bottleneck for route convergence time.
+
+## How It Works With ZMQ (Northbound)
```
haussli, line 57: Northbound is a meaningless term without context.  And, it still is not clear to what it refers.

doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
```diff
@@ -0,0 +1,155 @@
...
+
+If the scenario described in the earlier section (i.e fpmsyncd restarts between 2(b) and 2(c)) happens around a warm boot, the new version of ASIC\_DB would not have the route, while the old version (and the actual ASIC itself) would have the route. The syncd process would then find a “diff” between the old and new version and would appropriately remove the route from the ASIC. 
+
+So in a nutshell this scenario is not an issue for Warm Boot. It is only a problem for a process crash/restart.
+
+## Download Speed Results
+
+Route programming speed with and without ZMQ for various route table sizes:
```
haussli, line 144: Is this actual completion of the change on-chip or just placed in the queue?

doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
venkit-nexthop: > Can. you provide test results with Multi DB and ring buffer enabled for without ZMQ case?
> 
> | master | 500K | 337.576s | 292.593s | 1.15x |
> 
> This is less than 2K routes per second, which is very low.
> 
> Which platform and which chip do you use?
> 
> Also can you collect the log for the following performance measurements ?
> 
> https://github.com/sonic-net/SONiC/blob/master/doc/bgp_loading_optimization/bgp-loading-optimization-hld.md#how-to-analyze-the-performance-related-logs

In master (over 202505) we observed a significant degradation of performance. We need to investigate where that is coming from. But the speed is high on 202505
Thanks for perf measurement pointers. We will try to collect more info.

doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
venkit-nexthop: There is context in the Introduction section in the very beginning of the doc. This is not a term we came up with. This was done in an earlier effort described in this HLD => https://github.com/sonic-net/SONiC/blob/b578f6c1a5135d60c087427d8c421df9dc670e05/doc/orchagent_zmq/Improve_route_performance_with_zmq.md

doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
venkit-nexthop: The test ensures that the route count matches in the broadcom shell output, which queries the actual chip.

doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
```diff
@@ -0,0 +1,155 @@
...
+
+| Build | Route Count | Without ZMQ | With ZMQ | Improvement |
+| :---- | :---- | :---- | :---- | :---- |
```
mike-dubrovsky, line 147: Could you add DUT description (asic / CPU / RAM / model) ?

doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
```diff
@@ -0,0 +1,155 @@
...
+| 202505 | 100K | 45.626s | 13.401s | 3.40x |
+| 202505 | 250K | 162.503s | 49.563s | 3.28x |
+| 202505 | 500K | 491.928s | 92.719s | 5.31x |
+| master | 100K | 55.070s | 25.295s | 2.18x |
```
mike-dubrovsky, line 151: could you add sonic-buildimage sha or the date when the master was pulled ?

doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
```diff
@@ -0,0 +1,155 @@
...
+| master | 250K | 166.396s | 97.712s | 1.70x |
+| master | 500K | 337.576s | 292.593s | 1.15x |
+
+The route programming performance data in this document was collected using an internal benchmarking script. To ensure these measurements are reproducible and to make the tooling available to the community, we have opened a PR to upstream this script to sonic-mgmt ([PR \#22418](https://github.com/sonic-net/sonic-mgmt/pull/22418)).
```
mike-dubrovsky, line 155: there is also testplan 

https://github.com/sonic-net/sonic-mgmt/blob/master/docs/testplan/BGP-Convergence-Testplan-for-Benchmark-Performance.md

and the script 

https://github.com/sonic-net/sonic-mgmt/blob/0655d4408c7f01e0d5247e45afd15aced39d6a31/tests/snappi_tests/bgp/test_bgp_rib_in_convergence.py#L15

Maybe we can mention this info.

doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
```diff
@@ -0,0 +1,155 @@
...
+    description "Enable ZMQ feature on APPL_DB ROUTE tables.";
+    default "false";
+}
+```
+
+## The Restart Issue
+
+The fpmsyncd could crash/restart at any time. Normally this is not a problem because the restart of the fpmsyncd also causes the entire bgp container to restart. This would typically result in the control plane protocols relearning the routes and re-inserting them towards fpmsyncd. This will result in fpmsyncd writing the same routes into APP\_DB and towards orchagent.
+
+But it is possible that when the control plane protocols relearn the routes, one or more of the previously existing routes are no longer present. This will result in an inconsistent state for APP\_DB and potentially the ASIC, where the route exists (created by the previous incarnation of fpmsyncd).
```
mike-dubrovsky, line 105: this looks like a bug to me - on a cold bgp container restart - fpmsyncd shoud wipe out APPL_DB

doc/route_programming_optimization_zmq/HLD Northbound ZMQ.md
inder-nexthop: Hi @eddieruan-alibaba , I will test with ring buffer and multi-db and post updated results.

PR #2252, opened 2026-03-04
[chassis]: Add chassis provisioning HLD

liamkearney-msft:
Add HLD for automatic module provisioning. 
Introduces a new API and module operational states + pmon daemon to facilitate this within the sonic layer.

.md link with formatting: https://github.com/liamkearney-msft/SONiC/blob/liam/chassis-autoprovision/doc/chassis/module-provisioning/chassis-linecard-provisioning-hld.md


Javier-Tan (approved):
LGTM

doc/chassis/module-provisioning/chassis-linecard-provisioning-hld.md
```diff
@@ -0,0 +1,106 @@
...
+        provisioning to SONiC, and supports the various Provision operational states.
+
+        Returns:
+            bool: True if the request has been issued successfully, False if not
+        """
+        raise NotImplementedError
+```
+
+
+## New pmon daemon - sonic-provisiond
```
patrickmacarthur, line 96: Is it necessary to have a separate daemon for this? Can this functionality not be folded into chassisd?

liamkearney-msft: Decided to go this way after a discussion with @arlakshm 
Theres a few upsides of having it separate. If it's a separate daemon, disabling this feature becomes trivial. It also avoids creeping the scope for chassisd, and allows us to keep its role as "sync module state with statedb". It also simplifies keeping the state updated while conversion could potentially be running / allows us to block in `provision_module()` without having to block the chassisd thread (and avoid spinning out new threads in chassisd.)

So, yeah it could be folded in, but separating it out lets us avoid creeping the scope of chassisd & is more modular. If we are going through statedb anyway, there is no real need to strictly couple it with chassisd.

liamkearney-msft:
hi reviewers - PR for sonic-platform-common with API stubs / new states can be found here : https://github.com/sonic-net/sonic-platform-common/pull/635
cc @patrickmacarthur @kenneth-arista @arlakshm

doc/chassis/module-provisioning/chassis-linecard-provisioning-hld.md
```diff
@@ -0,0 +1,106 @@
...
+ - "ProvisionReady" – Linecard is available for conversion
+ - "ProvisionPending" – Linecard is undergoing conversion
+ - "Provisioned" – Linecard is converted but not fully available. Chassis may require reboot.
+
+```
+  ...
+    # Module state when module is detected, is able to run SONiC, but is not yet running SONiC.
+    # Modules in this state will be attempted to be converted to SONiC via calls to module.provision_module()
+    # This state & following "Provision" states should not be used if provision_module() is not implemented. 
+    MODULE_STATUS_PROVISION_READY = "ProvisionReady"
```
Staphylo, line 47: How is the vendor supposed to differentiate between `MODULE_STATUS_PROVISION_PRESENT` and `MODULE_STATUS_PROVISION_READY` ?
You cannot easily differentiate between a linecard that has been powered off and a linecard that you just inserted.
The mechanisms that I can think about would probably better live in the common infrastructure than the platform vendor API.

doc/chassis/module-provisioning/chassis-linecard-provisioning-hld.md
```diff
@@ -0,0 +1,106 @@
...
+## Background
+
+In a SONiC chassis, each linecard runs an independent instance of SONiC. If a linecard is replaced or a new linecard is inserted into an empty slot, there are no guarantees that this linecard has been provisioned to work with SONiC / has SONiC installed. Some vendors may require some steps / scripts / etc. to be executed on the supervisor card in order for the linecard to be available / operational / get to first boot.
+
+In other words, there is a use case for SONiC on the supervisor to have a hook into platform/vendor code when a new module is inserted / detected, to facilitate converting the module from factory state to SONiC.
+
+This document aims to introduce a unified mechanism to achieve this.
+
+## Requirements 
+ - When new linecards are inserted into a chassis, the supervisor card running SONiC is responsible for detecting the presence of these new modules. It is up to the vendors to implement a mechanism to detect this.
```
Staphylo, line 25: Sure vendors can implement a way to detect if a linecard is inserted but how would the daemon know about it if there is no common platform API to call?
For other components there is `get_change_event`
Short of doing this it means that the new daemon will have to use inefficient polling to see if something has changed. Having a platform API means that the platform vendor can either use its own polling logic or some event driven mechanism if they have one.

doc/chassis/module-provisioning/chassis-linecard-provisioning-hld.md
liamkearney-msft: They are very similar states, but PRESENT would indicate the state where the platform API does not support module provisioning / the module isn't ready for provisioning. Having a separate state allows the provisioning flow to be opt-in. 
Module detection is fundamentally platform specific - what would be the common mechanism for detecting this? All vendors would have to agree if we want to move this logic to the SONiC layer - my personal opinion is SONiC shouldnt mandate the implementation details for this.
I dont see why there would be difficulty in differentiating between a linecard which is powered off vs just inserted. I would expect the platform to be able to manage/monitor the power states of the modules & "presence" to be decoupled from power state.

doc/chassis/module-provisioning/chassis-linecard-provisioning-hld.md
liamkearney-msft: the daemon knows because it's reported via `get_oper_status()` / the state change is pushed to STATE_DB. This is the common API which is already used. The LC is being polled by chassisd using this API already (see https://github.com/sonic-net/sonic-platform-daemons/blob/master/sonic-chassisd/scripts/chassisd#L369).
This new daemon listens to changes in oper_state in STATE_DB, which is published by chassisd. We dont have to poll in the new daemon as we can just subscribe to the table in redis. This doesnt introduce any more polling than what is already there. 

The debate for having chassisd poll `get_oper_status()` vs having an event driven API to update the state reactively/dynamically is a whole other discussion, but this is simply leveraging the existing design & that change I think is out of scope (as it would require changes from everyone). In the future if we want to change the way module states are synced with STATE_DB, we can, and it shouldn't affect this new provisioning daemon. They've been decoupled deliberately.

liamkearney-msft:
Thanks for your comments @Staphylo. Ive added some responses to your questions. Let me know if these answer your concerns, and I can update the HLD to be more clear.

doc/chassis/module-provisioning/chassis-linecard-provisioning-hld.md
```diff
@@ -0,0 +1,106 @@
...
+ - A new platform API will be introduced as an entrypoint for vendor code to perform conversion on a module
+ - New module states will be introduced to represent the various provision states of a module
```
patrickmacarthur, line 27: Rather than implement this as new module states, could we introduce a new table in `CHASSIS_STATE_DB` called something like `MODULE_PROVISION_TABLE` that would look something like:

```
MODULE_PROVISION_TABLE|LINE-CARD4
{
    "state": "not_started", # or "started" or "complete"
    "timestamp": <unix timestamp>
}
```

It feels like either chassisd or the new provisiond service should be able to track whether or not provisioning has been started, instead of putting this burden onto the vendor platform driver.

In this model, chassisd would be responsible for keeping a persistent database of provisioned linecards, and when a linecard slot goes from Empty to Present or Online, it would determine if the linecard was previously provisioned in that slot, and if not it would create the `MODULE_PROVISION_TABLE` entry which would be detected by `provisiond`.

On first boot, the linecard would look for and update its `MODULE_PROVISION_TABLE` entry to `"complete"`.

patrickmacarthur: Discussed offline.

doc/chassis/module-provisioning/chassis-linecard-provisioning-hld.md
Staphylo: Thanks you for the detailed answer, under these conditions, `provisiond` is indeed event driven so we're not introducing new inefficiencies.

liamkearney-msft:
@kenneth-arista @patrickmacarthur @Staphylo 
If we are happy with the design as is, can I get some approvals so we can merge? 
thanks!

patrickmacarthur (approved):

arlakshm (approved):

PR #2259, opened 2026-03-16
HLD for platform specific config validation with YANG

rajath-nexthop:
## Summary
This PR introduces a comprehensive High Level Design document for Platform Specific Config Validation via YANG Models that provides **platform-aware configuration validation** at runtime without requiring static, one-size-fits-all YANG constraints.

## Key Features
- **Runtime YANG generation** using `jinja2` templates with platform-specific `feature_capabilities.json` files
- **Platform-aware validation** enabling different constraint ranges for different ASICs/platforms (e.g., Tomahawk5 vs Tomahawk6)
- **Seamless integration** with existing YANG validation flow via `postinst` hooks
- **Framework extensibility** allowing new features to easily add platform-specific checks

## Problem Solved
- **Platform-agnostic YANG limitations**: Current YANG models use static, global constraints that cannot accommodate platform-specific variations.
- **Silent failures**: Invalid configurations are accepted by CONFIG_DB but functionality fails, with errors only visible in syslogs.
- **Guardrail at validation level**: Reject out-of-bound values before writing to CONFIG_DB.

## Document Structure
- Complete SONiC HLD format following established conventions
- Placed in `doc/config_yang_validation/` directory
- Includes detailed design summary, framework architecture, and code snippets
- Provides concrete examples using buffer profile `dynamic_th` range validation

## Technical Highlights
- Leverages `feature_capabilities.json` for per-platform constraint definitions
- Uses jinja2 templates to generate deviation-based YANG modules at install time
- Integrates with `sonic_yang.py` and `sonic_yang_ext.py` for loading generated YANG
- Supports `config apply-patch`, `config reload`, `config replace`, and gNMI/RESTCONF paths

## Framework Flow
```
feature_capabilities.json + YANG template → Generated YANG → Runtime validation
```

## Review Focus Areas
- Framework design and `jinja2` template generation approach
- Integration with existing YANG validation infrastructure
- `postinst` hook implementation for platform-specific generation
- Extensibility guidelines for new feature owners

## PR for code changes 
[#26203 in sonic-buildimage](https://github.com/sonic-net/sonic-buildimage/pull/26203) - Framework for platform-specific validation.
[#27270 in sonic-buildimage](https://github.com/sonic-net/sonic-buildimage/pull/27270) - addition of `bpf` capabilities for Nexthop platforms.

### DIFF
--- a/doc/config_yang_validation/platform_config_validation.md
+++ b/doc/config_yang_validation/platform_config_validation.md
@@ -19,6 +19,7 @@
   - [8.1 Framework Testing](#81-framework-testing)
   - [8.2 Validation Testing](#82-validation-testing)
   - [8.3 Extending to New Features](#83-extending-to-new-features)
+- [9. Future Scope and long-term plans](#9-future-scope-and-long-term-plans)
 
 ## Revision
 
@@ -218,6 +219,8 @@ fi
 
 Note: The above snippet uses `PLATFORM_SUFFIX` and `PLATFORM_NAME` to dynamically determine the platform name from the postinst script name. This assumes the postinst script name follows a specific naming convention (e.g., `sonic-platform-nexthop-4010-r0.postinst`).
 
+**Note** - Failures in any `postinst` installation / `j2` file generation will be logged as part of the postinst logs themselves. If there is a failure here and the file generated is half-baked, we could potentially have errors in `YANG` model loading at config validation time, but the error would show what happened and could be corrected later.
+
 **Example feature_capabilities.json:**
 
 ```json
@@ -292,13 +295,14 @@ module sonic-buffer-profile-capability {
 1. `config apply-patch`
 2. `config reload -y`
 3. `config replace`
-4. `gNMI` related config changes (like `gnmi_set`)
+4. `gNMI` related config changes (like `gnmi_set`) through the native write path.
 
 **Cases where the new framework will NOT be triggered:**
 
 1. CLI based config - CLI can have validation checks in code itself if need be, depends on how the feature was implemented in the first place.
 2. Manual writes to CONFIG_DB - no validation is hit, invalid values only appear in syslogs
 3. `config load` - bypasses all validation, so YANG checks are never hit
+4. `RESTCONF / gNMI` translib write paths - these use the existing `go` based `CVL` validation flow.
 
 ## 6. YANG Model Enhancements
 
@@ -312,6 +316,8 @@ The generated YANG modules use the YANG `deviation` statement to override constr
 - Manual CONFIG_DB writes bypass validation
 - `config load` bypasses all validation
 - Platform must have `feature_capabilities.json` defined for platform-specific validation to apply
+- There could be possible confusion in the way `YANG` model maintenance changes due to this - since we add a new `platform-yang-models` directory with more `YANG` models to maintain. However, it is an acceptable change to incorporate robustness in platform-specific validation.
+- Upgrades to the image should ideally not affect these new `YANG` models, and if any features are disabled that created these models, they will remain redundant. There should be no operational difference or change in determinism with respect to config validation in this case.
 
 ## 8. Testing Requirements/Design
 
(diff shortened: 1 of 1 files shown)


venkatmahalingam:
I believe, any specific validations (for example, platform‑specific value ranges) should be implemented through custom validation. If we plan to add more comprehensive configuration validation in the future (apart from native YANG checks), introducing a custom validation handler would be beneficial.

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-templates/sonic-extension.yang.j2#L24

anilpannala:
Please find the recording of community review - https://zoom.us/rec/share/7apA5AkMuNqwzDTkoHOJZ_FqKnjhwRmCyxW_ch74S1SjKB1mxcbHhdwBQ2HC-tQ.fEjSfiCKpdNS7cDr?startTime=1774968801000

rajath-nexthop:
Review comments addressed and added into the doc (Some limitations, future scope of work and how this design fits in the broader scheme of things in config validation). Please take a look and review it now, thanks!

rookie-who:
Hi @rajath-nexthop — the 202605 branch cut is at end of May. This PR has no reviews yet. Could you suggest reviewers or request a review from the relevant maintainers?

rajath-nexthop:
Hi @rookie-who , sure!
I had an HLD meet for this, I will tag the relevant folks from that to check this PR out

@lguohan , @venkatmahalingam please take a look at the HLD once again, it has the comments from the meet addressed!

securely1g:
## Review Feedback (on behalf of @lguohan)

Hi @rajath-nexthop, nice work on this HLD — the problem is real and the deviation-based approach is clean. A few items to address:

### Design Concerns

1. **postinst platform-name derivation is fragile** — Parsing the script filename with `sed` substitutions (`nexthop-` → `nexthop_`) will break for platforms that don't follow that convention. Consider reading `/etc/sonic/sonic-platform` or using `sonic-cfggen -v platform` instead for a robust, canonical platform name.

2. **No validation of generated YANG** — If `j2` renders a syntactically broken file, it gets silently installed. The postinst should sanity-check the output (e.g., `pyang` lint or at minimum verify non-empty valid XML/YANG) before placing it in `/usr/local/platform-yang-models/`.

3. **CVL (Go path) gap** — The doc says RESTCONF/gNMI translib writes use CVL and won't pick up these deviations. Since gNMI is increasingly the primary config path, this is a significant gap. Please either describe how CVL will consume these constraints, or explicitly call it out as a known limitation with a timeline to close it.

4. **`config load` bypasses validation** — Listed as a limitation but not discussed. Many operators use `config load`. Is there a plan to close this gap?

### Suggestions

5. **Upgrade/downgrade story** — What happens when a new image ships different `feature_capabilities.json` values? Does the postinst re-run on upgrade? What about rollback? Needs more detail.

6. **`feature_capabilities.json` has no formal schema** — One example with `mmu_capabilities` but no contract. Can feature owners add arbitrary top-level keys? Consider defining a JSON schema.

7. **Testing section is sparse** — No negative test cases (corrupt JSON, missing keys, partial capabilities), no integration test outlines with actual `config apply-patch` on a real platform. Would be good to add at least test case descriptions.

8. **Section 9 (Future Scope)** — The SAI capability query idea is interesting but quite hand-wavy. Either flesh it out with a rough plan or trim to a brief "future direction" note.

### Positives
- YANG deviation statements are the right native approach
- Backward compatible — platforms without `feature_capabilities.json` are unaffected
- Mermaid flow diagram is clear
- Extensibility recipe in section 8.3 is helpful

Overall good direction, needs tightening on postinst robustness, CVL gap, and testing before merge.

CC: @vaibhavhd for review

securely1g (commented):
## Review: Platform Specific Config Validation with YANG

Nice design — clean separation of capability data, templates, and generated YANG. The deviation-based approach is non-invasive and backward compatible. A few observations:

### Questions / Suggestions

1. **Postinst platform name parsing is fragile** — Extracting platform from script name via `sed` regex will break for platforms with non-standard naming. Consider reading from `/etc/sonic/sonic_version.yml` or the `platform` field in `machine.conf` instead.

2. **No validation of generated YANG** — The HLD acknowledges "half-baked" files could cause YANG loading errors. A simple `pyang --lint` check post-generation in the postinst hook would catch template rendering bugs before they surface at config validation time.

3. **Install-time generation is rigid** — Generating at `postinst` means a `feature_capabilities.json` update requires package reinstall. Would runtime generation (e.g., at `config reload` or `sonic-cfggen` time) be more operationally flexible? Or is install-time intentional for immutability?

4. **`config load` bypass** — Worth calling out explicitly that this is a pre-existing SONiC limitation, not a gap introduced by this HLD.

5. **No versioning for `feature_capabilities.json`** — If the schema evolves between releases, how does the framework handle upgrades gracefully? A `version` field in the JSON would help.

6. **Testing section** — Could benefit from a concrete example of CI integration (e.g., a pytest that validates generated YANG against a mock `feature_capabilities.json`). Also, how would this be tested across multiple platforms in Elastictest?

7. **Section 9 (Future Scope)** — The SAI capability querying → STATE_DB → validation vision is compelling. The current approach is a pragmatic stepping stone.

Overall this is a solid design that addresses a real gap. The framework extensibility guidelines (Section 8.3) are clear and should make adoption straightforward for feature owners.

cc @lguohan for awareness

securely1g:
*Review posted on behalf of @lguohan*

rajath-nexthop:
Hi team,
Can I get an approval on this please? HLD meet was done, and comments addressed as well!
Thanks

PR #2260, opened 2026-03-16
[HLD] Match Condition Based DHCP Server (MATCH mode)

Blueve:
## Summary

HLD for adding a new `MATCH` mode to the SONiC DHCP server. This mode introduces a generalized match condition system where all matching criteria — including port identity (via Option 82 Circuit ID) — are treated uniformly as match conditions that can be composed freely.

## Key Design Points

- **New `MATCH` mode** for `DHCP_SERVER_IPV4` — cleanly separated from existing `PORT` mode
- **`DHCP_SERVER_IPV4_MATCH`** table — named, reusable match conditions (types: `circuit_id`, `option60`)
- **`DHCP_SERVER_IPV4_BINDING`** table — associates match condition(s) with IP pools per VLAN
- **Port is a match condition**, not a structural key — `circuit_id` is just another match type alongside `option60`
- **AND logic** via multiple matches per binding; **OR logic** via separate bindings pointing to the same pool
- **Specificity ordering** — bindings with more conditions take priority
- **Extensible** — new match types added by extending the enum, no structural changes needed

### DIFF
--- a/doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
+++ b/doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
@@ -111,6 +111,8 @@ Configuration of match condition feature can be done via:
 
 * **New `MATCH` mode.** The `DHCP_SERVER_IPV4` table's `mode` field gains a new enum value `MATCH`. When `mode=MATCH`, the DHCP server reads from `DHCP_SERVER_IPV4_MATCH` and `DHCP_SERVER_IPV4_BINDING` tables. When `mode=PORT`, behavior is unchanged. Different VLANs can use different modes independently.
 
+* **Mode switching.** When a VLAN's mode is changed from `PORT` to `MATCH`, the `DHCP_SERVER_IPV4_PORT` entries for that VLAN are ignored and the server is reconfigured using `BINDING` entries only. Active leases are not revoked — they expire naturally per their lease time. Operators should configure `BINDING` entries before switching modes to avoid a gap in service.
+
 * **Match conditions are the building blocks.** Each condition tests a single DHCP packet field. Conditions are composed via AND logic in bindings. This includes port identity — `circuit_id` is a match type, not a structural key.
 
 * **Unified design.** Port identification and DHCP option matching use the same `DHCP_SERVER_IPV4_MATCH` table and `DHCP_SERVER_IPV4_BINDING` table. There is no separate port-coupled table for match bindings.
@@ -178,6 +180,8 @@ When a VLAN has multiple bindings with overlapping conditions, more-specific bin
 
 For example, a "VendorA device on etp1" matches both a 2-condition binding [port_etp1, vendor_a] and a 1-condition binding [port_etp1]. The more-specific 2-condition binding takes priority.
 
+**Note:** Operators should avoid creating multiple bindings with the same number of conditions that can both match the same client. If such an overlap exists, the behavior is non-deterministic. This is considered a misconfiguration.
+
 ## DB Changes
 
 ### Config DB
@@ -200,7 +204,7 @@ Two new tables are added. One existing table is extended with a new mode value.
 | Field | Type   | Required | Description |
 |-------|--------|----------|-------------|
 | type  | enum   | Yes      | Match type. Currently: `circuit_id`, `option60`. Extensible for future types. |
-| value | string | Yes      | Value to match against (exact match). For `circuit_id`, this is the port alias (e.g., "etp1"). |
+| value | string | Yes      | Value to match against (exact match). For `circuit_id`, the user configures the **port alias** (e.g., "etp1"); the implementation constructs the full on-wire Circuit ID (`hostname:port_alias`) internally, consistent with existing `PORT` mode behavior. |
 
 **DHCP_SERVER_IPV4_BINDING** — Associates match condition(s) with an IP pool.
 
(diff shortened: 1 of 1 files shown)


doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
```diff
@@ -0,0 +1,565 @@
...
+}
+```
+
+Either VendorA or VendorB devices on etp1 will receive addresses from that pool.
+
+### Specificity Ordering
+
+When a VLAN has multiple bindings with overlapping conditions, more-specific bindings (more match conditions) take priority over less-specific ones:
+
+1. Bindings with more match conditions are evaluated first (most specific)
```
Blueve, line 178: Specificity ordering only defines that bindings with more match conditions take priority. What should happen when two overlapping bindings with the same number of conditions both match a client? Please define a deterministic tie-break rule or make equal-specificity overlaps invalid.

doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
```diff
@@ -0,0 +1,565 @@
...
+| `MATCH` | New. IP assignment based on `DHCP_SERVER_IPV4_MATCH` + `DHCP_SERVER_IPV4_BINDING` tables. |
+
+#### New Tables
+
+**DHCP_SERVER_IPV4_MATCH** — Defines named match conditions.
+
+| Field | Type   | Required | Description |
+|-------|--------|----------|-------------|
+| type  | enum   | Yes      | Match type. Currently: `circuit_id`, `option60`. Extensible for future types. |
+| value | string | Yes      | Value to match against (exact match). For `circuit_id`, this is the port alias (e.g., "etp1"). |
```
Blueve: For `circuit_id`, this says the value is the port alias (for example `etp1`), but the existing port-based DHCP server HLD shows the relay/Kea flow matching the full Circuit ID value like `hostname:etp1`. Please specify the exact on-wire format expected here so the configuration is unambiguous.

doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
```diff
@@ -0,0 +1,565 @@
...
+**DHCP_SERVER_IPV4_BINDING** — Associates match condition(s) with an IP pool.
+
+Key format: `<vlan>|<binding_name>`
+
+> `<binding_name>` is a user-defined unique label. It has no semantic meaning beyond making the key unique and providing a readable identifier. The actual match logic comes entirely from the `matches` field.
+
+| Field   | Type      | Required | Description |
+|---------|-----------|----------|-------------|
+| matches | leaf-list | Yes      | One or more references to `DHCP_SERVER_IPV4_MATCH` entries. Combined with AND logic. |
+| ips     | leaf-list | No       | Direct IP assignment. Mutually exclusive with `ranges`. |
```
Blueve: The table says `ips` and `ranges` are mutually exclusive, but both are optional. That appears to allow a binding with no pool at all. Should the model require exactly one of `ips` or `ranges`?

doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
```diff
@@ -0,0 +1,565 @@
...
+* JSON config input
+* SONiC CLI
+
+# Design
+
+## Design Considerations
+
+* This feature is an **additive extension** to the existing port-based DHCP server. All existing CLI commands and YANG models remain backward compatible.
+
+* **New `MATCH` mode.** The `DHCP_SERVER_IPV4` table's `mode` field gains a new enum value `MATCH`. When `mode=MATCH`, the DHCP server reads from `DHCP_SERVER_IPV4_MATCH` and `DHCP_SERVER_IPV4_BINDING` tables. When `mode=PORT`, behavior is unchanged. Different VLANs can use different modes independently.
```
Blueve, line 112: Since this introduces a new `MATCH` mode alongside existing `PORT` mode behavior, could you document the migration semantics for an existing VLAN switching from `PORT` to `MATCH`, especially what happens to existing `DHCP_SERVER_IPV4_PORT` entries and active leases during the transition?

doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
Blueve: yes. but seems like there is no enforcement we can do on YANG model. we will let the application fail if user not provide both of them.

doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
Blueve: Clarified in the doc: user configures port alias (e.g., \tp1\), the implementation constructs the full on-wire Circuit ID (\hostname:port_alias\) internally, consistent with existing PORT mode behavior.

doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
Blueve: Added migration semantics: when switching from PORT to MATCH, PORT entries are ignored, active leases expire naturally, and operators should configure BINDING entries before switching modes to avoid service gaps.

doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
Blueve: Added a note that equal-specificity overlapping bindings are non-deterministic and considered a misconfiguration. Operators should avoid creating ambiguous overlaps.

doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
Blueve: Updated the table to note that at least one of \ips\ or \
anges\ must be provided, enforced at application level since YANG cannot express this constraint.

doc/dhcp_server/match_condition_based_dhcp_server_high_level_design.md
```diff
@@ -0,0 +1,569 @@
...
+By making `circuit_id` a match condition type alongside `option60`, the design achieves full uniformity:
+- A port-only assignment is a binding with a single `circuit_id` match
+- A port + vendor class assignment is a binding with `circuit_id` + `option60` matches (AND logic)
+- Future match types are added by extending the enumeration — no structural changes required
+
+## Functional Requirements
+
+1. Support IP assignment based on configurable DHCP packet field matching.
+2. Support `circuit_id` (port identification via Option 82) as a match condition type.
+3. Support DHCP Option 60 (Vendor Class Identifier) as a match condition type.
```
sphuong-v3, line 93: It would be a good idea to write down in which DHCP packet this Option 60 and `circuit_id` should be present in. I think the easiest one is the `DHCPDISCOVER` message as this would be the first packet in a DHCP protocol that gets sent out. But it would be good to identify which DHCP message you want this option to be present in.

Blueve: Yes. The assumption here is DHCP DISCOVER where the client initiate the request

PR #2261, opened 2026-03-17
HLD: Eliminate D-Bus host-service dependency from sonic-gnmi

hdwhdw:
## Description
Adds a High Level Design document for eliminating the D-Bus IPC dependency between the sonic-gnmi container and sonic-host-services.

### Motivation
- The gnmi container already has host-level privileges; D-Bus through host-services is an unnecessary middleman
- sonic-host-services is single-threaded (GLib main loop) and blocks on subprocess calls — a single slow operation stalls all concurrent gNMI/gNOI requests
- KubeSonic ships the gnmi container independently against varying host OS versions where host-services D-Bus modules may differ or be missing

### Design
Replace 23 D-Bus operations with two tiers:
- **Tier 1 (14 ops):** Go-native API calls (godbus → systemd1, go-redis, os.*, net/http, Docker Go SDK)
- **Tier 2 (9 ops):** nsenter to SONiC CLI tools (config, sonic-installer, generate_dump)

New code lives in domain packages under `pkg/host/`. Migration proceeds per-gRPC handler — no feature flag.

### HLD Document
`doc/mgmt/gnmi/eliminate_dbus_host_service_hld.md`

### DIFF
--- a/doc/mgmt/gnmi/eliminate_dbus_host_service_hld.md
+++ b/doc/mgmt/gnmi/eliminate_dbus_host_service_hld.md
@@ -1,0 +1,466 @@
+# Eliminate D-Bus Host Service Dependency from sonic-gnmi
+
+# High Level Design Document
+
+#### Rev 0.1
+
+## Table of Contents
+- [Table of Contents](#table-of-contents)
+- [Revision](#revision)
+- [Scope](#scope)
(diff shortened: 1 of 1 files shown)


hdwhdw:
@sneelam20 @ndas7 PTAL. We can  present it in the next UMF meeting

PR #2269, opened 2026-03-29
OCS High Level Design

wni0812:
OCS High Level Design


doc/ocs/ocs_hld.md
```diff
@@ -0,0 +1,620 @@
...
+- Conditionally register CLI commands based on the switch_type in DEVICE_METADATA within CONFIG_DB.
+- Define L2/L3 as SONiC features and register the commands based on feature enable/disable.
+
+&nbsp;    
+## 10. Serviceability and Debug
+All output generated by OCS Orch agent are logged to local syslog and default log level is NOTICE.
+
+&nbsp;    
+## 11. Warm Boot Support
+Warm reboot is supported. OCS SDK performs a special initialization of OCS module after warm reboot.
```
swapnaNann, line 619: - Is the “special initialization” configurable via API?
- Does it preserve existing cross-connections (hardware/SDK persistence), or replay them from persistent configuration?
- Is this behavior specific to warm reboot, or does it apply to other SONiC reboot types as well?

doc/ocs/ocs_hld.md
```diff
@@ -0,0 +1,620 @@
...
+| src/sonic-utilities | sonic-utilities |  New OCS CLI commands     | 
+| src/sonic-mgmt-common | sonic-mgmt-common |  Update CVL for validation of OCS configuration    | 
+| src/sonic-yang-models | sonic-buildimage |  New definitions of OCS configuration and operational status    | 
+| src/sonic-yang-mgmt | sonic-buildimage |  Update tests for new OCS YANG definitions     | 
+| platform | sonic-buildimage |  Code specific to vendor device, e.g. vendor OCS SAI library, etc.     | 
+| device | sonic-buildimage |   Code specific to vendor device    | 
+
+&nbsp;
+## 6. High-Level Design
+### 6.1. Database Changes
```
swapnaNann, line 121: The table above indicates updates to sonic-gnmi and sonic-mgmt-common are needed. Could a section be added to summarize the OCS changes required for the NBI?

doc/ocs/ocs_hld.md
```diff
@@ -0,0 +1,620 @@
...
+```
+
+_Questions:_  
+* Since L2/L3 data forwarding CLI commands are not required for the OCS platform, the corresponding CLI commands can be excluded. The following are some possible approaches:
+- Conditionally register CLI commands based on the switch_type in DEVICE_METADATA within CONFIG_DB.
+- Define L2/L3 as SONiC features and register the commands based on feature enable/disable.
+
+&nbsp;    
+## 10. Serviceability and Debug
+All output generated by OCS Orch agent are logged to local syslog and default log level is NOTICE.
```
swapnaNann: It would be helpful to also describe the fault reporting mechanism. Are faults/alarms exposed through any structured interface (e.g., STATE_DB, gNMI/telemetry, or NBI) for external systems to consume?

doc/ocs/ocs_hld.md
```diff
@@ -0,0 +1,620 @@
...
+
+&nbsp;    
+## 5. Architecture Design
+To support OCS, the following design and changes will be implemented across the existing SONiC solution and containers:    
+* Database Tables
+  - New tables will be added to CONFIG_DB (for configuration data) and STATE_DB (for operational status data).
+* OCS Orch Implementation
+  - A new OCS Orchestration application (OCS Orch) will be included within the existing SWSS container's Orchagent. Main functions of this new agent will be the following items:
+    * Monitor Configuration: Register for Redis keyspace notifications on new OCS_PORT and OCS_CROSS_CONNECT tables in CONFIG_DB. This can be extended to database changes on other DB instances such as APPL_DB if needed for more scenarios, like dynamic case etc.  
+    * Process Changes: Handle and apply configuration updates for OCS ports and cross-connects upon changed in CONFIG_DB.
```
swapnaNann, line 92: Based on the use of Redis keyspace notifications and Orch callbacks, the flow looks asynchronous/event-driven, but this is not stated clearly. Please clarify whether a write to CONFIG_DB is only an intent submission, when the change is considered accepted vs successfully applied, whether ordering is guaranteed across multiple updates, and how failures/retries/partial success are reflected back to the client and in STATE_DB

doc/ocs/ocs_hld.md
```diff
@@ -0,0 +1,620 @@
...
+    * Update Status: Reflect the current operational status of OCS ports and cross-connects by updating the data into STATE_DB.
```
swapnaNann, line 93: Are batch commits processed atomically or may result in partial hardware programming ?

doc/ocs/ocs_hld.md
```diff
@@ -0,0 +1,620 @@
...
+```
+### 6.2. Orchestration Agent 
+OCS Orch is the core component for managing OCS configurations and updating corresponding operational status:    
+* Manage OCS switch capacity
+* Manage configuration and status of OCS ports
+* Manage configuration and status of OCS cross-connects
+* Manage operational data specific to OCS module
+
+#### 6.2.1. Management of OCS Switch Capacity    
+OCS device can have varying capacities from different hardware or module, defined as "number of A side ports" × "number of B side ports" (e.g., 64×64, 320×320, etc.). Cross-connects are provisioned between two ports.
```
swapnaNann, line 197: The current model assumes a single logical OCS module with A_SIDE and B_SIDE ports, but many OCS systems are modular. If multiple modules are present, does the design extend SAI to support multiple switch instances or does it aggregate them into one logical switch? This has implications on port identity, cross-connect scope and Orch-to-SAI mapping, which we might have to consider?

doc/ocs/ocs_hld.md
```diff
@@ -0,0 +1,620 @@
...
+
+**OCS_FACTORY_INSERTION_LOSS**
+```
+; Description: Inventory data for factory insertion loss measurements. This table uses a composite key.
+; Key: cross_connect_id|frequency_THz|temperature_C (e.g., "1A-2B|229.100|25.00")
+;
+OCS_FACTORY_INSERTION_LOSS:cross_connect_id|frequency_THz|temperature_C
+loss_dB                    = DECIMAL (2)         ; Factory insertion loss in dB (2 fraction digits)
+accuracy_dB                = DECIMAL (2)         ; Insertion loss measurement accuracy in dB (2 fraction digits)
+```
```
swapnaNann, line 188: It might be useful to consider adding a brief section on the role of ASIC_DB in the OCS flow, to clarify how CONFIG_DB intent is translated into SAI operations?

doc/ocs/ocs_hld.md
wni0812: Vendor SDK can restore from local or from device as vendor specific implementation. Also upper layer will replay it.

doc/ocs/ocs_hld.md
wni0812: Batch commit APIs (MULTI/EXEC) are common redis solution. sonic-gnmi uses this API to implement batch commit quite straightforward. sonic-mgmt-common is optional for the implementation to do customized validations. Later, PR for code submission will have more details.

doc/ocs/ocs_hld.md
wni0812: From gNMI to redis and keyspace notification callback, it is asynchronous processing. From Orchagent to SAI, it is sychronous processing. After operation, return result will be updated into STATE_DB.

doc/ocs/ocs_hld.md
wni0812: Yes. Batch commits via redis API are processed atomically.

doc/ocs/ocs_hld.md
wni0812: In OCS SAI, both port and cross-connect are string based definition. It is quite flexible to be extended for multiple instances and multiple modules scenarios. SONiC Multi ASIC can also be followed  for this case.

doc/ocs/ocs_hld.md
wni0812: Sequence diagram is provided about OCS operation workflow from NBI, CONFIG_DB, Orchagent, SAI/ASIC_DB, refer to section 6.2.4 for more details. Main logic focus on DB schema translations between CONFIG_DB and ASIC_DB.

doc/ocs/ocs_hld.md
swapnaNann: Thanks @wni0812. Should we consider making this behavior configurable, given that vendors may differ in how cross-connections are preserved or restored after reboot?

doc/ocs/ocs_hld.md
wni0812: SONiC current implementation use structured data in syslog for alert. The event and alarm management function(PR22617) in SONiC is still not merged. It can be considered as a reference.

doc/ocs/ocs_hld.md
swapnaNann: Makes sense to align with SONiC Multi-ASIC as a baseline.

I think we should also clarify the naming/identity model accordingly. For a modular system, the choice between multiple switch instances vs a single logical switch will affect how ports and cross-connects are represented. For example, in a multi-instance model, port identity may need to include module/switch context (for example module1:A1, module2:B4 or M1:A1 , M2:B4 ) and cross-connect identifiers may also need to be scoped per module. In a single logical-switch model, we may instead want a flat/global naming scheme.

So besides the architectural model, should we also define naming conventions for ports and cross-connect objects now, so Orch-to-SAI mapping stays unambiguous as we extend to multi-module systems?

doc/ocs/ocs_hld.md
wni0812: Replay logic is to ensure the consistency between the upper layer configurations and OCS device. Vendor SDK can determine whether to use upper layer configurations during restoration.

doc/ocs/ocs_hld.md
wni0812: Both port and cross-connect are string definition and have the flexibility to support different scenarios. If multi modules scenario from physical layer still with non blocking switch on all ports, then, 1A..320A and 1B..320B are suitable on this case. If multi OCS slot/instance/module are independent switching domain, then, the naming can be extended, like 1-1A, 3-20B etc., the first number is about slot/instance/module from logic view. Vendor SDK can translate it into the physical layer information.

dandoliska: The HLD for multi-asic describes an architecture where each ASIC has its own set of docker containers (database, syncd SAI/SDK). That seems overly complicated for managing multiple OCS modules.

doc/ocs/ocs_hld.md
```diff
@@ -0,0 +1,625 @@
...
+        SAI_BULK_OP_ERROR_MODE_STOP_ON_ERROR,
+        oid_list.data(), status_list.data()
+    );
+```
+
+&nbsp;    
+### 8. YANG Model
+Microsoft defines SONiC YANG model for OCS and YANG tree is as below. Ports are categorized into two types: A-side and B-side.    
+OCS cross-connects are only permitted between one A-side port and one B-side port.    
+As a string definition in YANG model, it is flexible to extend the naming format for different scenarios.
```
dandoliska, line 396: The model and underlying SAI design assume that ports are grouped as A-side and B-side and connections can only be made between A and B. This may not always be the case for future innovations or other types of architectures. For example:

1. an OCS module that instead of A-to-B architecture, one that can do any-to-any (any port can be connected to any other port other than itself).
2. in the case of multiple OCS modules, port 1-1A can not be connected to 2-3B. But nothing in the model indicates that this is not possible.
3. An A-to-B architecture is a 2-degree OCS. If a 3-degree OCS is developed (if that is even possible), not clear how this model could be used.

wni0812: As previous discussion on SAI PR and description in this HLD, port and cross-connect are string type. It is flexible for different scenarios. 

a_side and b_side are two attributes in YANG model should denote the two endpoints of cross-connect, they do not necessarily need to be on the A or B sides as two different port names (think of it as a-end and z-end). YANG model is meant to support cross-connects like ‘1A-2A' or ‘1A-5B’ or '1/1A-1/3B' too if the hardware supports it. Similar as the description about cross-connect needs to be configured between a_side and b_side two attributes. Port names for the value of attributes can be different or same as any-to-any.

For other scenarios like 3-dgree, if cross-connect is still two sides, current YANG attribute definition should be fine with new port naming on new technology. If cross-connect need more sides more than 2, then, YANG definition is extensible with new object or new list or new side.

doc/ocs/ocs_hld.md
wni0812: If multi modules scenario from physical layer still with non blocking switch on all ports, then solution using one group of: database, swss, syncd/SAI/SDK is suitable. If multi OCS slot/instance/module are independent switching domain, then SONiC multi ASIC can be considered.

swapnaNann: This topic came up in today’s meeting and it would be useful to revisit it with other vendors in the next session. Specifically: (1) Align on port and cross-connect naming conventions for both single-module and multi-module OCS's  (2) For multi module systems we can think of having more than 10 modules per CPU and having separate database, swss, syncd, redis instances is an overkill for Cpu/Mem and boot time should the ocs driver in syncd directly handle this ?

doc/ocs/ocs_hld.md
```diff
@@ -0,0 +1,625 @@
...
+
+#### 6.2.2. Management of OCS cross-connect    
+OCS Orch registers DB change notifications on OCS cross-connect table and processes DB changes.
+
+Performance Enhancement:    
+In current SONiC implementation, configuration change is committed to CONFIG_DB separately for each entry, Redis keyspace notifications trigger a callback for each individual DB entry change, Orch processes this single change and applies configuration to ASIC via sairedis/Syncd/SAI as a single SAI object operation. Overall processing flow is per entry implementation.
+ 
+To support high-performance on parallel/simultaneous configuration execution with full or multiple OCS cross-connects, a batch processing approach can be implemented with the following proposals:    
+* Batch DB Commit: NBIs do batch commit to CONFIG_DB via Redis transaction.
+* DB Change Processing: During callback handling, it compares cached entries against new CONFIG_DB entries then invokes bulk creation APIs for all CONFIG_DB cross-connect entries (via SAI error mode: SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR). After this bulk remove is needed to delete SAI related objects from logical view, instead of operations to module or hardware.
```
swapnaNann, line 210: Looks like we are using SAI_BULK_OP_ERROR_MODE_IGNORE_ERROR for bulk creation APIs what is the expected behavior for partial success/failure (e.g., what ends up in CONFIG_DB vs STATE_DB, rollback strategy, retry rules ..)?

wni0812: SONiC Bulk APIs do not support mixed scenarios, like mixed "create", "set", "remove" operations in one bulk API. So, configuration needs to use this mode to force overwrite cross-connects in syncd/SAI. If APIs return failure, STATE_DB will be updated with the running cross-connects from module.

doc/ocs/ocs_hld.md
wni0812: If multi modules scenario from physical layer still with non blocking switch on all ports, then the OCS driver can handle this need to translate from logical port to physical port.
If multi OCS slot/instance/module are independent switching domain, use solution either multi ASIC from SONiC as guideline or one instance in upper layer and OCS driver in syncd handles it.

PR #2272, opened 2026-03-31
SONiC gNMI support to get show command output on-demand

sharmavijay-ms:
Adds a new design/documentation markdown describing an approach to retrieve SONiC “show” command output on-demand via gNMI Get, including CLI→gNMI path mapping, security considerations, governance expectations, and a rollout/test plan.

Changes:
Introduces a new doc outlining gNMI-based on-demand “show” data retrieval goals, architecture, and flows.
Provides example CLI→gNMI path conversions and sample request/response payloads.
Describes governance (“gNMI-first”), migration strategy, and high-level testing/rollout plans.


doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,296 @@
...
+- [New design (HLD)](#new-design-hld)
+- [Details](#details)
+- [STATS update](#stats-update)
+- [Examples](#examples)
+- [Requirements](#requirements)
+- [Test](#test)
+- [Future plan](#future-plan)
+
+# Goals
+1. Provide a way to execute show CLI commands on device (on-demand) and get the output.
```
make1980: we already support this - i think the goal is to provide a framework for implementing show CLI commands using gNMI based API.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,296 @@
...
+2. Avoid interactive login to the device for command execution.
```
make1980: the goal is to provide a gNMI based API as a read only interface for retrieving SONiC device metadata, which can allow remote invocation without interactive user login.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,296 @@
...
+3. Return output in JSON format equivalent to CLI output.
```
make1980: Provide a structured response that can be consumed by application.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,296 @@
...
+4. Support throttling targets:
```
make1980: In general this is about supporting rate limiting - and the limit should be configurable.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,296 @@
...
+   - Up to 32 parallel command executions per device.
+   - Up to 100 maximum concurrent connections.
+
+# Problems to solve
+Existing tools (or new tools) that require device data for lifecycle workflows do not always have near real-time access to SONiC state.
+
+Today, data collection generally happens in two ways:
+1. Log in to the device and run CLI commands.
+   - This needs manual or scripted login workflows.
```
make1980: I would rephrase the problems for run CLI commands as:

1. Requires user account/password based authentication, instead of certificate based authentication
2. Introduces overhead for constructing the input and parsing the output of the CLI commands when used by application/automation.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,296 @@
...
+   - It introduces device access/security overhead.
+   - It is slower due to multiple execution steps.
+2. Use telemetry service output.
```
make1980: I don't feel that we need to mention telemetry service here.

sharmavijay-ms: Addressed all comments. Thank You.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,296 @@
...
+No parameter is required.
+
+```bash
+./gnmi_cli -client_types=gnmi \
+  -a <DEVICE-IP>:<PORT> \
+  -ca <path_to_CA_crt> \
+  -client_crt <path_to_client_crt> \
+  -client_key <path_to_client_key> \
+  -t OTHERS -logtostderr \
+  -qt p -pi 10s -q show/reboot_cause/history
```
zbud-msft: This is a subscribe request, perhaps we should use GET request instead. Do we plan on supporting subscribe requests for show commands?

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,296 @@
...
+**TODO:** Replace with final HLD diagram image/link.
+
+![HLD](https://microsoft.sharepoint.com/:i:/t/Aznet/IQCdYjwiFfnAQottXCMk2z94AeGgbtNZvsIwzRlswSU-iS0?e=tFk9u4)
+
+# Details
+Show commands retrieve data from multiple backends:
+- Redis
+- System files
+- Shell commands
+- vtysh
```
zbud-msft: Should we include a section on input-validation/injection for path keys feeding shell and vtysh backends?

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,296 @@
...
+## 1. Non-parameterized query
+No parameter is required.
+
+```bash
+./gnmi_cli -client_types=gnmi \
+  -a <DEVICE-IP>:<PORT> \
+  -ca <path_to_CA_crt> \
+  -client_crt <path_to_client_crt> \
+  -client_key <path_to_client_key> \
+  -t OTHERS -logtostderr \
```
zbud-msft: Should we use the actual SHOW path instead of OTHERS here?

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,296 @@
...
+    "Oper": "up",
+    "Admin": "up"
+  }
+}
+```
+
+  # Requirements
+  1. Support up to 32 parallel command executions per device.
+  2. Support up to 100 concurrent client connections.
+  3. Preserve AAA-based authentication and authorization controls for all requests.
```
zbud-msft: Can you add a section on what authorization model will exist for show commands?

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
sharmavijay-ms: done.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
sharmavijay-ms: ack. I will work on it.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
sharmavijay-ms: will address it by capturing cert info of existing telemetry.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Governance: Mandatory Review Gate
+
+- Introduce a **reviewer group for `sonic-utilities` repository**
+- Any new `show` CLI command:
+  - **Requires mandatory approval** from the reviewer group
+  - Must include:
+    - Corresponding **gNMI API implementation**
+    - Valid **CLI → gNMI path mapping**
+- CLI-only implementations will be **rejected**
```
sharmavijay-ms, line 203: Table review:- Adjust to consider: API can invoke py but with the cost of overhead. Not recommended.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
sharmavijay-ms: Done.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
sharmavijay-ms: done.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
sharmavijay-ms: Done.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
sharmavijay-ms: Done.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,310 @@
...
+  2. Functional tests
+    - Run representative commands (`show reboot-cause`, `show interface status`) and compare against expected output.
+    - Validate certificate-based client authentication and authorization behavior.
+  3. CLI test
+    - Existing CLI tests will be utilized to validate the functionality and correctness of the output.
+    - With the formatter in place, the gNMI API JSON payload combined with the formatter output must satisfy these existing test cases.
+  4. Nightly test
+    - A nightly end-to-end test suite will validate the complete execution path, including CLI-to-gNMI path translation by utility, gNMI API execution, and output formatting utility. The tests ensure that JSON responses generated by the gNMI APIs are correctly transformed into the expected human-readable CLI format, providing continuous validation of functional correctness and regression coverage.
+
+# Rollout plan
```
sharmavijay-ms: Formatting issue.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,365 @@
...
+Boolean flags → [key=True]
+
+- Argument Normalization: CLI arguments are normalized before gNMI path generation. This ensures that equivalent short-form and long-form options, such as `-v` and `--verbose`, are resolved to a common internal representation. The normalized representation is then used to generate a deterministic gNMI path. This preserves compatibility with the existing SONiC CLI behavior while ensuring that different forms of the same command generate an identical gNMI request.
+```bash
+show interfaces status -v
+show interfaces status --verbose
+```
+
+**SONiC CLI to gNMI Path Conversion Utility**
+* A CLI utility tool has been developed to translate legacy `show` commands into their corresponding gNMI paths using long-form options.
```
saravanan-i-CLS, line 150: Are these paths backed by a YANG model (OpenConfig / sonic-yang), or is the JSON hand-shaped per command? Please clarify the schema source of truth for the SHOW/... paths

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,365 @@
...
+  -a <DEVICE-IP>:<PORT> \
+  -ca <path_to_CA_crt> \
+  -client_crt <path_to_client_crt> \
+  -client_key <path_to_client_key> \
+  -t SHOW -logtostderr \
+  -qt p -pi 10s -q interface[interface=Ethernet0]/status
+```
+
+```json
+{
```
saravanan-i-CLS, line 189: How does the output converter will parse the gnmi output (say ietf-json) ? Whether jinja based rendering templates needs to be defined for each of the show commands as some show commands parse and display output in python in sonic-utilities ? Please clarify.

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
sharmavijay-ms: The SHOW/... paths are not derived from OpenConfig or sonic-yang models. They form a virtual, CLI-oriented gNMI namespace whose hierarchy is deterministically derived from the normalized SONiC show command.

For example:

show interfaces counters --period=5 detailed --verbose

becomes:

target: SHOW
path: interfaces/counters[period=5]/detailed[verbose=True]

The mapping rules are:

- Command and subcommand tokens become path elements.
- Options with values become gNMI path keys: --period=5 becomes [period=5].
- Valueless options become Boolean keys: --verbose becomes [verbose=True].
- Short and long options are normalized before constructing the path.

gNMI provides the transport and path representation, but the SHOW path itself is not a YANG data-tree path. Each supported path is registered against a server-side handler that retrieves information from different sources and returns command-specific structured JSON.

This is captured in HLD. Please let me know in case you any follow-up question.

michaeloso:
link to recording: https://zoom.us/rec/share/PlgwzldPVQE_ZSjaxWvyErnkF2Z76Ijn4r0YSYV1izoHSvfUCc1tCxsuptAIgZ8.xQQiC4wVgz3Pc6yQ?startTime=1790089039000

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
```diff
@@ -0,0 +1,449 @@
...
+| Key comparison / target | Option 1: CLI as a thin gNMI client | Option 2: CLI integration with gNMI interface via library |
+|---|---|---|
+| Contract enforcement | **Strong** | **Moderate** |
+| Efficiency | **Moderate** | **Strong** |
+| Robustness | **Strong** | **Moderate** |
+| Maintainability | **Strong** | **Moderate** |
+| Security and policy consistency | **Strong** | **Moderate** |
+| Local and remote behavior consistency | **Strong** | **Weak** |
+
+**Rating scale:** Strong = clear advantage; Moderate = acceptable with trade-offs; Weak = significant architectural or operational disadvantage.
```
prvattem, line 449: Why are we creating a second northbound API contract outside YANG?
Because after this proposal, a feature owner may need to maintain:

```
- YANG schema
- RESTCONF
- gNMI model-driven API
- SHOW gNMI API
- CLI formatter
```

That is effectively two gNMI worlds:
```
Model-driven gNMI
      vs
CLI-driven SHOW gNMI
```
The infrastructure is the same, but the contracts are different.

For SONiC feature developers, the biggest risk is:
Configuration/state lives in YANG.
Operational diagnostics live in SHOW gNMI.
Two separate API catalogs must be maintained.

If the feature is accepted - we should have a clear policy such as:
```
Configuration/state
    -> MUST be YANG based

Operational diagnostics
    -> MAY use SHOW gNMI framework

Any operational data already modeled in YANG
    -> MUST NOT be duplicated in SHOW gNMI
```

doc/sonic-show-cmd/SONiC-gNMI-Get-API.md
sharmavijay-ms: Thank you. Based on your comment, I added extra info in HLD for formatter. 

To answer your question/comment:
The output converter does not parse the existing human-readable CLI output. It receives structured JSON returned by the SHOW gNMI handler.

The processing flow is:

gNMI Get response
    -> extract TypedValue json_ietf_val/json_val
    -> decode JSON into a Python dictionary or list
    -> parse the original show command
    -> locate its registered CommandConfig
    -> dispatch to the corresponding formatter
    -> produce the existing SONiC CLI table or text output

If json_ietf_val is used, the client first extracts and JSON-decodes the TypedValue. Because the SHOW payload is not currently YANG-modeled, json_val may be a more accurate encoding designation unless the payload is made RFC 7951 compliant.

Jinja templates are not used by the current converter. The converter uses:

- A command registry and CommandConfig definitions.
- Generic Python formatters for simple table, list, or text responses.
- Command-specific formatter classes for complex output.

Consequently, a separate Jinja template is not required for every command. However, every supported command needs:

1. A registered CLI-command-to-gNMI-path mapping.
2. A defined JSON response shape.
3. A formatter selection.
4. A custom formatter when the existing sonic-utilities command performs command-specific parsing, calculations, sorting, filtering, or layout.

Complex Python behavior cannot be inferred automatically from the JSON. It must either be moved into the server-side Go handler as structured fields or reproduced in a command-specific output formatter. Existing CLI tests should then be run against the formatted gNMI result to ensure output compatibility.

PR #2273, opened 2026-03-31
Add Joint Mode support for SW-controlled CPO

tshalvi:
This PR extends the existing CPO HLDs to support SW-controlled CPO in Joint Mode.

The design reuses the existing CMIS host management architecture, with minimal changes to the generic logic and reliance on platform-level implementations.


doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
```diff
@@ -0,0 +1,242 @@
...
+to include the CPO module type 0x80 (xcvr_api_factory).
+
+```python
+def create_xcvr_api(self):
+    id = self._get_id()
+
+    id_mapping = {
+        0x18: (self._create_cmis_api, ()),
+        0x19: (self._create_cmis_api, ()),
+        0x80: (self._create_cmis_api, (CmisCpoMemoryMap)),
```
bgallagher-nexthop: 0x80 is a vendor-specific code right (one of the vendor codes in SFF-8024)? Should this be keyed off of vendor identifier / part number too?

In addition, one vendor may elect to map the ELSFP EEPROM into a different part of the vCPO module EEPROM than another vendor -- we would require per-vendor memory maps in that scenario, which would also require keying off of some vendor identifier to make sure the correct mem map is used.

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
```diff
@@ -0,0 +1,242 @@
...
+    if self.is_joint_mode:
+        cmis_cpo_manager = CmisCpoManagerTask(...)
+        cmis_cpo_manager.start()
+        self.threads.append(cmis_cpo_manager)
+```
+
+The only functional difference is the set of supported module types. While `CmisManagerTask` handles multiple module types (e.g., QSFP-DD, OSFP, QSFP+), `CmisCpoManagerTask` is restricted to CPO modules only:
+
+```python
+CMIS_MODULE_TYPES = ['CPO']
```
bgallagher-nexthop: CPO hardware could possibly have other non-CPO ports too, right? Maybe a management port or something like that with a CMIS/SFF transceiver.

Is the plan that two state machine threads would be instantiated in this scenario: one CmisCpoManagerTask for CPO ports and one SFF/CMIS state machine thread to handle other ports as appropriate?

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
```diff
@@ -0,0 +1,242 @@
...
+```
+
+This change also introduces a new memory map: **CmisCpoMemoryMap**.
+
+The EEPROM exposes both the standard CMIS data and additional ELS-related information. To support this, CmisCpoMemoryMap extends the existing CMIS memory map by:
+
+- Incorporating ELS-related fields based on the ELSFP specification.
+- Allowing vendor-specific fields to be defined and accessed through this memory map.
+
+Unlike existing generic CMIS memory maps, the CPO memory map supports vendor-specific fields. It accepts an optional dictionary of field definitions at init time, allowing each vendor to inject its own page layouts.
```
bgallagher-nexthop: This is interesting, and may address my above question. Would it be possible to add a brief code example of this, or more explanation to understand what this would look like more clearly?

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
```diff
@@ -0,0 +1,242 @@
...
+
+In Joint Mode, SONiC continues to operate using the existing CMIS host management architecture, using exactly the same xcvrd logic, without introducing changes to the overall control flow.
+
+As a result, the following extensions are needed:
+
+### 7.1 CMIS State Machine Thread for CPO Modules
+
+The CMIS state machine thread is responsible for module configuration. It orchestrates the bring-up of a CMIS transceiver, transitioning it from the inserted state to a ready-for-traffic state.
+
+A new thread, `CmisCpoManagerTask`, will be introduced in `xcvrd` to handle this flow specifically for CPO modules. This thread reuses the existing CMIS logic by inheriting from `CmisManagerTask`, with only minimal adjustments.
```
bgallagher-nexthop: How will `CmisCpoManagerTask` differ from `CmisManagerTask`? Later in this document it is mentioned that the CMIS state machine logic requires no changes. Is the assumption that all joint mode CPO hardware will handle things like the fiber check internally, without changes to software?

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
```diff
@@ -0,0 +1,242 @@
...
+| DOM    | Digital Optical Monitoring |
+
+<br>
+
+
+## 4. Scope
+
+This document defines the **SW-controlled CPO in Joint Mode**, where SONiC sees CPO modules and interacts with them through the CPO abstraction layer, leveraging the existing CMIS-based host management flows.
+
+Note: the **Separate Mode**, where the host system directly interacts with and manages the underlying optical hardware components is defined in *port_mapping_for_cpo.md* section 2.
```
bgallagher-nexthop, line 87: Something to consider: that HLD intends to provide support for both separate and joint modes. The same platform API is exposed for both modes (see "7.3.4 CPO Joint Mode" for more detail).

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
tshalvi: Indeed, 0x80 is a vendor-specific code defined in the spec. In the current design, the API object is instantiated based on the module type identifier only. At this stage, we do not see a need to key this selection off vendor ID or part number as well, since the SONiC handling path is determined by the module type.

Regarding the memory map, this scenario is addressed in Section 7.2 of the updated HLD, where vendor-specific extensions are described.

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
tshalvi: Yes, exactly. The intention is that in platforms that include both CPO ports and other CMIS/SFF-based ports, two state machine threads would be instantiated:

- A dedicated CmisCpoManagerTask for CPO ports only
- The existing CMIS/SFF state machine thread for the other supported transceiver types

In other words, the new CPO thread complements the existing state machine rather than replacing it, so non-CPO modules continue to follow the current flow

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
tshalvi: Good point. The HLD has been updated in the second revision to provide additional details on the memory map structure. Please refer to Section 7.3.2.

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
tshalvi: CmisCpoManagerTask differs from CmisManagerTask only in the set of module types it is responsible for. It reuses the existing CMIS state machine logic, but is dedicated to CPO modules only, while CmisManagerTask remains responsible for non-CPO CMIS modules.
This dedicated task is instantiated only in Joint Mode, controlled by the is_joint_mode flag in pmon_daemon_control.json (Please refer to section 7.1).

When the document mentions that the CMIS state machine logic requires no changes, the intention is that the existing logic can be reused as-is. The separation is introduced to keep CPO handling isolated from pluggable CMIS modules, rather than because CPO requires a different state machine flow.
Fiber check and similar hardware-specific procedures are expected to be handled internally in Joint Mode, without requiring additional changes in SONiC software.

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
```diff
@@ -0,0 +1,356 @@
...
+
+**CPO (Co-Packaged Optics)** is a system architecture in which optical components are integrated directly with the switch ASIC, rather than implemented as external pluggable transceivers (e.g., QSFP-DD, OSFP). This integration reduces electrical trace length and improves overall system power efficiency.
+
+At the hardware level, a CPO module is composed of:
+* **Optical Engine (OE)** — responsible for electrical-to-optical and optical-to-electrical signal conversion.
+* **External Laser Source (ELS)** — providing continuous laser light used by the Optical Engines for transmission.
+
+CPO systems support two operational models:
+* **Separate Mode**, where the host directly accesses and manages the underlying components (e.g., OEs and ELSs). See *port_mapping_for_cpo.md* section 2.
+* **Joint Mode**, where the host interacts with aCPO module abstraction, without directly managing the underlying components.
```
aditya-nexthop: nit: aCPO => a CPO

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
```diff
@@ -0,0 +1,356 @@
...
+
+To preserve compatibility with existing SONiC and SAI workflows, we introduce a **Virtual CPO Module (or vModule)** that mimcs the behavior of a trandtional optical module by providing a logical abstraction that exposes a single unified CMIS interface for both the integrated Optical Engine (OE) and External Laser Sources (ELS).
```
aditya-nexthop: nit: mimcs => mimics, trandtional => traditional

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
aditya-nexthop: Since 0x80 is in the vendor-specific area of the codes, I think it might be important to attach a vendor check of some sort with the code 0x80 lookup. By claiming that 0x80 is CPO-joint mode in community SONiC, we're saying that no other vendor can use 0x80 for anything else, defeating the purpose of it being defined as vendor-specified in the spec.

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
aditya-nexthop: > The separation is introduced to keep CPO handling isolated from pluggable CMIS modules, rather than because CPO requires a different state machine flow.

In this case, do we really need a separate `CmisCpoManagerTask` at all? If there is no difference in the flow or the code, just curious what isolation gets us.

Also, perhaps we should rename `is_joint_mode` to `is_cpo_joint_mode` because we're passing this as an argument to `xcvrd` which is not CPO-specific.

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
```diff
@@ -0,0 +1,356 @@
...
+        )
+```
+
+The `bank_id` parameter (0-3) identifies which 8-lane bank this SFP object represents within the 32-lane CPO module. It is used by `VmoduleCmisApi` for CDB commands that require a lane bank selector.
+
+This change also introduces a new memory map: **`VmoduleCmisMemMap`**, built on the vendor extension framework through a two-layer inheritance hierarchy:
+
+1. **`CmisExtendedMemMap`** (inherits from `CmisMemMap`) - the general-purpose vendor extension layer defined in [Vendor-Specific DOM Extensions for CMIS Modules](https://github.com/sonic-net/SONiC/pull/2291). It accepts an optional `vendor_ext` object at init time and dynamically injects vendor-specific field definitions into the memory map. This layer is not CPO-specific.
+
+2. **`VmoduleCmisMemMap`** (inherits from `CmisExtendedMemMap`) - adds the **ELSFP pages 0x1A/0x1B** (lane monitors, flags, thresholds, setpoints) as defined in the OIF-CMIS-ELSFP specification. This is the CPO vModule-specific layer.
```
aditya-nexthop: Is all the OE and ELSFP information combined in the standard CMIS pages and register locations - Pages 00L, 00H, 01H, 02H? More concretely, how would one expose different identifying information such as vendor names, part number, serial numbers for the OE and ELSFP? How about thresholds, specifically module-level? I would think that the thresholds for the OE would be different than the thresholds for the ELSFP.

The design seems to say that the only additional pages required for joint mode are 0x1A and 0x1B.
Given that there are 2 different devices behind the joint CMIS memory map, how are they able to express all their information through the memory map under this proposal?

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
tshalvi: Thanks for raising this, that’s a valid question.
The separation into CmisCpoManagerTask was introduced following prior design alignment, even though there is currently no difference in the CMIS state machine logic or flow.
I agree that from a purely functional perspective today, a single task could handle both CPO and non-CPO modules. However, since the current direction is to keep them separated, I’ve kept this structure in the design. Maybe @Prince can elaborate.
Regarding the naming- good point. I’ll update is_joint_mode to is_cpo_joint_mode for clarity.

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
tshalvi: Fixed all nits in the latest commit, thanks!

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
Natgerbi: In joint mode, we will use a single CMIS interface for both OE and ELS. The common pages(00L, 00H, 01H, 02H) will primarily expose the OE information, while in our implementation the ELS-specific information will be exposed through the vendor-specific section. In our implementation, pages 0xB0 to 0xB3 effectively replicate the common pages(00L, 00H, 01H, 02H).

aditya-nexthop: I see, thanks for explaining. Please can you share how your proposal would extract information from 0xB0 to 0xB3? Do you intend to just add all fields to the CmisExtendedMemMap as vendor-specific registers and poll them as per your [PR](https://github.com/sonic-net/SONiC/pull/2291) ? That would have two drawbacks:

1. We wouldn't be able to reuse the ELSFP standard CMIS registers in other platforms/designs without code duplication
2. It would be superflous to periodically poll all the static information (identification/thresholds) from 0xB0-0xB3 - we don't poll these pages in the current pluggable model either - we just get them once at the beginning.

Natgerbi: 1.In this design, ELSFP pages 1Ah and 1Bh will still be exposed to all vendors. Only the vendor-specific extensions will be handled through the vendor-specific mechanism. At this stage, we cannot reuse the common ELS pages because we do not know how each vendor will implement them on its own switch. This is why that part is considered vendor-specific.
2.Regarding the handling flow, the idea is not to treat everything as periodically polled vendor-specific data. Instead, we extend the relevant CmisApi functions according to where they are used. From our point of view, it is irrelevant whether the function is used for periodic polling or during a plug-in event.

aditya-nexthop: > ELSFP pages 1Ah and 1Bh will still be exposed to all vendors

Please can you explain how?
From the HLD: VmoduleCmisMemMap (inherits from CmisExtendedMemMap) - adds the ELSFP pages 0x1A/0x1B ... This is the CPO vModule-specific layer.

But VmoduleCmisMemMap is vendor specific for your implementation.

> At this stage, we cannot reuse the common ELS pages because we do not know how each vendor will implement them on its own switch.

Right, but since the ELS pages registers are standard, we can choose to define them as reusable by vendors and then allow vendors to compose them into a memory layout of their choosing based on their implementation as described in https://github.com/sonic-net/SONiC/pull/2207.
Your design effectively declares all registers from the ELSFP memory map - lower, 00h, 01h, 02h, 1Ah, 1Bh as vendor-custom in the VmoduleCmisMemMap even though they all conform to ELSFP-CMIS but just at a different page number, if I am understanding your design correctly.

> From our point of view, it is irrelevant whether the function is used for periodic polling or during a plug-in event.

I see, both this and https://github.com/sonic-net/SONiC/pull/2291 discuss DOM, so it was a bit confusing for me. Thanks for clarifying. Would you be able to add some text and an example or two for the non-DOM getters for completeness?

rookie-who:
Hi @tshalvi — the 202605 branch cut is at end of May. @aditya-nexthop left review comments in April. Please address them or let us know the status.

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
tshalvi: There will be a vendor-check inside the creation of the API object corresponding to module ID 0x80. Is it described in 'CPO API Wiring' section in the updated HLD.

doc/CPO/sw_controlled_cpo_joint_mode_community_hld.md
Natgerbi: the https://github.com/sonic-net/SONiC/pull/2291 is deprecated, the new approach is defined in this HLD.

abelamit (dismissed):

prgeor:
@tshalvi we need a new CPO manager and not reuse the existing CMIS manager for pluggable. That way we can maintain two seperate state machine as I anticipate state machine change between CPO and FRO optics. Eg. CMIS 5.5 or 5.6 we are introducing APSU + ILT which is NOT applicable for CPO

PR #2278, opened 2026-04-01
pmon/xcvr: xcvrd performance improvements HLD

Staphylo:
Propose a HLD to improve the overall performance of xcvrd.
Focusing initially on the DOM polling but opening the door for further enhancement to the management area.

PR lists to follow

### DIFF
--- a/doc/pmon/pmon-xcvrd-improvements.md
+++ b/doc/pmon/pmon-xcvrd-improvements.md
@@ -1,0 +1,486 @@
+# Xcvrd Performance Enhancements
+
+## Table of Content
+
+- [Xcvrd Performance Enhancements](#xcvrd-performance-enhancements)
+  - [Table of Content](#table-of-content)
+    - [1. Revision](#1-revision)
+    - [2. Scope](#2-scope)
+    - [3. Definitions/Abbreviations](#3-definitionsabbreviations)
+    - [4. Overview](#4-overview)
+    - [5. Requirements](#5-requirements)
+    - [6. Architecture Design](#6-architecture-design)
+    - [7. High-Level Design](#7-high-level-design)
+      - [7.1 sonic-platform-common changes](#71-sonic-platform-common-changes)
+        - [7.1.1 SfpBase.get\_io\_group\_name](#711-sfpbaseget_io_group_name)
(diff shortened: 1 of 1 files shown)


prgeor:
@liat-grozovik @eddyk-nvidia please review

doc/pmon/pmon-xcvrd-improvements.md
```diff
@@ -0,0 +1,486 @@
...
+configuration file.
+Proposed `--dom-use-io-threads` as a `boolean`.
+
+Even if it requires a new platform API to be implemented, we want to ensure
+that we can revert to a single thread as a mitigation that can be enabled in
+production.
+
+When starting, the xcvrd process will iterate over all the `SfpBase` objects of
+the platform library and construct a mapping keyed by the output of the new
+`SfpBase.get_io_group_name` platform API.
```
bgallagher-nexthop, line 245: Do you think it would make sense to store the `xcvr -> thread` mapping in configuration somewhere instead of code (`get_io_group_name`)? Maybe it would be better to add configuration in the platform or hwsku directory.

This could also provide the opportunity to re-define the mapping at runtime via config reload as opposed to the current approach requiring a code change, though I'm not sure how useful that would be since the optimal number of threads would probably end up being dictated by hardware topology (number of i2c controllers and how they are spread across the ports).

StormLiangMS:
hi @prgeor could you help to review?

doc/pmon/pmon-xcvrd-improvements.md
```diff
@@ -0,0 +1,486 @@
...
+   respectively
+ - CmisManager to handle newer optics adhering to the CMIS
+ - DomInfoUpdate to poll DOM information from modules
+ - DomThermalInfoUpdate to poll temperature information from modules
+
+Additionally this HLD will focus on a few additional inefficiencies that the
+xcvrd has.
+
+### 5. Requirements
+
```
shubhag-juniper, line 108: Today if we have 64 or 128 CMIS optics, the DPSM/MSM happens serially even though multiple set of optics are grouped under different I2C masters.
Is this HLD addressing the XCVR initialization during insertion or reboot/reload ? OR is it only addressing DOM ?

PR #2284, opened 2026-04-06
[HLD] Add support for PRBS per lane diagnostics

pavannaregundi:
This PR introduces PRBS per lane diagnostics feature in SONiC.

PRs:
| Repo | PR |
| :--- | :--- |
| sonic-swss-common | [Add DIAG_PORT_TABLE in application DB](https://github.com/sonic-net/sonic-swss-common/pull/1179) |
| sonic-swss | [Add PRBS suppot in orchagent and portmgrd](https://github.com/sonic-net/sonic-swss/pull/4510) |
| sonic-utilities | [Add CLI support for PRBS](https://github.com/sonic-net/sonic-utilities/pull/4485) |
| sonic-sairedis | [Add PRBS attribute serialization and deserialization support](https://github.com/sonic-net/sonic-sairedis/pull/1868) |


doc/prbs/PRBS_Per_Lane_HLD.md
sushilkdubey, line 1: What happens if user change port speed while PRBS is running?
- Should the CLI prevent this, or should SAI SDK handle the error?

sushilkdubey, line 1: Why disable RX before TX when stopping PRBS?
- Is this a hardware requirement or avoiding any pending in-flight transactions.

sushilkdubey, line 1: While PRBS is running, user/admin queries STATE_DB PORT_PRBS_RESULTS.
- There could race conditions while updating results to redis, is there any atomic apis planned to handle such race conditions?

sushilkdubey (commented):
Pls find my review comments.

lguohan:
@pavannaregundi , can you also present this in layer-1 subgroup?

doc/prbs/PRBS_Per_Lane_HLD.md
sushilkdubey, line 1: The HLD  proposes per-lane PRBS results, but it does not define how to handle cases where only port-level PRBS is supported.

Should you add a field to indicate whether the platform supports per-lane or port-level PRBS capability as part of pre requisite or cli?

pavannaregundi: Port operation during the PRBS test will interrupt the test and results of prbs will be unreliable. So such configuration should be avoided. CLI does not prevent this as list of checks to do can grow in future. Behaviour of SDK is vendor dependent.

pavannaregundi: RX side is the one doing the result computation. So if TX is disable first RX, no more sees the bit pattern hence will go into LOCK_LOST state. Hence the recommendation.

pavannaregundi:
> @pavannaregundi , can you also present this in layer-1 subgroup?

L1 WG review meeting scheduled for 04/15.

doc/prbs/PRBS_Per_Lane_HLD.md
pavannaregundi: While PRBS is still running PORT_PRBS_RESULTS tables will not yet created, as result capture happens only during the PRBS disable. So there is no race conditions.

pavannaregundi: @sushilkdubey Have updated the HLD to reflect the details. Also please wait for code review to be open, I feel your concerns will be addressed.

doc/prbs/PRBS_Per_Lane_HLD.md
```diff
@@ -0,0 +1,848 @@
+# PRBS Per-Lane Diagnostics Support
```
arpit-nexthop, line 1: How does the scenario play out if the system is reloaded or rebooted with PRBS in the config_db?

Currently we are putting it in the port table, at what time the port will be down (xcvr init can take time) and pollute the results.

Considering PRBS should run on already initialized datapath, do we need this as a configuration command or as an action command.

doc/prbs/PRBS_Per_Lane_HLD.md
```diff
@@ -0,0 +1,848 @@
...
+
+This feature extends SONiC to provide comprehensive PRBS diagnostics with per-lane visibility when the underlying SAI implementation exposes per-lane attributes, and with graceful degradation to port-aggregated results when only port-level PRBS statistics are supported, enabling operators to:
+- Configure PRBS tests on network interfaces with flexible mode (RX/TX/both) and polynomial pattern selection
+- Monitor running PRBS tests with elapsed time tracking
+- Capture and analyze per-lane results (lock status, error counts, and BER per lane) when **get** `SAI_PORT_ATTR_PRBS_PER_LANE_RX_STATE_LIST` succeeds, and port-level lock and error totals from `SAI_PORT_ATTR_PRBS_RX_STATE`
+- Identify lane-specific issues that may be masked in port-level statistics on platforms that report per-lane data
+
+The implementation follows an approach where PRBS tests run continuously, and results are captured only when the test is disabled, ensuring accurate and consistent measurements.
```
aditya-nexthop, line 66: We should have live reporting and not just on disable. It is very useful to see live PRBS data, especially when the test can run for a long time. Based on discussion in the meeting, the suggestion is to add more information to this HLD about how the design can be extended to accommodate this in the future.

doc/prbs/PRBS_Per_Lane_HLD.md
```diff
@@ -0,0 +1,848 @@
...
+Behavior:
+- Without `-i`: deletes all `PORT_PRBS_TEST*`, `PORT_PRBS_RESULTS*`, and `PORT_PRBS_LANE_RESULT*` keys from STATE_DB
+- With `-i <interface>`: validates interface exists in CONFIG_DB PORT table, then deletes `PORT_PRBS_TEST|<interface>`, `PORT_PRBS_RESULTS|<interface>`, and `PORT_PRBS_LANE_RESULT|<interface>*` keys
+- Reports number of records deleted
+
+**Show Commands:**
+
+1. **Show PRBS Status (all ports — no arguments)**
+```bash
+show interface prbs status [--json]
```
aditya-nexthop: It would be useful for summary to include some indication of the loss of lock and BER.
Raw error count is not as useful as the others. Per meeting discussion, we should explore the possibility of deriving this value from per-lane stats.

doc/prbs/PRBS_Per_Lane_HLD.md
```diff
@@ -0,0 +1,848 @@
...
+     5  Locked  LOCK_WITH_ERRORS       700  1.19 × 10⁻⁸   PASS
+     6  Locked  LOCK_WITH_ERRORS       225  1.18 × 10⁻¹⁴  PASS
+     7  Locked  LOCK_WITH_ERRORS      1247  1.26 × 10⁻⁸   PASS
+
+Summary:
+  Total Lanes:      8
+  Locked Lanes:     8/8 (100.0%)
+  Error-Free:       0/8 (0.0%)
+  Lock Status:      Locked
+  Total Errors:     15633
```
aditya-nexthop: Not sure if summary is super useful... we can get this information from the above table.

doc/prbs/PRBS_Per_Lane_HLD.md
```diff
@@ -0,0 +1,848 @@
...
+2. **Show PRBS Status (per-interface detailed view)**
+```bash
+show interface prbs status -i <interface> [--json]
+```
+
+**RX/both mode — per-lane results (after test completed):**
+```
+Interface: Ethernet0 | Mode: rx | Pattern: PRBS31 | Status: Completed | Duration: 00:01:41
+
+  Lane  Lock    RX Status           Errors  BER           Overall
```
aditya-nexthop: "Overall" result not very useful and may differ on a case by case basis... maybe we don't need it at all?

doc/prbs/PRBS_Per_Lane_HLD.md
pavannaregundi: ACK. Will make the changes.

doc/prbs/PRBS_Per_Lane_HLD.md
pavannaregundi: ACK. Will make the changes.

doc/prbs/PRBS_Per_Lane_HLD.md
pavannaregundi: Will upload the sequence diagram explaining the live reporting. This will be for future integration.

doc/prbs/PRBS_Per_Lane_HLD.md
pavannaregundi: @arpit-nexthop thanks for the comment. You are right. We tried the reboot with PRBS is config_db with CMIS capable SFP. We see that error counts are higher compared to normal non-reboot flow. So results seems not accurate in case of reboot.

We did propose the CLI as "debug interface prbs <interface> enable/disable" instead of "config interface prbs  ..." in community meeting, with following flow,

1) Debug CLI will directly write to APPL_DB, to a new table "PORT_DEBUG_TABLE".
              a) portmgrd will subscribe to "PORT_DEBUG_TABLE" and write to APPL_DB - "PORT_TABLE"
      debug CLI → APPL_DB:PORT_DEBUG_TABLE → portmgrd subscribes → writes to APPL_DB:PORT_TABLE
              b) "debug -h" is a supported option in SONiC.
2) Everything downstream is unchanged.
3) PRBS config is transient -  it won't survive config save / reboot.

Do you have any other suggestions?

doc/prbs/PRBS_Per_Lane_HLD.md
pavannaregundi: ACK. Will make changes.

doc/prbs/PRBS_Per_Lane_HLD.md
pavannaregundi: Feel that, it is good to introduce "diag -h" command line(say as "diag interface prbs <>"), as debug does not make much meaning in this case.

arpit-nexthop: I am not sure if such a chain exists in Sonic.

@prgeor @mihirpat1 Thoughts on this?

pavannaregundi: I have pushed the changes in HLD to give clear view. Please review. 
diag CLIs does not exist in SONiC as of today. However i see this is a candidate to introduce this.

pavannaregundi:
Code PRs are updated in the description. Please review the changes.

doc/prbs/PRBS_Per_Lane_HLD.md
pavannaregundi: HLD is updated with this ask.

pavannaregundi:
@prgeor, Can you help us review and get this HLD and code PRs merged.

PR #2294, opened 2026-04-16
Updates to BMC platform management and monitoring spec

judyjoseph:
Updates to pmon design spec based on comments and further discussion with partners.

## Summary

This PR consolidates Switch‑Host power‑control configuration into a unified schema, clarifies platform capability detection for liquid‑cooled systems, and updates default behaviors and CLI documentation for better consistency and predictability.

### Key Changes

*   **Add the RTC clock time synchronization approach**
    *  Plan to use this file "/usr/lib/clock-epoch" which is present when system boots
    * Later the chrony service sync with NTP server configured and updates the system time

*  **Apply the power_on_delay only if the device went through a fullpower cycle**
   * Check the last reboot cause using API get_reboot_cause() 
   * If it is because of Power loss (which I assume is cold boot/POR) - add delay to make sure liquid cooling is good.

*   **Unified CONFIG\_DB schema for Switch‑Host power control**
    *   Merged the previously separate `SWITCH_HOST_POWER_ON_DELAY|default` and `SWITCH_HOST_SHUTDOWN_TIMEOUT|default` tables into a single `CHASSIS_MODULE|SWITCH-HOST` entry.
    *   Introduced an explicit `admin_status` field for the Switch‑Host. The default is set to `down`, ensuring the Switch‑Host remains powered off on initial device boot unless explicitly enabled.

*   **Liquid‑cooling detection cleanup**
    *   Removed the `liquid_cooled=true` flag from `platform.env.conf`, since the same BMC platform/SKU can be reused across air‑cooled and liquid‑cooled chassis.
    *   Added explicit platform APIs to avoid implicit configuration:
        *   `is_bmc()` to identify whether the chassis has a BMC module.
        *   `is_liquid_cooled_chassis()` to indicate liquid or hybrid cooling capability.

*   **Power‑on delay default behavior**
    *   Updated the default `power_on_delay` from `-1` (Switch‑Host stays off) to `0`, so the Switch‑Host powers on immediately when `admin_status` is set to `up`.
    *   This makes the default behavior more intuitive while still preserving explicit administrative control via `admin_status`.

*   **CLI consistency and documentation updates**
    *   Standardized CLI command examples to use `SWITCH-HOST` (uppercase, fixed identifier) instead of the `<switch-host>` placeholder.
    *   Clarified CLI help text and descriptions for `power_on_delay` and `graceful_shutdown_timeout`.
    *   Updated documentation to reflect the new `admin_status` semantics and revised defaults.

### DIFF
--- a/doc/bmc/sonicBMC/pmon-bmc-design.md
+++ b/doc/bmc/sonicBMC/pmon-bmc-design.md
@@ -290,6 +290,16 @@ The Leak detection is applicable only to Liquid cooling platform. The action is 
 The general syslogs will be placed in /var/log/syslog where /var/log directory will be mounted on **tmpfs**. Syslogs will be sent to remote server as well.
 The Leak, Switch-Host state and interactions, Rack-manager interactions will be persistently stored on disk/eMMC in "/host/bmc/event.log" with log rotation enabled.
 
+#### 2.1.7 RTC Clock in BMC
+
+On most vendor platforms, the BMC RTC does not have a battery backup. As a result, the clock does not retain time across power cycles. When the BMC powers on, the system time is initialized using the following priority order:
+
+1. Use the file `/host/image-${IMAGE_VER}/rw/usr/lib/clock-epoch` if available, as the initial time source for the chrony service. (Coming in SONiC release 202605; installed by the switch upgrade service during software upgrade.)
+2. Use the platform API if supported, to retrieve the current time from the switch host to refine or override the initial time.
+3. Finally, the chrony systemd service synchronizes with remote NTP servers to obtain and maintain accurate time on the BMC.
+
+This sequence ensures a reasonable initial timestamp at boot, followed by progressively more accurate time sources.
+
 
 ### 2.2 BMC Platform Management
 
@@ -311,8 +321,9 @@ The bmc controller daemon "bmcctld" is started first in BMC pmon container. It a
 Detailed workflow below
 
 ```
-Sleep for power_on_delay configured in CHASSIS_MODULE|SWITCH-HOST (this is configurable value in config_db)
-This is to make sure the Rack Manager is up and Liquid flow rate is good. 
+if the previous reboot was a Cold Boot (Full Power Cycle)
+  - Sleep for power_on_delay configured in CHASSIS_MODULE|SWITCH-HOST (this is configurable value in config_db)
+  - This is to make sure the Rack Manager is up and Liquid flow rate is good.
 
 Check for any CRITICAL alert/leak in RACK_MANAGER_ALERT* tables or system SYSTEM_LEAK_STATUS table (device_leak_status == CRITICAL_SYSTEM_LEAK) in STATE_DB
 NO External/System LEAK present
@@ -607,6 +618,7 @@ This base class is already defined in sonic-platform-common.
 | Method | Present | Action |
 |---------|---------|----------|
 | get_all_modules() | Y | Fetch managed modules here, Switch-Host Module object |
+| get_reboot_cause()| Y | Fetch previous reboot cause, check if it is Cold Boot(Full Power Cycle) |
 | is_bmc() | New | Retrieves whether the sonic chassis instance is/has a BMC module |
 | is_liquid_cooled() | New | Is this chassis liquid/hybrid cooled ? |
 
--- a/doc/bmc/sonicBMC/pmon-bmc-design.md
+++ b/doc/bmc/sonicBMC/pmon-bmc-design.md
@@ -292,10 +292,12 @@ The general syslogs will be placed in /var/log/syslog where /var/log directory w
 The Leak, Switch-Host state and interactions, Rack-manager interactions will be persistently stored on disk/eMMC in "/host/bmc/event.log" with log rotation enabled.
 
 #### 2.1.7 RTC Clock in BMC
-On most vendor platforms, the BMC RTC does not have a battery backup. As a result, the clock does not retain time across power cycles. When the BMC powers on, the system time is initialized as follows:
+On most vendor platforms, the BMC RTC does not have a battery backup. As a result, the clock does not retain time across power cycles. 
 
-1. Use the clock epoch file at /usr/lib/clock-epoch as the initial system time, if available. This is read by systemd during startup. (Coming in SONiC release 202605)
-2. The chrony sysyemd service when it starts later synchronizes with remote NTP servers to obtain and maintain accurate time.
+When the BMC powers on, the system time is initialized as follows:
+
+1. Use the clock epoch file at "/usr/lib/clock-epoch" as the initial system time, if available. This is read by systemd during startup. (This file is updated regularly with a systemd timer service, this feature is coming in SONiC release 202605)
+2. The chrony systemd service when it starts later synchronizes with remote NTP servers to obtain and maintain accurate time.
 
 This sequence provides a reasonable initial timestamp at boot, followed by synchronization to an accurate time source via NTP.
 
--- a/doc/bmc/sonicBMC/pmon-bmc-design.md
+++ b/doc/bmc/sonicBMC/pmon-bmc-design.md
@@ -17,6 +17,7 @@
       * [2.1.4 BMC-Switch Host Interaction](#214-bmc-switch-host-interaction)
       * [2.1.5 BMC leak_detection_and_thermal policy](#215-bmc-leak-detection-and-thermal-policy)
       * [2.1.6 BMC event logging](#216-bmc-event-logging)
+      * [2.1.7 RTC clock in BMC](#217-rtc-clock-in-bmc)
     * [2.2 BMC Platform Management](#22-bmc-platform-management)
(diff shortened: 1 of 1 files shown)


judyjoseph:
@chander-nexthop f.y.i

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -311,8 +321,9 @@ The bmc controller daemon "bmcctld" is started first in BMC pmon container. It a
 Detailed workflow below
 
 ```
-Sleep for power_on_delay configured in CHASSIS_MODULE|SWITCH-HOST (this is configurable value in config_db)
-This is to make sure the Rack Manager is up and Liquid flow rate is good. 
+if the previous reboot was a Cold Boot (Full Power Cycle)
```
yxieca: how do we determine that this is a full chassis power cycle?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -292,6 +291,16 @@ The Leak detection is applicable only to Liquid cooling platform. The action is
 The general syslogs will be placed in /var/log/syslog where /var/log directory will be mounted on **tmpfs**. Syslogs will be sent to remote server as well.
 The Leak, Switch-Host state and interactions, Rack-manager interactions will be persistently stored on disk/eMMC in "/host/bmc/event.log" with log rotation enabled.
 
+#### 2.1.7 RTC Clock in BMC
+On most vendor platforms, the BMC RTC does not have a battery backup. As a result, the clock does not retain time across power cycles. 
+
+When the BMC powers on, the system time is initialized as follows:
+
+1. Use the clock epoch file at "/usr/lib/clock-epoch" as the initial system time, if available. This is read by systemd during startup. (This file is updated regularly with a systemd timer service, this feature is coming in SONiC release 202605)
```
roger-nexthop, line 299: How often would this be updated?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -626,19 +636,22 @@ CLI to enable user to graceful power on/off the Switch-Host, and to configure po
 Applicable to (LC, AC)
 
 ```
-config chassis modules startup <Switch-Host>
+config chassis modules startup SWITCH-HOST
    - This command is to POWER ON the Switch Host from BMC
+   - Sets the "admin_status to up
```
roger-nexthop: Just admin_status? Or double quotes

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -652,11 +665,13 @@ config chassis modules shutdown-timeout <Switch-Host> <seconds>
 ```
     "CHASSIS_MODULE": {
         "SWITCH-HOST": {
-            "admin_status": "up",
-            "power_on_delay": "300",              ; Time in secs BMC waits before powering on Switch-Host (default = -1, Switch-Host remain powered off)
-            "graceful_shutdown_timeout" : "120"   ; Time in secs BMC waits for graceful shutdown before forcing power-off (default = 120sec)
+            "admin_status": "up",                 ; admin_status up/down; default is down which keeps SWITCH-HOST powered down when device powers up.
```
roger-nexthop, line 668: Is there a show command to display this status for the switch-host? Something like show chassis modules status SWITCH-HOST?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -396,24 +406,22 @@ On an Event
...
+graceful_shutdown_timeout = integer                                            ; Time in secs BMC waits for graceful shutdown before forcing power-off (default = 120sec).
+                                                                               ; if set to 0, BMC will NOT do a graceful shutdown, instead will do POWER_OFF with platform API.
 
 key                       = HOST_STATE|switch-host                             ; STATE_DB on BMC to store state of Switch-Host
 ; field                   = value
-device_power_state        = POWER_ON | POWER_OFF| GRACEFUL_SHUT | POWER_CYCLE  ; What was the last action done on Switch-Host
-device_status             = ONLINE | OFFLINE                                   ; current oper status of device, can use the platform API module->get_oper_state()
+device_power_state        = POWERED_ON | POWERED_OFF | GRACEFUL_SHUTDOWN |     ; Represents the final and transitional power state of Switch-host
+                            POWER_CYCLE | POWERING_ON | POWERING_OFF |
```
roger-nexthop, line 420: To confirm, would GRACEFUL_SHUTDOWN be the state after a request for graceful shutdown completes? When would POWER_CYCLE be set?

doc/bmc/sonicBMC/pmon-bmc-design.md
manapalai: can we define the severity values...

This is what we have
typedef enum {
    SENSOR_OK_NO_LEAK,      /* inside normal band; no leak indication       */
    SENSOR_OK_MILD_LEAK,    /* in mild/advisory leak band                   */
    SENSOR_OK_HIGH_LEAK,    /* in high/critical leak band                   */
    SENSOR_DISCONNECTED,    /* sensor absent / open-circuit / not wired     */
    SENSOR_SHORT            /* electrical short / invalid ADC/faulted range */
} leak_sensor_state_t;

in general - let us define input/output of all these APIs

judyjoseph: @manapalai This already maps to what is defined today in platform API 

https://github.com/sonic-net/sonic-platform-common/blob/41af585ee6e359ea9adf6c5f9f422bbfd750f5bb/sonic_platform_base/liquid_cooling_base.py#L85

            LeakSeverity: LeakSeverity.CRITICAL or LeakSeverity.MINOR, or None if no leak

The other two should not come as a leak severity enum, it should just tell if sensor of faulty or not, mapped to another platform API

    def is_leak_sensor_ok(self) -> bool:

oleksandrivantsiv:
@Yakiv-Huryk, please review

doc/bmc/sonicBMC/pmon-bmc-design.md
manapalai: how it will tell sensor is faulty? we call this get_leak_severity(), this should return a value, say 'FAULT' or something?

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: @manapalai @vivekverma-arista @chander-nexthop Could you check on your platforms if we can identify the reboot cause of "POWER-LOSS", trying to see if we can identify the device is gone through a power cycle/cold-boot

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: it can be set based on usecase .. timer could be like daily. So even if there is a reboot -- the NTP/chrony should be able to take care of that jump !

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: yes I have updated it explicitly

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: We can check for is_leak_sensor_ok() first and based on sensor ok/faulty take or discard the get_leak_severity data ? Could you check this PR : https://github.com/sonic-net/sonic-platform-daemons/pull/776

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: yes. POWER_CYCLE can be send as a rack manager command - and we do a Switch-Host power cycle using API -- 

```
key                       = RACK_MANAGER_COMMAND|CMD_<command_id>         ; Commands from Rack Manager in STATE_DB in BMC
; field                   = value                                         ; e.g. ComputerSystem.Reset
command                   = POWER_ON | POWER_OFF | GRACEFUL_SHUT| POWER_CYCLE
```

<html>
<body>

do_power_cycle() | New | Performs a hardware power cycle of the Switch Host.




</body>
</html>

doc/bmc/sonicBMC/pmon-bmc-design.md
judyjoseph: @manapalai @oleksandrivantsiv @vivekverma-arista @fraserg-arista @chander-nexthop @roger-nexthop  I have also updated the wording a bit in latest commit -- to make sure it is clear that we take zone/location of leak sensor into consideration when getting the leak severity

Blueve (approved):

judyjoseph (commented):
NOT ready to merge yet,  as there could be some change in the initial state if Switch when the switch is powered on in Buildout

judyjoseph:
> NOT ready to merge yet, as there could be some change in the initial state if Switch when the switch is powered on in Buildout

Going ahead and merge as we are missing critical upates to the original design. Will create a followup PR as needed

PR #2297, opened 2026-04-16
HLD: CPO DOM Monitoring

bgallagher-nexthop:
Add a HLD that describes how the CPO Platform API introduced in https://github.com/sonic-net/SONiC/pull/2211 can be used to add DOM support to `xcvrd` for CPO hardware with a minimal amount of changes.

The proposal here will support both separate and joint mode CPO hardware platforms, since it leverages the platform API that supports both modes.


Natgerbi:
I think you should add some comment in the HLD that "OE pages"(lower pages, 00h,01h etc) are contain information also for the whole vmodule.

bgallagher-nexthop:
Closing this because it will be superseded by this HLD: https://github.com/sonic-net/SONiC/pull/2444

PR #2298, opened 2026-04-17
[HLD] Add support for port loopback

Keshavg-marvell:
Add support for port loopback


doc/port-loopback/Port_Loopback_HLD.md
```diff
@@ -0,0 +1,421 @@
+# Port Loopback Mode in SONiC
```
arpit-nexthop, line 1: Can we include gearbox also in the design, to avoid another HLD later. That way we can conclude on the CLI and basic methodology

Keshavg-marvell:
Capturing comments from community L1 meeting:
From @prgeor @eddyk-nvidia Port loopback capability may vary across individual ports and therefore should be published at the per‑port level, rather than at the switch level. The current mechanism using sai_query_attribute_enum_values_capability for SAI_PORT_ATTR_LOOPBACK_MODE exposes switch‑level capabilities

Notes from @keshav. Currently, SAI does not define a port level capability attribute for loopback modes, unlike other port capabilities such as FEC mode, which are exposed via attributes like SAI_PORT_ATTR_SUPPORTED_FEC_MODE. so 
  this require defining a new port attribute in SAI.

Keshavg-marvell:
Notes from @prgeor @eddyk-nvidia Querying capability per port on large-radix switches (128, 512, or even 1K ports) would be expensive, and suggested exploring to use SAI-level bulk capability query that returns all port capabilities in one call.

PR #2299, opened 2026-04-17
[HLD] User Defined Fields(UDF) feature in SONiC

srodd-nexthop:
Description:

Adds HLD for the UDF feature in SONiC, covering:

CONFIG_DB schema for UDF and UDF_SELECTOR tables
UdfOrch orchestration for SAI UDF Group, Match, and UDF object lifecycle
ACL integration with udf_field name resolution
YANG model validation, match deduplication, and dependency ordering

Yang changes: https://github.com/sonic-net/sonic-buildimage/pull/26874
Swss changes: https://github.com/sonic-net/sonic-swss/pull/4493
sairedis changes: https://github.com/sonic-net/sonic-sairedis/pull/1856
sonic-mgmt tests: https://github.com/sonic-net/sonic-mgmt/pull/24518

### DIFF
--- a/doc/udf/UDF_HLD.md
+++ b/doc/udf/UDF_HLD.md
@@ -8,7 +8,7 @@
   - [3.1 Quick Reference](#31-quick-reference)
 - [4. Core Concepts and Mental Model](#4-core-concepts-and-mental-model)
   - [4.1 What is UDF?](#41-what-is-udf)
-  - [4.2 The Three UDF Objects - Mental Model](#42-the-three-udf-objects---mental-model)
+  - [4.2 The Two UDF Config Objects - Mental Model](#42-the-two-udf-config-objects---mental-model)
   - [4.3 Packet Parsing Flow - Complete Example](#43-packet-parsing-flow---complete-example)
   - [4.4 Design Patterns and ACL Key Mapping](#44-design-patterns-and-acl-key-mapping)
   - [4.5 Group Index Mapping to ACL Key](#45-group-index-mapping-to-acl-key)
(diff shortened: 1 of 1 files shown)


doc/udf/UDF_HLD.md
```diff
@@ -0,0 +1,686 @@
...
+  // ACL integration
+  "ACL_TABLE_TYPE": { "ROCE_ACL_TYPE":         { "matches": ["IN_PORTS","BTH_RESERVED"] } },
+  "ACL_TABLE":      { "ROCE_TABLE":            { "type": "ROCE_ACL_TYPE", "stage": "ingress" } },
+  "ACL_RULE":       { "ROCE_TABLE|BLOCK_RSVD": { "BTH_RESERVED": "0x00/0x7F", "PACKET_ACTION": "DROP" } }
+}
+```
+
+### Object Roles(one-liner)
+
+- **UDF** = WHERE and how big (udf_field position/length in ACL key)
```
bingwang-ms, line 84: Where is `position` defined?

doc/udf/UDF_HLD.md
```diff
@@ -0,0 +1,686 @@
...
+
+### How It Fits Together
+
+Reading from ACL back to the packet — each layer answers one question:
+
+```json
+{
+  // UDF definition
+  "UDF":            { "BTH_RESERVED":          { "field_type": "GENERIC", "length": "1" } },
+  "UDF_SELECTOR":   { "BTH_RESERVED|roce_v1":  { "match": { "l2_type": "0x8915", "l2_type_mask": "0xFFFF", "priority": "100" }, "select": { "base": "L2", "offset": "22" } },
```
bingwang-ms: I suppose `base` defined the start point of UDF object. But I think `L2`, `L3` or `L4` could be misleading sometimes. For example, there could be encapsulated frames, including VXLAN, IPinIP, and different IP hearers.

doc/udf/UDF_HLD.md
```diff
@@ -0,0 +1,686 @@
...
+                      "BTH_RESERVED|roce_v2":  { "match": { "l3_type": "0x11",   "l3_type_mask": "0xFF",   "priority": "100" }, "select": { "base": "L4", "offset": "16" } } },
```
bingwang-ms: Can you please clarify how `match` and `select` work together? 
I suppose `match` + `select` = `When this packet type appears, extract bytes from HERE`. Is that correct?

doc/udf/UDF_HLD.md
```diff
@@ -0,0 +1,703 @@
...
+2. **UDF_SELECTOR** entries under `BTH_RESERVED` define when/how to extract bytes for each packet type
+3. **ACL_TABLE_TYPE** declares `BTH_RESERVED` in matches → attaches SAI UDF group to ACL table
+4. **ACL_RULE** matches against `BTH_RESERVED` directly by udf_field name
+
+### 5.3 Component Responsibilities
+
+| Component | Location | Responsibilities |
+|-----------|----------|------------------|
+| **UdfOrch** | `udforch.cpp/h` | • Manages UDF and UDF_SELECTOR objects<br/>• For each UDF_SELECTOR entry, creates both a SAI UDF Match and SAI UDF object internally<br/>• Resolves udf_field names to SAI OIDs via `getUdfGroupOid()`<br/>• Tracks ACL table references via `incrementGroupRefCount()` / `decrementGroupRefCount()`<br/>• Blocks udf_field removal when ACL ref count > 0 |
+| **AclOrch** | `aclorch.cpp` | • Resolves udf_field names in ACL_TABLE_TYPE matches to SAI OIDs<br/>• Attaches udf_fields to ACL tables and calls `incrementGroupRefCount()`<br/>• Calls `decrementGroupRefCount()` when an ACL table is removed<br/>• Applies UDF matching in ACL rules using udf_field name directly |
```
bingwang-ms, line 187: Since UDF is highly platform dependent, I suggest adding a capability check in aclorch.

doc/udf/UDF_HLD.md
bingwang-ms: What's the difference between `GENERIC` and `hash`?

doc/udf/UDF_HLD.md
```diff
@@ -0,0 +1,703 @@
...
+**File**: `sonic-udf.yang`
+
+| Table | Validation | Constraints |
+|-------|------------|-------------|
+| **UDF** | `field_type` (enum: GENERIC/HASH)<br/>`length` (1-20) | Mandatory fields |
+| **UDF_SELECTOR** | **match**: `l2_type`/`l3_type`/`gre_type`/`l4_dst_port` (hex string), masks, `priority` (0-255)<br/>**select**: `base` (enum: L2/L3/L4), `offset` (0-255) | `select.base` and `select.offset` mandatory; at least one of `l2_type`/`l3_type`/`gre_type`/`l4_dst_port` required; parent UDF entry must exist |
+
+### 8.2 CLI
+
+CLI configuration for UDF is **not implemented**. Configuration is applied directly via CONFIG_DB using `sonic-cfggen` or `redis-cli`:
```
bingwang-ms: GCU (Generic Config Update) should be the preferred mechanism if CLI is not in plan.

doc/udf/UDF_HLD.md
```diff
@@ -0,0 +1,703 @@
...
+    "ROCE_TABLE|BLOCK_OPCODE":{"priority": "100", "BTH_OPCODE": "0x04/0xFF",   "PACKET_ACTION": "DROP"}
+  }
+}
+```
+
+### 6.7 Design Rationale
+
+#### Why `select` and `match` are JSON values rather than flat fields
+
+CONFIG_DB stores each row as a flat Redis hash — one key, string-valued fields. Modeling `select` and `match` as nested containers would require splitting them into separate tables (e.g., `UDF_SELECTOR_SELECT`, `UDF_SELECTOR_MATCH`) keyed by the same `<udf>|<selector>`. We chose to keep the selector as a single atomic row so that a SET/DEL on one key represents one complete selector, and `match`/`select` cannot drift out of sync across tables. The tradeoff is that their internal structure lives inside a JSON value instead of the YANG tree.
```
bingwang-ms: Need to confirm and verify this `json-in-config_db` model is supported by GCU and other CLI, like sonic-cfggen.
YANG model validation can be another challenge as the json value will be handled as a pure string.

doc/udf/UDF_HLD.md
```diff
@@ -0,0 +1,703 @@
...
+### How It Fits Together
+
+Reading from ACL back to the packet — each layer answers one question:
+
+```json
+{
+  // UDF definition
+  "UDF":            { "BTH_RESERVED":          { "field_type": "GENERIC", "length": "1" } },
+  "UDF_SELECTOR":   { "BTH_RESERVED|roce_v1":  { "match": { "l2_type": "0x8915", "l2_type_mask": "0xFFFF", "priority": "100" }, "select": { "base": "L2", "offset": "22" } },
+                      "BTH_RESERVED|roce_v2":  { "match": { "l3_type": "0x11",   "l3_type_mask": "0xFF",   "priority": "100" }, "select": { "base": "L4", "offset": "16" } } },
```
bingwang-ms: May need some clarification on the `priority`, including
1. Is higher number = higher priority or lower?
2. What happens when two selectors have the SAME priority and both match?
3. Is priority globally unique required, or just per-field?

doc/udf/UDF_HLD.md
srodd-nexthop: "Position" here means the slot index in the ACL key where the extracted udf_field bytes are written, not a property of the UDF object itself. Hence its called out as `position in acl key`

doc/udf/UDF_HLD.md
srodd-nexthop: the base here are looking for the outer headers and for inner header we have to rely on the offset. Its been documented as an example in `sonic-sairedis/SAI/doc/ACL/UDF-based-ACL.md`

doc/udf/UDF_HLD.md
srodd-nexthop: yes, match acts as a filter and select acts as a pointer from where the field has to be extracted.

doc/udf/UDF_HLD.md
srodd-nexthop: GENERIC is used for ACL integration 
HASH is to define fields for ECMP hashing.

doc/udf/UDF_HLD.md
srodd-nexthop: Thanks for the question, I'll validate the behavior on HW and update the HLD with concrete answers.

doc/udf/UDF_HLD.md
srodd-nexthop: Thanks — your json-in-config_db comment pushed me to drop the JSON-string model. select and match are now flat CONFIG_DB fields (select_base, select_offset, match_l2_type, …), so GCU, sonic-cfggen, and YANG all work natively.

doc/udf/UDF_HLD.md
```diff
@@ -0,0 +1,721 @@
...
+
+## 10. Testing Requirements/Design
+
+### 10.1 Unit Tests
+- UdfGroup: Create/remove, validation (length, type)
+- UdfMatch: L2/L3/GRE matching, priority validation
+- Udf: Dependency handling, offset/base validation
+- UdfOrch: CONFIG_DB subscription, group OID resolution, ref-count tracking
+
+### 10.2 System Tests
```
prsunny, line 648: can you elaborate this with scenarios and expectation. Also, please update PR link for sonic-mgmt tests. 

e.g - https://github.com/sonic-net/SONiC/blob/master/doc/vxlan/Overlay%20ECMP%20with%20BFD.md#281-overlay-ecmp

srodd-nexthop: Thanks for the feedback @prsunny  I've updated the section with scenarios and expected results, and linked the sonic-mgmt tests in the PR description. Please take another look when you get a chance.

doc/udf/UDF_HLD.md
srodd-nexthop: Outdated. The `json-in-config_db` model was dropped (see the line 563 thread). Schema is now flat CONFIG_DB fields (`select_base`, `select_offset`, `match_l2_type`, `match_l3_type`, `match_gre_type`, `match_l4_dst_port`, `match_priority`), which GCU, sonic-cfggen, and YANG handle natively.

doc/udf/UDF_HLD.md
srodd-nexthop: Covered in §6.4 "Platform Capability Check (UdfOrch ctor)" — UdfOrch probes `sai_query_attribute_capability(SAI_OBJECT_TYPE_UDF_GROUP, SAI_UDF_GROUP_ATTR_TYPE)` at startup. If unsupported, config events are flushed and a NOTICE is logged.

PR #2301, opened 2026-04-20
Update link up sequence with SI settings sync between OA and xcvrd

arpit-nexthop:
Updated Interface link up sequence with SI settings sync between orchagent and xcvrd

| [[sonic-swss] [orchagent]: SI settings sync protocol between xcvrd and orchagent](https://github.com/sonic-net/sonic-swss/pull/4497) | GitHub issue/pull request detail | GitHub pull request check contexts |
| [SI settings sync protocol between xcvrd and orchagent](https://github.com/sonic-net/sonic-platform-daemons/pull/799) | GitHub issue/pull request detail | GitHub pull request check contexts |

### DIFF
--- a/doc/sfp-cmis/Interface-Link-bring-up-sequence.md
+++ b/doc/sfp-cmis/Interface-Link-bring-up-sequence.md
@@ -36,7 +36,7 @@ Deterministic Approach for Interface Link bring-up sequence
 | 0.8 | 02/16/2022  | Shyam Kumar                        | Updated feature-enablement workflow
 | 0.9 | 04/05/2022  | Shyam Kumar                        | Addressed review comments           |
 | 1.0 | 10/20/2023  | Mihir Patel                        | Added Port re-initialization during syncd/swss/orchagent crash section           |
-| 1.1 | 04/19/2026  | Arpit                              | Updated SI settings sync: two-field protocol (APPL_DB si_settings_notification + STATE_DB si_settings_ack), monotonic counter, CMIS_STATE_SI_SETTINGS_WAIT rename, timeout behavior |
+| 1.1 | 04/28/2026  | Arpit Kumar                        | Redesigned SI settings sync: two-field protocol (APPL_DB si_settings_notification + STATE_DB si_settings_ack), monotonic counter, CMIS_STATE_SI_SETTINGS_WAIT state; notification moved to CMIS_STATE_INSERTED and SffManagerTask gated by notify_si_settings flag; SI_SETTINGS_WAIT timeout triggers force_cmis_reinit |
 
 
 # About this Manual
@@ -225,7 +225,7 @@ A transceiver is classified as CMIS SM driven transceiver if its module type is 
 | Value                        | Written by                                                                                                                                                                          | Purpose                                                                                   |
 | ---------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------- |
 | `SI_SETTINGS_DEFAULT:<N>`    | 1\. SfpStateUpdateTask during transceiver removal (increments N)<br>2. xcvrd during boot-up initialization                                                                          | Signals to OA that transceiver was removed or SI settings are in default state             |
-| `SI_SETTINGS_NOTIFIED:<N>`   | 1\. SfpStateUpdateTask for non-CMIS SM driven transceivers<br>2. CmisManagerTask for CMIS SM driven transceivers (written in CMIS_STATE_AP_CONF)                                   | Signals to OA that SI settings have been written to PORT_TABLE (APPL_DB) for sequence N   |
+| `SI_SETTINGS_NOTIFIED:<N>`   | 1\. SffManagerTask for non-CMIS SM driven transceivers (on module insertion)<br>2. CmisManagerTask for CMIS SM driven transceivers (written in CMIS_STATE_INSERTED when `notify_si_settings=True`) | Signals to OA that SI settings have been written to PORT_TABLE (APPL_DB) for sequence N   |
 
 **STATE_DB PORT_TABLE `si_settings_ack`** — written by Orchagent, read by xcvrd  
 
@@ -239,10 +239,13 @@ A transceiver is classified as CMIS SM driven transceiver if its module type is 
 2. Update and notify SI settings to OA  
       The API `is_npu_si_settings_update_required` will return true if a module requires SI settings and `si_settings_notification` in PORT_TABLE (APPL_DB) is absent or equals `SI_SETTINGS_DEFAULT:<N>`. It will return false if `si_settings_notification == SI_SETTINGS_NOTIFIED:<N>` (i.e., already notified in this session).  
 
-      For non-CMIS SM driven transceivers, if `is_npu_si_settings_update_required` returns true, SfpStateUpdateTask will update SI settings in PORT_TABLE (APPL_DB) and write `si_settings_notification = SI_SETTINGS_NOTIFIED:<N>` to PORT_TABLE (APPL_DB). OA will be notified via the ConsumerStateTable mechanism.  
+      For non-CMIS SM driven transceivers, if `is_npu_si_settings_update_required` returns true, SffManagerTask will update SI settings in PORT_TABLE (APPL_DB) and write `si_settings_notification = SI_SETTINGS_NOTIFIED:<N>` to PORT_TABLE (APPL_DB) on module insertion. OA will be notified via the ConsumerStateTable mechanism.
 
-      For CMIS SM driven transceivers, if `is_npu_si_settings_update_required` returns true, CmisManagerTask will update SI settings in PORT_TABLE (APPL_DB) and write `si_settings_notification = SI_SETTINGS_NOTIFIED:<N>` to PORT_TABLE (APPL_DB) during CMIS_STATE_AP_CONF.  
-      The CMIS SM will then transition from CMIS_STATE_AP_CONF to CMIS_STATE_SI_SETTINGS_WAIT. If port doesn't require SI settings, CMIS SM will transition to CMIS_STATE_DP_INIT state directly.  
+      For CMIS SM driven transceivers, if `is_npu_si_settings_update_required` returns true, CmisManagerTask will update SI settings in PORT_TABLE (APPL_DB) and write `si_settings_notification = SI_SETTINGS_NOTIFIED:<N>` to PORT_TABLE (APPL_DB) during CMIS_STATE_INSERTED. This notification is gated by a `notify_si_settings` boolean flag in port_dict, which is set to True only on two events:
+      - Module insertion (TRANSCEIVER_INFO SET event from STATE_DB)
+      - SI_SETTINGS_WAIT timeout failure (force reinit after failed ack)
+
+      The flag is reset to False immediately after the notification fires, preventing re-notification on config-change reinits or state-machine retries. The CMIS SM will then transition through CMIS_STATE_DP_DEINIT → CMIS_STATE_AP_CONF → CMIS_STATE_SI_SETTINGS_WAIT. If SI settings were not notified (flag was False), CMIS SM will transition to CMIS_STATE_DP_INIT state directly from CMIS_STATE_AP_CONF.
 
 
 
@@ -253,8 +256,8 @@ A transceiver is classified as CMIS SM driven transceiver if its module type is 
 		- In case of failure, OA will log an error message and proceed to handling the next port
 	- If `si_settings_notification == SI_SETTINGS_DEFAULT:<N>` is received (transceiver removed), OA writes `si_settings_ack = SI_SETTINGS_DEFAULT:<N>` to STATE_DB (no serdes programming)
 
-4. CMIS_STATE_SI_SETTINGS_WAIT state will wait for `SI_SYNC_DONE:<N>` in STATE_DB `si_settings_ack`, where N must match the notification number sent. Upon match, CMIS SM transitions to CMIS_STATE_DP_INIT state.  
-There will be a timeout of 10s; upon timeout, CMIS SM proceeds to CMIS_STATE_DP_INIT with a warning (no reinit)
+4. CMIS_STATE_SI_SETTINGS_WAIT state will wait for `SI_SYNC_DONE:<N>` in STATE_DB `si_settings_ack`, where N must match the notification number sent. Upon match, CMIS SM transitions to CMIS_STATE_DP_INIT state.
+There will be a timeout of 10s; upon timeout, CmisManagerTask resets `si_settings_notification` to `SI_SETTINGS_DEFAULT:<N+1>` in APPL_DB, sets `notify_si_settings=True`, and calls `force_cmis_reinit(retries+1)`, transitioning back to CMIS_STATE_INSERTED. This consumes a retry (max 3) and causes re-notification on the next INSERTED entry. After exhausting retries, the SM transitions to CMIS_STATE_FAILED.
 
 5. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED” to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
 
@@ -312,11 +315,14 @@ sequenceDiagram
     end
 ```
 
-## SfpStateUpdateTask's role to notify SI settings to OA during xcvrd boot-up
+## SI settings notification during xcvrd boot-up
+
+SI settings notification is no longer performed by SfpStateUpdateTask. It is now split between CmisManagerTask (for CMIS SM driven transceivers) and SffManagerTask (for non-CMIS transceivers), triggered on module insertion events rather than during boot-up initialization.
+
+SfpStateUpdateTask is responsible only for posting transceiver info and DOM thresholds to the DB during boot-up:
 
 ```mermaid
 sequenceDiagram
-    participant OA
     participant APPL_DB as APPL_DB@asic_n
     participant STATE_DB as STATE_DB@asic_n
     participant SfpStateUpdateTask
@@ -325,14 +331,7 @@ sequenceDiagram
     Note over SfpStateUpdateTask: Following loop represents _post_port_sfp_info_and_dom_thr_to_db_once
     loop lport in logical_port_list
         alt post_port_sfp_info_to_db != SFP_EEPROM_NOT_READY
-             Note over SfpStateUpdateTask: post_port_dom_threshold_info_to_db
-            opt if not is_module_cmis_sm_driven and is_npu_si_settings_update_required
-                SfpStateUpdateTask ->> APPL_DB: Update SI params from media_settings.json to PORT_TABLE:<lport><br>si_settings_notification = SI_SETTINGS_NOTIFIED:<N>
-                APPL_DB -->> OA: Notify SI settings for ports (via ConsumerStateTable)
-                Note over OA: Disable admin status (update_host_tx_ready=false)<br>setPortSerdesAttribute
-                OA ->> STATE_DB: PORT_TABLE|<lport>.si_settings_ack = SI_SYNC_DONE:<N>
-                Note over OA: initHostTxReadyState
-            end
+            Note over SfpStateUpdateTask: post_port_dom_threshold_info_to_db
         else
             Note over SfpStateUpdateTask: retry_eeprom_set.add(lport)
         end
@@ -340,6 +339,34 @@ sequenceDiagram
     Note over SfpStateUpdateTask: _init_port_sfp_status_tbl<br>Subscribe to CONFIG_DB
     loop Wait for stop_event
         SfpStateUpdateTask ->> SfpStateUpdateTask: Handle config change event<br>retry_eeprom_reading()<br>_wrapper_get_transceiver_change_event
+    end
+```
+
+## SffManagerTask's role to notify SI settings to OA (non-CMIS modules)
+
+For non-CMIS SM driven transceivers, SffManagerTask notifies SI settings to OA upon module insertion:
+
+```mermaid
+sequenceDiagram
+    participant OA
+    participant APPL_DB as APPL_DB@asic_n
+    participant STATE_DB as STATE_DB@asic_n
+    participant SffManagerTask
+
+    Note over SffManagerTask: Subscribe to CONFIG_DB:PORT,<br>STATE_DB:TRANSCEIVER_INFO and STATE_DB:PORT_TABLE
+    loop Wait for stop_event
+        SffManagerTask ->> SffManagerTask: handle_port_update_event
+        opt xcvr_inserted (TRANSCEIVER_INFO SET) and not is_cmis_api
+            SffManagerTask ->> SffManagerTask: sfp.get_transceiver_info()
+            opt is_npu_si_settings_update_required
+                SffManagerTask ->> APPL_DB: Update SI params from media_settings.json to PORT_TABLE:<lport><br>si_settings_notification = SI_SETTINGS_NOTIFIED:<N>
+                APPL_DB -->> OA: Notify SI settings (via ConsumerStateTable)
+                Note over OA: Disable admin status (update_host_tx_ready=false)<br>setPortSerdesAttribute
+                OA ->> STATE_DB: PORT_TABLE|<lport>.si_settings_ack = SI_SYNC_DONE:<N>
+                Note over OA: initHostTxReadyState
+            end
+        end
+        Note over SffManagerTask: TX enable/disable based on host_tx_ready and admin_status
     end
 ```
 
@@ -355,6 +382,12 @@ stateDiagram-v2
     state if_state3 <<choice>>
     state if_state4 <<choice>>
     CMIS_STATE_INSERTED --> if_state
+    note right of CMIS_STATE_INSERTED
+        If notify_si_settings=True (module insertion or SI failure reinit):
+          Read transceiver info via api
+          Write SI params + si_settings_notification = SI_SETTINGS_NOTIFIED[N] to APPL_DB
+          Reset notify_si_settings = False
+    end note
     if_state --> CMIS_STATE_READY : if host_tx_ready != True or admin_status != up - Action disable TX
     if_state --> if_state2 : if host_tx_ready == True and admin_status == up
     if_state2 --> if_state4 : if appl >= 1 and host_lanes_mask > 0 and SI_lanes_mask > 0
@@ -369,19 +402,20 @@ stateDiagram-v2
     note left of CMIS_STATE_AP_CONF
         Ensure ModuleReady and DataPathDeactivated
         Configure ZR laser frequency
-        Apply SI settings from media_settings.json
-        Write si_settings_notification = SI_SETTINGS_NOTIFIED[N] to APPL_DB PORT_TABLE
+        Apply optics SI settings (optics_si_settings.json) via stage_custom_si_settings
         Call set_application
     end note
     CMIS_STATE_AP_CONF --> if_state3
-    if_state3 --> CMIS_STATE_SI_SETTINGS_WAIT : si_settings_notification == SI_SETTINGS_NOTIFIED[N]
+    if_state3 --> CMIS_STATE_SI_SETTINGS_WAIT : si_settings_notification == SI_SETTINGS_NOTIFIED[N] in APPL_DB
     if_state3 --> CMIS_STATE_DP_INIT : si_settings_notification not set or == SI_SETTINGS_DEFAULT
     CMIS_STATE_SI_SETTINGS_WAIT --> CMIS_STATE_DP_INIT : STATE_DB si_settings_ack == SI_SYNC_DONE[N]
-    CMIS_STATE_SI_SETTINGS_WAIT --> CMIS_STATE_DP_INIT : 10s timeout - proceed to DP_INIT with warning
+    CMIS_STATE_SI_SETTINGS_WAIT --> CMIS_STATE_INSERTED : 10s timeout - reset si_settings_notification to DEFAULT,\nset notify_si_settings=True, force_cmis_reinit(retries+1)
     note right of CMIS_STATE_SI_SETTINGS_WAIT
         Polls STATE_DB si_settings_ack for SI_SYNC_DONE[N]
         N must match the notification number sent in APPL_DB
-        On 10s timeout proceed to DP_INIT with warning
+        On 10s timeout: reset APPL_DB to SI_SETTINGS_DEFAULT[N+1],
+        set notify_si_settings=True, reinit with retries++
+        After CMIS_MAX_RETRIES exhausted: transition to CMIS_STATE_FAILED
     end note
     CMIS_STATE_DP_INIT --> CMIS_STATE_DP_TXON
     CMIS_STATE_DP_TXON --> CMIS_STATE_DP_ACTIVATE
@@ -396,11 +430,12 @@ sequenceDiagram
     participant OA
     participant APPL_DB as APPL_DB@asic_n
     participant CmisManagerTask
+    participant SffManagerTask
     participant SfpStateUpdateTask
 
     SfpStateUpdateTask ->> SfpStateUpdateTask : event = SFP_STATUS_REMOVED
     SfpStateUpdateTask -x STATE_DB : Delete TRANSCEIVER_INFO table for the port
-    par         CmisManagerTask, SfpStateUpdateTask
+    par CmisManagerTask, SfpStateUpdateTask
         CmisManagerTask ->> CmisManagerTask : Transition CMIS SM to CMIS_STATE_REMOVED
         SfpStateUpdateTask ->> APPL_DB : PORT_TABLE|<lport>.si_settings_notification = <br> SI_SETTINGS_DEFAULT:<N+1>
     end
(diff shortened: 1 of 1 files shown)


mihirpat1 (commented):
@arpit-nexthop Can you please update the PR description with the table similar to the below so that we can maintain the list of relevant code changes for this HLD change?
https://github.com/sonic-net/SONiC/pull/1432#issue-1819681969

benle7:
Hi @arpit-nexthop 

With this new design, do we prevent the admin state (and host TX signal) from coming up
until the host SI settings are configured?
If so, did you verify it in the ‘swss restart’ flow as well?

Thanks

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -193,10 +194,13 @@ if transceiver is not present:
...
-  Also, this key helps in preventing additional link flap which can happen if CMIS state machine initializes a port before NPU SI settings are applied (since application of NPU SI settings involves disable followed by enable of it).  
+  - SI settings synchronization uses a two-field protocol to communicate between xcvrd and Orchagent (OA):
+    - `si_sync_status` in PORT_TABLE (APPL_DB): written by xcvrd to signal SI settings state. OA reads this via its normal ConsumerStateTable flow. Values: `SI_SETTINGS_NOTIFIED:<N>` (xcvrd has notified SI settings for sequence N), `SI_SETTINGS_DEFAULT:<N>` (transceiver removed or no SI settings needed for sequence N).
```
mihirpat1: @arpit-nexthop Can you explain in the HLD on why we are using APPL_DB for `si_sync_status` whereas STATE_DB for `si_settings_sync_status`?

mihirpat1: @arpit-nexthop Can you please rename the fields and corresponding value names for the below? The current names are bit confusing.
1. si_sync_status -> Can change to something like si_settings_notification or anything more meaningful
2. si_settings_sync_status -> Can change to something like si_settings_ack or anything more meaningful

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -207,46 +211,56 @@ Pre-requisites for the infra to work:
...
-	- Disable port admin status
-	- Request SAI-SDK to apply the NPU SI settings via syncd
-		- If SAI-SDK returns success, OA will update the PORT_TABLE|\<port\>.NPU_SI_SETTINGS_SYNC_STATUS to NPU_SI_SETTINGS_DONE in STATE_DB
+      For CMIS SM driven transceivers, if `is_npu_si_settings_update_required` returns true, CmisManagerTask will update SI settings in PORT_TABLE (APPL_DB) and write `si_sync_status = SI_SETTINGS_NOTIFIED:<N>` to PORT_TABLE (APPL_DB) during CMIS_STATE_AP_CONF.  
+      The CMIS SM will then transition from CMIS_STATE_AP_CONF to CMIS_STATE_SI_SETTINGS_WAIT. If port doesn't require SI settings, CMIS SM will transition to CMIS_STATE_DP_INIT state directly.  
+
+
+
+3. The OA upon receiving SI settings (via `si_sync_status` update in APPL_DB) will  
+	- Disable port admin status (without updating host_tx_ready, to avoid triggering xcvrd)
```
mihirpat1: @arpit-nexthop Wouldn't disabling port admin status also cause the CMIS SM to start from CMIS_STATE_INSERTED?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -207,46 +211,56 @@ Pre-requisites for the infra to work:
...
+	- Request SAI-SDK to apply the SI settings via syncd
+		- If SAI-SDK returns success, OA will write `si_settings_sync_status = SI_SYNC_DONE:<N>` to PORT_TABLE|\<port\> (STATE_DB), where N matches the notification number received
 		- In case of failure, OA will log an error message and proceed to handling the next port
+	- If `si_sync_status == SI_SETTINGS_DEFAULT:<N>` is received (transceiver removed), OA writes `si_settings_sync_status = SI_SETTINGS_DEFAULT:<N>` to STATE_DB (no serdes programming)
 
-4. CMIS_STATE_NPU_SI_SETTINGS_WAIT state will wait for NPU_SI_SETTINGS_DONE and upon reaching to NPU_SI_SETTINGS_DONE, CMIS SM will transition to CMIS_STATE_DP_INIT state.  
-There will be a timeout of 5s for every retry
+4. CMIS_STATE_SI_SETTINGS_WAIT state will wait for `SI_SYNC_DONE:<N>` in STATE_DB `si_settings_sync_status`, where N must match the notification number sent. Upon match, CMIS SM transitions to CMIS_STATE_DP_INIT state.  
+There will be a timeout of 10s; upon timeout, CMIS SM proceeds to CMIS_STATE_DP_INIT with a warning (no reinit)
```
mihirpat1: @arpit-nexthop Why are we allowing CMIS initialization to progress without applying SI settings in case of timeout?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -207,46 +211,56 @@ Pre-requisites for the infra to work:
...
 
 5. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED" to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
 
 6. XCVRD will subscribe to PORT_TABLE in STATE_DB and trigger self-restart if the PORT_TABLE|Ethernet* is deleted for the namespace.
```
mihirpat1, line 272: @prgeor @arpit-nexthop 
Given that swss.sh already restarts xcvrd when media_settings.json is present [swss.sh#L429-L440](https://github.com/sonic-net/sonic-buildimage/blob/83560eba02c41a2ad9dae30fb0a78cd9646cd0ae/files/scripts/swss.sh#L429-L440), there's now redundancy between that workaround and this STATE_DB subscription mechanism.
Should the swss.sh workaround be removed as part of this change's implementation PRs?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -324,35 +334,39 @@ sequenceDiagram
...
-The below state machine is a high level flow and doesn't capture details for states other than CMIS_STATE_AP_CONF and CMIS_STATE_NPU_SI_SETTINGS_WAIT  
+The below state machine is a high level flow and doesn't capture details for states other than CMIS_STATE_AP_CONF and CMIS_STATE_SI_SETTINGS_WAIT  
 
 ```mermaid
 stateDiagram
     [*] --> CMIS_STATE_INSERTED
     state if_state <<choice>>
     state if_state2 <<choice>>
     state if_state3 <<choice>>
+    state if_state4 <<choice>>
```
mihirpat1, line 383: @arpit-nexthop Seems this section cannot be rendered. Can you please fix this?

Image

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -193,10 +194,13 @@ if transceiver is not present:
...
+  - SI settings synchronization uses a two-field protocol to communicate between xcvrd and Orchagent (OA):
+    - `si_sync_status` in PORT_TABLE (APPL_DB): written by xcvrd to signal SI settings state. OA reads this via its normal ConsumerStateTable flow. Values: `SI_SETTINGS_NOTIFIED:<N>` (xcvrd has notified SI settings for sequence N), `SI_SETTINGS_DEFAULT:<N>` (transceiver removed or no SI settings needed for sequence N).
+    - `si_settings_sync_status` in PORT_TABLE (STATE_DB): written by OA to acknowledge SI settings state. xcvrd polls this field. Values: `SI_SETTINGS_DEFAULT:<N>` (OA acknowledged removal/reset), `SI_SYNC_DONE:<N>` (OA successfully applied SI settings for sequence N).  
+    A monotonically increasing counter `<N>` embedded in both fields ensures xcvrd can distinguish stale acknowledgments from prior insertions. In case of warm reboot or xcvrd restart, these fields prevent re-application of SI settings if the settings are already applied. In case of transceiver insertion, SI settings will be applied irrespective of the current status.
```
mihirpat1: @arpit-nexthop Can you please clarify if the counter attached to `si_sync_status` will not be used by OA for making any decisions and is more of for record keeping and synchronizing?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -207,46 +211,56 @@ Pre-requisites for the infra to work:
...
+| ----------------------- | ----------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- |
+| `SI_SETTINGS_DEFAULT:0` | Orchagent during port initialization                                                 | Initial state; indicates SI settings have not been applied                                                    |
+| `SI_SETTINGS_DEFAULT:<N>` | Orchagent upon receiving `SI_SETTINGS_DEFAULT:<N>` from xcvrd via APPL_DB         | Acknowledges transceiver removal; used by xcvrd on restart to determine update is needed                      |
+| `SI_SYNC_DONE:<N>`      | Orchagent after successfully applying serdes settings for sequence N                | CmisManagerTask checks this in CMIS_STATE_SI_SETTINGS_WAIT to proceed to CMIS_STATE_DP_INIT; also used by `can_skip_cmis_init_after_restart` to verify SI settings are already applied |
```
mihirpat1: @arpit-nexthop What is `can_skip_cmis_init_after_restart`? Can you please clarify this in the HLD?

arpit-nexthop:
> Hi @arpit-nexthop
> 
> With this new design, do we prevent the admin state (and host TX signal) from coming up until the host SI settings are configured? If so, did you verify it in the ‘swss restart’ flow as well?
> 
> Thanks

@benle7 Do you mean warm restart here? swss restart in generally leads to restarting syncd and reprogramming the ASIC as well.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: That is an oversight, we can rename the variables accordingly.

arpit-nexthop: ACK, I will rename this in the HLD as well as the code.

arpit-nexthop: Resolving this as we are having same discussion in next comment.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: No, that happens because we set the host_tx_ready to false. Here we are not setting it to false when SI is present. This avoids the xcvrd going to CMIS_INSERTED state.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: I believe this is what we discussed in the meeting, where we wanted to wait for some time, reading the settings being applied. If not we should proceed rather than failing the state machine since SI notification does not happen in the cmis task.

We agreed not to add a notification for settings being programmed.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: ACK

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: @arpit-nexthop I agree that the CMIS reinit happens due to host_tx_ready toggling.
However, what do you mean by `Disable port admin status`? Wouldn't this cause the CMIS SM to be notified which would eventually result in a CMIS reinit per the below?
https://github.com/sonic-net/sonic-platform-daemons/blob/3bb3540f10f9efbaadfd98f424efad96d561682d/sonic-xcvrd/xcvrd/xcvrd_utilities/port_event_helper.py#L7
https://github.com/sonic-net/sonic-platform-daemons/blob/3bb3540f10f9efbaadfd98f424efad96d561682d/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py#L1196

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: ACK

mihirpat1:
@arpit-nexthop Please help in addressing the comments provided by Copilot as well.

benle7:
> > Hi @arpit-nexthop
> > With this new design, do we prevent the admin state (and host TX signal) from coming up until the host SI settings are configured? If so, did you verify it in the ‘swss restart’ flow as well?
> > Thanks
> 
> @benle7 Do you mean warm restart here? swss restart in generally leads to restarting syncd and reprogramming the ASIC as well.

I mean to "systemctl restart swss".
Today, PortOrch can set admin True (and open the host TX signal) before configuring the SI settings -
This order can lead to some module issues.
So I would like to understand if your design will change it and enforce configuring SI settings before opening
the host TX signal (set admin True).

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: It does have any logic in OA, OA just gets the number and updates the done field with same number once the SI is processed. I will make it clearer in the doc.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: This is not admin status in config_db, this is the call from OA to syncd/gbsyncd for disabling or enabling interface.

arpit-nexthop: `Disable port admin status` becomes `Disable host port`

bmridul:
@longhuan-cisco , pls review

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -324,35 +342,80 @@ sequenceDiagram
...
-    if_state2 --> CMIS_STATE_DP_DEINIT : if PORT_TABLE.port.CMIS_REINIT_REQUIRED == true or<br>is_cmis_application_update_required
-    note left of CMIS_STATE_READY : PORT_TABLE.port.CMIS_REINIT_REQUIRED = false
-    if_state2 --> CMIS_STATE_FAILED : if appl < 1 or <br>host_lanes_mask <= 0 or <br>NPU SI_lanes_mask <= 0
+    note right of CMIS_STATE_INSERTED
+        If notify_si_settings=True (module insertion or SI failure reinit):
+          Read transceiver info via api
+          Write SI params + si_settings_notification = SI_SETTINGS_NOTIFIED[N] to APPL_DB
+          Reset notify_si_settings = False
+    end note
+    if_state --> CMIS_STATE_READY : if host_tx_ready != True or admin_status != up - Action disable TX
```
bmridul, line 389: Is this correct? How do SI settings get applied when interface us made admin up?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -192,12 +193,16 @@ if transceiver is not present:
...
+  - SI settings synchronization uses a two-field protocol to communicate between xcvrd and Orchagent (OA):
+    - `si_settings_notification` in PORT_TABLE (APPL_DB): written by xcvrd to signal SI settings state. OA reads this via its normal ConsumerStateTable flow. Values: `SI_SETTINGS_NOTIFIED:<N>` (xcvrd has notified SI settings for sequence N), `SI_SETTINGS_DEFAULT:<N>` (transceiver removed or no SI settings needed for sequence N).
+    - `si_settings_ack` in PORT_TABLE (STATE_DB): written by OA to acknowledge SI settings state. xcvrd polls this field. Values: `SI_SETTINGS_DEFAULT:<N>` (OA acknowledged removal/reset), `SI_SYNC_DONE:<N>` (OA successfully applied SI settings for sequence N).  
+    A monotonically increasing counter `<N>` embedded in both fields ensures xcvrd can distinguish stale acknowledgments from prior insertions. In case of warm reboot or xcvrd restart, these fields prevent re-application of SI settings if the settings are already applied. In case of transceiver insertion, SI settings will be applied irrespective of the current status.
```
bmridul, line 202: Is the counter reset to 0 on xcvrd restart? or re-initialized from state db?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -450,24 +515,24 @@ sequenceDiagram
...
-| Swss restart | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
-| Syncd restart | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
-| config reload | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
-| Cold reboot | Y | Y | Y | NPU_SI_SETTINGS_DEFAULT | Y | N/A |
-| Warm reboot | N | Y | N | NPU_SI_SETTINGS_DONE | N | N |  
-| config interface shutdown | N | N | N | NPU_SI_SETTINGS_DONE | N | N/A |
-| config interface startup | N | N | N | NPU_SI_SETTINGS_DONE | N | N/A |
+| Event          | STATE_DB_<asic_n> cleared | Xcvrd restarted | SI settings renotify | si_settings_ack (STATE_DB) on xcvrd boot-up for initialized transceiver | CMIS re-init triggered | Link flap |
+| -------------- | ------------------------ | --------------- | -------------------- | -------------------------------------------------------------------------------- | ---------------------- | --------- |
+| Xcvrd restart | N | Y | N | SI_SYNC_DONE:<N> | N | N |
```
bmridul, line 519: Negative test case:
Check if State machine transitions to CMIS_FAILED if the orchagent is stuck/not acking.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -324,35 +342,80 @@ sequenceDiagram
...
+    if_state3 --> CMIS_STATE_SI_SETTINGS_WAIT : si_settings_notification == SI_SETTINGS_NOTIFIED[N] in APPL_DB
+    if_state3 --> CMIS_STATE_DP_INIT : si_settings_notification not set or == SI_SETTINGS_DEFAULT
+    CMIS_STATE_SI_SETTINGS_WAIT --> CMIS_STATE_DP_INIT : STATE_DB si_settings_ack == SI_SYNC_DONE[N]
+    CMIS_STATE_SI_SETTINGS_WAIT --> CMIS_STATE_INSERTED : 10s timeout - reset si_settings_notification to DEFAULT,\nset notify_si_settings=True, force_cmis_reinit(retries+1)
+    note right of CMIS_STATE_SI_SETTINGS_WAIT
+        Polls STATE_DB si_settings_ack for SI_SYNC_DONE[N]
+        N must match the notification number sent in APPL_DB
+        On 10s timeout: reset APPL_DB to SI_SETTINGS_DEFAULT[N+1],
+        set notify_si_settings=True, reinit with retries++
+        After CMIS_MAX_RETRIES exhausted: transition to CMIS_STATE_FAILED
```
bmridul, line 417: Is 10 second timeout big enough for orchagent restart/docker restart scenarios.

What is the recommended operator action if some Xcvr is in CMIS_STATE_FAILED?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: It is reinitialized from the STATE_DB

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: ACK, will add a unit test in xcvrd

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: Currently orchagent restart triggers a restart of xcvrd as well, that optimization is out of scope here, will be revisited later when we try to remove this dependency

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: OA turns shuts the port, applies SI settings and enables port and then sends the notification

bmridul: Question is whether state transition to CMIS_STATE_READY is correct here? (if host_tx_ready is not True or Admin status is Down)

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
bmridul: I believe it is not true for all scenarios (if media_settings.json is not used). Retries will help alleviate it though. So it might be ok.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1: @arpit-nexthop Can you please address this comment?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -207,46 +212,67 @@ Pre-requisites for the infra to work:
...
+**APPL_DB PORT_TABLE `si_settings_notification`** — written by xcvrd, read by Orchagent via ConsumerStateTable  
 
+| Value                        | Written by                                                                                                                                                                          | Purpose                                                                                   |
+| ---------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------- |
+| `SI_SETTINGS_DEFAULT:<N>`    | 1\. SfpStateUpdateTask during transceiver removal (increments N)<br>2. xcvrd during boot-up initialization                                                                          | Signals to OA that transceiver was removed or SI settings are in default state             |
```
mihirpat1: @arpit-nexthop Do you mean the xcvrd main process will perform this action?
`xcvrd during boot-up initialization`

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -207,46 +212,67 @@ Pre-requisites for the infra to work:
...
+	- Request SAI-SDK to apply the SI settings via syncd
+		- If SAI-SDK returns success, OA will write `si_settings_ack = SI_SYNC_DONE:<N>` to PORT_TABLE|\<port\> (STATE_DB), where N matches the notification number received
 		- In case of failure, OA will log an error message and proceed to handling the next port
+	- If `si_settings_notification == SI_SETTINGS_DEFAULT:<N>` is received (transceiver removed), OA writes `si_settings_ack = SI_SETTINGS_DEFAULT:<N>` to STATE_DB (no serdes programming)
+
+4. CMIS_STATE_SI_SETTINGS_WAIT state will wait for `SI_SYNC_DONE:<N>` in STATE_DB `si_settings_ack`, where N must match the notification number sent. Upon match, CMIS SM transitions to CMIS_STATE_DP_INIT state.
+There will be a timeout of 10s; upon timeout, CmisManagerTask resets `si_settings_notification` to `SI_SETTINGS_DEFAULT:<N+1>` in APPL_DB, sets `notify_si_settings=True`, and calls `force_cmis_reinit(retries+1)`, transitioning back to CMIS_STATE_INSERTED. This consumes a retry (max 3) and causes re-notification on the next INSERTED entry. After exhausting retries, the SM transitions to CMIS_STATE_FAILED.
```
mihirpat1, line 260: @arpit-nexthop Just to clarify, wouldn't calling `force_cmis_reinit(retries+1)` automatically trigger SI settings notification?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
mihirpat1, line 5: @arpit-nexthop 1.1?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -324,35 +342,80 @@ sequenceDiagram
...
+    participant STATE_DB as STATE_DB@asic_n
+    participant SffManagerTask
+
+    Note over SffManagerTask: Subscribe to CONFIG_DB:PORT,<br>STATE_DB:TRANSCEIVER_INFO and STATE_DB:PORT_TABLE
+    loop Wait for stop_event
+        SffManagerTask ->> SffManagerTask: handle_port_update_event
+        opt xcvr_inserted (TRANSCEIVER_INFO SET) and not is_cmis_api
+            SffManagerTask ->> SffManagerTask: sfp.get_transceiver_info()
+            opt is_npu_si_settings_update_required
+                SffManagerTask ->> APPL_DB: Update SI params from media_settings.json to PORT_TABLE:<lport><br>si_settings_notification = SI_SETTINGS_NOTIFIED:<N>
```
mihirpat1: @arpit-nexthop Shouldn't this be N + 1?

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -324,35 +342,80 @@ sequenceDiagram
...
+```
+
+## CMIS State machine with introduction of CMIS_STATE_SI_SETTINGS_WAIT state
 
-The below state machine is a high level flow and doesn't capture details for states other than CMIS_STATE_AP_CONF and CMIS_STATE_NPU_SI_SETTINGS_WAIT  
+The below state machine is a high level flow and doesn't capture details for states other than CMIS_STATE_AP_CONF and CMIS_STATE_SI_SETTINGS_WAIT  
 
 ```mermaid
-stateDiagram
+stateDiagram-v2
```
mihirpat1, line 378: @arpit-nexthop Can you please go through this diagram and remove the legacy content?

Image

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: ACK

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: This is no longer required as initial removal / insertion notification will take care of this

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: ACK

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: We would not want to keep setting the SI settings everytime we do a force cmis reinit, (called in shut/start flow as well). SI settings application will be re-triggered only if we fail to get the notification or an OIR event happens.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: ACK

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: This section of the mermaid template refers to the admin_down state transition

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: @mihirpat1  Added the PRE_INIT check as well. I would recommend viewing the file to see the diagram, the git unified view is not really showing the new diagram, just the old one being deleted.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: Done

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
arpit-nexthop: https://github.com/nexthop-ai/SONiC/blob/7fce6219a6de03d03143e4e1e29ce05b1a9da83a/doc/sfp-cmis/Interface-Link-bring-up-sequence.md#cmis-state-machine-with-introduction-of-cmis_state_si_settings_wait-state

snider-nokia:
@arpit-nexthop, all,
   Strongly suggest considering a slight alteration to this refactoring whereby orchagent does not actually bring the ASIC port admin up until such time as the xcvrd provided SI settings have actually arrived at Redis for orchagent use.   Even if SI settings are not in the mix for a given configuration on a given port, orchagent can still wait until the Redis notification flag explicitly indicates that the extant SI configuration provided by xcvrd (or not) is good-to-go.
   
Introducing this type of gate for orchagent, in advance of orchagent actually bringing the ASIC port admin up will prevent the churn that is currently seen when an ASIC port has already been enabled but then SI settings arrive at Redis thus causing orchagent to modify ASIC SerDes settings and consequently flap the ASIC port.

Waiting at orchagent until such time as the SI settings are confirmed to have arrived from xcvrd (even if there were no changes actually provided by xcvrd with the exception of Redis confirmation flag) would prevent ASIC port from being brought up while only partially configured (in advance of SI settings arrival).  Preventing this ASIC port churn/flap will help to clean up the critical enable path such that the ASIC port is only brought up once all provisioning has necessarily been completed and thus commensurately serve to minimize SI related flux within the machinery (as ASIC port(s) are transitioning down-then-up during provisioning time).

Am also copying this same comment to orchagent PR #4497.

snider-nokia:
For clarity, @benle7's suggestion/request earlier is essentially the same ask I made in my previous comment.

image

benle7:
@snider-nokia Correct.

@prgeor FYI
As I remember, you mentioned this HLD should resolve the issues we found with the AOI module:
- Open host TX before SI configured -> unstable host signal -> CDR LoL.
- When swss get restarted while optics datapath is already DpActivated, the module is not re-initialized.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -207,46 +212,67 @@ Pre-requisites for the infra to work:
...
 
-4. CMIS_STATE_NPU_SI_SETTINGS_WAIT state will wait for NPU_SI_SETTINGS_DONE and upon reaching to NPU_SI_SETTINGS_DONE, CMIS SM will transition to CMIS_STATE_DP_INIT state.  
-There will be a timeout of 5s for every retry
+5. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED” to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
 
-5. The CmisManagerTask thread will set “CMIS_REINIT_REQUIRED" to false after CMIS SM reaches to a steady state (CMIS_STATE_UNKNOWN, CMIS_STATE_FAILED, CMIS_STATE_READY and CMIS_STATE_REMOVED) for the corresponding port
+5a. **`can_skip_cmis_init_after_restart`**: When xcvrd restarts (process restart or warm reboot), CmisManagerTask checks whether full CMIS re-initialization can be skipped for a port. Skipping avoids an unnecessary link flap on already-configured ports. All four conditions must hold:
```
longhuan-cisco, line 264: This section describes can_skip_cmis_init_after_restart, including SI notification/ACK matching, but I do not see this helper or equivalent SI-ack based restart-skip logic in the current xcvrd implementation PR.

Current xcvrd appears to rely on is_cmis_application_update_required(), and SI notification is done earlier in INSERTED when notify_si_settings is set. Please either add the described restart-skip logic to the implementation or update the HLD to match the actual behavior.

doc/sfp-cmis/Interface-Link-bring-up-sequence.md
```diff
@@ -192,12 +193,16 @@ if transceiver is not present:
...
-  Also, this key helps in preventing additional link flap which can happen if CMIS state machine initializes a port before NPU SI settings are applied (since application of NPU SI settings involves disable followed by enable of it).  
-In case of continuous restart of xcvrd, both the keys will still hold the same value as before the restart. This would ensure that the port re-initialization is resumed from the last known state.  
+  - SI settings synchronization uses a two-field protocol to communicate between xcvrd and Orchagent (OA):
+    - `si_settings_notification` in PORT_TABLE (APPL_DB): written by xcvrd to signal SI settings state. OA reads this via its normal ConsumerStateTable flow. Values: `SI_SETTINGS_NOTIFIED:<N>` (xcvrd has notified SI settings for sequence N), `SI_SETTINGS_DEFAULT:<N>` (transceiver removed or no SI settings needed for sequence N).
```
longhuan-cisco, line 200: **Medium: document `SI_SETTINGS_UNAVAIL` as a third terminal protocol value**

The implementation PRs no longer use `SI_SETTINGS_DEFAULT:<N>` for a present port that has no applicable SI settings. xcvrd publishes the counterless value `SI_SETTINGS_UNAVAIL`; OA treats it as terminal, does not program SerDes, releases the admin-up gate, and writes `si_settings_ack = SI_SETTINGS_UNAVAIL`.

| Notification | Meaning | OA response |
|---|---|---|
| `SI_SETTINGS_NOTIFIED:<N>` | SI settings for this insertion are available | Program them and acknowledge `SI_SYNC_DONE:<N>` |
| `SI_SETTINGS_DEFAULT:<N>` | Removal/reset of the numbered transaction | Do not program; acknowledge `SI_SETTINGS_DEFAULT:<N>` |
| `SI_SETTINGS_UNAVAIL` | A present port has no applicable SI settings | Do not program; acknowledge `SI_SETTINGS_UNAVAIL` and release the gate |

The HLD currently lists only `NOTIFIED` and `DEFAULT`, and says `DEFAULT` can also mean "no SI settings needed." That no longer matches either implementation. It also hides the terminal-notification invariant: every gated present external port must eventually receive either `NOTIFIED:<N>` or `UNAVAIL`.

Could you add `UNAVAIL` to both field tables and the relevant flow/state diagrams, and reserve `DEFAULT:<N>` for removal/reset?

PR #2310, opened 2026-04-27
[pmon]: HLD: Enhance DPU Robustness in Smart Switch

vvolam:
#### What I did
Add a High Level Design document for DPU failure scenarios on SmartSwitch from the PMON (Platform Monitor) perspective.

#### Why I did it
SmartSwitch DPU lifecycle management requires clear specification of failure detection, DB state tracking, and recovery actions performed by `chassisd` and other PMON sub-daemons. This HLD documents all failure and planned operation scenarios to guide implementation.

#### How I did it
Added `doc/smart-switch/pmon/enhance-dpu-robustness.md` covering:

- **DPU software failures**: critical process restart, persistent failure, pmon/databasedpu crashes on NPU
- **DPU hardware failures**: complete DPU down, power failure, PCIe failure
- **NPU/switch-level failures**: kernel crash, memory exhaustion
- **Planned operations**: graceful shutdown, cold reboot, full SmartSwitch reboot
- **New DB fields**: `ready_status`, `recovery_status`, `reset_count`, `last_down_time`, `last_ready_time` in CHASSIS_STATE_DB
- **New feature flag**: `FEATURE|dpu-auto-recovery` in CONFIG_DB
- **DPU recovery state machine**: Mermaid diagram with state table
- **Timers and thresholds**: configurable via `platform.json`
- **Race condition handling**: concurrent operations on the same DPU

#### How to verify it
Review the HLD document for completeness and correctness of failure scenarios, DB state transitions, and recovery actions.

| Repo | PR Title / Link | Status |
|-------|---------------|--------|
| sonic-platform-daemons |[[chassisd]: Add DPU recovery state machine and new DB fields](https://github.com/sonic-net/sonic-platform-daemons/pull/815) | PR State |
| sonic-buildimage |[[smartswitch]: Add dpu-auto-recovery feature to SmartSwitch NPU default config](https://github.com/sonic-net/sonic-buildimage/pull/27385) | PR State |
| sonic-utilties |[[cli]: Add DPU recovery CLI commands for SmartSwitch](https://github.com/sonic-net/sonic-utilities/pull/4572) | PR State |

### DIFF
--- a/doc/smart-switch/pmon/enhance-dpu-robustness.md
+++ b/doc/smart-switch/pmon/enhance-dpu-robustness.md
@@ -14,8 +14,7 @@
   - [New DB entries](#new-db-entries)
 - [DPU Recovery State Machine](#dpu-recovery-state-machine)
 - [DPU Software Failures](#dpu-software-failures)
-  - [Process Restart on DPU](#process-restart-on-dpu)
-  - [Process Persistently Down on DPU](#process-persistently-down-on-dpu)
+  - [Process Crash/Restart on DPU](#process-crashrestart-on-dpu)
   - [pmon Crash on NPU](#pmon-crash-on-npu)
   - [databasedpu Crash on NPU](#databasedpu-crash-on-npu)
 - [DPU Hardware Failures](#dpu-hardware-failures)
@@ -127,8 +126,7 @@ All timers and thresholds used by PMON for DPU failure detection and recovery ar
 
 | Timer / Threshold | Default Value | Configurable | Used By | Description |
 | ----------------- | :-----------: | :----------: | ------- | ----------- |
-| `chassisd` health poll interval | 10 seconds | No | `chassisd` | Interval at which `chassisd` polls `dpu_control_plane_state`, `dpu_data_plane_state`, and `dpu_midplane_link_state` |
-| `dpu_auto_recovery_timeout` | 60 seconds | Yes (`platform.json`) | `chassisd` | Self-heal grace period after a software failure is first detected. `chassisd` waits this long for the DPU to recover on its own (e.g., container supervisor restarts the failing process). If `dpu_control_plane_state` or `dpu_midplane_link_state` is still `down` when this timer expires, `chassisd` issues a DPU power-cycle. |
+| `chassisd` health poll interval | 10 seconds | No | `chassisd` | Interval at which `chassisd` polls `dpu_control_plane_state`, `dpu_data_plane_state`, and `dpu_midplane_link_state`. As soon as `dpu_control_plane_state` or `dpu_midplane_link_state` is observed as `down`, `chassisd` initiates a DPU power-cycle (no self-heal grace period). |
 | `pcied` PCIe poll interval | 60 seconds | No | `pcied` | Interval at which `pcied` checks PCIe link status for all DPUs. A PCIe failure may go undetected for up to 60 seconds. |
 | `dpu_halt_services_timeout` | 60 seconds | Yes (`platform.json`) | `gnoi_reboot_daemon.py` | Maximum time to wait for DPU services to halt gracefully during reboot/shutdown |
 | `reset_limit` | 5 | Yes (`platform.json`) | `chassisd` | Maximum number of consecutive unplanned power-cycle attempts before marking DPU as unrecoverable |
@@ -239,9 +237,8 @@ stateDiagram-v2
     Booting --> Ready : All states up
 
     Ready --> SWFailure : Control plane down
-    SWFailure --> Ready : Self recovers before dpu_auto_recovery_timeout (60s)
-    SWFailure --> PowerCycle : dpu_auto_recovery_timeout (60s) expires [auto-recovery enabled]
-    SWFailure --> ManualIntervention : dpu_auto_recovery_timeout (60s) expires [auto-recovery disabled]
+    SWFailure --> PowerCycle : auto-recovery enabled
+    SWFailure --> ManualIntervention : auto-recovery disabled
 
     Ready --> PowerCycle : HW failure detected [auto-recovery enabled]
     Ready --> ManualIntervention : HW failure detected [auto-recovery disabled]
@@ -266,7 +263,7 @@ stateDiagram-v2
 | ----- | :------------: | :----------------: | ----------------- |
 | **Booting** | `false` | `recoverable` | `dpu_control_plane_state: down` |
 | **Ready** | `true` | `recoverable` | All three states `up` |
-| **SWFailure** | `false` | `recoverable` | `dpu_control_plane_state: down`, `dpu_midplane_link_state: up` |
+| **SWFailure** | `false` | `recoverable` | `dpu_control_plane_state: down`, `dpu_midplane_link_state: up`; transient state before `chassisd` selects `PowerCycle` or `ManualIntervention` based on the auto-recovery feature flag |
 | **PowerCycle** | `false` | `recoverable` | `chassisd` issuing power-cycle; `reset_count` incremented |
 | **ManualIntervention** | `false` | `recoverable` | DPU down; `FEATURE&#124;dpu-auto-recovery` `state` is `disabled` / `always_disabled`; no power-cycle issued; awaits operator action |
 | **Offline** | `false` | `recoverable` | `oper_status: Offline` |
@@ -276,43 +273,18 @@ stateDiagram-v2
 
 ## DPU Software Failures ##
 
-### Process restart on DPU ###
+### Process crash/restart on DPU ###
 
 **Description:**
-When any process crashes on the DPU, but the container supervisor successfully restarts the process and the DPU recovers on its own within `dpu_auto_recovery_timeout` (default: 60 seconds). No power-cycle is needed.
+Any process crashes on the DPU and `dpu_control_plane_state` transitions to `down`. `chassisd` does not wait for the container supervisor to self-heal — it issues a DPU power-cycle as soon as it observes `dpu_control_plane_state: down` on its next poll.
 
 **Detection (by PMON):**
 - `chassisd` on the NPU polls `dpu_control_plane_state` every 10 seconds and observes it as `down`.
 
 **PMON Action:**
 - `chassisd` sets `ready_status` to `false` and updates `last_down_time` for the corresponding DPU.
-- `chassisd` waits up to `dpu_auto_recovery_timeout` (60 seconds) for the DPU to self-recover.
-- Once `dpu_control_plane_state` transitions back to `up`, `chassisd` verifies all DPU states (midplane, control plane, data plane), sets `ready_status` back to `true`, and updates `last_ready_time`.
-
-**DB State Transition:**
-
-| DB Field | Before | During Failure | After Recovery |
-| -------- | :----: | :------------: | :------------: |
-| `dpu_control_plane_state` | `up` | `down` | `up` |
-| `ready_status` | `true` | `false` | `true` |
-| `last_down_time` | — | `<UTC timestamp>` | — |
-| `last_ready_time` | — | — | `<UTC timestamp>` |
-
----
-
-### Process persistently down on DPU ###
-
-**Description:**
-When any process crashes on the DPU and **remains down beyond `dpu_auto_recovery_timeout`** (i.e., the container supervisor cannot successfully restart it, or the process keeps crash-looping). Unlike a transient restart, this scenario indicates a persistent failure that requires a DPU power-cycle to recover.
-
-**Detection (by PMON):**
-- `chassisd` on the NPU polls `dpu_control_plane_state` every 10 seconds and observes it as `down`.
-- State remains `down` beyond `dpu_auto_recovery_timeout` (default: 60 seconds).
-
-**PMON Action:**
-- `chassisd` sets `ready_status` to `false` and updates `last_down_time` for the corresponding DPU.
-- When `dpu_auto_recovery_timeout` (default: 60 seconds, measured from the time failure is first detected) expires without recovery, `chassisd` issues a power-cycle of the DPU and increments `reset_count`.
-- Once `dpu_control_plane_state` transitions back to `up`, `chassisd` verifies all DPU states (midplane, control plane, data plane), sets `ready_status` back to `true`, and updates `last_ready_time`.
+- `chassisd` immediately issues a power-cycle of the DPU and increments `reset_count`.
+- After the DPU comes back, `chassisd` verifies all DPU states (midplane, control plane, data plane), sets `ready_status` back to `true`, and updates `last_ready_time`.
 - If `reset_count` reaches `reset_limit`, `chassisd` sets `recovery_status` to `"unrecoverable"` and stops further automatic power-cycle attempts.
 - **When auto-recovery is disabled:** `chassisd` skips the power-cycle. The DPU remains in **ManualIntervention** with `ready_status: false`; operator must reset the DPU manually.
 
@@ -389,7 +361,7 @@ A DPU completely fails due to hardware fault, thermal event, or unrecoverable er
 
 **PMON Action:**
 - `chassisd` sets `ready_status` to `false` and updates `last_down_time` for the corresponding DPU.
-- `chassisd` power-cycles the DPU **immediately** (skipping `dpu_auto_recovery_timeout` — the DPU is already confirmed non-functional via `oper_status: Offline`) and increments `reset_count`.
+- `chassisd` immediately power-cycles the DPU (the DPU is already confirmed non-functional via `oper_status: Offline`) and increments `reset_count`.
 - After power-cycle, DPU goes through full boot sequence: midplane attach → PCIe rescan → SONiC boot → container startup.
 - `chassisd` verifies all DPU states (midplane, control plane, data plane), sets `ready_status` back to `true`, and updates `last_ready_time`.
 - If `reset_count` reaches `reset_limit`, `chassisd` sets `recovery_status` to `"unrecoverable"` and stops further automatic power-cycle attempts.
@@ -419,7 +391,7 @@ The DPU loses power unexpectedly or shuts down without graceful notification (e.
 
 **PMON Action:**
 - `chassisd` sets `ready_status` to `false` and updates `last_down_time`.
-- `chassisd` power-cycles the DPU **immediately** (skipping `dpu_auto_recovery_timeout` — midplane and control plane are already confirmed down) and increments `reset_count`.
+- `chassisd` immediately power-cycles the DPU (midplane and control plane are already confirmed down) and increments `reset_count`.
 - After power-cycle, `chassisd` verifies all DPU states, sets `ready_status` back to `true`, and updates `last_ready_time`.
 - **When auto-recovery is disabled:** `chassisd` skips the power-cycle. The DPU remains in **ManualIntervention** with `ready_status: false`; operator must trigger recovery.
 
@@ -447,7 +419,7 @@ The PCIe bus between the NPU and a local DPU fails, making the DPU unreachable f
 
 **PMON Action:**
 - `chassisd` sets `ready_status` to `false` and updates `last_down_time`.
-- `chassisd` power-cycles the DPU **immediately** (skipping `dpu_auto_recovery_timeout` — midplane link loss and PCIe detach confirm the DPU is unreachable) and increments `reset_count`.
+- `chassisd` immediately power-cycles the DPU (midplane link loss and PCIe detach confirm the DPU is unreachable) and increments `reset_count`.
 - After power-cycle, PCIe rescan is performed:
   - Platform vendor API: `pci_reattach()` (provided by `sonic_platform`).
 - `chassisd` verifies all DPU states (midplane, control plane, data plane), sets `ready_status` back to `true`, and updates `last_ready_time`.
@@ -610,7 +582,7 @@ Planned reboot of the entire SmartSwitch (NPU + all DPUs) via CLI: `reboot`. All
 | DPU up after crash | up | up | true | Set `ready_status=true` after verifying all states |
 | DPU stuck (lost connectivity) | down | down | false | Power-cycle DPU; increment `reset_count` |
 | DPU up after losing connectivity / reboot | up | up | true | Set `ready_status=true` after verifying all states |
-| DPU control plane restart – critical services | down → up | up | false → true | Wait for auto-recovery; set `ready_status=true` on recovery |
+| DPU control plane restart – critical services | down → up | up | false → true | Power-cycle DPU; increment `reset_count`; set `ready_status=true` on recovery |
 | NPU/DPU OS upgrade | down → up | up | false → true | Re-poll DPU states on NPU recovery |
 | DPU dead – power cycle | down | down | false | Power-cycle DPU; increment `reset_count` |
 | DPU dead – unrecoverable | down | down | false | `reset_count` reached `reset_limit`; `recovery_status` set to `"unrecoverable"`; raise alert |
--- a/doc/smart-switch/pmon/enhance-dpu-robustness.md
+++ b/doc/smart-switch/pmon/enhance-dpu-robustness.md
@@ -92,9 +92,9 @@ This document enumerates all failure scenarios that can occur on a DPU or its su
 | chassisd | Chassis daemon running inside `pmon` on the NPU; monitors DPU health states, manages DPU power-cycle and reset operations |
 | pmon | Platform Monitor daemon on NPU; hosts `chassisd` and other hardware monitoring sub-daemons |
 | syncd | Sync daemon; manages SAI API calls to DPU ASIC |
-| control plane state | DPU SONiC is booted up, all containers are up, interfaces are up, and DPU is ready to accept configuration. Derived from SYSTEM_READY in STATE_DB. Values: `"unknown"`, `"up"`, `"down"`. |
-| midplane link state | The PCIe link between the NPU and DPU is operational. Monitored and updated by NPU pmon `chassisd` via the `is_midplane_reachable` platform API. Values: `"unknown"`, `"up"`, `"down"`. |
-| dataplane state | Configuration is downloaded, pipeline stages are up, and DPU hardware (port/ASIC) is ready to take traffic. Values: `"unknown"`, `"up"`, `"down"`. |
+| control plane state | DPU SONiC is booted up, all containers are up, interfaces are up, and DPU is ready to accept configuration. Derived from SYSTEM_READY in STATE_DB. Values: `"up"`, `"down"`. |
+| midplane link state | The PCIe link between the NPU and DPU is operational. Monitored and updated by NPU pmon `chassisd` via the `is_midplane_reachable` platform API. Values: `"up"`, `"down"`. |
+| dataplane state | Configuration is downloaded, pipeline stages are up, and DPU hardware (port/ASIC) is ready to take traffic. Values: `"up"`, `"down"`. |
 
 ---
 
@@ -259,7 +259,7 @@ stateDiagram-v2
 
 | State | `ready_status` | `recovery_status` | Key DB Indicators |
 | ----- | :------------: | :----------------: | ----------------- |
-| **Booting** | `false` | `recoverable` | `dpu_control_plane_state: unknown` |
+| **Booting** | `false` | `recoverable` | `dpu_control_plane_state: down` |
 | **Ready** | `true` | `recoverable` | All three states `up` |
 | **SWFailure** | `false` | `recoverable` | `dpu_control_plane_state: down`, `dpu_midplane_link_state: up` |
 | **PowerCycle** | `false` | `recoverable` | `chassisd` issuing power-cycle; `reset_count` incremented |
@@ -589,7 +589,7 @@ Planned reboot of the entire SmartSwitch (NPU + all DPUs) via CLI: `reboot`. All
 
 | DPU Scenario | `dpu_control_plane_state` | `dpu_midplane_link_state` | `ready_status` | PMON Action |
 | ------------ | :-----------------------: | :-----------------------: | :-----------: | ----------- |
-| DPU booting – initial state | unknown | unknown | false | `chassisd` polls; waiting for DPU to come up |
+| DPU booting – initial state | down | down | false | `chassisd` polls; waiting for DPU to come up |
 | DPU healthy and running – first boot | up | up | true | Set `ready_status=true` after verifying all states |
 | DPU crash / unplanned reboot | down | down | false | Power-cycle DPU; increment `reset_count` |
 | DPU up after crash | up | up | true | Set `ready_status=true` after verifying all states |
--- a/doc/smart-switch/pmon/enhance-dpu-robustness.md
+++ b/doc/smart-switch/pmon/enhance-dpu-robustness.md
@@ -1,6 +1,6 @@
 # Smart Switch: PMON: Enhance DPU Robustness #
 
-## Table of Content ##
+## Table of Contents ##
 
 - [Revision](#revision)
 - [Scope](#scope)
@@ -10,10 +10,12 @@
 - [Critical Processes for DPU Management](#critical-processes-for-dpu-management)
 - [Timers and Thresholds](#timers-and-thresholds)
 - [DPU Status DB Info](#dpu-status-db-info)
+  - [Existing DB entries](#existing-db-entries)
+  - [New DB entries](#new-db-entries)
 - [DPU Recovery State Machine](#dpu-recovery-state-machine)
 - [DPU Software Failures](#dpu-software-failures)
-  - [Critical Process Restart on DPU](#critical-process-restart-on-dpu)
-  - [Critical Process Persistently Down on DPU](#critical-process-persistently-down-on-dpu)
+  - [Process Restart on DPU](#process-restart-on-dpu)
+  - [Process Persistently Down on DPU](#process-persistently-down-on-dpu)
   - [pmon Crash on NPU](#pmon-crash-on-npu)
   - [databasedpu Crash on NPU](#databasedpu-crash-on-npu)
 - [DPU Hardware Failures](#dpu-hardware-failures)
@@ -46,7 +48,7 @@ This document covers the High Level Design for DPU failure scenarios on a SmartS
 
 The scope includes:
 
-- DPU software failures (critical process crashes and restarts on DPU; pmon and databasedpu crashes on NPU)
+- DPU software failures (process crashes and restarts on DPU; pmon and databasedpu crashes on NPU)
 - DPU hardware failures (complete DPU down, power failure / unexpected shutdown, PCIe failure)
 - NPU/switch-level failures (kernel crash, memory exhaustion)
 - DB state tracking for DPU failure detection and recovery (new and existing DB entries)
@@ -126,8 +128,7 @@ All timers and thresholds used by PMON for DPU failure detection and recovery ar
 | Timer / Threshold | Default Value | Configurable | Used By | Description |
 | ----------------- | :-----------: | :----------: | ------- | ----------- |
 | `chassisd` health poll interval | 10 seconds | No | `chassisd` | Interval at which `chassisd` polls `dpu_control_plane_state`, `dpu_data_plane_state`, and `dpu_midplane_link_state` |
-| DPU auto-recovery timeout | 60 seconds | Yes (`platform.json`) | `chassisd` | Time allowed for a DPU to recover from a critical process restart before escalating. If `dpu_control_plane_state` or `dpu_midplane_state` remains `down` beyond this timeout, `chassisd` initiates a DPU reset, if DPU is still up and running. |
-| DPU power-cycle timeout | 180 seconds | Yes | `chassisd` | Time `chassisd` waits for `dpu_control_plane_state` to return to `up` before issuing a power-cycle |
+| `dpu_auto_recovery_timeout` | 60 seconds | Yes (`platform.json`) | `chassisd` | Self-heal grace period after a software failure is first detected. `chassisd` waits this long for the DPU to recover on its own (e.g., container supervisor restarts the failing process). If `dpu_control_plane_state` or `dpu_midplane_link_state` is still `down` when this timer expires, `chassisd` issues a DPU power-cycle. |
 | `pcied` PCIe poll interval | 60 seconds | No | `pcied` | Interval at which `pcied` checks PCIe link status for all DPUs. A PCIe failure may go undetected for up to 60 seconds. |
 | `dpu_halt_services_timeout` | 60 seconds | Yes (`platform.json`) | `gnoi_reboot_daemon.py` | Maximum time to wait for DPU services to halt gracefully during reboot/shutdown |
 | `reset_limit` | 5 | Yes (`platform.json`) | `chassisd` | Maximum number of consecutive unplanned power-cycle attempts before marking DPU as unrecoverable |
@@ -238,13 +239,17 @@ stateDiagram-v2
     Booting --> Ready : All states up
 
     Ready --> SWFailure : Control plane down
-    SWFailure --> Ready : Self recovers within 60s
-    SWFailure --> PowerCycle : 180s timeout expires
+    SWFailure --> Ready : Self recovers before dpu_auto_recovery_timeout (60s)
+    SWFailure --> PowerCycle : dpu_auto_recovery_timeout (60s) expires [auto-recovery enabled]
+    SWFailure --> ManualIntervention : dpu_auto_recovery_timeout (60s) expires [auto-recovery disabled]
 
-    Ready --> PowerCycle : HW failure detected
+    Ready --> PowerCycle : HW failure detected [auto-recovery enabled]
+    Ready --> ManualIntervention : HW failure detected [auto-recovery disabled]
 
     PowerCycle --> Booting : Power cycle issued
     PowerCycle --> Unrecoverable : reset count >= reset limit
+
+    ManualIntervention --> Booting : Operator power-cycle / module startup
 
     Ready --> PlannedShutdown : CLI module shutdown
     Ready --> PlannedReboot : CLI reboot DPU
@@ -263,6 +268,7 @@ stateDiagram-v2
 | **Ready** | `true` | `recoverable` | All three states `up` |
 | **SWFailure** | `false` | `recoverable` | `dpu_control_plane_state: down`, `dpu_midplane_link_state: up` |
 | **PowerCycle** | `false` | `recoverable` | `chassisd` issuing power-cycle; `reset_count` incremented |
+| **ManualIntervention** | `false` | `recoverable` | DPU down; `FEATURE&#124;dpu-auto-recovery` `state` is `disabled` / `always_disabled`; no power-cycle issued; awaits operator action |
 | **Offline** | `false` | `recoverable` | `oper_status: Offline` |
 | **Unrecoverable** | `false` | `unrecoverable` | `reset_count` ≥ `reset_limit` |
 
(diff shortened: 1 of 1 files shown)


gpunathilell:
Regarding the 
databasedpu crash on NPU section, there are two DBs here, one is CHASSIS_STATE_DB to which the DPU is sending data from the chassisd running on the DPU, but this is not the same as databasedpuN dockers running on the switch, these dockers are the ones which orchagent on the DPU is writing/reading from, and chassisd is not involved here, databasedpuN issues in itself should not be possible to be detected, but assuming this is caused by failure in midplane, then either way `CHASSIS_STATE_DB` is also inaccessible to the DPU

gpunathilell:
We also need a sonic-mgmt section, about how the regular smartswitch tests are planned to be executed (with/without autorecovery or not)

gpunathilell:
@vvolam it is not clear from the document as to when exactly the autorecovery is triggered, the control plane goes down even when we execute shutdown/reboot, please mention in the document about the exact scenario when the autorecovery is triggered, and if we have a timeout configured for it

vvolam:
@gpunathilell I have updated the document addressing the comments. Please review

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,645 @@
...
+**Description:**
+Any process crashes on the DPU and `dpu_control_plane_state` transitions to `down`. `chassisd` does not wait for the container supervisor to self-heal — it issues a DPU power-cycle as soon as it observes `dpu_control_plane_state: down` on its next poll.
+
+**Detection (by PMON):**
+- `chassisd` on the NPU polls `dpu_control_plane_state` every 10 seconds and observes it as `down`.
+
+**PMON Action:**
+- `chassisd` sets `ready_status` to `false` and updates `last_down_time` for the corresponding DPU.
+- `chassisd` immediately issues a power-cycle of the DPU and increments `reset_count`.
+- After the DPU comes back, `chassisd` verifies all DPU states (midplane, control plane, data plane), sets `ready_status` back to `true`, and updates `last_ready_time`.
```
gpunathilell: is this based on a timeout?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,645 @@
...
+
+    PlannedShutdown --> Offline : gNOI HALT then power down
+    PlannedReboot --> Booting : gNOI HALT then power cycle
+
+    Offline --> Booting : CLI module startup
+
+    Unrecoverable --> Booting : chassisd reset on NPU
+```
+
+| State | `ready_status` | `recovery_status` | Key DB Indicators |
```
gpunathilell, line 258: @vvolam the current state, ready_state and recovery_state and the reset_count should also be visible to user (Because we are doing reboot of the dpus) so I think we may need to add CLI for this

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,645 @@
...
+**Description:**
+The PCIe bus between the NPU and a local DPU fails, making the DPU unreachable from the NPU. The DPU may still be running internally but is disconnected from the NPU.
+
+**Detection (by PMON):**
+- `pcied` detects PCIe link down and updates `PCIE_DETACH_INFO|DPU<dpu_index>` in STATE_DB with `dpu_state: detached`.
+- Independently, `chassisd` detects midplane loss via `is_midplane_reachable()` polling and updates `dpu_midplane_link_state` → `down` in CHASSIS_STATE_DB.
+
+**PMON Action:**
+- `chassisd` sets `ready_status` to `false` and updates `last_down_time`.
+- `chassisd` immediately power-cycles the DPU (midplane link loss and PCIe detach confirm the DPU is unreachable) and increments `reset_count`.
```
gpunathilell: We only do reboot of the DPU if the DPU is currently online, I think if we detect DPU offline (because of unplanned event) we should wait to see if it auto recovers, and if not the dpu should be powered on right?

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: Thank you! I planned for this but missed updating HLD. Updated in the latest document

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: To keep the design simple, it is better to not power-cycle the DPU if it is already offline.

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: No, It is only the detection delay of polling interval of 10 secs

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,665 @@
...
+
+## DPU Recovery State Machine ##
+
+The following diagram shows the state transitions managed by `chassisd` for a single DPU. Each box represents a `chassisd`-observed DPU state; edges show the triggers and actions.
+
+```mermaid
+stateDiagram-v2
+    [*] --> Booting : DPU power on
+
+    Booting --> Ready : All states up
```
gpunathilell, line 231: @vvolam correct me if I am wrong, but if there is an image issue (i.e. broken service which prevents control plane to go up on startup) we will be in booting stage continously and the autorecovery wont fire?

gpunathilell: this would be the same case in case of this transition Ready -> SWFailure -> Booting. We would be stuck in booting, that was my question regarding if we have a timeout configured on when the ready state is expected

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,665 @@
...
+| Full SmartSwitch reboot (planned) | down → up | down → up | false → true | gNOI halt; power-cycle; re-verify |
+
+---
+
+## CLI ##
+
+The new DPU recovery state fields are exposed to operators via the existing `show chassis modules status` command, extended to include recovery information:
+
+```
+admin@sonic:~$ show chassis modules status
```
gpunathilell, line 486: Does it make sense to create a new CLI? this is too many columns and wont fit in one screen (just a NIT, not super important(

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: Yes, same mechanism applies. After `SWFailure → PowerCycle → Booting`, the `dpu_boot_timeout` timer starts. If the DPU doesn't reach Ready within 300s, it gets power-cycled again (tracked by `reset_count`). This is explicitly documented in the Timers table and in the Booting state description in the State Table. For planned reboots as well, there's a note in the DPU Cold Reboot section confirming the boot timeout applies post-reboot too.

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: Agreed, already addressed — I split it into two commands: `show chassis modules status` (existing, extended with just Ready-Status) and a new `show chassis modules recovery` command for the detailed recovery fields (Recovery-Status, Reset-Count, Last-Down-Time, Last-Ready-Time). See the updated CLI section.

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+
+## Scope ##
+
+This document covers the High Level Design for DPU failure scenarios on a SmartSwitch from the PMON (Platform Monitor) perspective — specifically focused on detection, DB state management, and recovery actions performed by `chassisd` and other PMON sub-daemons.
+
+The scope includes:
+
+- DPU software failures (process crashes and restarts on DPU; pmon and databasedpu crashes on NPU)
+- DPU hardware failures (complete DPU down, power failure / unexpected shutdown, PCIe failure)
+- NPU/switch-level failures (kernel crash, memory exhaustion)
```
rameshraghupathy: Why are calling out only "NPU/switch-level failures (kernel crash, memory exhaustion)"?  This applies to DPU as well.

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+
+| Term | Meaning                                                 |
+| ---- | ------------------------------------------------------- |
+| API  | Application Programming Interface                       |
+| ASIC | Application-Specific Integrated Circuit                 |
+| CLI  | Command-Line Interface                                  |
+| DB   | Redis Database                                          |
+| DPU  | Data Processing Unit                                    |
+| gNOI | gRPC Network Operations Interface                       |
+| gRPC | Google Remote Procedure Call                             |
```
rameshraghupathy: gRPC Remote Procedure Call.

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+---
+
+## Terminology ##
+
+| Term | Explanation |
+| ---- | ----------- |
+| chassisd | Chassis daemon running inside `pmon` on the NPU; monitors DPU health states, manages DPU power-cycle and reset operations |
+| pmon | Platform Monitor daemon on NPU; hosts `chassisd` and other hardware monitoring sub-daemons |
+| syncd | Sync daemon; manages SAI API calls to DPU ASIC |
+| control plane state | DPU SONiC is booted up, all containers are up, interfaces are up, and DPU is ready to accept configuration. Derived from SYSTEM_READY in STATE_DB. Values: `"up"`, `"down"`. |
```
judsonwilson-nvidia:
```suggestion
| control plane state | Indicates whether DPU SONiC is booted up, all containers are up, interfaces are up, and DPU is ready to accept configuration. Derived from SYSTEM_READY in STATE_DB. Values: `"up"`, `"down"`. |
```

judsonwilson-nvidia: This is a nit, but I think it reads more cleanly.

Same for midplane link state and dataplane link state, below.

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: @rameshraghupathy they are covered under section "[DPU Kernel Panic / Memory Exhaustion (HW Watchdog)](https://github.com/vvolam/SONiC/blob/de5e500dca46401d162a45fa1e04560962298fab/doc/smart-switch/pmon/enhance-dpu-robustness.md#dpu-kernel-panic--memory-exhaustion-hw-watchdog)"

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+
+## Revision ##
+
+|  Rev  |        Author       | Change Description                     |
+| :---: |  :----------------: | -------------------------------------- |
+|  0.1  |  Vasundhara Volam   | Initial Version                        |
+
+---
+
+## Scope ##
```
rameshraghupathy, line 39: @vvolam This HLD needs to be reviewed in the context of SmartSwitch HA deployments. The current document discusses only standalone recovery behavior and does not explain how HA DPU switchover impacts the design.

In an HA deployment, all of the failure scenarios described here is expected to trigger a DPU switchover. **In such case, consider if this design  can be simplified just by power cycling the failed DPU in the HA DPU switchover contest itself?**

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+| Process | Container | Role | Failure Impact |
+| ------- | --------- | ---- | -------------- |
+| `gnoi_reboot_daemon.py` | `gnmi` | Sends gNOI Reboot RPCs to DPUs for graceful shutdown / reboot | Graceful shutdown and planned reboot operations fail; DPU cannot be halted cleanly before power-cycle |
+| `sysmgr` | Host | Routes DPU planned shutdown and reboot requests to host services for execution | Planned DPU reset operations cannot be carried out |
+
+
+---
+
+## Timers and Thresholds ##
+
```
rameshraghupathy, line 118: @vvolam There are two aspects that needs to captured.

(1)  This document in its current state is not taking the HA cases. It needs to explain the states of the Peer and how this logic is applicable to the peer as well.

(2) There are multiple corner case windows that can cause false positives and please explain how this design addresses them.

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+> **Auto-recovery trigger vs. planned operations:** `chassisd` auto-recovery is triggered **only** for unplanned failures. During planned operations (graceful shutdown via `config chassis module shutdown` or DPU reboot via `reboot -d`), the `state_transition_in_progress` field in `CHASSIS_MODULE_TABLE|DPU<dpu_index>` is set to `True` **before** the DPU control plane goes down. When `chassisd` observes `dpu_control_plane_state: down`, it checks `state_transition_in_progress`: if `True`, `chassisd` skips auto-recovery because the shutdown/reboot is intentional. Auto-recovery is only initiated when `dpu_control_plane_state` or `dpu_midplane_link_state` transitions to `down` **and** no planned transition is in progress (`state_transition_in_progress == False`). There is no additional timeout configured for this check — the distinction is purely flag-based.
```
rameshraghupathy, line 132: @vvolam What is the coordination between planned and unplanned operations.  How will the auto recovery feature distinguish between an unintentional control plane down event versus a programmatic restart of the control plane or a similar event?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+}
+```
+
+| Field | Default | Description |
+| ----- | ------- | ----------- |
+| `state` | `enabled` | Enable or disable the DPU auto-recovery feature. When `disabled` or `always_disabled`, `chassisd` will not automatically power-cycle DPUs on failure. |
+| `auto_restart` | `enabled` | Standard SONiC FEATURE table field — enables `systemd` to restart the feature's associated service if it crashes. |
+| `high_mem_alert` | `disabled` | Standard SONiC FEATURE table field — high memory usage alert threshold. |
+
+> **Note:** `dpu-auto-recovery` is **not** a separate service or container. It is a feature flag entry in CONFIG_DB's `FEATURE` table, read by `chassisd` (running inside the `pmon` container) to determine whether automatic DPU power-cycle recovery is enabled. The `auto_restart` and `high_mem_alert` fields are standard SONiC FEATURE table fields required by the feature infrastructure; they do not govern `chassisd` itself. When `state` is `disabled`, `chassisd` still monitors and updates DPU states in CHASSIS_STATE_DB, but will not initiate automatic power-cycle recovery. Manual intervention is required to recover failed DPUs.
```
judsonwilson-nvidia: So what does auto_restart and high_mem_alert actually do here?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+
+    Unrecoverable --> Booting : Operator module startup or chassisd restart
+```
+
+| State | `ready_status` | `recovery_status` | Key DB Indicators |
+| ----- | :------------: | :----------------: | ----------------- |
+| **Booting** | `false` | `recoverable` | `dpu_control_plane_state: down`; `chassisd` starts `dpu_boot_timeout` timer — if DPU does not reach Ready before timeout, triggers PowerCycle (or ManualIntervention if auto-recovery disabled) |
+| **Ready** | `true` | `recoverable` | All three states `up` |
+| **SWFailure** | `false` | `recoverable` | `dpu_control_plane_state: down`, `dpu_midplane_link_state: up`; transient state before `chassisd` selects `PowerCycle` or `ManualIntervention` based on the auto-recovery feature flag. If **both** control plane and midplane are down, this is treated as a HW failure — skips SWFailure and goes to **WaitForWatchdog** (if auto-recovery enabled). |
+| **PowerCycle** | `false` | `recoverable` | `chassisd` issuing power-cycle; `reset_count` incremented |
```
judsonwilson-nvidia: How is this state detected?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+
+    PlannedShutdown --> Offline : gNOI HALT then power down
+    PlannedReboot --> Booting : gNOI HALT then power cycle
+
+    Offline --> Booting : CLI module startup
+
+    Unrecoverable --> Booting : Operator module startup or chassisd restart
+```
+
+| State | `ready_status` | `recovery_status` | Key DB Indicators |
```
judsonwilson-nvidia, line 258: A bit unclear what all is packed in the "Key DB Indicators" column.

It would be good to state here very explicitly how each state is detected. I think it's mostly there but maybe not complete?

judsonwilson-nvidia: Also, future work, would be nice to have a decision tree flow chart of how to detect what the current state is?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+| **Ready** | `true` | `recoverable` | All three states `up` |
+| **SWFailure** | `false` | `recoverable` | `dpu_control_plane_state: down`, `dpu_midplane_link_state: up`; transient state before `chassisd` selects `PowerCycle` or `ManualIntervention` based on the auto-recovery feature flag. If **both** control plane and midplane are down, this is treated as a HW failure — skips SWFailure and goes to **WaitForWatchdog** (if auto-recovery enabled). |
+| **PowerCycle** | `false` | `recoverable` | `chassisd` issuing power-cycle; `reset_count` incremented |
+| **WaitForWatchdog** | `false` | `recoverable` | Midplane down detected; `chassisd` waiting up to `dpu_boot_timeout` (600s) for DPU HW watchdog to power-cycle and bring DPU back. On DPU return, validates `dpu_reboot_cause` — accepts `Kernel Panic`, `Memory Exhaustion`, or `Watchdog`. |
+| **ManualIntervention** | `false` | `recoverable` | DPU down; `FEATURE&#124;dpu-auto-recovery` `state` is `disabled` / `always_disabled`; no power-cycle issued; awaits operator action |
```
judsonwilson-nvidia: "awaits operator action"  - is it waiting for user to reenable? Or are there other actions?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,793 @@
...
+| **Unrecoverable** | `false` | `unrecoverable` | `reset_count` ≥ `dpu_reset_limit`; operator can recover via `config chassis module startup DPU<x>` which resets `recovery_status` to `recoverable` and `reset_count` to 0 |
+
+---
+
+## DPU Software Failures ##
+
+### Process crash/restart on DPU ###
+
+**Description:**
+Any process crashes on the DPU and `dpu_control_plane_state` transitions to `down`. `chassisd` does not wait for the container supervisor to self-heal — it issues a DPU power-cycle as soon as it observes `dpu_control_plane_state: down` on its next poll.
```
judsonwilson-nvidia: A short justification would be helpful.

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+    PlannedReboot --> Booting : gNOI HALT then power cycle
+
+    Offline --> Booting : CLI module startup
+
+    Unrecoverable --> Booting : Operator module startup or chassisd restart
+```
+
+| State | `ready_status` | `recovery_status` | Key DB Indicators |
+| ----- | :------------: | :----------------: | ----------------- |
+| **Booting** | `false` | `recoverable` | `dpu_midplane_link_state: up`, `dpu_control_plane_state: down`; `dpu_boot_timeout` timer running |
```
judsonwilson-nvidia: Does this make more sense?

```suggestion
| **Booting** | `false` | `recoverable` |  `dpu_boot_timeout` timer running `AND` `NOT` (`dpu_midplane_link_state: up` `AND` `dpu_control_plane_state: up`) ```

File: doc/smart-switch/pmon/enhance-dpu-robustness.md [hunk 18]
```diff
@@ -0,0 +1,557 @@
...
+| **Ready** | `true` | `recoverable` | `dpu_midplane_link_state: up`, `dpu_control_plane_state: up`, `dpu_data_plane_state: up` |
```
judsonwilson-nvidia, line 261: Does this require dpu_data_plane_state: up ?   Is the data plane state part of the state machine at all?

doc/smart-switch/pmon/enhance-dpu-robustness.md
judsonwilson-nvidia: Sorry, this chain belongs with the table in the previous section, DPU Recovery State Machine

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+| **WaitForSelfRecovery** | `false` | `recoverable` | `dpu_control_plane_state: down` OR `dpu_midplane_link_state: down`; `dpu_self_recovery_timeout` timer running |
+| **PowerCycle** | `false` | `recoverable` | `dpu_midplane_link_state: down`, `dpu_control_plane_state: down`; `reset_count` incremented; power-cycle in progress |
+| **ManualIntervention** | `false` | `recoverable` | `dpu_control_plane_state: down` OR `dpu_midplane_link_state: down`; `FEATURE|dpu-auto-recovery` `state`: `disabled` / `always_disabled` |
```
judsonwilson-nvidia: The `FEATURE|dpu-auto-recovery`  text isn't playing nice with the markup render in preview:

Image

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: No, as mentioned in the HLD, dpu_data_plane_state is only needed to mark the ready status but don't trigger recovery flow

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: Yes it requires user to recover DPU

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: These are not needed as I will use config_DB to enable the feature, since this is not a service

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: Auto recovery detects planned operations by reading state_transition_in_progress flag and skips recovery in case of planned operations

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: Please check latest simplified version. Thank you

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: @rameshraghupathy This is chassisd behavior and requirement. It is independent of HA flow. HA will the consumer of the DB information if needed.

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+
+```
+DPU_STATE|DPU<dpu_index>:
+{
+  "dpu_control_plane_state": "up" | "down",
+  "dpu_control_plane_time":  "<UTC timestamp>",
+  "dpu_data_plane_state":    "up" | "down",
+  "dpu_data_plane_time":     "<UTC timestamp>",
+  "dpu_midplane_link_state": "up" | "down",
+  "dpu_midplane_link_time":  "<UTC timestamp>"
```
judsonwilson-nvidia, line 152: Does dpu_midplane_link_time still exist?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+  "last_down_time":                    "<UTC timestamp>",
+  "last_ready_time":                   "<UTC timestamp>"
+}
+```
+
+| Field | Description | Set by | Cleared by |
+| ----- | ----------- | ------ | ---------- |
+| `ready_status` | Set to `"true"` when the DPU is fully up and ready (midplane, control plane, data plane all up). Set to `"false"` when the DPU goes down or undergoes a reset. | `chassisd` | `chassisd` (set to `"false"` on failure/reset) |
+| `recovery_status` | Set to `"recoverable"` on initialization. Set to `"unrecoverable"` when `reset_count` reaches `dpu_reset_limit`. Reset back to `"recoverable"` (and `reset_count` to 0) on: (1) `chassisd` restart (pmon crash / NPU reboot), or (2) operator-initiated `config chassis module startup DPU<x>` on an unrecoverable DPU. | `chassisd` | `chassisd` (reset to `"recoverable"` on chassisd restart or operator module startup) |
+| `reset_count` | Number of unplanned DPU resets. Reset to 0 on `chassisd` reset on NPU (e.g., NPU reboot, `pmon` restart). | `chassisd` | `chassisd` |
```
judsonwilson-nvidia, line 202: Nit:
The recovery_status description says that it can also reset to zero in some cases when `config chassis module startup DPU<x>` is called?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+
+---
+
+## DPU Recovery State Machine ##
+
+The following diagram shows the state transitions managed by `chassisd` for a single DPU. Each box represents a `chassisd`-observed DPU state; edges show the triggers and actions.
+
+```mermaid
+stateDiagram-v2
+    [*] --> Booting : DPU power on
```
judsonwilson-nvidia, line 229: It's unclear how startup should work on chassid start.  I assume it should do nothing if admin state = down. If admin state is up, should it always go to booting? What if it was already in a state where the midplane and control plane were up -- should it wait for the dataplane to be up?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+    PlannedReboot --> Booting : gNOI HALT then power cycle
+
+    Offline --> Booting : CLI module startup
+
+    Unrecoverable --> Booting : Operator module startup or chassisd restart
+```
+
+| State | `ready_status` | `recovery_status` | Key DB Indicators |
+| ----- | :------------: | :----------------: | ----------------- |
+| **Booting** | `false` | `recoverable` | `dpu_boot_timeout` timer running AND NOT (`dpu_midplane_link_state: up` AND `dpu_control_plane_state: up`) |
```
judsonwilson-nvidia:
```suggestion
| **Booting** | `false` | `recoverable` | `dpu_boot_timeout` timer running AND NOT (`dpu_midplane_link_state: up` AND `dpu_control_plane_state: up` AND `dpu_data_plane_state: up`) |
```

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+
+| DB Field | Before | During WaitForSelfRecovery | After Power-Cycle Recovery |
+| -------- | :----: | :------------------------: | :------------------------: |
+| `dpu_control_plane_state` | `up` | `down` | `up` |
+| `dpu_midplane_link_state` | `up` | `down` | `up` |
+| `ready_status` | `true` | `false` | `true` |
+| `last_down_time` | — | `<UTC timestamp>` | — |
+| `last_ready_time` | — | — | `<UTC timestamp>` |
+| `reset_count` | N | N | N+1 |
+| `recovery_status` | `recoverable` | `recoverable` | `recoverable` (or `unrecoverable` if N+1 ≥ `dpu_reset_limit`) |
```
judsonwilson-nvidia: Nit:
The "After Power-Cycle Recovery" column doesn't make sense for the "unrecoverable" state... 
The "recovery_status" can't be "unrecoverable" if it recovered. And if it does go to "unrecoverable", we almost certainly won't have 2 up states, ready = true, etc.

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+
+> **Note:** `chassisd` polls `dpu_data_plane_state` alongside `dpu_control_plane_state` and `dpu_midplane_link_state`, but `dpu_data_plane_state` alone does not trigger recovery actions. The `dpu_data_plane_state` is used solely to determine full DPU readiness for setting `ready_status` to `true`.
+
+> **Note:** During **WaitForSelfRecovery**, if a planned operation (`config chassis module shutdown`) is requested, `chassisd` cancels the self-recovery timer, powers down the DPU, and transitions to **Offline**.
```
judsonwilson-nvidia: This is missing from the mermaid. Also what about planned reboot?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+
+**Detection (by PMON):**
+- `chassisd` on the NPU polls `dpu_control_plane_state`, `dpu_data_plane_state`, and `dpu_midplane_link_state` every 10 seconds.
+- A failure is detected when either `dpu_control_plane_state` or `dpu_midplane_link_state` transitions to `down`.
+
+**PMON Action:**
+1. `chassisd` sets `ready_status` to `false` and updates `last_down_time` for the affected DPU.
+2. `chassisd` enters the **WaitForSelfRecovery** state and starts the `dpu_self_recovery_timeout` timer (default 300 seconds). This grace period allows the DPU to recover from transient failures (e.g., process restart, HW watchdog-triggered reboot) without external intervention.
+3. `chassisd` continues polling the DPU state every 10 seconds during this period.
+4. **If the DPU self-recovers** (midplane comes back up OR control plane comes back up) within the `dpu_self_recovery_timeout`:
```
judsonwilson-nvidia, line 285: Couple of questions:
1) Is it OK to wait for data plane in this case (which happens in Booting state)? I assume its OK, but just want to check.
2) If it's control plane, is it OK to not power cycle? I thought we always want to power cycle in this case? Or is there something else that will ensure this happens (dpu detects stale state and that does something???)

I think when I was suggesting that we consolidate SW and HW failures, that after the timeout period we would always power cycle and never actually keep going after self recovery.  I was under the impression we might need to allow this time for kernel panics to be recorded properly, or something to this affect, rather than just power-cycle immediately. (It would be good to document this if it's true.)

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+   - `transition_start_time`: `<UTC timestamp>`
+   - `transition_type`: `shutdown`
+3. `chassisd` updates CHASSIS_STATE_DB:
+   - `DPU_STATE|DPU<dpu_index>`: `ready_status`: `false`, `last_down_time`: `<UTC timestamp>`
+4. `gnoi_reboot_daemon.py` detects the transition and sends gNOI Reboot RPC (Method: `HALT`) to DPU.
+5. DPU gracefully shuts down all services via `reboot -p`.
+6. NPU polls `gnoi_client -rpc RebootStatus` until `active=false` (services terminated).
+7. `state_transition_in_progress` set to `False`.
+8. `module_base.py` calls platform API `power_down()` to power off DPU.
+9. PCIe detach: platform vendor API `pci_detach()`.
```
judsonwilson-nvidia: i think 7, 8 and 9 are in reversed order? 10 needs reordering as well?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+- If `pcied` detects a PCIe failure and updates `PCIE_DETACH_INFO` at the same time `chassisd` initiates a power-cycle due to midplane loss: `chassisd` holds a per-DPU lock during the power-cycle sequence. `pcied` updates `PCIE_DETACH_INFO` independently (no lock contention). `chassisd` reads `PCIE_DETACH_INFO` during its power-cycle flow and performs PCIe rescan if `dpu_state` is `detached`. No conflicting actions occur because `pcied` is read-only from `chassisd`'s perspective — it only updates state, while `chassisd` acts on it.
+
+---
+
+### DPU Cold Reboot ###
+
+**Description:**
+Reboot a DPU with full power-cycle via CLI: `reboot -d <DPU_ID>`.
+
+**PMON Sequence:**
```
judsonwilson-nvidia, line 404: Unless you changed it, I think this doesn't involve pmon for the most part.

Also, not sure if it matters, but this doesn't mention the graceful shutdown module-state-transition flag.

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+- If switch reboot is requested during module shutdown: graceful shutdown completes; switch reboot proceeds.
+- Concurrent startup/shutdown on the same module: fails; user retries later.
+- If `config chassis module shutdown` is issued while `chassisd` is in the middle of an auto-recovery power-cycle for the same DPU: `chassisd` detects the admin-down request, aborts the auto-recovery loop, and proceeds with the graceful shutdown sequence.
+- If `pcied` detects a PCIe failure and updates `PCIE_DETACH_INFO` at the same time `chassisd` initiates a power-cycle due to midplane loss: `chassisd` holds a per-DPU lock during the power-cycle sequence. `pcied` updates `PCIE_DETACH_INFO` independently (no lock contention). `chassisd` reads `PCIE_DETACH_INFO` during its power-cycle flow and performs PCIe rescan if `dpu_state` is `detached`. No conflicting actions occur because `pcied` is read-only from `chassisd`'s perspective — it only updates state, while `chassisd` acts on it.
+
+---
+
+### DPU Cold Reboot ###
```
judsonwilson-nvidia, line 399: This seems mostly out of scope of this document, except that it could race with the new state machine during power-cycle / reboot?  Should the state machine use the graceful-shutdown module-state-transition flag to synchronize with admin->up/down and reboot?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+- If DPU stuck: hardware watchdog triggers reset (vendor-specific).
+
+---
+
+### Full SmartSwitch Reboot ###
+
+**Description:**
+Planned reboot of the entire SmartSwitch (NPU + all DPUs) via CLI: `reboot`. All DPUs are gracefully shut down in parallel before the NPU reboots.
+
+**PMON Sequence:**
```
judsonwilson-nvidia, line 437: Again, I don't think this is PMON unless it's being changed?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+1. NPU sends gNOI Reboot RPC (Method: `HALT`) to **all** DPUs in parallel (multiple threads).
+2. NPU polls gNOI `RebootStatus` for each DPU until `active=false` and `Status=STATUS_SUCCESS`.
+3. Timeout per DPU: `dpu_halt_services_timeout` (default from `platform.json`, typically 60 seconds).
+4. For each DPU: PCIe detach via platform vendor API `pci_detach()`.
+5. NPU proceeds with its own reboot sequence.
+6. On NPU boot, PCIe enumeration discovers all DPUs.
+7. `chassisd` power-cycles each DPU and performs PCIe reattach.
+8. Each DPU boots: midplane attach → SONiC boot → container startup → reports `dpu_control_plane_state=up`.
```
judsonwilson-nvidia: Is this true? The behavior of chassisd after the NPU comes up looks similar to the ungraceful NPU reboot case above - I'm not understanding if it's different or why. If it's different, it would be good to point out how and why.

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+> **Note — Multiple DPU recovery:** When multiple (or all) DPUs need recovery simultaneously, `chassisd` issues power-cycles sequentially (one DPU at a time) to avoid power-rail overload and PCIe bus contention. The `dpu_boot_timeout` is tracked per-DPU independently. If a platform supports parallel DPU power-cycle (declared in `platform.json` via `parallel_dpu_recovery: true`), `chassisd` may issue power-cycles in parallel batches.
+
+**DB State Transition:**
+
+| DB Field | Before | During Reboot | After Recovery |
+| -------- | :----: | :-----------: | :------------: |
+| `ready_status` (all DPUs) | `true` | `false` | `true` (per DPU) |
+| `last_down_time` (all DPUs) | — | `<UTC timestamp>` | — |
+| `last_ready_time` (all DPUs) | — | — | `<UTC timestamp>` (per DPU) |
+| `PCIE_DETACH_INFO` `dpu_state` (per DPU) | `reattached` | `detaching` → `detached` | `reattached` |
```
judsonwilson-nvidia, line 456: Nit:
I don't think the DPUs are guaranteed to reattach before the NPU goes down. I think we skip it (but hotplug can trigger it.)

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+    ManualIntervention --> Booting : Operator power-cycle / module startup
+
+    Booting --> Offline : CLI module shutdown during boot
+    Ready --> PlannedShutdown : CLI module shutdown
+    Ready --> PlannedReboot : CLI reboot DPU
+
+    PlannedShutdown --> Offline : gNOI HALT then power down
+    PlannedReboot --> Booting : gNOI HALT then power cycle
+
+    Offline --> Booting : CLI module startup
```
judsonwilson-nvidia: Can PlannedShutdown and Offline be merged into a single state called AdminDown where the state machine is just waiting for admin state up again? After it detects admin state up it can then maybe go to booting?  (All the actual planned shutdown actions are happening separately in the existing PMON flow, right?)

judsonwilson-nvidia (approved):
See my previous comments. Approving with goal of getting the code out there so we can see if it works. If there are serious problems, we can hopefully just disable the feature until it is fixed.

Reiterating my top concerns:

1. It seems like PowerCycle state should set the module-state-transition flag so as not to have races with graceful shutdown? Handling these races with other power up/down/reboot events has been my strongest concern with these design, and the module-state-transition and the "graceful shutdown" procedure seemed like the way we handle this? You know better than I do.

2. Is it OK for the DPU to self-recover if there is a control plane failure, or should we just force a reboot in all failure scenarios after some amount of time (allowing for any necessary watchdog periods, etc.)?

gpunathilell (dismissed):
@vvolam please specify what is done during power cycle (exact operations). other steps look good for now. need to verify code

vvolam:
> @vvolam please specify what is done during power cycle (exact operations). other steps look good for now. need to verify code

Sure @gpunathilell, I will address this as a follow-up PR. Thank you

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: yes, it does

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: That's right, on a planned power-cycle, we would like to reset the recover_status

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: That's right. If admin-state is down, it has nothing to do with the state diagram. Once admin-state is up, chassisd will keep polling for ready_status which is true only when all three states are up

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: Yes, waiting for data plane in Booting state is intentional — `dpu_data_plane_state`: up is required for full readiness (`ready_status=true`), but it does NOT trigger recovery on its own if it goes down.

The self-recovery path is by design. The `dpu_self_recovery_timeout `serves two purposes:

- Allow kernel panics/watchdog to fire and record crash data
- Allow the DPU to self-recover from transient failures (e.g., critical process restart by supervisord, short-lived hardware glitch)

If the DPU self-recovers (control plane or midplane comes back up), we transition to Booting and still enforce `dpu_boot_timeout` for full readiness. If it doesn't reach Ready within that timeout, we power-cycle anyway. So a partially-recovered DPU that's stuck will still get power-cycled — just via the boot timeout path rather than immediately.

This avoids unnecessary power-cycles for transient issues that `supervisord` or the DPU's own watchdog can handle autonomously.

doc/smart-switch/pmon/enhance-dpu-robustness.md
judsonwilson-nvidia: It would be good to clarify that this does not involve the state machine, and describe how it doesn't race?

vvolam: Yes, the state machine uses the `state_transition_in_progress `flag to synchronize. During a planned cold reboot, chassisd sets `state_transition_in_progress=True` which suppresses auto-recovery. The flag is cleared after the DPU halts services (step 5), so if the DPU fails to come back after the power-cycle, `dpu_boot_timeout `kicks in normally.

This section is included because chassisd owns the reboot sequence (gNOI HALT → PCIe detach → power-cycle → reattach) and because the boot-timeout recovery applies post-reboot. But agreed it's largely existing behavior — the new addition is just the interaction with the recovery state machine.

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: This is PMON — gnoi_reboot_daemon.py (running inside pmon container) handles the gNOI HALT and chassisd handles the power-cycle/PCIe sequence (steps 4-6) and boot verification (step 8).

Regarding `state_transition_in_progress`: it's already mentioned in the Note below the sequence — the flag is cleared after DPU halts services so that `dpu_boot_timeout `recovery is not suppressed post-reboot.

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: no, just added for completeness

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: Modified.

vvolam:
> > @vvolam please specify what is done during power cycle (exact operations). other steps look good for now. need to verify code
> 
> Sure @gpunathilell, I will address this as a follow-up PR. Thank you

@gpunathilell I have addressed. Please review top 2 commits to the PR.

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: That's right, both are identical, except in case of ungraceful NPU reset, DPUs will be reset as they are out of sync.

doc/smart-switch/pmon/enhance-dpu-robustness.md
vvolam: Agreed — fixed in latest push. Step 7 now explicitly does a full cycle (power_down() → pci_detach() → power_up() → pci_reattach()) so it doesn't assume DPUs are in any particular PCIe state on NPU boot. Whether they auto-reattached via hotplug or not, the power-cycle resets everything cleanly.

gpunathilell (approved):

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+
+    Ready --> WaitForSelfRecovery : Control plane OR midplane down [auto-recovery enabled]
+    Ready --> ManualIntervention : Control plane OR midplane down [auto-recovery disabled]
+
+    WaitForSelfRecovery --> Booting : DPU self-recovered (midplane up OR control plane up)
+    WaitForSelfRecovery --> PowerCycle : self-recovery timeout expired
+    WaitForSelfRecovery --> AdminDown : CLI module shutdown
+    WaitForSelfRecovery --> Booting : CLI reboot DPU (cancel timer, power-cycle)
+
+    PowerCycle --> Booting : Power cycle issued
```
judsonwilson-nvidia, line 243: ` PowerCycle --> Booting : Power cycle issued`

I don't think this exists in the code (and probably shouldn't?). I think it's a self loop back to PowerCycle when the physical power cycling is issued?

doc/smart-switch/pmon/enhance-dpu-robustness.md
```diff
@@ -0,0 +1,557 @@
...
+    PowerCycle --> Unrecoverable : reset count >= dpu_reset_limit
+    PowerCycle --> AdminDown : CLI module shutdown
+
+    ManualIntervention --> Booting : Operator power-cycle / module startup
```
judsonwilson-nvidia, line 247: Is there also a transition from ManualIntervention to AdminDown that is missing? I believe this is what the code does.

Issue #106, opened 2017-09-07
Boardcom sai seems not support attribute SAI_PORT_ATTR_SUPPORTED_SPEED


tieguoevan:
Hi,
I'm using Broadcom sai 0.9.4, but it seems not support sai attribute SAI_PORT_ATTR_SUPPORTED_SPEED. 

> Sep  7 14:27:56.869544 sonic WARNING orchagent: :- validatePortSpeed: Unable to validate speed for port 1000000000025. Not supported by platform
Sep  7 14:27:56.871497 sonic NOTICE orchagent: :- doPortTask: Set port Ethernet30 admin status to up
Sep  7 14:27:56.872011 sonic NOTICE orchagent: :- doPortTask: Set port Ethernet30 MTU to 9216
Sep  7 14:27:56.874333 sonic WARNING syncd: :- processEvent: get API for key: SAI_OBJECT_TYPE_PORT:oid:0x1000000000023 op: get returned status: SAI_STATUS_ATTR_NOT_SUPPORTED_0
Sep  7 14:27:56.875314 sonic ERR orchagent: :- meta_sai_get_oid: get status: -327680
Sep  7 14:27:56.875314 sonic WARNING orchagent: :- validatePortSpeed: Unable to validate speed for port 1000000000023. Not supported by platform
Sep  7 14:27:56.877580 sonic NOTICE orchagent: :- doPortTask: Set port Ethernet56 speed to 25000
Sep  7 14:27:56.877868 sonic NOTICE orchagent: :- doPortTask: Set port Ethernet56 admin status to down
Sep  7 14:27:56.878517 sonic NOTICE orchagent: :- doPortTask: Set port Ethernet56 MTU to 1500

stcheng:
correct. this is not supported on v0.9.4. will it affect anything?

tieguoevan:
The problem is that if we set port speed, the orchagent will crash.

And the file will set port speed if we use it to generate minigraph.xml.
 https://github.com/Azure/sonic-mgmt/blob/master/ansible/templates/topo/dev_metadata.j2 

> Sep 11 03:14:05.286789 sonic NOTICE swssconfig: :- main: Loading config from JSON file:/etc/swss/config.d/ports.json...
Sep 11 03:14:05.292007 sonic ERR orchagent: :- meta_sai_get_oid: get status: -327680
[...]

PR #2313, opened 2026-04-28
[dhcp_relay] Fix HLD: align agent_relay_mode values with ISC CLI aliases

### LINKED ISSUE #24052: Bug:  [dhcp_relay] agent_relay_mode value set is inconsistent across YANG schema, dhcp_relay CLI, and dhcp4relay daemon
### Is it platform specific

generic

### Importance or Severity

Critical

## Summary
 
 The `agent_relay_mode` field of the `DHCPV4_RELAY` table is declared
 with one set of enum values in the YANG schema, accepted with a
 different set by the `dhcp_relay` CLI, and honored with yet another
 set by the `dhcp4relay` daemon. Filing this for the dhcp_relay /
 sonic-dhcpv4-relay owners to track and fix; I hit it as a downstream
 side-effect while triaging an unrelated test failure and don't have
 the area expertise to drive the cleanup.
 
 ## Symptom (where I tripped over it)
 
 Running `tests/dhcp_relay/test_dhcpv4_relay.py::test_dhcp_relay_agent_mode`
 on a 720dt testbed produces a libyang validation error during the
 post-test config snapshot / teardown phase:
 

libyang[0]: Invalid value "append" in "agent_relay_mode" element.

 
 The test itself parametrizes `agent_relay_mode` with `discard`,
 `replace`, `append` (the values the CLI accepts and the daemon
 honors), but the YANG typedef does not list any of those — it lists
 `forward_and_append`, `forward_and_replace`, `forward_untouched`,
 `discard`. So writing the CLI's value into CONFIG_DB makes the next
 `config save` / sonic-cfggen pass fail YANG validation.
 
 I attempted a minimal additive fix in
 sonic-net/sonic-buildimage#26886 (now closed — see "Why I'm not
 fixing this myself" below).
 
 ## Where each layer stands today
 
 | Layer | File | Accepted / honored values |
 |---|---|---|
 | CLI plugin | `sonic-buildimage/dockers/docker-dhcp-relay/cli/config/plugins/dhcp_relay.py` | `click.Choice(["discard", "append", "replace"])` (×4) |
 | Daemon | `sonic-dhcp-relay/dhcp4relay/src/dhcp4relay.cpp` ~L593-610 | hard-coded `strcmp` against `"append"` and `"replace"`; **everything else** (incl. `"discard"`, the `forward_*` strings, an unset/empty field, or any typo) falls through to the else-branch and the packet is dropped |
 | YANG schema | `sonic-buildimage/src/sonic-yang-models/yang-templates/sonic-types.yang.j2` typedef `relay-agent-mode` | `forward_and_append`, `forward_and_replace`, `forward_untouched`, `discard` (default `forward_untouched`) |
 | HLD doc | `SONiC/doc/DHCPv4_relay/DHCPv4-relay-agent-High-Level-Design.md` | uses the YANG (long-form) values throughout — typedef snippet, sample CONFIG_DB, CLI help, example commands, show output |
 | Test docstring | `sonic-mgmt/tests/dhcp_relay/test_dhcpv4_relay.py::test_dhcp_relay_agent_mode` | docstring lists long-form values; `@pytest.mark.parametrize` actually uses `discard`/`replace`/`append` |
 
 The schema-allowed set and the daemon-honored set are disjoint on the
 values that matter (`append`, `replace`).
 
 ## Important nuance: when `agent_relay_mode` is consulted at all
 
 The daemon consults `agent_relay_mode` **only for relay-of-relay
 traffic** — DHCP packets that already carry an Option 82 from another
 upstream relay agent. For ordinary client packets (no Option 82), the
 daemon always forwards and adds its own Option 82 regardless of
 `agent_relay_mode`.
 
 For relay-of-relay traffic specifically:
 
 | `agent_relay_mode` value | Daemon behaviour |
 |---|---|
 | `"append"` | Forward + append our own Option 82 |
 | `"replace"` | Strip existing Option 82, add our own, forward |
 | `"discard"` | Drop |
 | unset / `"forward_untouched"` / `"forward_and_append"` / `"forward_and_replace"` / typo | Drop (else-branch fall-through) |
 
 Two consequences worth flagging:
 
 1. The YANG default `forward_untouched` is misleading — the name
    implies "forward unchanged" but the daemon actually *drops* such
    packets. The daemon's true no-config behaviour falls into the
    *discard* branch.
 2. The three `forward_*` enums are dead values: nothing in the
    codebase ever sets them and the daemon never honored them.
 
 Source: [`dhcp4relay.cpp` ~L593-610](https://github.com/sonic-net/sonic-dhcp-relay/blob/master/dhcp4relay/src/dhcp4relay.cpp#L593-L610).
 
 ## Suggested fix (for the dhcp_relay owners)
 
 Pick `discard` / `append` / `replace` as the single source of truth
 across YANG, CLI, daemon, HLD doc, and test docstring, and change the
 YANG default to `discard` to match the daemon's actual no-config
 behaviour. Org-wide grep shows zero production references to the
 `forward_*` values (no daemon, CLI, orchagent, dbmigrator,
 golden-config, ansible playbook, or YAML fixture sets them) — they
 exist only in the YANG typedef itself, the YANG sample CONFIG_DB,
 the HLD doc, and the test docstring above. So removing them from the
 typedef is safe.
 
 cc / FYI: dhcpv4-relay maintainers (please re-assign as needed).

### Steps to Reproduce

On any switch running an image with the new SONiC DHCPv4 relay (has_sonic_dhcpv4_relay=true):

 # 1. Pick any VLAN that has a DHCPV4_RELAY entry (or create one).
 VLAN=Vlan1000
 DHCP_SERVER=192.168.20.100
 
 # 2. Use the dhcp_relay CLI to set agent_relay_mode to a value the CLI accepts.
 sudo config dhcp_relay ipv4 helper add ${VLAN} ${DHCP_SERVER}
 sudo config dhcp_relay ipv4 helper update ${VLAN} --agent-relay-mode append
 
 # 3. Confirm the value landed in CONFIG_DB.
 sonic-db-cli CONFIG_DB HGET "DHCPV4_RELAY|${VLAN}" agent_relay_mode
 # -> "append"
 
 # 4. Trigger any YANG validation pass.
 sudo config save -y
 #   - or -
 sudo config reload -y
 #   - or -
 sonic-cfggen -d -y /etc/sonic/init_cfg.json -t /usr/share/sonic/templates/...

Equivalent reproducer via the existing test:

 cd sonic-mgmt
 ./run_tests.sh -n <testbed-name> -c "dhcp_relay/test_dhcpv4_relay.py::test_dhcp_relay_agent_mode"
 # Observe libyang error during the post-test config snapshot / teardown phase.

### Actual Behavior and Expected Behavior

Actual behavior

The YANG schema rejects the CLI's value, so step 4 above (or the post-test config snapshot) fails with:

 libyang[0]: Invalid value "append" in "agent_relay_mode" element.

The same error occurs for replace. Symmetrically, the schema accepts forward_and_append, forward_and_replace, and forward_untouched, but the daemon does not honor any of them — they all silently fall into the discard else-branch (dhcp4relay.cpp ~L593-610), so even setting one of those "valid" values does not produce the behaviour its name suggests.

In a sonic-mgmt test context this surfaces as a teardown-time YANG validation failure (observed in test_dhcp_relay_agent_mode and adjacent tests on 720dt testbeds), with PRs #24043 / #24044 needing additional cleanup workarounds purely to keep the YANG validator happy.

Expected behavior

YANG should accept exactly the values the CLI emits and the daemon implements:

 - discard — drop relay-of-relay packets (also the daemon's actual no-config default, so this should be the YANG default as well).
 - append — forward + append our own Option
  82.
 - replace — strip existing Option 82, add ours, forward.

config save, config reload, and sonic-cfggen should all complete without libyang errors after any CLI-driven --agent-relay-mode change. The legacy forward_and_append / forward_and_replace / forward_untouched strings should either be removed or, at minimum, documented as deprecated aliases that map to the daemon-honored values (preferably the former — they
are dead values today, with zero references in any production code path across the sonic-net org).

### Relevant log output

```shell

```

### Output of `show version`

```shell

```

### Attach files (if any)

_No response_

ashutosh-agrawal:
Update `DHCPv4-relay-agent-High-Level-Design.md` to align all `agent_relay_mode`
references with the short-form ISC CLI aliases (`append`, `replace`, `forward`,
`discard`) that the daemon and CLI actually use.

The HLD was using long-form ISC internal enum names (`forward_and_append`,
`forward_and_replace`, `forward_untouched`) which were never accepted by the CLI
and never honored by the daemon, causing confusion and contributing to the
YANG validation failures reported in sonic-net/sonic-mgmt#24052.

Changes: CLI help text, example commands, YANG snippet default, CONFIG_DB
sample, show output table, migration table note.

| Repo | PR |
| --- | --- |
| sonic-net/sonic-dhcp-relay | [#106 — daemon fix](https://github.com/sonic-net/sonic-dhcp-relay/pull/106) |
| sonic-net/sonic-buildimage | [#27029 — YANG + CLI fix](https://github.com/sonic-net/sonic-buildimage/pull/27029) |

### DIFF
--- a/doc/DHCPv4_relay/DHCPv4-relay-agent-High-Level-Design.md
+++ b/doc/DHCPv4_relay/DHCPv4-relay-agent-High-Level-Design.md
@@ -421,7 +421,7 @@ module sonic-dhcpv4-relay {
                 leaf agent_relay_mode {
                     description "How to forward packets that already have a relay option";
                     type stypes:relay-agent-mode;
-                    default forward_untouched;
+                    default replace;
                 }
 
                 leaf max_hop_count {
@@ -458,7 +458,7 @@ A new table, named DHCPV4_RELAY, will be introduced in the config-db to define D
             "link_selection": "enable",
             "vrf_selection": "enable",
             "server_id_override": "enable",
-            "agent_relay_mode": "forward_untouched"
+            "agent_relay_mode": "replace"
         }
     }
 }
@@ -521,9 +521,9 @@ sudo config dhcp_relay ipv4 helper del 1000 7.7.7.7
           --link-selection <enable|disable>          Enable link selection
           --vrf-selection  <enable|disable>          Enable VRF selection
           --server-id-override <enable|disable>      Enable server id override
-          --agent-relay-mode <forward_and_append|
-                              forward_and_replace|
-                              forward_untouched|
+          --agent-relay-mode <append|
+                              replace|
+                              forward|
                               discard>              How to forward packets that already have a relay
                                                     option
           --max-hop-count <1..16>                   Maximum hop count for relayed packets
@@ -575,7 +575,7 @@ sudo config dhcp_relay ipv4 helper del 1000 7.7.7.7
     - Update dhcpv4 relay configuration to replace relay-options if it receives a packet that already contains relay-options
 
         ```
-        sudo config dhcp_relay ipv4 helper update Vlan12 --agent-relay-mode forward_and_replace
+        sudo config dhcp_relay ipv4 helper update Vlan12 --agent-relay-mode replace
         ```
 
 #### 9.2 Show CLI
@@ -608,7 +608,7 @@ root@sonic:/home/cisco# show dhcp_relay ipv4 helper
 root@sonic:/home/cisco# show dhcp_relay ipv4 helper
 NAME    SERVER VRF    SOURCE INTERFACE    LINK SELECTION    VRF SELECTION    SERVER ID OVERRIDE    AGENT RELAY MODE       MAX HOP COUNT  DHCPV4 SERVERS
 ------  ------------  ------------------  ----------------  ---------------  --------------------  -------------------  ---------------  ----------------
-Vlan12  Vrf01         Loopback0           enable            enable           enable                forward_and_replace                4  192.168.12.1
+Vlan12  Vrf01         Loopback0           enable            enable           enable                replace                4  192.168.12.1
                                                                                                                                          192.168.12.2
 ```
 
@@ -706,7 +706,7 @@ The table below lists the various isc-dhcp command-line options currently used i
 
 | Existing isc cmd line arg | New Configuration | Comments |
 |-|-|-|
-| -m discard | --agent-relay-mode with the same default value| |
+| -m discard | --agent-relay-mode discard | ISC default was replace; new default is also replace (matches ISC) |
 | -a %%h:%%p | Not required | Always insert agent-relay-options in hostname:int-alias:vlan-id format |
 | %%P | Not Required | Always insert remote-id in the agent-relay-options |
 | --name-alias-map-file | Not Required | Interface Alias is dynamically retrieved from ConfigDB |
--- a/doc/DHCPv4_relay/DHCPv4-relay-agent-High-Level-Design.md
+++ b/doc/DHCPv4_relay/DHCPv4-relay-agent-High-Level-Design.md
@@ -421,7 +421,7 @@ module sonic-dhcpv4-relay {
                 leaf agent_relay_mode {
                     description "How to forward packets that already have a relay option";
                     type stypes:relay-agent-mode;
-                    default replace;
+                    default discard;
                 }
 
                 leaf max_hop_count {
@@ -458,7 +458,7 @@ A new table, named DHCPV4_RELAY, will be introduced in the config-db to define D
             "link_selection": "enable",
             "vrf_selection": "enable",
             "server_id_override": "enable",
-            "agent_relay_mode": "replace"
+            "agent_relay_mode": "discard"
         }
     }
 }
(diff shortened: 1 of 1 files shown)


doc/DHCPv4_relay/DHCPv4-relay-agent-High-Level-Design.md
```diff
@@ -706,7 +706,7 @@ The table below lists the various isc-dhcp command-line options currently used i
 
 | Existing isc cmd line arg | New Configuration | Comments |
 |-|-|-|
-| -m discard | --agent-relay-mode with the same default value| |
+| -m discard | --agent-relay-mode discard | ISC default was replace; new default is also replace (matches ISC) |
```
Xichen96: where do you get isc default replace?

Xichen96: In the isc template it says discard

command=/usr/sbin/dhcrelay -d -m discard -a %%h:%%p %%P --name-alias-map-file /tmp/port-name-alias-map.txt -id {{ vlan_name }}

ashutosh-agrawal: You're right — the SONiC ISC template always passed `-m discard` explicitly, so the effective default was `discard`, not the ISC binary's own internal default of `replace`. Fixed: changed the YANG default to `discard` in both `sonic-dhcpv4-relay.yang` and the HLD, updated the migration table comment, and fixed `tests/files/sample_config_db.json` which still had the old `forward_untouched` value (causing the CI YANG test failure).

ashutosh-agrawal:
@Xichen96 Please help with review.

Xichen96 (approved):

PR #2317, opened 2026-05-04
Align transceiver CLIs formatting for common and platform-specific technology extensions

benle7:
This HLD defines the minimal changes required to align the 'sfputil' and 'show interface transceiver' CLIs
to include new common and platform-specific data in the output (CPO as reference use case).

### DIFF
--- a/doc/transceiver_clis_key_formatting/platform_specific_key_formatting.md
+++ b/doc/transceiver_clis_key_formatting/platform_specific_key_formatting.md
@@ -1,0 +1,179 @@
+# Align transceiver CLIs formatting for common and platform-specific technology extensions
+
+## 1. Revision
+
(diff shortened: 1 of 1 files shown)


abelamit (dismissed):

abelamit (approved):

prgeor (commented):
@benle7 can you refer to section 4 in this HLD https://github.com/sonic-net/SONiC/blob/master/doc/platform_api/CMIS_Diagnostic_Monitoring_Overview_in_SONiC.md. ?

The above HLD is for pluggable world but can you come up with similar for CPO? We should be able to isolate OE (module aspect vs datapath aspect0 vs ELSP (module aspect vs datapath aspect)

We need better CLI than the existing sfputil output for CPO. lets develop new debug tool for CPO system

PR #2324, opened 2026-05-07
high level design document for ptp

mgeng-nokia:
Add HLD for a PTP feature using linuxptp for its protocol stack.

The first use case in the HLD  is to run SONiC devices in a data center as PTP boundary clocks.

### DIFF
--- a/doc/ptp/ptp-hld.md
+++ b/doc/ptp/ptp-hld.md
@@ -1,0 +1,515 @@
+# PTP Feature
+
+### High Level Design document
+
+### Table of Contents
+
+- [PTP Feature](#ptp-feature)
+    - [High Level Design document](#high-level-design-document)
+    - [Table of Contents](#table-of-contents)
+    - [Revision](#revision)
+    - [About this manual](#about-this-manual)
+    - [Scope](#scope)
+    - [Abbreviations](#abbreviations)
+- [1 Introduction](#1-introduction)
+- [2 Feature Design](#2-feature-design)
+  - [2.1 Operational Flow](#21-operational-flow)
+  - [2.2 PTP Container](#22-ptp-container)
+  - [2.3 Configuration Templates](#23-configuration-templates)
+  - [2.4 Time Distribution Network Variants](#24-time-distribution-network-variants)
+    - [2.4.1 Ordinary Clock](#241-ordinary-clock)
+    - [2.4.2 Boundary Clock](#242-boundary-clock)
+    - [2.4.3 Transparent Clock](#243-transparent-clock)
+    - [2.4.4 IPv4/IPv6 Unicast transports](#244-ipv4ipv6-unicast-transports)
+    - [2.4.5 IPv4/IPv6 Multicast transports](#245-ipv4ipv6-multicast-transports)
+    - [2.4.6 L2 transport](#246-l2-transport)
+  - [2.5 Hardware Features Support](#25-hardware-features-support)
+    - [2.5.1 Hardware Timestamping Configuration](#251-hardware-timestamping-configuration)
+    - [2.5.2 Static Delay Asymmetry Configuration](#252-static-delay-asymmetry-configuration)
+    - [2.5.3 SyncE](#253-synce)
+    - [2.5.4 G.8275.1 Support](#254-g82751-support)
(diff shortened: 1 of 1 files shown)


securely1g:
## HLD Review

**Summary:** HLD for running SONiC devices as PTPv2 boundary/ordinary/transparent clocks using linuxptp (ptp4l). Phase 1 targets G.8275.2 profile with one-step HW timestamping on single-ASIC pizza boxes for 202611 release.

---

### Strengths

- Clean architecture — ptp4l is a well-proven open-source stack, smart to not reinvent the wheel
- Per-ASIC-namespace container model works well for multi-ASIC/chassis
- Leverages existing SAI PTP attributes (`SAI_SWITCH_ATTR_PORT_PTP_MODE`) — no SAI changes needed
- Phased roadmap is realistic (BC unicast first, multi-ASIC later)
- STATE_DB model follows RFC 8575 which is a good standard to align with

### Concerns

1. **Too many TBDs for a merged HLD** — COUNTERS_DB schema is "TBD", APPL_DB ptp configuration is "TBD", testing HLD is "TBD", multi-ASIC coupled-PHC design is "TBD". At minimum the APPL_DB and COUNTERS_DB schemas should be fleshed out since they are part of the core data flow.

2. **No YANG model** — The HLD mentions CLI commands but provides no YANG model for PTP configuration. Given SONiC's push toward YANG-validated config, this seems like an oversight. Will this be a separate PR?

3. **Restart-on-config-change is heavy-handed** — §2.2 says any config change regenerates ptp4l.cfg and relaunches the process. This means a `config ptp interface add` causes a full PTP re-convergence on all ports. For a BC with many downstream slaves, that's disruptive. Has on-the-fly reconfiguration via ptp4l's UDS/pmc interface been considered for non-disruptive changes (e.g., adding a port)?

4. **No CoPP details** — §5.2 mentions "a copp rule added to handle PTP packets" but doesn't specify the trap group, policer rate, or priority. PTP event messages are time-sensitive — if they share a queue with general protocol traffic, timestamping accuracy degrades. This deserves its own subsection with concrete values.

5. **Missing warmboot/fastboot section** — The SONiC HLD template requires a warmboot/fastboot impact section. When orchagent restarts, does the PTP mode get re-applied? Does ptp4l survive a swss restart? What about dataplane PTP timestamping continuity during warmboot?

6. **PHC-to-port association assumption** — §2.2 states "each ASIC to have one independently adjustable PHC" and "one PHC must be associated with all ports attached to the ASIC." This is a strong hardware assumption. What happens on platforms that don't meet this (e.g., external PHY with its own PHC per port-group)? Should there be a platform capability check or graceful fallback?

7. **No ACL / packet steering details** — How do PTP packets reach the ptp4l process in the container? For L2 transport or multicast, there needs to be a trap or ACL rule steering PTP ethertype (0x88F7) to CPU. For unicast UDP (ports 319/320), similar. The data path from ASIC → kernel → ptp4l container namespace isn't fully described.

8. **CONFIG_DB schema inconsistencies** — `PTP_UNICAST_MASTER_TABLE` uses `|` separator with IP embedded in the key, but `PTP_Interface` uses a different naming convention. Should be `PTP_INTERFACE` to match SONiC conventions. Also the `ptp-mode` field in `SWITCH|switch` uses hyphens — SONiC convention is typically underscores in CONFIG_DB field names.

9. **Security considerations missing** — PTP has known attack vectors (delay attacks, spoofing). No mention of authentication (IEEE 1588 Annex K / integrated security TLV) or any filtering of PTP messages from untrusted ports.

10. **No mention of interaction with existing NTP/chrony** — If both PTP and NTP are running, which disciplines the system clock? Is there a priority mechanism or does the operator need to disable NTP manually?

### Minor nits

- §6.1.4.1 heading number is reused for both "PTP Status" and "PTP Port Statistics"
- §2.4.6 says "PTP packets may transmit PTP packets" — wording is circular
- The `config ptp interface add` CLI shows `[--unicast-listen]` but doesn't explain what it does

---

**Verdict:** The architecture direction is sound (ptp4l in a container, per-namespace, leveraging existing SAI), but there are too many unresolved sections and missing standard HLD requirements (YANG, warmboot, CoPP details, security) to be merge-ready. Would suggest addressing at least items 1-5 before merge.

PR #2327, opened 2026-05-11
HLD for PIC support with Protection NHG

prashanth-nexthop:
This is an HLD that describes the orchagent design to support PIC with Protection Next Hop Group.  It works with the Protection NHG infra added by this PR https://github.com/sonic-net/sonic-swss/pull/4390#issuecomment-4228074250 and extends it to support N:M cardinality for Protection Next Hop Groups.

### DIFF
--- a/doc/pic/hld_pic_with_pnhg_orcagent.md
+++ b/doc/pic/hld_pic_with_pnhg_orcagent.md
@@ -40,6 +40,7 @@ Rev v1.0
 This document is a High-Level Design (HLD) for orchagent's support of *protection* next-hop groups (NHGs).  This describes the platform implementation of the PIC feature covered by the [https://github.com/sonic-net/SONiC/pull/2292](https://github.com/sonic-net/SONiC/pull/2292), where NHGs whose member set is partitioned into a **primary** subset and a **standby** subset, with hardware/SDK-assisted switching between the two.
 
 The design is platform-capability driven: orchagent uses the protection-NHG path only when SAI advertises support; otherwise the route falls back to ECMP over the primary subset.  
+
 ---
 
 ## 3\. Scope
--- a/doc/pic/hld_pic_with_pnhg_orcagent.md
+++ b/doc/pic/hld_pic_with_pnhg_orcagent.md
@@ -1,0 +1,690 @@
+# Protection NHG for Primary/Standby ECMP Routes
+
+Rev v1.0
+
+## Table of Contents
+
+1. [Revision History](#1-revision-history)  
+2. [About this Manual](#2-about-this-manual)  
+3. [Scope](#3-scope)  
+4. [Definitions and Abbreviations](#4-definitions-and-abbreviations)  
+5. [References](#5-references)  
+6. [5\. Overview](#5-overview)  
+7. [6\. Requirements](#6-requirements)  
+8. [7\. Architecture and Design](#7-architecture-and-design)  
+9. [8\. Behavior Scenarios](#8-behavior-scenarios)  
+10. [9\. SAI Contract](#9-sai-contract)  
+11. [10\. Data Model and DB Schema](#10-data-model-and-db-schema)  
+12. [11\. Switchover Trigger Sources](#11-switchover-trigger-sources)  
+13. [12\. Capability Detection and Eligibility](#12-capability-detection-and-eligibility)  
+14. [13\. Configuration and Manageability](#13-configuration-and-manageability)  
+15. [14\. Warm Boot and Fast Boot](#14-warm-boot-and-fast-boot)  
+16. [15\. Restrictions and Limitations](#15-restrictions-and-limitations)  
+17. [16\. Test Plan](#16-test-plan)  
+18. [17\. Open Questions and Future Work](#17-open-questions-and-future-work)
+
+---
+
+## 1\. Revision History
+
+## 
+
+| Rev | Date | Author(s) | Notes |
+| :---- | :---- | :---- | :---- |
+| v1.0 | 2026-05-06 | Prashanth (Nexthop AI) | Capability-driven N:M protection NHG; covers expected behavior in all paths. |
+
+---
+
+## 2\. About this document
+
+This document is a High-Level Design (HLD) for orchagent's support of *protection* next-hop groups (NHGs).  This describes the platform implementation of the PIC feature covered by the [https://github.com/sonic-net/SONiC/pull/2292](https://github.com/sonic-net/SONiC/pull/2292), where NHGs whose member set is partitioned into a **primary** subset and a **standby** subset, with hardware/SDK-assisted switching between the two.
+
+The design is platform-capability driven: orchagent uses the protection-NHG path only when SAI advertises support; otherwise the route falls back to ECMP over the primary subset.  
+---
+
+## 3\. Scope
+
+In scope:
+
+- The orchagent control plane: how routes carrying primary/standby intent are translated into protection NHGs, how member adds/removes are wired into liveness events, and how the active subset is selected.  
+- The producer contracts on `APPL_DB.ROUTE_TABLE` that triggers protection-NHG handling.  
+- The SAI contract that orchagent assumes (attributes, capability query, expected semantics).  
+- Behavior under primary loss, primary recovery, standby loss, and degenerate "all paths dead" cases.
+
+Out of scope:
+
+- The route producer (BGP/FRR or any other control-plane source). Producer-side logic is the responsibility of that component; this design only describes what producers must place in `APPL_DB`.  
+- SAI driver / SDK implementation. This document specifies the SAI surface that orchagent uses; how a particular SAI implementation realizes the contract on its hardware (for example, FEC pairs vs. ECMP-replace) is out of scope.  
+- Tunnel/overlay next hops as protection-NHG members. The current scope is plain underlay next hops only. Tunnel/overlay support is anticipated as a follow-up.
+
+---
+
+## 4\. Definitions and Abbreviations
+
+## 
+
+| Term | Definition |
+| :---- | :---- |
+| NHG | Next Hop Group |
+| ECMP | Equal Cost Multi-Path |
+| NH | Next Hop |
+| FEC | Forwarding Equivalence Class — hardware forwarding entry |
+| SAI | Switch Abstraction Interface |
+| OID | SAI Object Identifier |
+| APPL\_DB | Application database (Redis) populated by route producers (e.g., BGP/FRR) |
+| ASIC\_DB | ASIC database (Redis) populated by orchagent after SAI programming |
+| BFD | Bidirectional Forwarding Detection |
+| Primary | A next hop intended to carry traffic when reachable |
+| Standby | A next hop that takes over when the primary subset becomes unreachable |
+| Active subset | The role (primary or standby) currently selected for forwarding by the protection NHG |
+| Switchover | A transition of the active subset from primary to standby |
+| Switchback | A transition of the active subset from standby back to primary |
+
+## 5\. Overview
+
+A traditional ECMP NHG treats every member as an equal peer; loss of a member only redistributes flows across the survivors. While it provides high-availability, it does not provide a mechanism to create a hierarchy of these members where one set of paths is preferred and a backup set to take over on failure of the preferred set.
+
+A **protection NHG** introduces that framework. The route producer marks the first *N* of the route's *N+M* next hops as **primary** and the remaining *M* as **standby**. When all primaries are unreachable, traffic is steered to the standbys; when at least one primary becomes reachable again, traffic returns to the primaries. Since these NHGs are shared by multiple routes, on failure it provides a faster convergence to all the routes using this group.
+
+This design supports any **N:M** ratio, including 1:1, subject only to the platform's NHG capacity limit. The transition from primary to standby and back can be either HW driven or SW driven based on the capability of underlying HW/SDK.  
+
+For example, if the HW/SDK can support monitoring objects like interface state or BFD session state and use that to trigger switchover in HW itself.  However if the HW does not support some or all of the features required, then we can switchover by issuing `SAI_NEXT_HOP_GROUP_ATTR_SET_SWITCHOVER` to the affected group based on the nature of the failure.
+
+---
+
+## 6\. Requirements
+
+As mentioned earlier, this HLD is the platform complement of the feature described in the HLD [https://github.com/sonic-net/SONiC/pull/2292](https://github.com/sonic-net/SONiC/pull/2292)
+
+### 6.1 Functional
+
+- Support arbitrary **N:M** primary:standby ratios, including 1:1.  
+- Activate the protection-NHG path **only when SAI advertises support** for `SAI_NEXT_HOP_GROUP_TYPE_PROTECTION` or `SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION`. On platforms that advertise neither, fall back to plain ECMP over the primary subset; never spread BGP traffic across standbys when the producer asked for protection.  
+- When **all primary** next hops become unreachable, switch the active subset to **standby**.  
+- When **any primary** next hop becomes reachable again, switch the active subset **back to primary**.  
+- While at least one primary is reachable (i.e., not switched over), neighbor events on individual primaries shall **shrink or grow the primary subset programmed in SAI** without changing the switchover state.  
+- While not switched over, neighbor events on individual standbys shall keep the standby subset programmed in SAI in sync with their reachability state, so that a future switchover lands on a current set.  
+- While switched over (standbys active), neighbor events on individual standbys are **deferred** — orchagent does not perturb the standby set in SAI in the current design. This would cause traffic to blackhole.  But given that this is a second order failure and the fact that it leads to a degenerate case where there are no primary or secondary paths with similar results, this is not a concern.  Since this is solution for faster convergence, the expectation is that this is just a transistor state and that control plane would eventually come and update the routes with best paths available.  
+- Multiple routes whose primary and standby sets are identical shall **share** a single protection NHG (deduplication keyed on the member tuple).
+
+### 6.2 Configuration
+
+- The producer signals protection-NHG intent by adding a `primary_nh_count` field on the existing `APPL_DB.ROUTE_TABLE` entry. No new tables, no schema migration. The field is optional; absent or `0` or if `primary_nh_count` is the same as the size of nexthop list, means the route is plain ECMP.
+
+### 6.3 Scalability
+
+- There is no change to scale numbers due to this feature.  The number of distinct protection NHGs in the system is bounded by the SAI/SDK NHG capacity.  
+- The max member count per group is also similar to the ECMP behavior, except in case of Protection NHGs, the max member count is enforced on the total size including primary and standby members.    
+- Routes with identical primary+standby sets share one NHG (refcounted).
+
+### 6.4 Boot and Replay
+
+- Warm boot semantics are **inherited** from regular ECMP NHGs — the design does not introduce new persistence requirements. See §15.
+
+---
+
+## 7\. Architecture and Design
+
+### 7.1 Component Interaction
+
+```
+   +-----------------------------+        +-----------------------+
+   | Route Producer (BGP / FRR)  |        |     PortsOrch         |
+   +--------------+--------------+        +-----------+-----------+
+                  |                                   |
+                  | set primary_nh_count              | port oper status
+                  v                                   v
+   +-----------------------------+        +-----------------------+
+   |   APPL_DB.ROUTE_TABLE       |        |      NeighOrch        |
+   |   + primary_nh_count        |        |  (NH liveness oracle) |
+   +--------------+--------------+        +-----------+-----------+
+                  |                                   |
+                  | notify                            | setNextHopFlag IFDOWN
+                  v                                   | clearNextHopFlag IFDOWN
+   +-----------------------------+                    | addNextHop / removeNextHop
+   |          RouteOrch          |                    |
+   +--------------+--------------+                    |
+                  |                                   |
+                  | createProtNhg                     |
+                  | decProtNhgRefCount                |
+                  | removeProtNhg                     |
+                  v                                   v
+   +------------------------------------------------------+
+   |                       NhgOrch                        |
+   |  owns m_protNhgs    fan-out: validateNextHop /       |
+   |                     invalidateNextHop                |
+   +--------------+---------------------------------------+
+                  |
+                  | validateMember / invalidateMember /
+                  | updateSwitchoverState
+                  v
+   +------------------------------------------------------+
+   |    ProtNhg     (per-route key, refcounted)           |
+   +--------------+---------------------------------------+
+                  |
+                  | create_next_hop_group
+                  | create_next_hop_group_member
+                  | set_next_hop_group_attribute
+                  v
+   +------------------------------------------------------+
+   |                        SAI                           |
+   +------------------------------------------------------+
+```
+
+### 7.1 Lifecycle of a Protection NHG
+
+This design assumes that the PR that is currently being upstreamed provides the infrastructure necessary to create and manage the lifecycle of a Protection NHG.  The PR is available at [https://github.com/sonic-net/sonic-swss/pull/4390\#issuecomment-4228074250](https://github.com/sonic-net/sonic-swss/pull/4390#issuecomment-4228074250)
+
+However, additional changes may be necessary to support this use case of N:M cardinality and initiating switchover via the control plane.  
+   
+
+```
+                     producer sets primary_nh_count
+                     and route is eligible
+                              |
+                              v
+                  +-----------------------+
+                  |        Created        |
+                  +-----------+-----------+
+                              |
+                              | ProtNhg::sync, SAI NHG OID assigned
+                              v
+                  +-----------------------+
+                  |   Primary_Active      | <-----------------+
+                  |  (primary subset is   |                   |
+                  |   the active subset)  |                   | any primary
+                  +-----------+-----------+                   | recovers
+                              |                               | (SET_SWITCHOVER
+                              |  last live primary lost       |  = false)
+                              |  (SET_SWITCHOVER = true)      |
+                              v                               |
+                  +-----------------------+                   |
+                  |   Standby_Active      | ------------------+
+                  |  (standby subset is   |
+                  |   the active subset)  |
+                  +-----------+-----------+
+                              |
+              route deleted or replaced (either state) -> refcount-- ; if 0:
+                              v
+                  +-----------------------+
+                  |        Removed        |  --> SAI NHG / members destroyed
+                  +-----------------------+
+```
+
+The state lives in the `ProtNhg` instance. The `Primary_Active` ↔ `Standby_Active` transition is the *switchover*; everything else is membership and lifecycle.
+
+### 7.2 RouteOrch Eligibility Decision
+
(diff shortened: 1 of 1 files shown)


doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,690 @@
...
+### 10.1 APPL\_DB.ROUTE\_TABLE
+
+A single new optional field is added to the existing entry:
+
+```
+Key:    ROUTE_TABLE:<vrf>:<prefix>
+Fields:
+    nexthop          = "<ip1>,<ip2>,...,<ipN>"   (N+M total)
+    ifname           = "<a1>,<a2>,...,<aN>"      (N+M total)
+    weight           = "<w1>,...,<wN>"           (optional)
```
sushilkdubey, line 540: How "weight" as optional field is useful?

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,690 @@
...
+3. The route is a plain underlay route (not overlay, not SRv6, not blackhole) — these flavors are out of scope for Phase 1\.  
+4. `len(nexthop) <= NHGRP_MAX_SIZE`.
+
+If (2) and (3) hold but (1) or (4) fails, the route is programmed as ECMP **over the primary subset only** and a WARN is emitted that names the failing condition. Standbys are *never* spread into a plain-ECMP fallback, because doing so would defeat the producer's intent.
+
+---
+
+## 13\. Configuration and Manageability
+
+There are no CLI knobs introduced by this design. The feature is gated by SAI capability.
```
sushilkdubey: How about having a knob to enable/control this feature on supported ASIC?

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,690 @@
...
+### 7.2 RouteOrch Eligibility Decision
+
+When a `ROUTE_TABLE` entry is processed, RouteOrch classifies it based on `primary_nh_count`:
+
+| `primary_nh_count` | Outcome |
+| :---- | :---- |
+| equal to `len(nexthop)` | Existing path: ECMP, single-NH, blackhole, or interface route as appropriate. |
+| `> len(nexthop) or 0` | Invalid: warn, skip the row. |
+| `0 < primary_nh_cout < len(nexthop)` and platform supports it | **Protection-NHG path**: first `primary_nh_count` nexthops are primary, rest are standby. |
+| `0 < pc < len(nexthop)` and platform does **not** support it | Truncate to primaries only, program as ECMP. Logged at WARN. |
```
sushilkdubey: Missing definition of pc. Pls explain what is this?

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,690 @@
...
+   |                       NhgOrch                        |
+   |  owns m_protNhgs    fan-out: validateNextHop /       |
+   |                     invalidateNextHop                |
+   +--------------+---------------------------------------+
+                  |
+                  | validateMember / invalidateMember /
+                  | updateSwitchoverState
+                  v
+   +------------------------------------------------------+
+   |    ProtNhg     (per-route key, refcounted)           |
```
sushilkdubey, line 165: ProtNhg- new daemon?

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,690 @@
...
+---
+
+## 6\. Requirements
+
+As mentioned earlier, this HLD is the platform complement of the feature described in the HLD [https://github.com/sonic-net/SONiC/pull/2292](https://github.com/sonic-net/SONiC/pull/2292)
+
+### 6.1 Functional
+
+- Support arbitrary **N:M** primary:standby ratios, including 1:1.  
+- Activate the protection-NHG path **only when SAI advertises support** for `SAI_NEXT_HOP_GROUP_TYPE_PROTECTION` or `SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION`. On platforms that advertise neither, fall back to plain ECMP over the primary subset; never spread BGP traffic across standbys when the producer asked for protection.
```
sushilkdubey: A couple of clarifications would strengthen the HLD -
- what deployment scale and use cases (DC, carrier edge, or both) is this feature targeting? Is not this convergence benefit is most relevant at large routing scales.
- Another curious question- how does adding standby members impact ASIC NHG capacity, given that both primary and standby members count toward the NHG member limit?

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+## 6\. Requirements
+
+As mentioned earlier, this HLD is the platform complement of the feature described in the HLD [https://github.com/sonic-net/SONiC/pull/2292](https://github.com/sonic-net/SONiC/pull/2292)
+
+### 6.1 Functional
+
+- Support arbitrary **N:M** primary:standby ratios, including 1:1.  
+- Activate the protection-NHG path **only when SAI advertises support** for `SAI_NEXT_HOP_GROUP_TYPE_PROTECTION` or `SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION`. On platforms that advertise neither, fall back to plain ECMP over the primary subset; never spread BGP traffic across standbys when the producer asked for protection.  
+- When **all primary** next hops become unreachable, switch the active subset to **standby**.  
+- When **any primary** next hop becomes reachable again, switch the active subset **back to primary**.
```
abdosi, line 106: we need to add test-case to cover the back to back flaps of primary path (up->down-up->down) and constrain/stress it can bring on OA and Hw-resources

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+- When **any primary** next hop becomes reachable again, switch the active subset **back to primary**.  
+- While at least one primary is reachable (i.e., not switched over), neighbor events on individual primaries shall **shrink or grow the primary subset programmed in SAI** without changing the switchover state.  
+- While not switched over, neighbor events on individual standbys shall keep the standby subset programmed in SAI in sync with their reachability state, so that a future switchover lands on a current set.  
+- While switched over (standbys active), neighbor events on individual standbys are **deferred** — orchagent does not perturb the standby set in SAI in the current design. This would cause traffic to blackhole if the traffic was being hashed to the standby path that is down.  But given that this is a second order failure and the fact that it leads to a degenerate case where there are no primary or secondary paths with similar results, this is not a concern.  Since this is solution for faster convergence, the expectation is that this is just a transit state and that control plane would eventually come and update the routes with best paths available.
```
abdosi: what is the reason OA can be modify to handle this case ? We can treat this as ECMP Acceleration of Stand-by ECMP group ?

abdosi: especially when we have ECMP for stand-by we should avoid traffic black-hole

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+The field is optional; no schema migration is needed. Producers that do not understand the field continue to publish plain-ECMP rows and orchagent processes them unchanged.
+
+Here is an example of a route with 2 primary and 2 standby members
+
+```
+$ redis-cli -n 0 hgetall 'ROUTE_TABLE:100.0.0.0/24'
+   1) "nexthop"
+   2) "10.0.0.1,10.0.0.2,10.0.0.3,10.0.0.5"
+   3) "ifname"
+   4) "PortChannel102,PortChannel103,Ethernet68,PortChannel105"
```
abdosi, line 560: please add note here the expectation with this schema is the primary path are always sorted/ordered such that they are in front of the list.  Producer can be FRR/any other Routing Stack/Controller.

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+   NeighOrch    NhgOrch    ProtNhg    SAI
+       |           |          |        |
+       | invalidateNextHop(p2)  ** p2 is the last live primary **
+       +---------->|          |        |
+       |           | invalidate primary p2
+       |           +--------->|        |
+       |           |          | ** detect: would empty live primaries **
+       |           |          | set_next_hop_group_attribute(SET_SWITCHOVER=true)
+       |           |          +------->|
+       |           |          |  ** standbys are now the active subset **
```
abdosi: as discussed in meeting can you clarify what will be the behaviour when all primary are down and stay down permanently than how FRR/control plane will produce APP_DB

At that moment my understanding is that Route will be set as `primary_nh_count` as 0. Can you explain what will be behaviors when this happen. (Case where Route from FRR perspective only have stand-by paths)

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+       |           | invalidate standby s1 (not switched over)
+       |           +--------->|        |
+       |           |          | remove_next_hop_group_member(s1)
+       |           |          +------->|
+       |           |          | ** switchover state unchanged **
+```
+
+### 8.6 Standby flap while switched over (deferred)
+
+While the standbys are currently active, the design does **not** add or remove standbys from SAI on individual standby NeighOrch events. Rationale: doing so would shrink the active subset on the fly, potentially down to zero, while the active path is in use. The current design accepts that the standby subset programmed in SAI may diverge from NeighOrch's view while it is active, and re-syncs on the next switchover. See §16, §18.
```
abdosi: as noted above not sure what is complexity here. this should be same as inValidate/validate nexthop member in nexthop group

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+This document does not require any of the non-port sources to be wired up as part of the initial deliverable; the design simply does not preclude them.
+
+---
+
+## 12\. Capability Detection and Eligibility
+
+At orchagent boot, `NhgOrch` performs one capability query against SAI for the `SAI_NEXT_HOP_GROUP_ATTR_TYPE` enum and caches two booleans:
+
+- `sw_protection_supported` — true iff `SAI_NEXT_HOP_GROUP_TYPE_PROTECTION` is in the returned enum.  
+- `hw_protection_supported` — true iff `SAI_NEXT_HOP_GROUP_TYPE_HW_PROTECTION` is in the returned enum.
```
abdosi: Remove the reference for this  based on community feedback to have clarity that HLD support Protection Group with Switch over triggered by OA and not by HW itself.

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+
+```c
+sai_object_type_get_availability(switch_id, SAI_OBJECT_TYPE_NEXT_HOP_GROUP, ...);
+sai_query_attribute_enum_values_capability(switch_id, SAI_OBJECT_TYPE_NEXT_HOP_GROUP,
+                                           SAI_NEXT_HOP_GROUP_ATTR_TYPE, &values);
+```
+
+The orchagent inspects the returned enum list for:
+
+- `SAI_NEXT_HOP_GROUP_TYPE_PROTECTION` — software-controlled active subset (orchagent issues `SET_SWITCHOVER`).
```
abdosi: @manamand2020
it seems this HLD comment: https://github.com/sonic-net/SONiC/pull/2346/changes#diff-6f108e99b1118f2d3b6820f9144d3d9287220db7a9802d273f4ccf753a9331aeR87 does not looks correct. I don;t think we will flip `CONFIGURED_ROLE` in any case . Can you please double check. `SET_SWITCHOVER` should be trigger for flip.

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+| equal to `len(nexthop)` | Existing path: ECMP, single-NH, blackhole, or interface route as appropriate. |
+| `> len(nexthop)` | Invalid: warn, skip the row. |
+| `0 < primary_nh_cout < len(nexthop)` and platform supports it | **Protection-NHG path**: first `primary_nh_count` nexthops are primary, rest are standby. |
+| `0 < pc < len(nexthop)` and platform does **not** support it | Truncate to primaries only, program as ECMP. Logged at WARN. |
+
+Once classified into the protection-NHG path, RouteOrch builds a deterministic key (canonicalized over the primary and standby sets) and asks `NhgOrch` to materialize it. Distinct routes with the same key share the same SAI NHG OID.
+
+### 7.3 NhgOrch Ownership
+
+`NhgOrch` owns the protection-NHG objects in the same way it owns regular ECMP NHGs: keyed by a deterministic string, refcounted by the routes referencing them, created lazily, and destroyed when the refcount reaches zero. The protection-NHG object encapsulates:
```
abdosi: don't think this  statement is correct. For ROUTE_TABLE schema that contains flat list of nexthops ECMP NHG are maintained by routeOrch itself. Only in case of Route pointing to Nexthop Group Id it is owned by NhgOrch.

So with ProtoNhg we expect to modify routeOrch also accordingly ?

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+       | invalidateNextHop(s1)  ** switched over, s1 dies **
+       +---------->|          |
+       |           | invalidate standby s1 (switched over)
+       |           +--------->|
+       |           |          | ** no-op (deferred) **
+```
+
+### 8.7 Last primary down with no live standbys ("all paths dead")
+
+When the last live primary goes down and the standby subset programmed in SAI has zero members (because all standbys were unreachable for long enough that they were already removed from SAI per §9.5), the SAI `SET_SWITCHOVER=true` call is rejected by SAI/SDK because the destination active subset would be empty. The design responds gracefully:
```
abdosi: what is the error/return code we are expecting here from SAI ?  Does it need SAI spec changes to make sure all SAI vendors return same code accordingly ?  API can fail for multiple reason and it might be hard to say failure is because of this condition

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+
+- the SAI rejection is propagated up;  
+- `m_switched_over` is **not** flipped;  
+- the dead primary is **not** removed from SAI.
```
abdosi: what does it means dead primary will already be removed as that is condition for switchover ?

abdosi (changes requested):
need to close on open comments

abdosi:
@prashanth-nexthop Also based on community feedback we need to make current CRM resources define here: https://github.com/opencomputeproject/SAI/blob/master/inc/saiswitch.h#L1243 are good for this use case and do we need any SAI spec enhancement

doc/pic/hld_pic_with_pnhg_orcagent.md
manamand2020: Yes that was not correct, I have fixed that.

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+- The orchagent control plane: how routes carrying primary/standby intent are translated into protection NHGs, how member adds/removes are wired into liveness events, and how the active subset is selected.  
+- The producer contracts on `APPL_DB.ROUTE_TABLE` that triggers protection-NHG handling.  
+- The SAI contract that orchagent assumes (attributes, capability query, expected semantics).  
+- Behavior under primary loss, primary recovery, standby loss, and degenerate "all paths dead" cases.
+
+Out of scope:
+
+- The route producer (BGP/FRR or any other control-plane source). Producer-side logic is the responsibility of that component; this design only describes what producers must place in `APPL_DB`.  
+- SAI driver / SDK implementation. This document specifies the SAI surface that orchagent uses; how a particular SAI implementation realizes the contract on its hardware (for example, FEC pairs vs. ECMP-replace) is out of scope.  
+- Tunnel/overlay next hops as protection-NHG members. The current scope is plain underlay next hops only. Tunnel/overlay support is anticipated as a follow-up.
```
apannerselva: do we support NHGs managed by NhgOrch ?

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+       |           |          | create_next_hop_group_member(...)
+       |           |          | and/or set_next_hop_group_attribute(SET_SWITCHOVER=...)
+       |           |          +------->|
+```
+
+### 8.8 Route update that changes the NH set
+
+If a route is updated such that its primary or standby set changes, RouteOrch builds a **new** key and asks `NhgOrch` to materialize the new protection NHG. The route is repointed to the new NHG OID. The old NHG's refcount is decremented; if it hits zero, the old NHG is destroyed in SAI. Routes with the same NH set are not perturbed.
+
+If only `primary_nh_count` flips between zero and non-zero (i.e., the route transitions in or out of protection-NHG semantics), RouteOrch tears down the protection NHG (decRef) and reprograms the route as plain ECMP, or vice versa.
```
apannerselva: What happens when routes are already learned with a certain primary_nh_count, and later this value is changed, so that new routes are learned with a different primary next-hop count?
In this situation, how does the system behave if the same next-hop appears in both the primary path and the secondary path due to the change?

doc/pic/hld_pic_with_pnhg_orcagent.md
```diff
@@ -0,0 +1,691 @@
...
+- While not switched over, neighbor events on individual standbys shall keep the standby subset programmed in SAI in sync with their reachability state, so that a future switchover lands on a current set.  
+- While switched over (standbys active), neighbor events on individual standbys are **deferred** — orchagent does not perturb the standby set in SAI in the current design. This would cause traffic to blackhole if the traffic was being hashed to the standby path that is down.  But given that this is a second order failure and the fact that it leads to a degenerate case where there are no primary or secondary paths with similar results, this is not a concern.  Since this is solution for faster convergence, the expectation is that this is just a transit state and that control plane would eventually come and update the routes with best paths available.  
+- Multiple routes whose primary and standby sets are identical shall **share** a single protection NHG (deduplication keyed on the member tuple).
```
apannerselva, line 110: Suppose I initially learn some routes using a NHG, and later I enable primary_nh_count. How will the system behave after this change?

doc/pic/hld_pic_with_pnhg_orcagent.md
apannerselva: can this feature be supported with ARS/ALB ?

prashanth-nexthop: You are right — this was misleading. Traditional ECMP groups (flat nexthop list) are RouteOrch-owned; only routes pointing at an explicit next-hop-group index are NhgOrch-owned. The new protection NHGs are stored/owned by NhgOrch, but their lifecycle — create, refcount, destroy — is driven by RouteOrch. Fixed the document accordingly.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: Agreed.  In the absence of a specific SAI error, determining this condition will be ambiguous.  Will change the design such that Orchagent prechecks this.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: Clarified the ordering. Normal last-primary case (§8.3): OA issues `SET_SWITCHOVER=true` first and removes the dead primary only after it succeeds (switchover-then-remove, so the active subset is never momentarily empty). All-paths-dead case (§8.7): the pre-check finds the standby subset empty, so OA does not switch over and therefore does not remove the last dead primary — the operation is aborted and the group stays `Primary_Active` (black-holing) until any path recovers. So "dead primary not removed" applies only to the aborted all-dead case, not the successful §8.3 flow.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: Done. Scoped the HLD to the OA-driven model: switchover is triggered by orchagent via `SET_SWITCHOVER` based on NeighOrch liveness. Removed the `HW_PROTECTION` type, `hw_protection_supported`, and the `MONITORED_OBJECT`/`ADMIN_ROLE` (hardware-driven) references throughout (§5, §6.1, §9.1–9.4, §10.2, §12). Capability detection now keys solely on `SAI_NEXT_HOP_GROUP_TYPE_PROTECTION`. HW/SDK-driven switchover is listed as out of scope / future work (§17).

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: Agreed. While switched over, the standby subset is the active forwarding set, so leaving a dead standby in it needlessly black-holes flows that hash to it. Changing the design: standby member churn while switched over is now handled symmetrically to the primary subset — a dead standby is removed from the active set (ECMP shrinks) via the same invalidate/validate NH-member path, guarded so we never remove the last live member (that would empty the active subset, which folds into the §8.7 all-paths-dead handling). This removes the deferral and the associated restriction. Updated §6.1, §8.6, §15, §17.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: You're right — no added complexity. It's the same invalidate/validate NH-member path already used for the primary subset. Since I have now added a non-empty-active-subset guard, this would be simpler.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: The feature is already gated by the producer by setting appropriate value for primary_nh_count.  The producers like BGP will have necessary config knobs to control this behavior.  Orchagent does this based on capability.  So, a switch level knob seems unnecessary.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: `pc` was shorthand for `primary_nh_count`. Updated the document.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: No — `ProtNhg` is a C++ class/object inside the `orchagent` process, owned by `NhgOrch`, not a new daemon.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: Added a producer-contract note in §10.1: the producer (FRR or any other routing stack / controller) MUST place the primary next hops at the front of the `nexthop`/`ifname` lists; role is positional (the first `primary_nh_count` entries are primaries).

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: Added a §16.2 test case: back-to-back primary flaps (rapid up->down->up->down) verifying bounded OA/SAI resource churn

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: `weight` is the existing ECMP/WCMP weight field, not introduced by this design, so it stays optional exactly as it already is in `ROUTE_TABLE`. When present, weights apply to members within the currently active subset.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: Not currently — a protection NHG is a distinct SAI NHG type and can't be combined with ARS/ALB on the same group in this design. Added it to §17 (Future Work) as a possible follow-up that would need separate design.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: Not in the current scope. This design covers flat-list `ROUTE_TABLE` routes carrying `primary_nh_count` (see the ownership section). Routes pointing at an explicit NhgOrch next-hop-group index (`NEXT_HOP_GROUP_TABLE`) as protection NHGs are listed as out of scope and a possible follow-up.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: When all primaries go down, orchagent reacts to it and updates the data plane to use the standby set of nexthops.  Subsequently BGP/FRR updates the ROUTE_TABLE with new nexthops based on the path recalculation.  primary_nh_count will never be 0.  If the currently programmed standby is the only path available even after convergence, ROUTE_TABLE would be updated such that all the formerly standby paths are now primary paths and primary_nh_count reflects appropriate value. It is possible that it is a regular ECMP at this point with no standbys. Hope that clarifies.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: When primary_nh_count changes for a programmed route, orchagent recomputes the primary/standby partition and re-keys the route to the corresponding protection NHG — repointing it and releasing the old NHG (refcount--). As for the same NH being in the active and standby set: within a given route it is the producer's responsibility to keep the primary and standby sets mutually exclusive. However, a NH that is primary in one NHG can be standby in another, and vice versa.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: Yes — that's the plain-ECMP → protection transition (§8.8 case 1).  When a route's adjacency changes from plain-ECMP to protection-NHG orchagent builds a new protection-NHG with a new OID and the route-entry is reprogrammed, and the previously created ECMP is released after all the routes pointing to it stop using it.  Same is true when an adjacency transitions from protection NHG to plain ECMP.

doc/pic/hld_pic_with_pnhg_orcagent.md
prashanth-nexthop: For deployment and use cases, please refer to the companion BGP-PIC HLD — https://github.com/sonic-net/SONiC/pull/2292 — which this document complements as its platform/orchagent counterpart.

On scale and NHG capacity: orchagent bounds the total nexthop list (primary + standby) by `NHGRP_MAX_SIZE`, but how much hardware next-hop-group resource the standby subset actually consumes while inactive depends on the hardware architecture and vendor SAI/SDK implementation. So the effective scale is vendor-dependent.

prashanth-nexthop:
> @prashanth-nexthop Also based on community feedback we need to make current CRM resources define here: https://github.com/opencomputeproject/SAI/blob/master/inc/saiswitch.h#L1243 are good for this use case and do we need any SAI spec enhancement

Separate resource accounting for protection groups can be achieved without any SAI spec changes — SAI_NEXT_HOP_GROUP_ATTR_TYPE is already @isresourcetype. It would, however, need vendors to support type-based availability counts. And if a vendor doesn't actually maintain separate pools per type, tracking them separately from the orchagent side would be misleading

abdosi (approved):
LGTM

PR #1234, opened 2023-01-26
Isis configuration support

cchoate54:
Here is the documentation to describe the initial changes proposed to introduce FRR-ISIS configuration to SONiC.

PR #2328, opened 2026-05-12
[orchagent] HLD: Improve Resiliency of Orchagent Route Programming

venkit-nexthop:
## Summary
- Adds a new HLD document under `doc/orchagent/` describing a redesign of the orchagent route programming hot path.
- The redesign updates `m_toSync` directly from `mqPollThread`, splits `drain()` into three independently schedulable stages (`toBulkTask`, `flushTask`, `responseHandlingTask`) with yield/resume semantics and bounded time quanta, and makes the route state publish asynchronous.
- Documents the new `show orchagent tasks` / `sonic-clear orchagent tasks` instrumentation.
- Diagrams are checked in as external PNGs under `doc/orchagent/images/orchagent_route_redesign/` (no inline base64).

## Why
Tracked under [sonic-net/sonic-buildimage#27098](https://github.com/sonic-net/sonic-buildimage/issues/27098): the current serial `pops()` / `addToSync()` / `drain()` flow in the orchagent main thread lets ZMQ buffers and intermediate non-coalescing queues back up under churn and starves other orch agents of CPU.

## Test plan
- [ ] HLD reviewed against the [SONiC HLD template](https://github.com/sonic-net/SONiC/blob/master/doc/guidelines/hld_template.md).
- [ ] Diagrams render correctly on GitHub via the relative `images/orchagent_route_redesign/*.png` links.
- [ ] Doc-only change; no build/test impact.


venkit-nexthop:
@prabhataravind @qiluo-msft @prsunny @vivekrnv @deepak-singhal0408 for viz

securely1g:
## HLD Review

**TL;DR:** Solid HLD that addresses a real pain point — orchagent's route programming path starving other orch agents under churn. The design is well-structured and the incremental rollout approach is pragmatic. A few concerns below.

---

### Strengths

- Problem statement is crystal clear — serial processing + non-coalescing queues = starvation under load
- Incremental implementation plan (initial PRs hold lock for full `drain()` duration, same as today, then relax later)
- No SAI/syncd/DB schema changes — purely internal to orchagent
- Good observability story with the per-Executor stats (median/q1/q3/max/outliers)
- Warmboot/fastboot impact properly addressed (neutral to positive)
- Already has companion PRs in flight: swss-common#1187, swss#4564, swss#4437, swss#4563, utilities#4536

### Concerns / Questions

1. **Lock contention details are thin** — §7.2 says "locks protect concurrent access to `m_toSync`" but doesn't specify the lock granularity. With `mqPollThread` writing and `toBulkTask` reading/removing concurrently, what's the locking strategy? Per-bucket? A single mutex with short critical sections? This is the crux of the performance story and deserves more detail.

2. **Time quantum values not specified** — §7.5 says "fixed time quantum" but never states what the default is (or if it's configurable). Is it 10ms? 50ms? 500ms? The right value depends heavily on the platform and scale. Should be at least mentioned as a tunable with a sensible default.

3. **Self-notification for reschedule** — §7.4 says tasks reschedule by "posting a notification event to self." What's the priority/ordering of these self-notifications vs. other Selectables? If route tasks keep self-notifying, could they still starve lower-priority orch agents through round-robin effects?

4. **`unordered_map` iteration order** — §7.5 says `toBulkTask` walks "some bounded number of entries" from `m_toSync`. With `unordered_map`, iteration order is non-deterministic. Under sustained churn, could some prefixes get consistently deprioritized (starvation at the entry level)? Maybe a cursor/epoch approach would be safer.

5. **Failure retry semantics** — §7.5 says `responseHandlingTask` writes failed entries back to `m_toSync` "if the key does not exist there." But what if the failure is persistent (e.g., SAI resource exhaustion)? The entry will bounce between `m_toSync` → `toBulk` → flush → fail → `m_toSync` indefinitely. Is there a retry limit or backoff?

6. **Testing gaps** — No mention of testing the lock correctness under concurrent access (e.g., TSAN/thread sanitizer). Given this introduces real multi-threaded mutation of `m_toSync`, that seems important. Also no chaos/fault-injection test for the yield/resume checkpointing.

7. **Memory bound on `toBulk`** — If `toBulkTask` accumulates entries across multiple quanta before `flushTask` runs, `toBulk` could grow large. Is there a cap on `toBulk` size that triggers a flush, or is it purely time-driven?

### Minor nits

- §9.2 table formatting is slightly broken in the markdown (missing closing `|` on the header row) — cosmetic
- The open items (§14) about `fpmsyncd` back-pressure feel important enough to be tracked as a follow-up issue rather than just a bullet

---

Overall this is a well-thought-out HLD for a meaningful improvement. The main gap is more specifics on the concurrency model (lock strategy, quantum values, starvation prevention). Would approve with those addressed.

doc/orchagent/orchagent_route_redesign.md
```diff
@@ -0,0 +1,471 @@
...
+`orchagent` between the ZMQ consumer and `RouteOrch`.
+
+### 6.1. Existing design
+
+![Existing orchagent route programming flow](images/orchagent_route_redesign/existing_design.png)
+
+1. `zmq_recv()` reads from the ZMQ socket buffer into `m_buffer`.
+2. `BinarySerializer::deserializeBuffer()` is called.
+3. Data is written into `m_receivedOperationQueue`.
+4. Notify the main thread.
```
qiluo-msft, line 133: Could you explain the notify mechanism? what is the provider/consumer side behavior?
Ideally the consumer side should see a Selctable, so it can combine all the Selectable(s) in an arry and use single command to get notified by any event happening.

doc/orchagent/orchagent_route_redesign.md
```diff
@@ -0,0 +1,471 @@
...
+- [sonic-swss#4564](https://github.com/sonic-net/sonic-swss/pull/4564)
+
+- Eliminate the intermediate non-coalescing queues described above and update
+  the coalescing `m_toSync` directly from `mqPollThread`.
+- This ensures ZMQ socket buffers are drained on time.
+- More coalescing happens when `drain()` is slower.
+- The following operations are performed inside `mqPollThread`:
+  - `pops()`
+  - `addToSync()`
+- Locks protect concurrent access to `m_toSync`.
```
qiluo-msft: This lock on big and frequent changing object `m_toSync` will hurt the performance of route programming.

The HLD targeting issue may be easily resolved by changing data structure m_receivedOperationQueue from `queue<KeyOpFieldsValuesTuple>` to `SyncMap` (type of m_toSync). Could you explore along this direction?

deepak-singhal0408:
**On Qi's `SyncMap` direction — **

+1 to Qi's suggestion of a coalescing intermediate rather than locking `m_toSync`. To be clear, this isn't "lock-free" — it **relocates** the lock: instead of holding a lock on the hot `m_toSync` across `doTask`, the ingest side fills a small intermediate under a **brief O(1) handoff**, and the main thread **swaps** it out and programs the snapshot **unlocked**. Two design reasons:

**Result — two things the goal hinges on:**
- *Keep the socket drained* so the producer never hits HWM. A lock on `m_toSync` held during `doTask` makes ingest and programming contend, so the socket isn't drained while a burst is being programmed; a separate intermediate lets drain and programming run **concurrently**.
- *Bound the buffer.* Because the intermediate is a `SyncMap` (coalesced by key), a flapping prefix collapses to a **single** entry — so a sustained burst stays bounded by the number of **distinct keys** rather than growing **per message** toward OOM (as a `queue<KCO>` would). Coalescing is the memory-safety win, not just a throughput one.

**Architecture:** it keeps `m_toSync` single-threaded, preserving orchagent's existing "main loop isn't thread-safe" contract — a sharing property a shorter lock hold wouldn't restore.

**On "won't the later yield/resume split shorten the lock anyway?"** — it reduces the *duration*, but the intermediate is what makes that split **correct**: you can't hold a lock across a yield, so a resumed `drain()` needs a **stable snapshot** to iterate while new arrivals accumulate elsewhere — otherwise you're programming a map ingest is mutating mid-pass. The intermediate isn't an alternative to the split; it's the boundary the split depends on.

deepak-singhal0408:
@venkit-nexthop, could you check and address comments..
This design change is critical to Scaled systems stability under stress/churn. 

JFYI, I will be tracking this PR set actively and will try to help coordinate reviews/discussions/merges as needed.

doc/orchagent/orchagent_route_redesign.md
venkit-nexthop: That is how it is implemented. If you refer to https://github.com/sonic-net/sonic-swss-common/pull/1187

```
    /* Wakes the orch main loop. Called by ZmqServer::mqPollThread after a
     * burst (EAGAIN), and may be called by the ingress callback mid-burst when
     * m_toSync exceeds gMaxBulkSize. */
    void notifyPending() override
    {
      m_selectableEvent.notify();
    }
 ```

File: doc/orchagent/orchagent_route_redesign.md (hunk 2 above)
Comment (venkit-nexthop) [reply doc/orchagent/orchagent_route_redesign.md]: Yes, thanks for the suggestion. A similar feedback was given during the review as well.
I will update the HLD and the PR to reflect this.

**Comment by venkit-nexthop** [comment]
> @venkit-nexthop, could you check and address comments.. This design change is critical to Scaled systems stability under stress/churn.
> 
> JFYI, I will be tracking this PR set actively and will try to help coordinate reviews/discussions/merges as needed.

I am working on this.

File: doc/orchagent/orchagent_route_redesign.md [hunk 3]
```diff
@@ -0,0 +1,471 @@
...
+  allows concurrent access.
+- `toBulkTask` does not walk through the entire `m_toSync` collection like it
+  does today. It walks some bounded number of entries (until a fixed time
+  quantum is over).
+- While walking, it removes those items from `m_toSync` and updates the
+  `toBulk` data structure.
+- `flushTask` does not need access to `m_toSync` at all.
+- `responseHandlingTask` normally does not need access to `m_toSync` if there
+  are no failures. If there is a failure, it does the following:
+  - Write the entry back to `m_toSync`, if the key does not exist there.
```
anarasimhan-upscale, line 291: is there any risk of reordering? consider a case `m_toSync` had set followed by del for a prefix. If the set operation failed, will the set now be inserted in `m_toSync`after the del, for retry?

doc/orchagent/orchagent_route_redesign.md
```diff
@@ -0,0 +1,471 @@
...
+  state and calls the appropriate function. The checkpoint information helps
+  it start from where it left off.
+
+### 7.5. Task processing adheres to time quanta
+
+- To keep `m_toSync` independent and allow `mqPollThread` to keep updating
+  (and coalescing) it, `toBulkTask` consults `m_toSync` in a manner that
+  allows concurrent access.
+- `toBulkTask` does not walk through the entire `m_toSync` collection like it
+  does today. It walks some bounded number of entries (until a fixed time
```
anarasimhan-upscale, line 279: how is the `fixed time` configured? should be considering have the number of entries processed as upper bound?

doc/orchagent/orchagent_route_redesign.md
venkit-nexthop: The next few lines say the following

`If the key exists in  m_toSync , drop the failed entry because there is a more recent update.`

This ensures that there is no re-ordering. The failed entry is simply dropped in favor of the latest state update.

doc/orchagent/orchagent_route_redesign.md
venkit-nexthop: A time quantum is better than a fixed number of entries because it is scalable across all processor types. A faster processor may update the fixed number of entries quicker, while a slower processor may end up taking more time. Using a time quantum is better. This is only to yield and allow other orch agents a chance to run. It will come back and resume this once rescheduled.

deepak-singhal0408 (dismissed):
Thanks for addressing the feedbacks. Overall Design LGTM.

venkit-nexthop:
@prsunny do you know who can help merge this HLD?

prsunny:
@deepak-singhal0408 , can you review/signoff?

deepak-singhal0408 (commented):
A few references still describe the earlier threading model: the PR body/ToC/§7.11 show direct `mqPollThread` → `m_toSync`, while §7.2/§13.1 use `m_toSyncMutex` and place `addToSync()` under the lock. Could these be aligned with the implemented `m_ingressMutex` handoff, narrow the 50 ms statement to PR 1234’s drain-to-empty condition, add the **O(distinct in-flight route keys)** `m_ingress` bound in §11, and record the latest §7.3–§7.5 rewrite in the revision table?

venkit-nexthop:
@deepak-singhal0408 thanks — all addressed in rev 0.3 (commit a6b5a2f9), and I took the chance to reconcile the rest of §7 with the merged code:

- **Threading model** (ToC §7.2, §7.11, and the PR body): replaced the stale "`mqPollThread` → `m_toSync` directly" wording with the staging-map (`m_ingress`) → `execute()` merge model.
- **§7.2 / §13.1**: `m_toSyncMutex` → `m_ingressMutex` (guards `m_ingress` only); `execute()` moves `m_ingress` out under that lock and calls `addToSync()` **after** releasing it, so `m_toSync` is never accessed under the lock. §7.4 tightened to match.
- **50 ms**: narrowed to #1234's drain-to-empty condition — the flush is evaluated once per drain-to-empty pass (recv `EAGAIN`) with 50 ms as the dirty-age cutoff, i.e. "within ~50 ms once the socket empties," not a hard bound; a producer that keeps the socket continuously non-empty under `gMaxBulkSize` defers only for one drain pass.
- **§11**: added the `m_ingress` memory bound — O(distinct in-flight route keys), not churn volume.

Also aligned to the implementation while I was in there:
- `responseHandlerTask` (correct name), and the `RouteDrainBatch` `kIdle/kStaging/kFlushing/kResponding` phase machine with its per-phase resume points (scan-erase, `gRouteBulker.flush_until`, `respondCursor`).
- §7.5: the configurable `-Q` quantum (default 20 ms; `-Q 0` is debug/test only) and the warm-restore drain-to-completion exception (`setWarmStartDrainToCompletion`, NOS-16421), also cross-referenced in §10.
- §7.1: it's a templated `RouteSyncMap`/`OverwriteMerge` for the route consumer (others keep `SyncMap`/`multimap`).
- §7.6: async publish is `-A`-gated (`publishAsync`/`publishAsyncBatch`).
- §7.7: the CLI auto-scales duration units.

Revision table updated (rev 0.3). PTAL.

deepak-singhal0408 (approved):

venkit-nexthop:
@prsunny could you help merge this HLD

Issue #16, opened 2016-09-24
Shared test infrastructure


dplore:
- jenkins building and uploading docker  containers to dockerhub
- jenkins building and uploading to public apt repo setup
- linux kernel build and upload to public apt repo
- network switch hardware accessible to https://sonic-jenkins.westus.cloudapp.azure.com/
- jenkins projects to deploy to public switch hward
- jenkins projects created to execute tests

PR #2329, opened 2026-05-12
[bfd] Add SBFD hardware offload HLD document

yuezhoujk:
Add High Level Design document for Seamless BFD (SBFD) hardware offload as an extension to BFD-Syncd HLD. Covers SBFD Initiator, Echo, and Reflector session types with SRv6 encapsulation, SR-TE Policy integration, SAI-level BFD-coupled NHG fast switchover, and CONFIG_DB/APPL_DB/STATE_DB schema extensions.

### DIFF
--- a/doc/bfd/SBFD_HW_Offload_HLD.md
+++ b/doc/bfd/SBFD_HW_Offload_HLD.md
@@ -1,0 +1,1563 @@
+# Seamless BFD (SBFD) Hardware Offload — BFD-Syncd Extension
+## High Level Design Document
+### Rev 0.1
+
+# Table of Contents
+
+  * [Revision](#revision)
+  * [About this Manual](#about-this-manual)
+  * [Scope](#scope)
+  * [Definitions/Abbreviation](#definitionsabbreviation)
+  * [1 SBFD Overview](#1-sbfd-overview)
+    * [1.1 What is SBFD](#11-what-is-sbfd)
+    * [1.2 SBFD Session Types](#12-sbfd-session-types)
+    * [1.3 Role of SRv6 Encapsulation](#13-role-of-srv6-encapsulation)
+  * [2 Requirements](#2-requirements)
+    * [2.1 Functional Requirements](#21-functional-requirements)
+    * [2.2 Scalability Requirements](#22-scalability-requirements)
+  * [3 Architecture Design](#3-architecture-design)
+    * [3.1 Relationship with BFD-Syncd](#31-relationship-with-bfd-syncd)
+    * [3.2 Overall Architecture](#32-overall-architecture)
+    * [3.3 Configuration Sources](#33-configuration-sources)
+  * [4 Design Details](#4-design-details)
+    * [4.1 SBFD Initiator](#41-sbfd-initiator)
+    * [4.2 SBFD Echo](#42-sbfd-echo)
+    * [4.3 SBFD Reflector](#43-sbfd-reflector)
+  * [5 Component Changes](#5-component-changes)
+    * [5.1 bfdd SBFD Extension](#51-bfdd-sbfd-extension)
+    * [5.2 bfdsyncd SBFD Extension](#52-bfdsyncd-sbfd-extension)
+    * [5.3 BFDOrch SBFD Extension](#53-bfdorch-sbfd-extension)
+    * [5.4 Convergence Acceleration — SAI-level BFD-coupled NHG Fast Switchover](#54-convergence-acceleration--sai-level-bfd-coupled-nhg-fast-switchover)
+  * [6 SR-TE Policy Integration](#6-sr-te-policy-integration)
+    * [6.1 pathd SBFD Session Management](#61-pathd-sbfd-session-management)
+    * [6.2 Segment-List Liveness Detection](#62-segment-list-liveness-detection)
+    * [6.3 Policy State Propagation](#63-policy-state-propagation)
+  * [7 Database Schema](#7-database-schema)
+    * [7.1 CONFIG_DB](#71-config_db)
+    * [7.2 APPL_DB](#72-appl_db)
+    * [7.3 STATE_DB](#73-state_db)
+  * [8 SAI API](#8-sai-api)
+  * [9 Configuration and Management](#9-configuration-and-management)
+    * [9.1 FRR CLI Commands](#91-frr-cli-commands)
+    * [9.2 CONFIG_DB Configuration Examples](#92-config_db-configuration-examples)
+    * [9.3 Show Commands](#93-show-commands)
+    * [9.4 YANG Models](#94-yang-models)
+  * [10 Warm Restart](#10-warm-restart)
+  * [11 Limitations and Constraints](#11-limitations-and-constraints)
+  * [12 Test Requirements](#12-test-requirements)
+  * [13 References](#13-references)
+
+# Revision
+
+| Rev | Date       | Author       | Change Description                           |
+|:---:|:----------:|:------------:|----------------------------------------------|
+| 0.1 | 2026-05-12 | Kang Jiang   | Initial version                              |
+
+# About this Manual
+
+This document is an **SBFD extension document** for the [BFD-Syncd HLD](BFD_Syncd_HLD.md). Readers should first read the BFD-Syncd HLD to understand the overall architecture of classic BFD hardware offload (bfdd → bfdsyncd → APPL_DB → BFDOrch → SAI → ASIC data path). This document only describes the design additions for SBFD and does not repeat content already covered by classic BFD.
+
+# Scope
+
+This document covers:
+- Hardware offload design for three SBFD session types: Initiator, Echo, and Reflector
+- SRv6 encapsulation for path-level liveness detection
+- Two configuration paths: SR-TE Policy (pathd) auto-trigger and bfdd CLI manual configuration
+- SBFD extensions for bfdd, bfdsyncd, and BFDOrch components
+- SR-TE Policy integration (segment-list level liveness detection)
+- CONFIG_DB, APPL_DB, STATE_DB schema extensions
+- SBFD usage of SAI BFD API
+
+This document does not cover:
+- Classic BFD hardware offload (see [BFD-Syncd HLD](BFD_Syncd_HLD.md))
+- Non-SRv6 SBFD (e.g., MPLS SBFD)
+- Not-yet-implemented features: IPv4-only SBFD, ADMIN DOWN logic, Echo functionality within Initiator session
+- SRv6 data plane internal implementation
+- SR-TE Policy pathd internal design (candidate path selection algorithm)
+
+# Definitions/Abbreviation
+
+| Term            | Definition                                                                                         |
+|-----------------|----------------------------------------------------------------------------------------------------|
+| SBFD            | Seamless Bidirectional Forwarding Detection (RFC 7880)                                            |
+| BFD             | Bidirectional Forwarding Detection (RFC 5880)                                                     |
+| Initiator       | SBFD endpoint that sends BFD control packets to a Reflector to detect path liveness                |
+| Reflector       | SBFD endpoint that reflects received SBFD probe packets back to the Initiator, without maintaining per-session state |
+| Echo            | SBFD mode where the Initiator sends packets that traverse an SRv6 path and loop back to itself     |
+| bfdd            | BFD daemon running in the BGP container, extended to support SBFD                                  |
+| bfdsyncd        | Daemon bridging bfdd and BFDOrch via APPL_DB                                                      |
+| BFDOrch         | SONiC orchestration agent responsible for hardware BFD/SBFD sessions                               |
+| SRv6            | Segment Routing over IPv6 (RFC 8986)                                                              |
+| Sidlist         | SRv6 Segment Identifier List — an ordered list of IPv6 addresses defining an SRv6 path             |
+| SRH             | Segment Routing Header — IPv6 extension header carrying the segment list                          |
+| SR-TE           | Segment Routing Traffic Engineering                                                                |
+| pathd           | FRR daemon managing SR-TE Policies                                                                 |
+| Discriminator   | Unique identifier for a BFD/SBFD session endpoint (RFC 5880 §4.1)                                |
+| bfd-name        | Named identifier for an SBFD session, used as a key field in APPL_DB/STATE_DB keys and notification messages |
+| SAI             | Switch Abstraction Interface                                                                       |
+| ASIC            | Application Specific Integrated Circuit                                                            |
+| Distributed BFD | FRR protocol for offloading BFD to external data planes; bfdd and bfdsyncd communicate via Unix domain socket (see `bfdd/bfddp_packet.h`) |
+
+# 1 SBFD Overview
+
+## 1.1 What is SBFD
+
+SBFD (Seamless BFD, RFC 7880) is a simplified version of BFD that eliminates most of the negotiation process, providing rapid deployment and flexible path monitoring capabilities.
+
+**Core differences from classic BFD:**
+
+| Feature | Classic BFD (RFC 5880) | SBFD (RFC 7880) |
+|---------|------------------------|-----------------|
+| Session establishment | Three-way handshake (Down → Init → Up) | No handshake, Initiator sends probes directly |
+| Discriminator exchange | Negotiated during handshake | Initiator must know the Reflector's discriminator in advance |
+| Peer state | Both sides maintain session state | Only Initiator has state, Reflector is stateless |
+| Timer negotiation | Bidirectional negotiation taking maximum | No negotiation, Initiator unilaterally decides |
+
+**SBFD Initiator State Machine (RFC 7880):**
+
+```
+                     +--+
+        ADMIN DOWN,  |  |
+        TIMER        |  V
+                   +------+   UP                +------+
+                   |      |-------------------->|      |----+
+                   | DOWN |                     |  UP  |    | UP
+                   |      |<--------------------|      |<---+
+                   +------+   ADMIN DOWN,       +------+
+                              TIMER
+```
+
+- If the Initiator does not receive a response within the detection time, the session transitions to DOWN
+- If a response is received: if the response state is ADMINDOWN, the session transitions to DOWN; if the response state is UP, the session transitions to UP
+
+The Reflector is stateless: it only replies with a response packet when it receives a valid SBFD probe packet (Your Discriminator field matches).
+
+## 1.2 SBFD Session Types
+
+| Type | RFC | Direction | Purpose | SRv6 Encap | UDP Dst Port | remote_discr |
+|------|-----|-----------|---------|-------------|--------------|--------------|
+| Initiator | RFC 7880 | Initiator → Reflector | Detect reachability of remote Reflector via a specific SRv6 path | Yes | 7784 | Reflector's discriminator (pre-configured) |
+| Echo | RFC 7880 §4 | Self → Self (via SRv6 path) | Detect SRv6 path liveness via loopback | Yes | 3785 | = local_discr (self-addressed) |
+| Reflector | RFC 7880 §4 | Passive response | Reflect received SBFD probe packets back to Initiator | No | 3784 | 0 (don't care) |
+
+**Key difference between Initiator and Echo:**
+
+- **Initiator mode**: The probe packet reaches the remote Reflector, which actively responds. This tests **end-to-end reachability**, including the Reflector's ability to process and respond. The remote end must be configured with an SBFD Reflector.
+- **Echo mode**: The probe packet's destination address is the local device's own IP. After traversing the SRv6 path to the remote end, the remote end forwards it back via normal IP routing. The remote end requires no SBFD configuration — it only needs SRv6 End decapsulation and IP forwarding capability. This tests **path liveness only**.
+
+## 1.3 Role of SRv6 Encapsulation
+
+SBFD itself verifies node reachability. To extend to **path-level verification**, SBFD packets are encapsulated in SRv6 tunnels, causing packets to be forwarded hop-by-hop along the specified segment list. This enables detection of forwarding capability at each hop along a specific SRv6 path.
+
+The following example shows how an SBFD Initiator uses SRv6 encapsulation to detect the A→B→D path (all nodes support SRv6):
+
+```
+                    +------------C-----------+
+                   /                           \
+                 A---------------B---------------D
+                 ^               ^               ^
+                 |               |               |
+           End: 100::A       End: 100::B        End: 100::D
+      Loopback: 200::A                     Loopback: 200::D
+   BFD Discrim: 123                     BFD Discrim: 456
+```
+
+A is the SBFD Initiator, D is the SBFD Reflector.
+
+**Step 1: A sends SBFD probe packet to B**
+
+```
+IPv6(src="200::A", dst="100::B", nh=43)
+  /IPv6ExtHdrSegmentRouting(addresses=["100::D"], nh=41, segleft=1)
+  /IPv6(src="200::A", dst="200::D")
+  /UDP(dport=7784)
+  /BFD(my_dis=123, your_disc=456, state=UP)
+```
+
+**Step 2: B performs SRv6 End action**
+
+B's dst is 100::B (its own End address), performs SRv6 End action, switches dst to the next hop in the segment list, 100::D:
+
+```
+IPv6(src="200::A", dst="100::D", nh=41)
+  /IPv6(src="200::A", dst="200::D")
+  /UDP(dport=7784)
+  /BFD(my_dis=123, your_disc=456, state=UP)
+```
+
+**Step 3: D decapsulates and processes**
+
+D's dst is 100::D (its own End address), performs SRv6 End decapsulation, remaining inner packet:
+
+```
+IPv6(src="200::A", dst="200::D")
+  /UDP(dport=7784)
+  /BFD(my_dis=123, your_disc=456, state=UP)
+```
+
+This packet's destination address is D's loopback 200::D, which is delivered to the kernel protocol stack. The SBFD Reflector service processes it and replies:
+
+```
+IPv6(src="200::D", dst="200::A")
+  /UDP(sport=7784)
+  /BFD(my_dis=456, your_disc=123, state=UP)
+```
+
+The response packet is routed back to A via normal IP routing (possibly via D→B→A or D→C→A).
+
+**Echo mode packet format difference:**
+
+In Echo mode, the inner IPv6 destination address is **the local device itself** (200::A), the UDP port is 3785, and `your_disc` = `my_disc`:
+
+```
+IPv6(src="200::A", dst="100::B", nh=43)
+  /IPv6ExtHdrSegmentRouting(addresses=["100::D"], nh=41, segleft=1)
+  /IPv6(src="200::A", dst="200::A")        ← destination address is self
+  /UDP(dport=3785)                          ← BFD Echo port
+  /BFD(my_dis=123, your_disc=123, state=UP) ← self-addressed
+```
+
+After the packet reaches D via the A→B→D path, D decapsulates the inner packet, and the destination address 200::A is forwarded back to A via normal routing. The remote end D requires no SBFD configuration.
+
+# 2 Requirements
+
+## 2.1 Functional Requirements
+
+| ID   | Requirement                                                                                                   |
+|------|---------------------------------------------------------------------------------------------------------------|
+| FR-1 | Support SBFD Initiator sessions with SRv6 encapsulation for path-level reachability detection                  |
+| FR-2 | Support SBFD Echo sessions with SRv6 encapsulation for path-level loopback liveness detection                  |
+| FR-3 | Support SBFD Reflector sessions that respond to SBFD probes from remote Initiators                             |
+| FR-4 | Offload all three SBFD session types to ASIC hardware for fast failure detection                               |
+| FR-5 | Support SR-TE Policy (pathd) triggered SBFD sessions for automatic segment-list liveness detection             |
+| FR-6 | Support bfdd CLI direct configuration of SBFD sessions                                                         |
+| FR-7 | Support IPv4 and IPv6 inner-layer addresses for SBFD Initiator and Echo sessions                               |
+| FR-8 | Support SRv6 sidlist sharing (reference counting) among multiple SBFD sessions using the same segment list      |
+| FR-9 | Propagate SBFD state changes from hardware back to bfdd and SR-TE Policy for path selection decisions          |
+
+## 2.2 Scalability Requirements
+
+| Parameter                    | Value         | Notes                                       |
+|------------------------------|---------------|--------------------------------------------|
+| Maximum SBFD sessions        | Shared with BFD HW session pool (typically 4000) | SBFD sessions consume the same ASIC BFD resources as classic BFD |
+| Maximum SBFD Reflectors      | ASIC-dependent | Each Reflector consumes one SAI BFD session object |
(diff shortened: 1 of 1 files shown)


doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+### 3.3.1 Path A — SR-TE Policy (pathd) Triggered (color + endpoint)
+
+When an SR-TE Policy is configured with color and endpoint, SBFD is enabled directly under the policy:
+
+```
+! SBFD Echo mode
+segment-routing
+ traffic-eng
+  policy color 100 endpoint 200::D
+   sbfd echo 3 source-address 200::A
```
rajshekhar-nexthop, line 323: The config model shown enables SBFD only at the policy level. In production, operators commonly want one of three scopes: **(a) global** — enable SBFD for all SR-TE policies by default; **(b) per-policy** — enable for one policy (shown here); **(c) per-candidate-path** — enable for a specific candidate-path while leaving sibling CPs unmonitored. The current schema only supports (b). Could you extend the model to express all three, with a precedence rule (CP overrides policy overrides global) so operators don't have to repeat the same `sbfd` block under every policy?

rajshekhar-nexthop, line 323: Both example forms inline the detect-multiplier (`3`) and `source-address` directly in the `sbfd` command. SONiC already has a first-class `BFD_PROFILE` table (`sonic-bfd-profile.yang`, landing via PR sonic-net/sonic-buildimage#27229) carrying `detect_multiplier`, `receive_interval`, `transmit_interval`, `echo_interval`, `echo_mode`, `passive_mode`, `minimum_ttl`. Recommend the SBFD CLI accept a `profile <name>` reference instead of (or in addition to) inline parameters, so SBFD timers can reuse the same profile abstraction as classic BFD and so the global/per-policy/per-CP scopes from comment #1 can share a profile by name.

Note: to serve SBFD specifically, `BFD_PROFILE` needs two additional optional leaves — `mode` (`bfd` / `sbfd_init` / `sbfd_echo`, default `bfd`) and `local_address` — so that mode and a shared inner-source address can be templated. See comment #16 for the full CONFIG_DB schema, including the `profile` leafref on `BFD_PEER`.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+  !
+ !
+```
+
+Configuration chain:
+
+```
+CONFIG_DB SRTE_POLICY → frrcfgd → pathd (vtysh)
+  → pathd internal sr_config_sbfd_apply()
+  → ZAPI ZEBRA_SBFD_DEST_REGISTER
```
rajshekhar-nexthop, line 341: The design introduces a new ZAPI opcode `ZEBRA_SBFD_DEST_REGISTER` distinct from the existing `ZEBRA_BFD_DEST_REGISTER` (`bfdd/ptm_adapter.c:476`). Could you justify the new opcode rather than extending the existing `ZEBRA_BFD_DEST_REGISTER` payload with SBFD-specific fields (session type, segment list, encap source IPv6, bfd-name)? A second parallel opcode doubles the registration paths zebra/bfdd have to keep in sync (DEREGISTER, peer-down, peer-replay), and FRR upstream review tends to push back on duplicate opcodes when an extension is sufficient.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+    BFDOrch->>BFDOrch: Create SRv6 sidlist (if not exists)<br/>or reuse existing sidlist (ref count +1)
+    BFDOrch->>SAI: sai_create_bfd_session<br/>(TYPE=INITIATOR, ENCAP=SRV6, sidlist_oid)
+    SAI-->>BFDOrch: Session created successfully
+    BFDOrch->>BFDOrch: Update STATE_DB (bfd_hw_state=up)
+    BFDOrch-->>bfdsyncd: STATE_DB keyspace notification
+    bfdsyncd-->>bfdd: BFD_STATE_CHANGE (up)
+    bfdd->>bfdd: Hardware offload confirmed<br/>After BFD_XMTDEL_DELAY_TIMER delay<br/>Stop software TX
+```
+
+**Software-to-hardware switchover:** bfdd initially sends SBFD packets (with SRv6 encapsulation) via raw socket. Upon receiving `BFD_STATE_CHANGE(up)` from bfdsyncd, bfdd continues software TX for a `BFD_XMTDEL_DELAY_TIMER` delay period to ensure overlap, then stops software TX. During the overlap window, the remote end may receive duplicate packets, which is harmless per RFC 7880.
```
rajshekhar-nexthop, line 418: Is software to hardware switchover is vendor ASIC specific requirement for S-BFD sessions?

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
++---------------------------+
+| UDP (dst_port=7784)       |
++---------------------------+
+| BFD Control Packet        |
+|   my_discr = local_discr  |
+|   your_discr = remote_discr|
++---------------------------+
+```
+
+The SRv6 sidlist is created in BFDOrch via `sai_srv6_api->create_srv6_sidlist()` with type `SAI_SRV6_SIDLIST_TYPE_ENCAPS`, and associated with the BFD session via `SAI_BFD_SESSION_ATTR_SRV6_SIDLIST_ID`. Multiple SBFD sessions using the same segment list share a single SAI sidlist object (reference counting management, see §5.3.2).
```
rajshekhar-nexthop, line 447: BFDOrch directly calls `sai_srv6_api->create_srv6_sidlist()` / `remove_srv6_sidlist()` and maintains its own `m_bfdSidlist` reference-count table. `Srv6Orch::createUpdateSidList()` and `Srv6Orch::deleteSidList()` already exist in `sonic-buildimage/src/sonic-swss/orchagent/srv6orch.h:182-183` and own SRv6 sidlist lifecycle for the data path. Recommend BFDOrch delegate to SRv6Orch (or call a shared helper) rather than maintaining a parallel sidlist table. Otherwise a segment-list used by both an SR-TE policy data path and an SBFD session has two orchagents racing on the same SAI object's lifetime.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+};
+```
+
+SBFD Echo maps to `SAI_BFD_SESSION_TYPE_ASYNC_ACTIVE` with additional `SAI_BFD_SESSION_ATTR_ECHO_ENABLE = true`.
+
+When type is `SBFD_INITIATOR` or `SBFD_ECHO`, BFDOrch sets `entry.isSbfd = true`.
+
+### 5.3.2 SRv6 Sidlist Management
+
+BFDOrch creates SRv6 sidlist objects for SBFD Initiator and Echo sessions. Multiple sessions using the same segment list share a single SAI sidlist object.
```
rajshekhar-nexthop, line 718: Restating the same concern as comment #6 at the section that designs it: §5.3.2 makes BFDOrch the owner of sidlist create/delete and the reference-count table. The grounded recommendation is to factor `createBfdSidList()` / `deleteSidList()` out into Srv6Orch and have BFDOrch call into it by segment string, so SAI SRv6 object lifecycle stays in one orchagent.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+
+4. **Common attribute**: All SBFD sessions set `SAI_BFD_SESSION_ATTR_OFFLOAD_TYPE = SAI_BFD_SESSION_OFFLOAD_TYPE_SUSTENANCE`
+
+### 5.3.4 Session State Notification and DOWN Handling
+
+BFDOrch receives ASIC `bfd_session_state_change` events via SAI notification callback:
+
+1. Look up APPL_DB key from `m_bfdSAIIdToDB` mapping via SAI OID
+2. If state is **DOWN**:
+   - Call `removeBfdPeer()` to delete SAI session
```
rajshekhar-nexthop, line 762: §5.3.4 calls `removeBfdPeer()` (delete SAI session), `deleteSidList()` (decrement sidlist ref), and clears `m_bfdPeers`/`m_bfdSAIIdToDB` on every ASIC `bfd_session_state_change` event that reports DOWN. A transient flap (segment-list briefly unreachable, then recovers) would then require a full SAI session re-create plus possible sidlist re-create on every recovery. Is teardown on every DOWN required for correctness, or could BFDOrch leave the SAI session in place (state-only update) and only tear down on explicit user removal? The latter matches how classic BFD handles peer-down notifications and avoids SAI churn under flap.

A second concrete failure mode: **bfdd restart while a session is Down loses recovery entirely.** If bfdd crashes / warm-restarts during a Down state, the SAI session is already gone (BFDOrch deleted it), bfdd has no record to re-issue, and the ASIC retains no state. The operator must recreate the session. This is worse than classical BFD HW offload's behavior, where the SAI session survives both Down transitions and bfdd restarts.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+   - If `isSbfd == true`, call `deleteSidList()` to decrement sidlist reference count
+   - Clear records from `m_bfdPeers` and `m_bfdSAIIdToDB`
+3. Update `bfd_hw_state` in STATE_DB (updated for both UP and DOWN)
+
+The design of deleting the SAI session on DOWN allows bfdd to rebuild the hardware session by resending `DP_ADD_SESSION` when needed.
```
rajshekhar-nexthop, line 767: This sentence makes the split-authority explicit: BFDOrch deletes the SAI session autonomously on ASIC DOWN, and bfdd reconstructs it by resending `DP_ADD_SESSION`. Two issues with that contract: (1) bfdd's view of "session exists" temporarily diverges from BFDOrch's, leading to race windows where a delete+recreate from bfdd's side could collide with BFDOrch's auto-delete; (2) it's harder to reason about who owns the SAI object's lifetime. Recommend a cleaner contract: BFDOrch only reports state to STATE_DB on DOWN (does not delete), bfdd is the sole issuer of both `DP_DELETE_SESSION` and `DP_ADD_SESSION`. That keeps SAI lifecycle authority in one place.

This also **diverges from the classical BFD HW Offload HLD**, which keeps the SAI session in place through Down/Up transitions and only deletes on explicit operator removal. SBFD shouldn't introduce a different lifecycle pattern without explicit justification — operators, observability tools, and downstream consumers all rely on classical BFD's "session persists, only state field changes" model.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+
+    CP1 -->|"pathd auto-creates"| S1["SBFD Initiator/Echo<br>via SL1"]
+    CP2 -->|"pathd auto-creates"| S2["SBFD Initiator/Echo<br>via SL2"]
+
+    style POLICY fill:#e1f5fe,stroke:#01579b
+    style CP1 fill:#c8e6c9,stroke:#2e7d32
+    style CP2 fill:#fff9c4,stroke:#f9a825
+```
+
+pathd creates SBFD sessions via `sr_config_sbfd_apply()`, setting BFD timers, segment list, echo mode, and SRv6 encapsulation source address. The SBFD session lifecycle is fully managed by pathd (create/modify/delete SBFD sessions correspondingly when segment-lists are added/modified/deleted).
```
rajshekhar-nexthop, line 1039: §3.3.1 (line 341) makes the pathd→bfdd transport explicit (ZAPI `ZEBRA_SBFD_DEST_REGISTER` via zebra). §6.2.1 just says "pathd creates SBFD sessions via `sr_config_sbfd_apply()` … fully managed by pathd," which reads as if `sr_config_sbfd_apply()` is a direct internal API. Please restate here that create/modify/delete from pathd go over ZAPI (whichever opcode is settled on after comment #4), so the IPC isn't ambiguous in this section.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+
+| ID   | Requirement                                                                                                   |
+|------|---------------------------------------------------------------------------------------------------------------|
+| FR-1 | Support SBFD Initiator sessions with SRv6 encapsulation for path-level reachability detection                  |
+| FR-2 | Support SBFD Echo sessions with SRv6 encapsulation for path-level loopback liveness detection                  |
+| FR-3 | Support SBFD Reflector sessions that respond to SBFD probes from remote Initiators                             |
+| FR-4 | Offload all three SBFD session types to ASIC hardware for fast failure detection                               |
+| FR-5 | Support SR-TE Policy (pathd) triggered SBFD sessions for automatic segment-list liveness detection             |
+| FR-6 | Support bfdd CLI direct configuration of SBFD sessions                                                         |
+| FR-7 | Support IPv4 and IPv6 inner-layer addresses for SBFD Initiator and Echo sessions                               |
```
rajshekhar-nexthop, line 234: FR-7 (line 234): "Support IPv4 and IPv6 inner-layer addresses for SBFD Initiator and Echo sessions." §11 Limitations: "IPv4-only SBFD not yet implemented: Currently only IPv6 scenarios are supported. Pure IPv4 packets (without SRv6) SBFD is not yet implemented." These two statements are inconsistent: the requirements list claims IPv4 inner-layer support as in-scope, the limitations table says it isn't implemented. Please reconcile — either downgrade FR-7 to IPv6-only (matching the v1 implementation reality) or move it out of the v1 requirements list with a forward reference, so reviewers can tell what this PR is actually committing to deliver.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+
+**SBFD Initiator / Echo:** Set `sbfd` flag to 1, extract `segment` and `encap_src_ipv6` fields. Other fields (timers, discriminator, ports, etc.) are read directly from the `DP_ADD_SESSION` message.
+
+### 5.2.2 APPL_DB Entry Construction
+
+`BfdSync::onBfdMsg()` generates the APPL_DB key and writes to `BFD_SESSION_TABLE`.
+
+**Key format:**
+```
+BFD_SESSION_TABLE:{{dst_ip}}|{{local_discr}}|{{bfd_name}}|{{sequence_id}}
```
rajshekhar-nexthop, line 657: The APPL_DB `BFD_SESSION_TABLE` key includes `sequence_id` as a fourth tuple field (line 657, defined at line 663 as "monotonically increasing sequence number for deduplication and ordering"). In the full hardware-offload model this PR describes, `sequence_id` is an ASIC-vendor-specific deduplication workaround and not a property of the SONiC SBFD design — the design's authoritative session identity is `(dst_ip, local_discr, bfd_name)`, which is exactly what the STATE_DB key uses (line 1181) and what bfdsyncd parses on the reverse path (line 680). Recommend removing `sequence_id` from the documented APPL_DB key here so the HLD reflects the architecture, not a transient vendor mitigation. If a specific ASIC needs sequence-tagging, that belongs in vendor-specific notes, not in the canonical schema.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+    autonumber
+    participant CONFIG as CONFIG_DB / pathd
+    participant bfdd
+    participant bfdsyncd
+    participant APPL as APPL_DB
+    participant BFDOrch
+    participant SAI as SAI / ASIC
+
+    CONFIG->>bfdd: Configure SBFD Initiator<br/>(bfd-mode sbfd-init, remote-discr, srv6-encap-data)
+    bfdd->>bfdd: Create SBFD Initiator session<br/>Allocate local discriminator<br/>Start software TX
```
rajshekhar-nexthop, line 405: §5.4 fast-switchover depends on the `SAI_NEXT_HOP_ATTR_BFD_DISCRIMINATOR` value embedded by pathd in the route encap (§5.4.3) matching the `local_discr` of an SBFD session on this device. The doc currently implies dynamic allocation by bfdd (line 405: "Allocate local discriminator") but exposes no configuration surface — FRR CLI (§9.1), pathd CLI, and the proposed CONFIG_DB `BFD_SESSION_TABLE` (§7.1) all lack a `local-discr` field for Initiator/Echo, and no bfdd→pathd return path for the allocated value is described. Recommend adopting the **static-configuration model already in use for the Reflector** (line 569: "administratively allocated by the operator, globally unique on the device"): expose `local-discr <value>` on the sbfd-init/sbfd-echo FRR CLI and as a `local_discr` field in CONFIG_DB `BFD_PEER` (per comment #16, which recommends reusing `BFD_PEER` rather than introducing a new `BFD_SESSION_TABLE`), drop the "Allocate local discriminator" wording on line 405, and document that pathd's embedded discriminator must match the bfdd-configured `local_discr`. This avoids introducing a new ZAPI return path for allocated IDs and keeps the design uniform across all three SBFD session types.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+1. **SBFD DOWN** → `segment_list_down_handle()` → segment-list marked as DOWN → candidate path re-evaluation
+2. **SBFD UP** → `segment_list_up_handle()` → segment-list marked as UP → `sbfd_refresh_policy_state()` re-evaluates policy state
+3. If all candidate paths are DOWN → policy is marked DOWN
+4. pathd selects the best active candidate path and installs the corresponding route, achieving sub-second path switchover
+
+# 7 Database Schema
+
+## 7.1 CONFIG_DB
+
+### BFD_SESSION_TABLE Table (extended for SBFD)
```
rajshekhar-nexthop, line 1097: The CONFIG_DB schema in §7.1 introduces a new `BFD_SESSION_TABLE` keyed by `{name}` with `mode`, `remote_discr`, `segment_list`, `source_ipv6`, etc. Recommend dropping this table and instead reusing the existing `BFD_PEER` model already handled by `frrcfgd.bfd_handler` (`src/sonic-frr-mgmt-framework/frrcfgd/frrcfgd.py:2826`), extended along these lines:

**`BFD_PEER` (existing, extend):**
```
key = BFD_PEER|<bfd_name>
  enabled           : true|false
  mode              : bfd|sbfd_init|sbfd_echo      ; default "bfd"
  peer              : ipv4/v6
  local_address     : ipv4/v6                       ; per-session override of profile
  vrf               : <vrf_name>
  multihop          : true|false
  remote_discr      : uint32                        ; sbfd_init only
  local_discr       : uint32                        ; static, see comment #15
  srv6_encap_data   : <ipv6-csv>                    ; sbfd_init / sbfd_echo
  profile           : leafref -> BFD_PROFILE        ; optional
```

**`BFD_PROFILE` (existing, extend per PR sonic-net/sonic-buildimage#27229):** add two optional leaves so fields shared across a fleet of S-BFD sessions are declared once and inherited:
```
  mode              : bfd|sbfd_init|sbfd_echo       ; NEW, default "bfd"
  local_address     : ipv4/v6                        ; NEW, optional
```
Per-session values on `BFD_PEER` override profile-supplied values. Timer changes on a bound profile are applied in place by `BfdOrch::update_bfd_session` via `set_bfd_session_attribute`; `mode`/`local_address` changes require session re-create (CREATE_ONLY SAI attributes).

**`srv6_source_ipv6` is intentionally not added to either table.** The outer-header source IPv6 should be read from FRR's existing global `segment-routing srv6 encap-source-address`, which already serves every other SRv6 feature on the node. Per-session/per-profile duplication of a node-wide value invites drift.

**`SBFD_REFLECTOR` stays as proposed** (its own table keyed by discriminator; Reflector has no peer).

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+
+The Reflector is stateless: it only replies with a response packet when it receives a valid SBFD probe packet (Your Discriminator field matches).
+
+## 1.2 SBFD Session Types
+
+| Type | RFC | Direction | Purpose | SRv6 Encap | UDP Dst Port | remote_discr |
+|------|-----|-----------|---------|-------------|--------------|--------------|
+| Initiator | RFC 7880 | Initiator → Reflector | Detect reachability of remote Reflector via a specific SRv6 path | Yes | 7784 | Reflector's discriminator (pre-configured) |
+| Echo | RFC 7880 §4 | Self → Self (via SRv6 path) | Detect SRv6 path liveness via loopback | Yes | 3785 | = local_discr (self-addressed) |
+| Reflector | RFC 7880 §4 | Passive response | Reflect received SBFD probe packets back to Initiator | No | 3784 | 0 (don't care) |
```
rajshekhar-nexthop, line 141: The §1.2 session-type table lists the SBFD Reflector with `UDP Dst Port = 3784`, and §5.2.1 fills the same default in bfdsyncd. RFC 7881 §3 mandates UDP port **7784** for SBFD Control packets (3784 is the classical BFD port per RFC 5881). The HLD already uses 7784 correctly in the §1.3 packet walkthroughs (lines 173, 184, 195), so this is an internal contradiction, not a doctrine disagreement. Replace 3784 with 7784 in the §1.2 table and the §5.2.1 reflector defaults;

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+1. **SRv6 encapsulation**: When `entry.isSbfd && entry.segmentOid != SAI_NULL_OBJECT_ID`:
+   - `SAI_BFD_SESSION_ATTR_BFD_ENCAPSULATION_TYPE` = `SAI_BFD_ENCAPSULATION_TYPE_SRV6`
+   - `SAI_BFD_SESSION_ATTR_SRV6_SIDLIST_ID` = sidlist OID
+   - `SAI_BFD_SESSION_ATTR_TUNNEL_SRC_IP_ADDRESS` = `encap-src-ipv6`
+
+2. **Echo mode**: When `dst_port == 3785`, set `SAI_BFD_SESSION_ATTR_ECHO_ENABLE = true`
+
+3. **Reflector/Initiator validation exemption**: For `SAI_BFD_SESSION_TYPE_REFLECTOR` and `SAI_BFD_SESSION_TYPE_INITIATOR`, skip `validate()` check (do not require both src/dst IP pairs to be valid, since Reflector's dst_ip is `::`)
+
+4. **Common attribute**: All SBFD sessions set `SAI_BFD_SESSION_ATTR_OFFLOAD_TYPE = SAI_BFD_SESSION_OFFLOAD_TYPE_SUSTENANCE`
```
rajshekhar-nexthop, line 754: SBFD has no handshake (RFC 7880 §3), so there's no establishment to do in software. Can you share more details why we need sustenance mode for S-BFD? Is it vendor ASIC specific?

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+  → pathd internal sr_config_sbfd_apply()
+  → ZAPI ZEBRA_SBFD_DEST_REGISTER
+  → zebra → bfdd
+  → bfdd creates SBFD session
+  → bfdd Distributed BFD → bfdsyncd
+```
+
+pathd manages SBFD sessions at **segment-list granularity**. Each segment-list can be bound to one SBFD Initiator or Echo session. When a segment-list is added, modified, or deleted, pathd correspondingly creates, updates, or deletes the associated SBFD session.
+
+### 3.3.2 Path B — bfdd CLI Direct Configuration (color only)
```
rajshekhar-nexthop, line 349: §3.3.2 and §6.2.2 use the term "color only" SR Policy to mean "policy without an explicit endpoint." RFC 9256 §2.1 already standardizes this concept as a **null-endpoint** SR Policy (`0.0.0.0` for IPv4, `::` for IPv6); the §9.2 CONFIG_DB examples (lines 1347-1369) confirm this — the `SRV6_POLICY` key `1|::` encodes color + null endpoint. Recommend using "null-endpoint SR Policy (RFC 9256 §2.1)" consistently throughout instead of the local term "color only," and stating that `endpoint = ::` / `0.0.0.0` in CONFIG_DB / FRR CLI represents the null-endpoint case. The current upstream FRR pathd CLI (`pathd/path_cli.c:613`) makes `endpoint` mandatory; if this PR proposes making it optional with null as the implicit default, please call that out explicitly.

doc/bfd/SBFD_HW_Offload_HLD.md
rajshekhar-nexthop, line 657: §5.2.2 makes `bfd_name` part of the APPL_DB key but doesn't define its uniqueness contract: must it be globally unique on the DUT, unique within a `(dst_ip, local_discr)` tuple, or unspecified? pathd-allocated names (e.g., synthesized from `color|endpoint|preference|sl_name`) and operator-typed names (free strings) could collide. Need an explicit uniqueness invariant plus the enforcement point (CONFIG_DB YANG `must`, frrcfgd validation, or bfdsyncd-side reject).

**Note:** comment #16's `BFD_PEER` schema (keyed by `bfd_name`) makes this question even more load-bearing, since the table key *is* `bfd_name` — collisions become silent overwrites.

doc/bfd/SBFD_HW_Offload_HLD.md
```diff
@@ -0,0 +1,1563 @@
...
+- SBFD usage of SAI BFD API
+
+This document does not cover:
+- Classic BFD hardware offload (see [BFD-Syncd HLD](BFD_Syncd_HLD.md))
+- Non-SRv6 SBFD (e.g., MPLS SBFD)
+- Not-yet-implemented features: IPv4-only SBFD, ADMIN DOWN logic, Echo functionality within Initiator session
+- SRv6 data plane internal implementation
+- SR-TE Policy pathd internal design (candidate path selection algorithm)
+
+# Definitions/Abbreviation
```
rajshekhar-nexthop, line 78: Nit: Terms used in the body but absent from the definitions table: NHG, NHGM (next-hop group member), RIB, FIB, SAI OID, OCP.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: Thank you for pointing that out. We introduced ZEBRA_SBFD_DEST_REGISTER because the implementation of SBFD for SRv6 TE differs from classic BFD in several aspects, for example, it needs to carry the segment list. That said, it should also be possible to extend the existing ZEBRA_BFD_DEST_REGISTER to support this use case. I agree with your suggestion, and I will review the implementation accordingly.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: Yes, because on some ASICs, a BFD session transitioning from down to up does not trigger an up-state notification.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: For the cpath scenario configured with bfd-name, the segment lists managed by BfdOrch and Srv6Orch are independent at the upper layer, so there is no conflict here. However, for the scenario where SBFD is enabled under a policy, the race condition you pointed out does indeed exist.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: For the first point, our initial implementation actually worked as you described: the SAI session was not torn down when BFD went down. The issue is that on some ASICs, when the BFD session transitions from up to down, the SAI session reports the down event, but when the BFD session transitions from down to up, no up event is reported. As a result, if we keep the SAI session in place, the BFD session state in software can become inconsistent with the hardware state. Because of this limitation, we were forced to adopt the current approach. That said, I agree with your point that we may not need to accommodate the limitations of certain ASICs and could instead implement a more generic solution.

Our current design does not have the issue you described. If the session goes down, the hardware session is deleted. If bfdd restarts at that point, it will reload the BFD configuration after coming back up and re-establish the session. Once the BFD session comes back up, it will program the hardware again.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: Fundamentally, this still comes back to that limitation. I agree with your point that the SAI session should be kept on a Down event. That said, I would like to clarify one aspect of the current implementation: after BFDOrch autonomously deletes the SAI session when the ASIC reports DOWN, the session still exists from bfdd’s point of view. The difference is that, at that point, BFD protocol packet processing is handled by the bfdd process in software, i.e., packets are sent from software rather than offloaded to hardware. In essence, BFDOrch is not competing with bfdd; BFDOrch is managed by bfdd. Even after BFDOrch removes the hardware session due to the session going down, whether and when the session is reprogrammed into hardware is still controlled by bfdd.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: Sure, thank you for the careful review.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: Sorry, the description here is inaccurate. What I intended to say is that this HLD only supports SBFD over SRv6, and does not support classical SBFD without an outer SRv6 header. I’ll update it.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: Sounds good, I agree.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: Sorry, this was not described clearly enough in the HLD. The local discriminator is a randomly generated value created by FRR when the BFD session is instantiated. When SRv6 pathd binds the BFD session, it records that local discriminator and then programs it down to hardware through the pathd → zebra (ZAPI) → fpmsyncd → APP_DB → nhgorch/srv6orch → SAI pipeline. The purpose is to allow the hardware nexthop to associate with the corresponding BFD session, so that hardware-assisted BFD-triggered fast switchover can be achieved.

I will refine this part of the HLD to make the flow and ownership clearer.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: Not exactly. S-BFD still requires the Reflector end to send reply packets before the session can transition to Up. In that sense, this issue is caused by that limitation—specifically, the lack of notification for the down-to-up state transition.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: Okay, agreed. Thank you.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: For bfd_name, different BFD sessions must have different names. If you configure the same name twice, FRR will reject it. Going forward, we also plan to enforce this through CONFIG_DB YANG validation.

If the bfd_name values are different, then the local_discr values will also necessarily be different, because local_discr, like bfd_name, is also unique. For BFD sessions generated by pathd, the bfd_name will be automatically set to None, so in practice there will be no conflict.

doc/bfd/SBFD_HW_Offload_HLD.md
yuezhoujk: OK,We were not using BFD_PROFILE before, so we directly modified BFD_SESSION_TABLE. Going forward, I will build on your approach and extend the BFD_PROFILE table as well.

PR #2338, opened 2026-05-15
[HLD] MAC Move Guard

sudheer-nexthop:
HLD for the MAC Move Guard feature.
This feature detects mac-move notification storms and applies configurable action to mitigate the notification storm. The feature is part of Orchagent.

Implementation PRs:
macmoveguardorch:     https://github.com/sonic-net/sonic-swss/pull/4602
Yang model changes:    https://github.com/sonic-net/sonic-buildimage/pull/27530
sonic-mgmt tests:         https://github.com/sonic-net/sonic-mgmt/pull/24839

### DIFF
--- a/doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
+++ b/doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
@@ -1,0 +1,675 @@
+# SONiC MAC Move Guard High Level Design
+
+## Table of Contents
+
+- [1. Revision](#1-revision)
+- [2. Scope](#2-scope)
+- [3. Definitions/Abbreviations](#3-definitionsabbreviations)
+- [4. Overview](#4-overview)
+  - [4.1 Detection](#41-detection)
+  - [4.2 Mitigation actions](#42-mitigation-actions)
+  - [4.3 Mitigation duration](#43-mitigation-duration)
+- [5. Requirements](#5-requirements)
+- [6. Module Design](#6-module-design)
+  - [6.1 Overall design](#61-overall-design)
+  - [6.2 Configuration and control flow](#62-configuration-and-control-flow)
+    - [6.2.1 Detection: native SAI MAC move](#621-detection-native-sai-mac-move)
+    - [6.2.2 Detection: synthesized move from AGED + LEARNED](#622-detection-synthesized-move-from-aged--learned)
+    - [6.2.3 Mitigation: DISABLE_PORT](#623-mitigation-disable_port)
+    - [6.2.4 Mitigation: DISABLE_LEARN_ON_MAC_WITH_ACL](#624-mitigation-disable_learn_on_mac_with_acl)
+    - [6.2.5 Recovery](#625-recovery)
+    - [6.2.6 Orchagent restart handling](#626-orchagent-restart-handling)
+  - [6.3 Data structures](#63-data-structures)
+  - [6.4 SWSS and syncd changes](#64-swss-and-syncd-changes)
+- [7. Configuration and Management](#7-configuration-and-management)
+  - [7.1 CONFIG_DB](#71-config_db)
+  - [7.2 DB and Schema changes](#72-db-and-schema-changes)
+  - [7.3 YANG model](#73-yang-model)
+  - [7.4 CLI examples](#74-cli-examples)
+- [8. Warmboot](#8-warmboot)
+- [9. Memory Consumption](#9-memory-consumption)
+- [10. Restrictions/Limitations](#10-restrictionslimitations)
+- [11. Testing Requirements](#11-testing-requirements)
+  - [11.1 Unit tests (one-liners)](#111-unit-tests-one-liners)
+  - [11.2 System tests](#112-system-tests)
+- [12. Open/Action items](#12-openaction-items)
+
+## 1. Revision
+Rev | Date | Author | Change Description
+----|------|--------|-------------------
+|v0.1|2026-05-25|Sudheer Y R(Nexthop)|Initial version of MAC Move Guard HLD
+
+## 2. Scope
+This document describes the high level design of the MAC Move Guard feature in SONiC. MAC Move Guard detects abnormally high rates of MAC address moves between Layer-2 ports within a VLAN/bridge domain and applies a configurable mitigation action against the offending MAC, for a period of configured interval. The feature is implemented entirely inside `orchagent` as a new orchestrator (`MacMoveGuardOrch`) that subscribes to FDB events emitted by `FdbOrch` and drives `PortsOrch` / SAI FDB / SAI VLAN / SAI Bridge APIs to apply mitigation.
+
+## 3. Definitions/Abbreviations
+Definitions/Abbreviation|Description
+------------------------|-----------
+MAC Move | An FDB event where a previously-learned MAC address is re-learned on a different bridge port within the same VLAN
+Bad MAC | A (VLAN, MAC) whose number of moves within `detect_interval` has exceeded the configured `threshold`
+Pinned port | The one port to which a bad MAC stays anchored under the `DISABLE_PORT` action; all other ports the MAC was bouncing on are admin-disabled
+Detect interval | Sliding window (seconds) over which moves are counted
+Action interval | Time (seconds) the mitigation action stays in effect before recovery
+`bv_id` | SAI bridge-vlan OID identifying the L2 broadcast domain a MAC belongs to
+`MacKey` | Tuple of (`MacAddress`, `bv_id`) — uniquely identifies a tracked MAC
+FDB | Forwarding DataBase (the bridge MAC address table)
+SAI | Switch Abstraction Interface
+FSM | Finite State Machine
+
+## 4. Overview
+In a stable L2 network, a host's MAC is normally seen on a single bridge port. Pathological conditions — L2 forwarding loops, a misbehaving host, or a duplicated MAC across two endpoints — cause the same MAC to bounce between two or more ports at very high rates. The data plane "MAC move storm" that results is harmful in several ways: the CPU is saturated by FDB notifications, forwarding to the affected MAC becomes effectively random because the FDB entry is overwritten constantly, and it can mask the real source of the problem by spreading the impact across many ports.
+
+MAC Move Guard provides a contained, automatic mitigation that is split into: detection and mitigation for a user specified amount of time.
+
+### 4.1 Detection
+Per (VLAN, MAC), the orchestrator maintains a sliding window of move timestamps. On each move event it prunes entries older than `detect_interval` and re-derives the move count. When the count exceeds `threshold`, the MAC enters the **Bad MAC** state.
+
+Two FDB-event paths feed detection:
+1. **Native** SAI MAC move (`SAI_FDB_EVENT_MOVE`) is delivered as-is through `SUBJECT_TYPE_MAC_MOVE`.
+2. **Synthesized** move — for SDKs that only emit `AGED` then `LEARNED` instead of a native MOVE, the orch reconstructs the MOVE from a residual cache entry on the LEARN path.
+
+### 4.2 Mitigation actions
+Two mitigation actions are supported. They are mutually exclusive — only one is in effect at a time, configured by the `action` field in CONFIG_DB.
+
+Action | Effect on the bad MAC
+-------|----------------------
+`DISABLE_PORT` | Pin the MAC to one selected port; admin-disable every other port it was bouncing on
+`DISABLE_LEARN_ON_MAC_WITH_ACL` | Install a pre-ingress ACL entry matching `(vlan, smac)` with `SAI_ACL_ACTION_TYPE_SET_DO_NOT_LEARN` so the bad MAC stops being (re)learned while forwarding continues via the existing FDB entry
+
+`DISABLE_LEARN_ON_MAC_WITH_ACL` requires `SAI_ACL_ACTION_TYPE_SET_DO_NOT_LEARN` to be supported at the `PRE_INGRESS` stage. The orchestrator queries that capability at config time; on platforms that do not advertise it, the action is **soft-disabled** — the configuration is accepted, bad MACs are still detected and tracked, but no ACL table/entries are programmed. The action becomes active automatically once running on a platform that supports it.
+
+### 4.3 Mitigation duration
+A `SelectableTimer` fires every 30 s. For every bad MAC whose `action_expiry_time` has passed, the orchestrator reverses the action that was applied (re-enable the port for `DISABLE_PORT`, remove the per-MAC ACL entry for `DISABLE_LEARN_ON_MAC_WITH_ACL`). Action state is reference-counted, so a shared target (e.g. a port disabled because two bad MACs were both bouncing on it) is reverted only when the last bad MAC releases it. Both the bad-MAC identity and its recovery context are mirrored to STATE_DB so the recovery can complete even if `orchagent` restarts in the middle of the action interval (see [§6.2.6](#626-orchagent-restart-handling)).
+
+## 5. Requirements
+- Detect MAC moves on a per (VLAN, MAC) basis using a sliding-window threshold
+- Support two mitigation actions: `DISABLE_PORT` and `DISABLE_LEARN_ON_MAC_WITH_ACL`
+- Mitigation stays for a configurable action interval
+- CONFIG_DB-driven configuration via a single `GLOBAL` row, validated by a new YANG model
+- Persist bad-MAC identities and recovery context in STATE_DB so an `orchagent` restart cleanly completes (or aborts) any in-flight mitigation
+- Discover `SAI_ACL_ACTION_TYPE_SET_DO_NOT_LEARN` capability at the `PRE_INGRESS` stage; soft-disable the ACL action on platforms that do not support it
+
+## 6. Module Design
+### 6.1 Overall design
+- Management framework writes the `MAC_MOVE_GUARD|GLOBAL` row to CONFIG_DB using the YANG model in [§7.3](#73-yang-model)
+- `MacMoveGuardOrch` (new) in orchagent subscribes to the `MAC_MOVE_GUARD` table and configures itself
+- `MacMoveGuardOrch` attaches as an `Observer` on `FdbOrch` and consumes `SUBJECT_TYPE_MAC_LEARN` / `SUBJECT_TYPE_MAC_MOVE`
+- Mitigation is applied by calling `PortsOrch::setPortAdminStatusByAlias()` (for `DISABLE_PORT`) and `sai_acl_api` directly (for `DISABLE_LEARN_ON_MAC_WITH_ACL`, which uses a single pre-ingress ACL table bound to `SAI_SWITCH_ATTR_PRE_INGRESS_ACL`)
+- Bad-MAC identities and their recovery context are mirrored to STATE_DB (`MAC_MOVE_GUARD`) so the orchestrator can reconstruct in-flight mitigation across a restart (see [§6.2.6](#626-orchagent-restart-handling))
+- A 30 s `SelectableTimer` is used to apply the mitigation action for the configured duration, and to manage the MAC state
+- syncd / SAI: no changes
+
+The picture below groups the relationships by *role*. The numbered arrows describe the lifetime of one bad-MAC episode.
+
+```mermaid
+%%{init: {'theme':'base','themeVariables':{'primaryColor':'#EAF1F8','primaryBorderColor':'#5B7BAB','primaryTextColor':'#2C3E50','lineColor':'#7A8B9F','secondaryColor':'#F4F0E8','tertiaryColor':'#F9FAFC','fontSize':'12px'},'flowchart':{'useMaxWidth':false,'nodeSpacing':25,'rankSpacing':35}}}%%
+flowchart TB
+    HWIN:::invis
+    subgraph DETECT["DETECTION PLANE"]
+        direction TB
+        FDB["FdbOrch"]
+    end
+
+    subgraph MGMT["MANAGEMENT PLANE"]
+        direction TB
+        CLI["sonic config utilities"]
+        CFGDB["CONFIG_DB<br/>MAC_MOVE_GUARD|GLOBAL"]
+        CLI --> CFGDB
+    end
+
+    subgraph CTRL["CONTROL PLANE"]
+        direction TB
+        MMG["<b>MacMoveGuardOrch</b><br/>observer + 30s timer"]
+    end
+
+    subgraph ACTION["ACTION PLANE"]
+        direction TB
+        PO["PortsOrch"]
+        SAIACL["sai_acl_api"]
+        SAISW["sai_switch_api<br/>(PRE_INGRESS bind)"]
+    end
+    STATE["STATE_DB<br/>MAC_MOVE_GUARD"]
+    HWOUT:::invis
+
+    HWIN -->|"raw FDB events from hardware<br/>(SAI notif via syncd)"| FDB
+    CFGDB -->|"(1) config"| MMG
+    FDB -->|"(2) notify"| MMG
+    MMG -->|"(3) action"| PO
+    MMG -->|"(3) action"| SAIACL
+    MMG -->|"(3) action"| SAISW
+    MMG <-->|"(3a) persist /<br/>restore on restart"| STATE
+    ACTION -.->|"(4) SAI calls to hardware<br/>(via syncd)"| HWOUT
+
+    classDef invis fill:transparent,stroke:transparent,color:transparent
+```
+
+`MacMoveGuardOrch` holds:
+- maps: `m_macTrackingState`, `m_disabledPorts`, `m_learntMac`
+- ACL state: `m_learnDisableAclTable`, `m_learnDisableAclEntryCount`, `m_aclSetDoNotLearnSupported`
+- STATE_DB handle: `m_stateBadMacTable` (`MAC_MOVE_GUARD` table)
+- timer: `m_recoveryTimer` (30 s)
+- handlers: `handleMacLearn()`, `handleMacMove()`, `checkRecovery()`, `restoreBadMacState()`
+
+**Numbered flows (one bad-MAC episode):**
+1) Administrator writes `MAC_MOVE_GUARD|GLOBAL` to CONFIG_DB; `MacMoveGuardOrch::doTask(Consumer&)` consumes it and (when `action=DISABLE_LEARN_ON_MAC_WITH_ACL`) reconciles the pre-ingress ACL table lifecycle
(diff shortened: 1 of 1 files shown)


pbrisset:
Before accepting / merging this feature, remote MAC handling and EVPN MH synch. MAC handling must be added.
Also, how does the new logic tell EVPN MH about MAC freeze in HW. How FRRouting will react to that?

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,674 @@
...
+`DISABLE_LEARN_ON_MAC_WITH_ACL` | Install a pre-ingress ACL entry matching `(vlan, smac)` with `SAI_ACL_ACTION_TYPE_SET_DO_NOT_LEARN` so the bad MAC stops being (re)learned while forwarding continues via the existing FDB entry
+
+If the orchestrator cannot program `DISABLE_LEARN_ON_MAC_WITH_ACL`, it is flagged `action_not_supported`: configuration is still accepted and bad MACs are still tracked and persisted to STATE_DB, but no ACL table or entry is programmed.
+
+### 4.3 Mitigation duration
+A `SelectableTimer` fires every 30 s. For every bad MAC whose `action_expiry_time` has passed, the orchestrator reverses the action that was applied (re-enable the port for `DISABLE_PORT`, remove the per-MAC ACL entry for `DISABLE_LEARN_ON_MAC_WITH_ACL`). Action state is reference-counted, so a shared target (e.g. a port disabled because two bad MACs were both bouncing on it) is reverted only when the last bad MAC releases it. Both the bad-MAC identity and its recovery context are mirrored to STATE_DB so the recovery can complete even if `orchagent` restarts in the middle of the action interval (see [§6.2.6](#626-orchagent-restart-handling)).
```
venkatmahalingam: Applying mitigation every 30 seconds for a continuously moving MAC is costly. Can the mitigation interval be dynamically extended based on sustained MAC move behavior?

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,674 @@
...
+Two FDB-event paths feed detection:
+1. **Native** SAI MAC move (`SAI_FDB_EVENT_MOVE`) is delivered as-is through `SUBJECT_TYPE_MAC_MOVE`.
+2. **Synthesized** move — for SDKs that only emit `AGED` then `LEARNED` instead of a native MOVE, the orch reconstructs the MOVE from a residual cache entry on the LEARN path.
+
+### 4.2 Mitigation actions
+Two mitigation actions are supported. They are mutually exclusive — only one is in effect at a time, configured by the `action` field in CONFIG_DB.
+
+SAI Action | Effect on the bad MAC
+-----------|----------------------
+`DISABLE_PORT` | Pin the MAC to one selected port; admin-disable every other port it was bouncing on
```
venkatmahalingam, line 78: In scenarios where multiple MACs are moving within a VLAN, if a port has already been pinned for one MAC, should that port also be used for the other moving MACs?

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
sudheer-nexthop: @venkatmahalingam , we do not apply mitigation every 30s. We only evaluate if the mitigation is in effect long enough, and we should come out of the mitigation. For dynamic interval, the issue is mac moves stop once the mitigation is applied. The user can configure it for a longer interval to begin with

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
sudheer-nexthop: Yes. Thats right. The code tries to do that.

sudheer-nexthop:
> Before accepting / merging this feature, remote MAC handling and EVPN MH synch. MAC handling must be added. Also, how does the new logic tell EVPN MH about MAC freeze in HW. How FRRouting will react to that?

@pbrisset , we decided to support EVPN mac moves in phase-2. Phase-1 support is only for local-macs. I have updated the document with this plan.

prsunny:
@pbrisset , @venkatmahalingam , please check response. Would you sign off on the PR?

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+- `MacMoveGuard` (new) added in [`orchagent/macmoveguard.{h,cpp}`](orchagent/macmoveguard.h). It is owned by `FdbOrch` via composition (`std::unique_ptr<MacMoveGuard> m_macMoveGuard`); it does not inherit from `Orch` or `Observer`. It registers its `MAC_MOVE_GUARD` config-table `Consumer` and recovery `SelectableTimer` with the parent `FdbOrch`'s executor list (via `Orch::addExecutor`, reached through a `friend class MacMoveGuard;` declaration on `FdbOrch`)
+- `FdbOrch` constructor takes one new required parameter (`DBConnector* configDb`) so it can construct the guard; it is wired in [`orchdaemon.cpp`](orchagent/orchdaemon.cpp) where `gFdbOrch` is built. STATE_DB access reuses the existing `TableConnector` already passed to `FdbOrch`. `FdbOrch::doTask(Consumer&)` and `FdbOrch::doTask(SelectableTimer&)` route the guard's tasks back via `m_macMoveGuard->doConfigTask()` / `doRecoveryTimerTask()`. `FdbOrch`'s `SAI_FDB_EVENT_LEARNED` / `SAI_FDB_EVENT_MOVE` handlers call `m_macMoveGuard->onMacLearn(...)` / `onMacMove(...)` inline. The previous standalone `MacMoveGuardOrch` Orch and the `SUBJECT_TYPE_MAC_LEARN` / `SUBJECT_TYPE_MAC_MOVE` observer subjects are removed
```
judsonwilson-nvidia: I think this sentence is unintentional / LLM generated and references early commits that never made it to master branch. So I think it should be removed - no need to mention early drafts:

"The previous standalone `MacMoveGuardOrch` Orch and the `SUBJECT_TYPE_MAC_LEARN` / `SUBJECT_TYPE_MAC_MOVE` observer subjects are removed..."

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+    MMG->>SAI: remove_acl_table
+    Note over MMG: in-memory state empty<br/>no action_interval continuity
+```
+
+</div>
+
+Edge cases and invariants:
+
+- **Foreign pre-ingress ACL table**: if a different feature has bound a table to the pre-ingress slot, its signature will not match ours and the cleanup sweep leaves it alone. (The guard's normal-path `ensureLearnDisableAclTable()` also refuses to overwrite a non-null binding, so the two features cannot coexist on the pre-ingress slot regardless.)
+- **Stale rows from an older schema**: STATE_DB rows written by the previous (per-bad-MAC) schema have keys other than `HW_RESOURCES`; the cleanup sweep ignores them. They are not actively cleaned, but they are not consulted either.
```
judsonwilson-nvidia: Is this comment relevant to the community? It sounds like potentially unintended LLM generated text referring to an internal experimental version that didn't make it to master? I think it should be removed in that case?:

Introduced in [ffef8163](https://github.com/sonic-net/sonic-swss/commit/ffef81636d89165bd1eb0a345f66241bb053bcdb)

Removed/replaced in [f8871283](https://github.com/sonic-net/sonic-swss/commit/f887128386cecdcd2a5611090817587c6edfe89d)

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+- STATE_DB: new `MAC_MOVE_GUARD` table, at most one row (`HW_RESOURCES`) holding a `disabled_ports` CSV; absent when no ports are guard-disabled. Schema documented in [§7.2](#72-db-and-schema-changes)
+- syncd / SAI: no changes
+
+### 6.5 Interaction with Notification Dedup
+A separate SONiC effort ([sonic-net/SONiC#2334](https://github.com/sonic-net/SONiC/pull/2334)) introduces a Notification Dedup mechanism that coalesces repeated SAI notifications — including FDB events — at the syncd / swss layer and publishes aggregated counters for the dedup'd events. When that feature is deployed alongside MAC Move Guard, the two form a layered defence:
```
judsonwilson-nvidia: I think that characterizing them as a "layered defense" is too optimistic - suggesting that they are generally additive. I think coalescing is mostly unhelpful - when it kicks in, it prevents MacMoveGuard from working correctly.

I would call out your suggested mitigations and say that it's future work to split up the fdb event message handling in the Dedup feature so that MOVE/etc. can be handled differently - as I think dedup is currently an all-or-nothing feature for fdb events?  (Please verify my claims for yourself - this was my quick interpretation.)

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+
+**Tuning recommendation.** With Notification Dedup enabled, the per-MAC event rate that reaches `FdbOrch` (and therefore the guard) is a coalesced count, not the raw hardware rate. The operator should configure a **lower `threshold` over a larger `detect_interval`** in this case — for example `threshold ≈ 50` with `detect_interval ≈ 10s` instead of the defaults. The lower per-window count is appropriate because each surviving event already represents many hardware-level events. The longer window gives Notification Dedup a representative aggregate to act on. The net effect is that MMG trips earlier on a representative sample of the storm and applies mitigation at the hardware level *before* the storm has had time to broaden — which is precisely what minimizes both control-plane load and forwarding disruption. The exact values depend on the dedup window and counter semantics defined in the Notification Dedup HLD.
+
+## 7. Configuration and Management
+### 7.1 CONFIG_DB
+Configure MAC Move Guard by creating the single `MAC_MOVE_GUARD|GLOBAL` row:
```
judsonwilson-nvidia: Maybe continue the comment to say to "see the schema description in 7.2, below."

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+
+```
+; Defines schema for MAC Move Guard configuration attributes
+key                 = MAC_MOVE_GUARD:GLOBAL          ; only the GLOBAL key is permitted
+; field             = value
+ENABLED             = "true" / "false"               ; default false
+THRESHOLD           = 1*10DIGIT                      ; max moves in detect_interval (default 1000)
+DETECT_INTERVAL     = 1*4DIGIT                       ; seconds, 1..3600 (default 5)
+ACTION_INTERVAL     = 1*5DIGIT                       ; seconds, 120..86400 (default 600)
+ACTION              = action_value                   ; default DISABLE_PORT
```
judsonwilson-nvidia: Should this recommend DISABLE_LEARN_ON_MAC_WITH_ACL for switches that support it?

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+; STATE_DB MAC_MOVE_GUARD table — at most one row, key fixed to HW_RESOURCES
+key                 = MAC_MOVE_GUARD:HW_RESOURCES
+; field             = value
+disabled_ports      = port_alias *("," port_alias)        ; comma-separated; field/row absent when empty
+```
+
+> Note: Refer to swss-schema.md for general BNF conventions used across SONiC documents.
+
+### 7.3 YANG model
+The CONFIG_DB schema is backed by a YANG module added under `src/sonic-yang-models/yang-models/sonic-mac-move-guard.yang`.
```
judsonwilson-nvidia: The yang model section needs to be updated to match what was merged in the code. The differences are significant.

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-mac-move-guard.yang

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+                    default DISABLE_PORT;
+                }
+            }
+        }
+    }
+}
+```
+
+Notes:
+- `max-elements 1` enforces the "only `GLOBAL`" constraint, mirroring the orch's runtime check
```
judsonwilson-nvidia: Update this to match the new model  (remove the max-elements 1 comment?)

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+- Range constraints keep sonic config utilities from accepting nonsensical values such as zero or negative seconds
```
judsonwilson-nvidia: The comment about range constraints is stating the obvious, I think.

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+- Defaults are aligned with the orch's in-class defaults (`m_threshold = 1000`, `m_durationSeconds = 5`, `m_recoverySeconds = 600`, `m_action = DISABLE_PORT`)
+- Adding a new action means adding an `enum` value to `mac-move-guard-action` and a `case` in `MacMoveGuard::doConfigTask()`
+
+### 7.4 CLI examples
+No new CLI commands are introduced. Configuration is applied through sonic config utilities, which validate the input against the YANG model in [§7.3](#73-yang-model) before writing CONFIG_DB.
+
+Enable with the port-shut action:
```
judsonwilson-nvidia: "port-shut" is new terminology. Why not "Enable the feature, configured to use the DISABLE_PORT action:"

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+
+Persist across reboot:
+```bash
+config save -y
+```
+
+Operational visibility is provided through orchagent syslog at NOTICE level, including bad-MAC detection, port admin-down/up transitions, ACL table/entry creation and removal for `DISABLE_LEARN_ON_MAC_WITH_ACL`, `action_not_supported` reporting, restart cleanup actions (port re-enable, stale ACL table tear-down), and action-interval expiry.
+
+## 8. Warmboot
+- Warmboot is not supported
```
judsonwilson-nvidia: Nit: "Warmboot is not supported" seems a little bit heavy handed.  The user can warmboot, but this feature will reset in a way that shouldn't disrupt traffic for flows on ports / MACS that are well behaved.

Just a suggestion, which you can ignore if you like:

"MAC Move Guard state is not preserved across warmboot. During an orchagent restart, MAC Move Guard follows its cleanup-on-restart behavior: active mitigations are reverted—as applicable, ports administratively disabled by the guard are re-enabled, and guard-owned pre-ingress ACL entries and table are removed. Per-MAC detection history and action-expiry state are discarded, and detection restarts from an empty state using the retained CONFIG_DB configuration. A continuing MAC-move storm must cross the configured threshold again before mitigation is reapplied."

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+- `orchagent` restart is supported via a one-shot cleanup sweep: any leftover hardware state from the previous run (admin-disabled ports listed in STATE_DB, plus the pre-ingress ACL table identified by signature match) is reverted in the `MacMoveGuard` constructor. The `action_interval` countdown is not preserved across the restart; a still-flapping MAC will re-trip the threshold and the mitigation will be re-applied (see [§6.2.6](#626-orchagent-restart-handling))
+
+## 9. Memory Consumption
```
judsonwilson-nvidia: An explicit example memory estimate would be useful.

At startup do we expect to learn every MAC for all active devices on the vlans and hold temporary state for all of them?  Do you have any sense for what the typical scale might be?   I think a startup burst could be mitigated easily, but I have no idea if that would be worth the effort.

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+- Minimal control-plane state in orchagent
+- `m_macTrackingState` and `m_learntMac` are bounded in steady state by the SDK's MAC-table size. Quiet MACs are garbage-collected from `m_macTrackingState` by `checkRecovery()` after one detection-interval of silence, and `m_learntMac` entries are pruned by `pruneLearntMacCache()` on the same schedule
+- `m_disabledPorts` and the per-bad-MAC ACL entry count (`m_learnDisableAclEntryCount`) are bounded by the number of bad MACs currently in effect
+- STATE_DB footprint: at most a single row (`MAC_MOVE_GUARD|HW_RESOURCES`) with a `disabled_ports` CSV; zero rows when no ports are guard-disabled. DLOMWA leaves no STATE_DB footprint
+- No growth when feature disabled
+
+## 10. Restrictions/Limitations
```
judsonwilson-nvidia: I would mention that Notification Dedup feature can interfere with this feature.

judsonwilson-nvidia: I would also mention that this feature in `DISABLE_PORT` mode does not take into consideration the CONFIG_DB administrative state of ports - it assumes that it is UP. That should be true if the feature takes any action to set it to admin DOWN, but if the user then also sets it to DOWN in CONFIG_DB, there can be race conditions - the MAC Move Guard may then set it back to UP, without checking the current CONFIG_DB state.

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+13) YANG validation: sonic config utilities reject out-of-range `detect_interval`, invalid `action`, and non-`GLOBAL` keys before they reach CONFIG_DB
+14) Restart cleanup — `DISABLE_PORT`: pre-populate `MAC_MOVE_GUARD|HW_RESOURCES` with a `disabled_ports` CSV and mark those ports admin-down in the SAI mock; constructing the guard re-enables each port, clears the STATE_DB row, and leaves all in-memory bad-MAC tracking empty
+15) Restart cleanup — pre-ingress ACL table: pre-stage a bound pre-ingress ACL table with our signature (`stage=PRE_INGRESS`, `FIELD_SRC_MAC=true`, `FIELD_OUTER_VLAN_ID=true`, `action_list=[SET_DO_NOT_LEARN]`) and a few entries; constructing the guard enumerates the entry list, deletes each entry, unbinds the switch attr, and deletes the table
+16) Restart cleanup — foreign ACL table left alone: a pre-ingress table whose signature does not match (e.g. different action type) is **not** touched
+17) Restart cleanup — stale schema tolerance: a STATE_DB row left over from the older per-bad-MAC schema is ignored; the constructor does not throw and starts with empty in-memory state
```
judsonwilson-nvidia: Is this test for leftovers from the pre-merge-to-master design needed?

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+   - Wait until at least one ACL entry whose `SAI_ACL_ENTRY_ATTR_TABLE_ID` points at that table OID appears in ASIC_DB. Parse the entry's `SAI_ACL_ENTRY_ATTR_FIELD_SRC_MAC` to extract the tracked MAC, and verify `SAI_ACL_ENTRY_ATTR_FIELD_OUTER_VLAN_ID` matches the storm VLAN and the entry has the `SET_DO_NOT_LEARN` action.
+   - With the storm still running, disable rsyslog rate-limiting in the swss container, poll orchagent syslog for further "BAD MAC ... continues to move" lines for the tracked MAC, and require that no new matching line appears for a sustained quiescence window — proof that the ACL entry has taken effect.
+   - Stop the storm, reapply CONFIG_DB with a short action interval, and assert that the ASIC_DB ACL entry is removed within the new interval plus a grace period.
+   - Teardown: delete the CONFIG_DB row, restore rsyslog rate-limiting, remove the dedicated VLAN and its members, and flush the FDB.
+
+## 12. Open/Action items
+- `show mac-move-guard` CLI for operational visibility (bad MACs currently tracked, ports admin-disabled by the guard, installed ACL entries)
```
judsonwilson-nvidia: Also mention potential Notification Dedup improvements? (E.g. split the fdb event handling in Dedup to handle MAC MOVE events in a different way from other fdb event dedup?)

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+## 5. Requirements
+### Phase-1: Local MAC move support
+- Detect MAC moves on a per (VLAN, MAC) basis using a sliding-window threshold
+- Support two mitigation actions: `DISABLE_PORT` and `DISABLE_LEARN_ON_MAC_WITH_ACL`
+- Mitigation stays for a configurable action interval
+- CONFIG_DB-driven configuration via a single `GLOBAL` row, validated by a new YANG model
+- On `orchagent` restart, revert any hardware state left behind from a previous run (admin-disabled ports and the pre-ingress ACL table) so it does not leak; the in-flight `action_interval` is not preserved across the restart
+- Flag `DISABLE_LEARN_ON_MAC_WITH_ACL` as `action_not_supported` when it cannot be programmed; continue to track bad MACs but skip ACL programming
+
+### Phase-2: EVPN MAC move support
```
judsonwilson-nvidia: Mark this as "Future Work" here, since it's not currently in this HLD at all?

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+
+## 2. Scope
+This document describes the high level design of the MAC Move Guard feature in SONiC. MAC Move Guard detects abnormally high rates of MAC address moves between Layer-2 ports within a VLAN/bridge domain and applies a configurable mitigation action against the offending MAC, for a period of configured interval. The feature is implemented entirely inside `orchagent` as a new helper class `MacMoveGuard` that lives in [`macmoveguard.{h,cpp}`](orchagent/macmoveguard.h) and is owned by `FdbOrch` via composition. `FdbOrch` calls into the guard inline when it emits MAC learn/move updates; the guard drives `PortsOrch` / `sai_acl_api` / `sai_switch_api` to apply mitigation.
+
+## 3. Definitions/Abbreviations
+Definitions/Abbreviation|Description
+------------------------|-----------
+MAC Move | An FDB event where a previously-learned MAC address is re-learned on a different bridge port within the same VLAN
+Bad MAC | A (VLAN, MAC) whose number of moves within `detect_interval` has exceeded the configured `threshold`
+Pinned port | The one port to which a bad MAC stays anchored under the `DISABLE_PORT` action; all other ports the MAC was bouncing on are admin-disabled
```
judsonwilson-nvidia: Be more specific about "admin-disabled'. It's in PortsOrch, but not CONFIG_DB?

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+- `MacMoveGuard` is owned by `FdbOrch` via composition (`std::unique_ptr<MacMoveGuard> m_macMoveGuard`). It registers its `MAC_MOVE_GUARD` config-table `Consumer` and its recovery `SelectableTimer` with `FdbOrch`'s executor list; `FdbOrch::doTask(Consumer&)` and `FdbOrch::doTask(SelectableTimer&)` route those tasks back to the guard
+- `FdbOrch` invokes `m_macMoveGuard->onMacLearn(...)` and `onMacMove(...)` inline as part of emitting MAC-update notifications — there is no observer indirection
+- Mitigation is applied by calling `PortsOrch::setPortAdminStatusByAlias()` (for `DISABLE_PORT`) and `sai_acl_api` directly (for `DISABLE_LEARN_ON_MAC_WITH_ACL`, which uses a single pre-ingress ACL table bound to `SAI_SWITCH_ATTR_PRE_INGRESS_ACL`)
+- Only the `DISABLE_PORT` action leaves a STATE_DB footprint: a single row (`MAC_MOVE_GUARD|HW_RESOURCES`) holding a CSV of admin-disabled port aliases. DLOMWA writes nothing — its pre-ingress ACL table is rediscovered on restart by signature-matching the table bound to `SAI_SWITCH_ATTR_PRE_INGRESS_ACL`. The guard's constructor runs a one-shot cleanup sweep that reverts any such hardware state (see [§6.2.6](#626-orchagent-restart-handling))
```
judsonwilson-nvidia: Can you add a little more here explaining why?   I think the answer is that this is the minimum amount of database state needed to cleanup after a crash, and section 6.2.6 explains this, but the current wording in this paragraph was not helping me connect the dots.

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+    MMG -->|"(3) action"| PO
+    MMG -->|"(3) action"| SAIACL
+    MMG -->|"(3) action"| SAISW
+    MMG <-->|"(3a) DISABLE_PORT only:<br/>write disabled_ports CSV;<br/>cleanup sweep on startup"| STATE
+    ACTION -.->|"(4) SAI calls to hardware<br/>(via syncd)"| HWOUT
+
+    classDef invis fill:transparent,stroke:transparent,color:transparent
+```
+
+`MacMoveGuard` holds:
```
judsonwilson-nvidia: Nit:
```suggestion
`MacMoveGuard` holds the following (see [§6.3](...)):
```

doc/mac_move_guard/MAC_MOVE_GUARD_HLD.md
```diff
@@ -0,0 +1,683 @@
...
+};
+```
+
+Per-orch maps:
+
+Map | Type | Purpose
+----|------|--------
+`m_macTrackingState` | `std::map<MacKey, MacMoveTrackingState>`                            | Per-MAC sliding window + bad-MAC state
+`m_disabledPorts`    | `std::map<std::string, std::set<MacKey>>`                           | Reference count for `DISABLE_PORT`: port alias → bad MACs holding it down
+`m_learntMac`        | `std::unordered_map<MacKey, LearntMacEntry, MacKeyHash>`            | Last-known port per (VLAN, MAC); used to synthesize MOVE from AGED+LEARNED SDKs (see [§6.2.2](#622-detection-synthesized-move-from-aged--learned)). Pruned by `pruneLearntMacCache()` after one detection window of silence
```
judsonwilson-nvidia: I think we can stop populating this once m_nativeMovesSeen is true? It should naturally stop updating and clear out after that.

Issue #2035, opened 2025-07-08
Suppoer for RadSec and TACACS over TLS


AnantKishorSharma:
Currently there is lack of security for large parts of Radius and Tacacs packet payload. Specifically, RADIUS/TACACS security is based on the MD5 algorithm, which has been proven to be insecure.
 
This feature enhancement will add support for moving them to TLS(RadSec and TACACS over TLS) to secure the communication between RADIUS/TACACS peers on the

zhangyanzhao:
Any HLD PR or code PR for this feature?

AnantKishorSharma:
Opened the RadSec (RADIUS over TLS/DTLS) HLD as a draft PR for the RADIUS half of this issue: sonic-net/SONiC#2400. It complements the TACACS+ over TLS HLD (#2352). Feedback welcome.

PR #2352, opened 2026-05-25
[doc][aaa] Add TACACS+ TLS HLD

Rod-Persky:
This PR adds a high-level design for TACACS+ over TLS in SONiC using a KubeSONiC-managed local TACACS+ forwarding service.

The design keeps existing SONiC TACACS+ behavior valid while adding a TLS-capable upstream path for deployments that need stronger management-plane transport security, connection reuse, ordered failover, and clearer diagnostics. Existing PAM, shell, audit, accounting, and CLI TACACS+ consumers continue to use their current TACACS+ client path through a host-local loopback proxy; they do not need to implement TLS or consume TLS-specific configuration.

This HLD supports issue #2035.

## Design Shape

The HLD defines two local consumption paths:

- Existing SONiC code with native TACACS+ libraries uses a loopback TACACS+ TCP proxy.
- New SONiC code, or code written in languages without native TACACS+ libraries, may use a host-local UDS/gRPC/protobuf IPC surface when direct integration is needed.

The local forwarding service owns upstream TACACS+ transport, including TCP/TLS connection setup, TLS credential handling, single-connection reuse, ordered failover, preferred-server recovery, configuration hot reload, and serviceability.

## Configuration Model

The design keeps existing TACACS+ configuration separate from new TLS forwarding configuration:

- Existing `TACPLUS` and `TACPLUS_SERVER` rows remain the compatibility view used by existing SONiC TACACS+ consumers.
- New `TACPLUS_SERVER_TLS` rows describe upstream TACACS+ over TLS servers.
- New `TACPLUS_FORWARDER` rows describe forwarder-local settings such as the loopback listener address and port.

The forwarding service filters its own loopback entry out of `TACPLUS_SERVER`, combines the remaining TCP candidates with `TACPLUS_SERVER_TLS` candidates, and selects upstream servers by configured priority. CLI commands manage TCP and TLS rows independently; a TCP server row is not transformed into a TLS server row.

## TLS and Credential Coverage

The HLD covers TACACS+ over TLS 1.3 following RFC 9887, including:

- Server certificate validation.
- Mutual TLS.
- TLS 1.3 PSK.
- Configured TLS cipher suite preference lists using RFC 8446 names.
- PSK-DHE group selection using RFC 8446 `NamedGroup` labels.
- Explicit PSK-only interop mode for servers that require TLS 1.3 `psk_ke`.
- DER-encoded X.509 certificate material for trust anchors, pinned server certificates, and mTLS client certificates.
- Protected references or files for private keys and TLS PSK material rather than raw secrets in ConfigDB.

The design also defines how TACACS+ shared-secret obfuscation behaves through the proxy. Already-obfuscated TACACS+ packet bodies pass through unchanged, unobfuscated packets may be obfuscated using the selected upstream configuration, and the proxy does not translate TACACS+ shared-secret obfuscation keys.

## Compatibility

This is an additive HLD and does not remove or replace the existing SONiC TACACS+ implementation.

When the forwarding feature is disabled, absent, unhealthy, or not selected by priority, existing TACACS+ behavior remains available. Operators may keep existing non-TLS TACACS+ servers later in the `TACPLUS_SERVER` list as migration fallback.

The HLD does not propose changes to authentication coverage, command authorization coverage, accounting coverage, or current root OS AAA plugins.

## Testing and Review Areas

The document defines expected unit, system, security, sonic-mgmt, and backward compatibility test coverage for:

- Existing TACACS+ behavior with the feature disabled.
- Loopback listener behavior and non-network exposure.
- Separation between compatibility rows and TLS forwarding rows.
- TCP and TLS upstream candidate selection by priority.
- TLS server authentication, mTLS, and TLS PSK.
- PSK-DHE, explicit PSK-DHE group lists, and PSK-only interop mode.
- Hot reload and invalid configuration handling.
- Failover, preferred-server recovery, and single-connection reuse.
- Secret masking in logs and show commands.

The main open review areas are KubeSONiC deployment mechanics, ConfigDB schema names, CLI syntax, YANG alignment with RFC 9950, credential provisioning through ACMS or an equivalent SONiC-approved flow, and sonic-mgmt test planning.

# Performance Numbers

This benchmark captures TACACS+ authorization throughput across the main execution paths used by the project: direct server access, direct server access with dedicated per-request sessions, the raw TACACS+ proxy, and the gRPC client API. The purpose is to show both absolute throughput and scaling behavior as client-side concurrency increases.

Each result is reported as requests per second (RPS). The three result sets represent load-test runs with 1, 3, and 10 parallel clients. Direct + Dedicated disables TACACS+ single-connection mode and pays per-request session setup cost. Direct uses TACACS+ single-connection mode directly against the upstream server. Proxy and gRPC use centralized forwarding, with the intermediary reusing upstream TACACS+ single-connection mode where supported.

## Benchmark Results

**1 Parallel Client**

| Transport | Direct + Dedicated | Direct  | Proxy  | GRPC   |
| --------- | ------------------ | ------- | ------ | ------ |
| TCP       | 315.68             | 1691.38 | 798.5  | 884.46 |
| TLS       | 18.69              | 1691.1  | 757.72 | 905.01 |
| TLS-PSK   | 18.12              | 1612.63 | 716.69 | 883.9  |

**3 Parallel Clients**

| Transport | Direct + Dedicated | Direct  | Proxy   | GRPC    |
| --------- | ------------------ | ------- | ------- | ------- |
| TCP       | 581.4              | 3853.82 | 1645.2  | 2046.01 |
| TLS       | 52.95              | 3254.4  | 1413.57 | 1934.24 |
| TLS-PSK   | 51.63              | 3240.77 | 1358.59 | 2037.42 |

**10 Parallel Clients**

| Transport | Direct + Dedicated | Direct  | Proxy  | GRPC   |
| --------- | ------------------ | ------- | ------ | ------ |
| TCP       | 1083.2             | 10104.4 | 1387.4 | 2963.7 |
| TLS       | 160.7              | 12140.1 | 1537.7 | 3060.0 |
| TLS-PSK   | 152.6              | 9745.3  | 1497.5 | 3106.3 |

## Analysis

The direct single-connection path remains the performance ceiling and scales most strongly with additional parallel clients. Across the tested transports it grows from roughly 1.6k to 1.7k RPS at 1 client, to roughly 3.2k to 3.9k RPS at 3 clients, and reaches roughly 9.7k to 12.1k RPS at 10 clients. This confirms that the upstream TACACS+ implementation and transport stack can sustain high throughput when multiplexed connection reuse is active.

The Direct + Dedicated baseline scales much more slowly because every request pays dedicated TACACS+ session setup cost. Plain TCP improves from 315.68 RPS at 1 client to 1083.2 RPS at 10 clients, but TLS and TLS-PSK remain much lower, rising only from about 18 RPS to about 150 to 160 RPS. This spread is the clearest evidence that repeated TLS session setup dominates dedicated-mode cost.

The proxy path benefits substantially from upstream connection reuse, especially for TLS-based transports. At 1 client it reaches about 717 to 799 RPS, at 3 clients it reaches about 1.36k to 1.65k RPS, and at 10 clients it remains around 1.39k to 1.54k RPS. That means proxy throughput improves sharply from 1 to 3 clients, but it does not continue scaling with the direct single-connection path at 10 clients. The likely remaining cost is in the intermediary forwarding path: downstream per-request TCP handling, packet translation, scheduling, request fan-in/fan-out, and proxy service overhead rather than upstream TACACS+ transport capacity.

The gRPC path is the strongest centralized mode in these runs. It grows from about 884 to 905 RPS at 1 client, to about 1.93k to 2.05k RPS at 3 clients, and then to about 2.96k to 3.11k RPS at 10 clients. gRPC still lands well below the direct single-connection ceiling, but it scales more consistently than the raw proxy path and demonstrates that centralized TACACS+ forwarding can deliver substantially better throughput than dedicated direct operation while preserving centralized configuration, TLS or PSK policy, credential isolation, observability, and future policy enforcement.

Overall, the centralized paths are clearly viable for management-plane authorization workloads and provide substantial throughput above dedicated direct operation, particularly for TLS and TLS-PSK. The scaling results also show where the remaining optimization work sits: direct single-connection mode proves that the upstream can go much faster, while proxy and gRPC results show the cost of the intermediary path as concurrency rises.

### DIFF
--- a/doc/aaa/TACACS+ TLS with Central Agent.md
+++ b/doc/aaa/TACACS+ TLS with Central Agent.md
@@ -1,8 +1,8 @@
-# TACACS+ TLS with KubeSONiC Local Forwarding
+# TACACS+ TLS with Central Agent
 
 ## Table of Content
 
-- [TACACS+ TLS with KubeSONiC Local Forwarding](#tacacs-tls-with-kubesonic-local-forwarding)
+- [TACACS+ TLS with Central Agent](#tacacs-tls-with-central-agent)
   - [Table of Content](#table-of-content)
   - [1. Revision](#1-revision)
   - [2. Scope](#2-scope)
@@ -18,7 +18,7 @@
     - [5.7 Exemptions and Not Supported Items](#57-exemptions-and-not-supported-items)
   - [6. Architecture Design](#6-architecture-design)
     - [6.1 Existing SONiC TACACS+ Architecture](#61-existing-sonic-tacacs-architecture)
-    - [6.2 Proposed KubeSONiC Architecture](#62-proposed-kubesonic-architecture)
+    - [6.2 Proposed Central Agent Architecture](#62-proposed-central-agent-architecture)
     - [6.3 Component Responsibilities](#63-component-responsibilities)
     - [6.4 Repositories and Modules](#64-repositories-and-modules)
   - [7. High-Level Design](#7-high-level-design)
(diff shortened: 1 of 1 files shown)


Rod-Persky:
This HLD is in support of #2035 for TACACS+ over TLS

PR #2357, opened 2026-06-01
[doc] Add HLD for ASIC SerDes precoding configuration

Natgerbi:


abelamit (dismissed):

doc/pmon/precoding_via_media_settings_hld.md
```diff
@@ -0,0 +1,179 @@
...
+APSU stands for Autonomous Path Start-Up. It is part of IEEE 802.3dj and is used during link bring-up to coordinate link training and precoding across the link. 
+
+When APSU is enabled, the link can negotiate the required behavior automatically; when it is disabled, software or firmware must configure precoding manually. 
+
+APSU is a capability defined in IEEE 802.3dj and is still being adopted across both ASICs and modules. Until APSU is supported end to end across the full link path, the system may need to rely on explicit manual precoding configuration instead of autonomous negotiation. 
+
+---
+
+### 5. Requirements
+
```
prgeor, line 74: @Natgerbi What about the case where module also needs to honor precoding or no-precoding?

Natgerbi: The expectation is to have the exact same precoding configurations in media_settings.json as in the module.
If no precoding is needed at all, simply skip the precoding configuration in media_settings.json, and the behavior will remain the same as it is today.

doc/pmon/precoding_via_media_settings_hld.md
```diff
@@ -0,0 +1,179 @@
...
+| 0.1 | 2026-05-12 | Natanel Gerbi | Initial draft.                                                          |
+
+
+---
+
+### 2. Scope
+
+This HLD covers ASIC/SerDes precoding configuration on platforms where module firmware does not yet support APSU.
+
+In scope:
```
prgeor, line 37: @Natgerbi can you specify the scope is limited to precoding at TP0/ASIC. Module precoding is outside of this scope?

Natgerbi: yes the scope is ASIC/Serdes only and Module precoding is out of scope.
I marked it above here: "This HLD covers ASIC/SerDes precoding configuration on platforms where module firmware does not yet support APSU."

do you want me to add sentence that say explicitly that module precoding is out of scope for this HLD?

doc/pmon/precoding_via_media_settings_hld.md
```diff
@@ -0,0 +1,179 @@
...
+| Term      | Definition |
+| --------- | ---------- |
+| CMIS      | Common Management Interface Specification |
+| APSU      | Autonomous Path Start-Up (IEEE 802.3dj Annex 178B) — end-to-end protocol that extends iLT with in-band precoding negotiation |
+| iLT       | Initial Link Training — per-lane electrical handshake between switch SERDES and module host CDR |
+| SAI       | Switch Abstraction Interface |
+
+---
+
+### 4. Background
```
prgeor, line 60: @Natgerbi please add a section Why we need to take this approach instead of following the IEEE 802.3dj spec for APSU + ILT where both the link partner can negotiate the Tx and Rx precoding.

Natgerbi: it is exist here in background below:
"APSU is a capability defined in IEEE 802.3dj and is still being adopted across both ASICs and modules. **Until APSU is supported end to end across the full link path, the system may need to rely on explicit manual precoding configuration instead of autonomous negotiation.** "

prgeor: @Natgerbi ok. Can you add a subsection `Why precoding is needed?` The explanation remains same as you pointed out.

doc/pmon/precoding_via_media_settings_hld.md
prgeor: @Natgerbi can you add a sub-section "Out-of-Scope" . That transceiver module precoding is NOT part of this HLD

Natgerbi: Done

doc/pmon/precoding_via_media_settings_hld.md
Natgerbi: Done.

prgeor (approved):

PR #2358, opened 2026-06-02
TACACS+ TraceId documentation and update HLD

Rod-Persky:
## Summary

Adds a High-Level Design (HLD) document at `doc/aaa/TACACS+ Trace ID.md` describing how SONiC propagates a client-supplied SSH `TraceId` environment variable into TACACS+ command authorization requests issued by `bash_tacplus`.

## Related implementation PR

- sonic-net/sonic-buildimage#27667 — `[tacacs]: Add SSH TraceId authorization attribute` (the code change implementing this HLD)
- sonic-net/sonic-host-services#407 — `[tacacs]: Add SSH TraceId authorization attribute support`
- sonic-net/sonic-utilities#4743 — `[tacacs]: Add SSH client TraceId configuration CL`

## Motivation

Operators routinely need to reconstruct what happened during a specific access session or automation workflow — not just what commands were observed on a device. Today, TACACS+ command authorization records identify the user and command but carry no end-to-end correlation value supplied by the client workflow. This results in disjoint audit trails between the device-side TACACS+ logs and external systems such as access proxies, session recorders, or automation runners.

`TraceId` provides a shared join key. An access proxy or automation system generates a `TraceId`, injects it into the SSH session, preserves the same value in its own logs, and SONiC copies the validated value into every TACACS+ command authorization request for that session. A TACACS+ server or downstream log pipeline can then correlate device command authorization records with external access-session, session-recording, incident-response, or automation-run records.

## Design overview

```text
SSH client  --SetEnv/SendEnv TraceId-->  sshd (AcceptEnv TraceId)
   --> bash startup marks TraceId readonly
   --> bash_tacplus validates TraceId
   --> tac_add_attrib(&attr, "traceid", trace_id)
   --> TACACS+ authorization server
```

- Standard OpenSSH `AcceptEnv` is used to receive `TraceId`; no new SSH client option is introduced.
- The user's initial interactive bash shell marks `TraceId` `readonly` as a best-effort accidental-change guard (not a security boundary).
- `bash_tacplus` validates the value (length, character set) and either sends the exact value as a `traceid=<value>` AV pair, or omits the attribute entirely.
- Behavior is **unchanged** when `TraceId` is absent or invalid, and for all non-authorization TACACS+ flows.

## Scope

**In scope**
- Accept an SSH client-supplied `TraceId` environment variable.
- Preserve it as `readonly` in the initial interactive bash shell.
- Validate before attaching to TACACS+ requests.
- Add `traceid=<value>` to TACACS+ command authorization requests only.
- Unit, integration, and regression tests for the authorization path.

**Out of scope**
- Other TACACS+ packet types or other AAA flows.
- Local Linux audit record changes.
- A new `ssh -o TraceId=<value>` client option.
- A tamper-resistant root-owned per-session state store.
- SAI, SWSS, syncd, ASIC, or platform driver changes.

## Compatibility

- SSH login continues to work for clients that do not send `TraceId`.
- TACACS+ servers that ignore unknown AV pairs continue to work unchanged.
- Existing TACACS+ server configuration and AAA CLI behavior are untouched.
- Existing OpenSSH hardening settings remain intact.

## Document

- `doc/aaa/TACACS+ Trace ID.md` — full HLD following the SONiC HLD template (sections 1-16).

## Type of change

- [x] Documentation (HLD)

This PR is documentation-only; the implementation lives in sonic-net/sonic-buildimage#27667.


tirupatihemanth:
Any plans to show some sort of indication in either CLI (or elsewhere) to indicate that we've TraceId implementation/support present. 
NOTE: This is a non-blocker for this HLD approval

sarabjeetsingh1509:
Hi Rod,

One recommendation is to check if any testcases needs to be added into https://github.com/sonic-net/sonic-mgmt/tree/master/tests/tacacs

Regards

doc/aaa/TACACS+ Trace ID.md
```diff
@@ -0,0 +1,541 @@
...
+| HLD | High Level Design. |
+
+## 5. Overview
+
+SONiC supports TACACS+ command authorization through `bash_tacplus`. When a TACACS+ user runs a command in bash, the plugin builds a TACACS+ authorization request and sends command metadata such as `task_id`, `protocol`, `service`, `cmd`, and `cmd-arg`. This device-side record can then be correlated with external access proxy or automation records when a shared `TraceId` is present.
+
+This design adds one optional authorization attribute:
+
+```text
+traceid=<trace-id>
```
venkatmahalingam, line 85: What if we don't want to honor the trace-id in the SONiC switch, do we have a config option? I'm thinking from security standpoint.

Rod-Persky: Hi @venkatmahalingam, I've added [yang configuration which is exposed in redis](https://github.com/Rod-Persky/SONiC/blob/ff684ff366e22b4749a8abd77182de5d4f5ec61c/doc/aaa/TACACS%2B%20Trace%20ID.md?plain=1#L440C1-L466C4) that can be used to disable this feature. This also can be leveraged to identify the feature is available.

Rod-Persky:
@tirupatihemanth, regarding

> Any plans to show some sort of indication in either CLI (or elsewhere) to indicate that we've TraceId implementation/support present. NOTE: This is a non-blocker for this HLD approval

I've added [yang configuration which is exposed in redis](https://github.com/Rod-Persky/SONiC/blob/ff684ff366e22b4749a8abd77182de5d4f5ec61c/doc/aaa/TACACS%2B%20Trace%20ID.md?plain=1#L440C1-L466C4) that can be used to identify the feature is available and active.

doc/aaa/TACACS+ Trace ID.md
Rod-Persky: Hey @venkatmahalingam I've also added https://github.com/sonic-net/sonic-utilities/pull/4743 which adds a CLI config for this too, and it's now default disabled

PR #2362, opened 2026-06-02
Add pmon configuration resolution HLD

aditya-nexthop:
Add the High Level Design for how pmon daemons resolve their runtime tunables
from the per-platform `pmon_daemon_control.json`, replacing the per-tunable
command-line-flag mechanism.

Today a tunable is plumbed end-to-end as a flag: the platform sets it in
`pmon_daemon_control.json`, `sonic-cfggen` loads that file while rendering
`docker-pmon.supervisord.conf.j2`, the template flattens it into `--flag value`,
`argparse` re-parses it, and the daemon constructor grows a parameter. Adding one
knob means editing four places, and it has to be redone per daemon — `thermalctld`
carries five tunables this way today, `xcvrd` several.

The design defines a shared `PmonDaemonConfig` resolver in `sonic_py_common` that
owns everything not specific to a daemon: locating the file (hwsku over platform,
mirroring `docker_init.j2`), extracting the daemon's section, layering it over the
built-in defaults, coercing types, validating ranges, and degrading to defaults on
any error. A daemon adopts it by declaring a dataclass subclass with a section name
and one field per tunable. Adoption is per-daemon and independent, so a daemon that
has not migrated is unaffected. `xcvrd` is the first adopter; `thermalctld` is
written out as a worked example.

Adding a tunable becomes one field plus one `FieldSpec` entry, and changing a value
only requires restarting the daemon rather than having pmon regenerate the whole
supervisord template.

## Rev 0.3 update

Rev 0.3 generalizes the mechanism further so a daemon section can express related
tunables as **one-level-deep subsections** instead of a flat list of keys, and
adds a first-class backward-compatibility path:

- **Subsections (`SUBSECTIONS`).** A daemon schema may nest child schemas exactly
  one level deep. `xcvrd`'s tunables are regrouped under `dom`, `cmis_mgr`,
  `sff_mgr`, and `cpo_mgr`, so each manager owns its own `enabled` toggle and
  `dom` owns its poll/update intervals.
- **Legacy aliases (`LEGACY_ALIASES`).** The old flat keys and the top-level
  manager keys are retained as deprecated aliases for a compatibility window. When
  the nested form is absent, the old key still takes effect and logs a one-time
  deprecation warning; when both are present, the nested form wins.
- **Manager toggles move daemon-side.** `skip_xcvrd_cmis_mgr` →
  `xcvrd.cmis_mgr.enabled` (inverted) and `enable_xcvrd_sff_mgr` →
  `xcvrd.sff_mgr.enabled`, both retained as `scope='file'` aliases. `cpo_mgr.enabled`
  is net-new (the template never emitted `--skip_cpo_mgr`, so there is no prior
  top-level key to preserve).
- **0-boundary semantics.** Flat legacy keys apply a `v or None` transform to
  reproduce the old Jinja truthy gate (a flat `0` maps to the daemon default),
  while the nested `dom.*` form treats `0` as continuous polling.
- **Worked JSON examples.** §7.9 and §9.3 now show the concrete nested
  `pmon_daemon_control.json` a platform owner writes (for both `xcvrd` and the
  illustrative `thermalctld`), alongside the deprecated form still honored during
  the compatibility window.

## Also covered

- **Validation.** Ranges are enforced after type coercion, with a documented type,
  valid range, default, and boundary semantics per tunable. Coercion alone is not
  validation: a negative interval coerces to a perfectly good `int` but is not a
  valid cadence, and consumers today disagree about what happens next. An
  out-of-range value keeps the built-in default and logs a warning — never fatal,
  so a bad tunable cannot keep a daemon down.
- **Where the shared resolver lives.** `sonic_py_common` is already a universal
  dependency of every pmon daemon. Alternatives considered and rejected: keeping
  the resolver private to `xcvrd`, a new shared wheel in `sonic-platform-daemons`,
  folding it into `DaemonBase`, and sourcing tunables from Config DB.
- **Top-level keys vs the daemon section (§9.3).** Top-level `skip_*` / `delay_*` /
  capability keys are pmon orchestration settings consumed by the supervisord
  template; the nested `"<daemon>"` object is the daemon's own runtime config.
  `skip_xcvrd` and `delay_xcvrd` gate process startup and can never move daemon-side.
- **Warmboot/fastboot impact, memory, restrictions, unit and system test plans, and
  open items** (including the compatibility window before the legacy aliases are
  removed).

The doc lives at `doc/pmon/pmon_daemon_config_hld.md` and is linked from
`sonic_docs_toc.md` under the pmon section, matching its pmon-wide scope.

## Revision history

- **v0.1** — initial (`xcvrd`-specific `XcvrdConfig`).
- **v0.2** — generalizes the resolver to `PmonDaemonConfig` and adds per-field
  range validation, following review feedback.
- **v0.3** — adds one-level-deep subsections (`SUBSECTIONS`) and generic
  `LEGACY_ALIASES` support; regroups `xcvrd` tunables under
  `dom`/`cmis_mgr`/`sff_mgr`/`cpo_mgr`; relocates the manager enable/disable
  toggles into the `xcvrd` section behind a compatibility window; adds example
  nested `pmon_daemon_control.json` sections.

## Associated code PRs (to be merged in order)

- sonic-net/sonic-buildimage#28859
- sonic-net/sonic-platform-daemons#854
- sonic-net/sonic-buildimage#28306


aditya-nexthop:
@prgeor

moshemos (dismissed):

Junchao-Mellanox:
Hi @aditya-nexthop , it is a good idea to simplify the flow of adding parameters to xcvrd. Based on your proposal, I think we could do it more generic. How about changing XcvrdConfig to PmonDaemonConfig? Other daemons like thermalctld, psud could also be beneficial from this new desing.

prgeor:
> Hi @aditya-nexthop , it is a good idea to simplify the flow of adding parameters to xcvrd. Based on your proposal, I think we could do it more generic. How about changing XcvrdConfig to PmonDaemonConfig? Other daemons like thermalctld, psud could also be beneficial from this new desing.

@Junchao-Mellanox I have requested @aditya-nexthop for a community review

doc/xrcvd/xcvrd_config_hld.md
```diff
@@ -0,0 +1,320 @@
...
+
+### 7.3. Merge and coercion
+
+- Keys not declared as `XcvrdConfig` fields are ignored with a log notice
+  (forward-compatible: a newer platform file may carry keys an older daemon
+  does not know).
+- A `None` value never overrides a lower layer.
+- A `_FIELD_CASTERS` map coerces values to the declared type (e.g. a value given
+  as the string `"20"` becomes `int 20`), mirroring the old `argparse type=int`.
+  An uncoercible value is ignored with a log warning and the default is kept.
```
tshalvi: Should `XcvrdConfig` also validate the allowed range of each tunable after type coercion? or is it validated somewhere else? For example, a negative polling interval can be successfully coerced to an int, but may not be a valid configuration value. It may be useful to document the valid range for each field.

aditya-nexthop: yes, added this, please check now.

aditya-nexthop:
I moved the file to the pmon directory as we have increased the scope.

keboliu:
Hi @aditya-nexthop is this HLD finalized, or are you still updating it? Do you have a plan to release the new PRs(I see you mentioned that the PRs are out-of-date)

aditya-nexthop:
> Hi @aditya-nexthop is this HLD finalized, or are you still updating it? Do you have a plan to release the new PRs(I see you mentioned that the PRs are out-of-date)

Yes, I have updated the PRs - please take a look @keboliu.
We will have to figure out the proper merge order or split the PRs, but the code should all be there now.

PR #2366, opened 2026-06-05
BMP External Collector Support in SONiC

kalash-nexthop:
Adds a High-Level Design document for configurable BMP collectors in SONiC. Today `bgpd` streams BMP only to the on-box `openbmpd` (hard-coded `127.0.0.1:5000`); this feature lets operators stream to one or more **external** collectors of their choice and control the feed, avoiding the on-box collector's memory cost at high route scale.

The HLD covers:
- CONFIG_DB + YANG model: `BMP_TARGET`, `BMP_TARGET_COLLECTOR`, `BMP_TARGET_AFI_SAFI`, and global `mirror-buffer-limit`
- Per-target, per-AFI/SAFI monitoring (Adj-RIB-In pre/post, Loc-RIB), mirroring, stats, retry timers, and source-interface
- Rendering onto FRR's native BMP config via both `bgpcfgd` and `frrcfgd`, preserving the existing on-box default for backward compatibility

Related PR: https://github.com/sonic-net/sonic-buildimage/pull/24965

### DIFF
--- a/doc/bmp/bmp-external-collector-hld.md
+++ b/doc/bmp/bmp-external-collector-hld.md
@@ -1,0 +1,406 @@
+# BMP External Collector Support in SONiC
+
+## High Level Design Document
+
+# Table of Contents
+
+  * [Revision](#revision)
+
+  * [About this Manual](#about-this-manual)
+
+  * [Definitions/Abbreviation](#definitionsabbreviation)
+
+  * [1 Requirements Overview](#1-requirements-overview)
+    * [1.1 Functional requirements](#11-functional-requirements)
+    * [1.2 CLI requirements](#12-cli-requirements)
+    * [1.3 Scalability and Default Values](#13-scalability-and-default-values)
+    * [1.4 Warm Restart requirements](#14-warm-restart-requirements)
+  * [2 Architecture Design](#2-architecture-design)
+    * [2.1 High-Level Architecture](#21-high-level-architecture)
+    * [2.2 BMP Object Model](#22-bmp-object-model)
+    * [2.3 BGP/FRR Configuration Mapping](#23-bgpfrr-configuration-mapping)
+    * [2.4 Configuration Rendering (bgpcfgd and frrcfgd)](#24-configuration-rendering-bgpcfgd-and-frrcfgd)
+    * [2.5 Backward Compatibility](#25-backward-compatibility)
+    * [2.6 Config DB and YANG Schema](#26-config-db-and-yang-schema)
+    * [2.7 Information Flow](#27-information-flow)
+  * [3 CLI](#3-cli)
+  * [4 Resource usage and Test plan](#4-resource-usage-and-test-plan)
+
+###### Revision
+
+| Rev |     Date    |       Author       | Change Description                |
+|:---:|:-----------:|:------------------:|-----------------------------------|
+| 0.1 |  06/04/2026 |   Kalash Nainwal   | Initial version                   |
+
+
+# About this Manual
+
+SONiC already supports BMP (BGP Monitoring Protocol, RFC 7854) for monitoring on-box BGP state. That feature — described in the existing design document [BMP for monitoring SONiC BGP info](https://github.com/sonic-net/SONiC/blob/master/doc/bmp/bmp.md) — brings up a `bmp` container running two daemons: `bmpcfgd`, which watches the `BMP` table in CONFIG_DB and controls which tables are populated, and `openbmpd`, an **oon-box** BMP collector that accepts the BMP stream from `bgpd`, parses it, and writes BGP state into `BMP_STATE_DB` for consumption over GNMI / streaming telemetry. To feed `openbmpd`, `bgpd` is started with the BMP module (`-M bmp`) and its configuration template emits a single, hard-coded BMP target connecting to `127.0.0.1:5000`. The only operator-facing knobs (`BMP|table` — `bgp_neighbor_table`, `bgp_rib_in_table`, `bgp_rib_out_table`) control *which tables `openbmpd` populates*; there is no way to point `bgpd` at any collector other than the local one, and no way to tune what each target monitors.
+
+However, some customers want to stream BMP to one or more **external** collectors of their own choosing, rather than depending on the locally-running OpenBMP instance in the `bmp` container. Running that collector on the switch also carries a heavy memory cost in high route-scale environments. Shipping BMP directly to purpose-built external collectors alleviates this memory concern.
+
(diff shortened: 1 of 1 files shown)


doc/bmp/bmp-external-collector-hld.md
```diff
@@ -0,0 +1,406 @@
...
+    * [1.3 Scalability and Default Values](#13-scalability-and-default-values)
+    * [1.4 Warm Restart requirements](#14-warm-restart-requirements)
+  * [2 Architecture Design](#2-architecture-design)
+    * [2.1 High-Level Architecture](#21-high-level-architecture)
+    * [2.2 BMP Object Model](#22-bmp-object-model)
+    * [2.3 BGP/FRR Configuration Mapping](#23-bgpfrr-configuration-mapping)
+    * [2.4 Configuration Rendering (bgpcfgd and frrcfgd)](#24-configuration-rendering-bgpcfgd-and-frrcfgd)
+    * [2.5 Backward Compatibility](#25-backward-compatibility)
+    * [2.6 Config DB and YANG Schema](#26-config-db-and-yang-schema)
+    * [2.7 Information Flow](#27-information-flow)
```
asechoud, line 25: Thanks for coming up with this document. Please find my review comments.

Since the document is to address external collectors, it is important to address security aspects, mainly,:
1. is the data in plaintext, TLS, secure tunnel?
2. Any ACL/allowlist to restrict traffic from/to collector
3. data steers through mgmt eth0 or data-plane

doc/bmp/bmp-external-collector-hld.md
```diff
@@ -0,0 +1,406 @@
...
+
+This document proposes **extending** that support so that, in addition to (or instead of) the on-box collector, an operator can stream BMP to one or more **external** collectors and fully control the BMP feed. Concretely, this feature adds the ability to configure: multiple BMP **targets**; one or more **collectors** per target (by IP/hostname and TCP port); per-target, per-AFI/SAFI selection of Adj-RIB-In pre-policy, Adj-RIB-In post-policy, and Loc-RIB; per-target route mirroring and statistics interval; per-collector connection-retry timers and source interface; and a global mirror buffer limit. The underlying capability already exists in `sonic-frr`'s `bgpd`; this design exposes it through SONiC's configuration model (CONFIG_DB + YANG), while preserving today's behavior for deployments that configure nothing new.
+
+The on-box collector path (`openbmpd`, `BMP_STATE_DB`, the GNMI/telemetry consumption, and the `BMP|table` controls) is unchanged by this design.
+
```
asechoud, line 47: This document doesn't talk about multi-asic platforms. It will be good to put a small note or defer it.

doc/bmp/bmp-external-collector-hld.md
```diff
@@ -0,0 +1,406 @@
...
+| `mirror`                 | boolean                            | `false`       |
+| `adj-rib-in-pre/post`, `loc-rib` | boolean                    | `false`       |
+| target `name`            | string, length `1..64`             | n/a           |
+
+The number of targets/collectors is bounded by what `bgpd` and the platform can sustain; no artificial SONiC-side cap is imposed. The monitored AFI/SAFI set is limited to the seven enumerated values above.
+
+## 1.4 Warm Restart requirements
+
+No special handling for Warm restart. BMP is a monitoring/export feature outside the data-forwarding path. BMP configuration is ordinary persistent CONFIG_DB state and is reapplied through the standard render/daemon path after a reboot or container restart; sessions re-establish according to the configured retry timers.
+
```
asechoud, line 103: since it is monitoring traffic, it will be good to touch base on packet treatment, as cs6/7 management or best-effort or any other.

doc/bmp/bmp-external-collector-hld.md
```diff
@@ -0,0 +1,406 @@
...
+Operational visibility of the collected data continues to use the existing commands (unchanged):
+
+```
+1. Command: `show bmp bgp-neighbor-table`
+2. Command: `show bmp bgp-rib-out-table`
+3. Command: `show bmp bgp-rib-in-table`
+4. Command: `show bmp tables`
+```
+
+Dedicated `config bmp collector …` / `show bmp collectors` click commands for ergonomic editing and display of the new tables are a possible follow-up but are not required by this design.
```
asechoud: show bmp collector command is important from debugability and even security perspective. If not not supported in initial phase, corresponding frr command may be listed here.

doc/bmp/bmp-external-collector-hld.md
```diff
@@ -0,0 +1,406 @@
...
+| 0.1 |  06/04/2026 |   Kalash Nainwal   | Initial version                   |
+
+
+# About this Manual
+
+SONiC already supports BMP (BGP Monitoring Protocol, RFC 7854) for monitoring on-box BGP state. That feature — described in the existing design document [BMP for monitoring SONiC BGP info](https://github.com/sonic-net/SONiC/blob/master/doc/bmp/bmp.md) — brings up a `bmp` container running two daemons: `bmpcfgd`, which watches the `BMP` table in CONFIG_DB and controls which tables are populated, and `openbmpd`, an **oon-box** BMP collector that accepts the BMP stream from `bgpd`, parses it, and writes BGP state into `BMP_STATE_DB` for consumption over GNMI / streaming telemetry. To feed `openbmpd`, `bgpd` is started with the BMP module (`-M bmp`) and its configuration template emits a single, hard-coded BMP target connecting to `127.0.0.1:5000`. The only operator-facing knobs (`BMP|table` — `bgp_neighbor_table`, `bgp_rib_in_table`, `bgp_rib_out_table`) control *which tables `openbmpd` populates*; there is no way to point `bgpd` at any collector other than the local one, and no way to tune what each target monitors.
```
asechoud: oon-box->on-box

asechoud (changes requested):

doc/bmp/bmp-external-collector-hld.md
kalash-nexthop: Thanks for the review @asechoud . Please find my comments inline (I will also update the hld to clearly state all these points).

1. Plaintext. RFC 7854 BMP is cleartext TCP, and FRR's BMP implementation has no TLS/DTLS. Someone worried about security should have an IPSec tunnel interface for example and set that as `source-interface <tunnel-if> on the BMP_TARGET_COLLECTOR row`, but BMP by itself doesn't provide any secure channel.

2. Inbound is a non-issue (bgpd is the client; there's no listening BMP socket on the switch, so nothing to allowlist on the "from" side). On the outbound side, bgpd only ever opens a BMP session to the exact (ip, port) pairs in BMP_TARGET_COLLECTOR.  But otherwise there is no ACL configuration to further restrict traffic to a collector.

3. Using `source-interface` one can specify mgmt eth0 or data-plane interface and then bmp updates are steered through that interface.

doc/bmp/bmp-external-collector-hld.md
kalash-nexthop: I will add this, thanks.

doc/bmp/bmp-external-collector-hld.md
kalash-nexthop: BMP update pkts don't get any special treatment by default, so they should inherit default behavior; if an operator wants CS6 treatment, it probably can be done via a QoS/ACL rule, out of scope here.

doc/bmp/bmp-external-collector-hld.md
kalash-nexthop: I will add a note here. The FRR vtysh command is `show bmp` that displays targets, connect sessions, state, and counters. We can have a sonic show command around that to help with observability.

doc/bmp/bmp-external-collector-hld.md
kalash-nexthop: Oops. Will fix.

PR #2367, opened 2026-06-05
Protocol Route-Map HLD

kalash-nexthop:
Adds a HLD for `PROTOCOL_ROUTE_MAP` — a CONFIG_DB table that binds a route-map to a `(vrf, addr_family, source-protocol)` tuple, so operators can selectively prevent specific routes of a given protocol from being installed into the FIB. Backed by FRR's native `ip|ipv6 protocol <P> route-map <RM>` zebra route filtering. The main use-case of this feature is to allow configuring fib route filtering.

The HLD covers:
- YANG model and CONFIG_DB schema for `PROTOCOL_ROUTE_MAP`
- Boot-time render via a Jinja template included from `frr.conf.j2`
- Runtime apply via new `protocol_route_map_handler` in `frrcfgd`, routed via `mgmtd`
- A `set_src` route-map leaf on `ROUTE_MAP_SET_LIST` so the classic `RM_SET_SRC` behavior is expressible via CONFIG_DB — feature parity with what bgpcfgd's `ZebraSetSrc` manager provides today

Related PR: https://github.com/sonic-net/sonic-buildimage/pull/27192

### DIFF
--- a/doc/protocol-route-map/protocol-route-map-hld.md
+++ b/doc/protocol-route-map/protocol-route-map-hld.md
@@ -1,0 +1,459 @@
+# Protocol Route-Map Support in SONiC
+
+## Table of Contents
+
+1. [Revision](#1-revision)
+2. [Scope](#2-scope)
+3. [Definitions / Abbreviations](#3-definitions--abbreviations)
+4. [Overview](#4-overview)
+5. [Requirements](#5-requirements)
+6. [Architecture Design](#6-architecture-design)
+7. [High-Level Design](#7-high-level-design)
+   - 7.1 [YANG Schema](#71-yang-schema)
+   - 7.2 [Boot-time Rendering (Jinja Template)](#72-boot-time-rendering-jinja-template)
+   - 7.3 [Runtime Apply — frrcfgd](#73-runtime-apply--frrcfgd)
+   - 7.4 [FRR Zebra Route-Map Evaluation](#74-frr-zebra-route-map-evaluation)
+8. [CLI](#8-cli)
+   - 8.1 [CONFIG_DB Example](#81-config_db-example)
+   - 8.2 [Rendered FRR CLI](#82-rendered-frr-cli)
+9. [Information Flow Diagrams](#9-information-flow-diagrams)
+   - 9.1 [Apply Flow](#91-apply-flow)
+   - 9.2 [Remove Flow](#92-remove-flow)
+10. [Warmboot and Fastboot Design Impact](#10-warmboot-and-fastboot-design-impact)
+11. [Test Plan](#11-test-plan)
+    - 11.1 [Unit Tests](#111-unit-tests)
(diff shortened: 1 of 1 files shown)


doc/protocol-route-map/protocol-route-map-hld.md
```diff
@@ -0,0 +1,454 @@
...
+```
+
+---
+
+## 9. Information Flow Diagrams
+
+### 9.1 Apply Filter Flow
+
+```
+  Operator                CONFIG_DB        runtime manager           FRR (zebra)
```
eddieruan-alibaba: ”runtime manager"?  it is frrcfgd, right?

doc/protocol-route-map/protocol-route-map-hld.md
```diff
@@ -0,0 +1,454 @@
...
+     │                       │                  │                       │
+     │ sonic-db-cli HSET     │                  │                       │
```
eddieruan-alibaba, line 405: Why not keep the same syntax as vtysh, since it is route-map anyway.

doc/protocol-route-map/protocol-route-map-hld.md
kalash-nexthop: Thanks for looking @eddieruan-alibaba . The next line clarifies this:
```
The runtime manager is `bgpcfgd.FibRouteFilterMgr` when `bgpcfgd` owns FRR config translation, and `frrcfgd.fib_route_filter_handler` (routed through `mgmtd`) when `frrcfgd` is the active translator.
```

doc/protocol-route-map/protocol-route-map-hld.md
kalash-nexthop: I don't have any strong reason to pick `FIB_ROUTE_FILTER|<vrf>|<afi>|<protocol> → route_map=<RM>`, other than that I didn't find following vtysh cmd very intuitive (i.e, the syntax makes it hard to tell this is filtering routes at zebra level and not at individual protocol level):

`ip[v6] protocol PROTOCOL route-map ROUTEMAP`

I thought `FIB_ROUTE_FILTER` here would be better than something like `IP_PROTOCOL_FILTER` in conveying that this feature is filtering routes from fib rather than say applying bgp route maps to filter routes from rib itself.

kalash-nexthop: @eddieruan-alibaba I will change it to `PROTOCOL_ROUTE_MAP` (from `FIB_ROUTE_FILTER`) and update the PR.

doc/protocol-route-map/protocol-route-map-hld.md
kalash-nexthop: I will change it to frrcfgd. There is no point supporting this in bgpcfgd, as the model supports vrf, and in bgpcfgd mode many vrf features are not supported.

PR #2371, opened 2026-06-08
[bfd] Add SAI-level BFD-coupled NHG fast switchover HLD

yuezhoujk:
Add High Level Design document for SAI-level BFD-coupled Next Hop Group fast switchover. This feature enables ASIC hardware to autonomously remove or add ECMP group members based on BFD session state changes without control plane involvement, achieving microsecond-level convergence.

Key design points:
- New SAI attribute SAI_NEXT_HOP_ATTR_BFD_DISCRIMINATOR
- ASIC behavioral spec for autonomous ECMP member management
- Primary/standby role-based make-before-break switchover
- SONiC control plane changes (fpmsyncd, NhgOrch, Srv6Orch, BfdOrch)
- SAI SDK implementation reference

### DIFF
--- a/doc/bfd/BFD_SAI_Triggered_NHG_Fast_Switchover_HLD.md
+++ b/doc/bfd/BFD_SAI_Triggered_NHG_Fast_Switchover_HLD.md
@@ -1,0 +1,1015 @@
+# SAI-Level BFD-Coupled Next Hop Group Fast Switchover
+## High Level Design Document
+### Rev 0.1
+
+# Table of Contents
+
+  * [Revision](#revision)
+  * [About this Manual](#about-this-manual)
+  * [Scope](#scope)
+  * [Definitions/Abbreviation](#definitionsabbreviation)
+  * [1 Motivation](#1-motivation)
+    * [1.1 Problem Statement](#11-problem-statement)
+    * [1.2 Convergence Bottleneck Analysis](#12-convergence-bottleneck-analysis)
+  * [2 Requirements](#2-requirements)
+    * [2.1 Functional Requirements](#21-functional-requirements)
+    * [2.2 Non-Functional Requirements](#22-non-functional-requirements)
+  * [3 Architecture Design](#3-architecture-design)
+    * [3.1 Overall Architecture](#31-overall-architecture)
+    * [3.2 Dual-Path Convergence Model](#32-dual-path-convergence-model)
+  * [4 SAI Attribute Specification](#4-sai-attribute-specification)
+    * [4.1 SAI_NEXT_HOP_ATTR_BFD_DISCRIMINATOR](#41-sai_next_hop_attr_bfd_discriminator)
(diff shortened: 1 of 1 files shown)


doc/bfd/BFD_SAI_Triggered_NHG_Fast_Switchover_HLD.md
```diff
@@ -0,0 +1,1015 @@
...
+            Else:
+                Add directly (no BFD protection)
+
+    Step 2 — Break: Remove non-target_role members from ECMP forwarding
+        For each member with configured_role != target_role:
+            Remove from ECMP forwarding
+            Mark as inactive
+```
+
+### 4.3.4 BFD UP — Restore Forwarding
```
rajshekhar-nexthop, line 342: Is this safe operation? One scenario is routing protocols implement route flap damping/ hold-off timers and might enforce a 30-second wait before declaring link viable again. If the hardware autonomously re-adds the member to the ECMP group the millisecond BFD goes UP, it will blackhole traffic on a flapping link and completely ignore the control plane's stability policies.

doc/bfd/BFD_SAI_Triggered_NHG_Fast_Switchover_HLD.md
```diff
@@ -0,0 +1,1015 @@
...
+void NextHopGroup::onBfdDown(uint32_t disc)
+{
+    std::set<NextHopKey> members;
+    for (auto& mbr_it : m_members) {
+        if (mbr_it.first.srv6_bfd_disc == disc && mbr_it.second.isSynced()) {
+            members.insert(mbr_it.first);
+        }
+    }
+    if (!members.empty()) {
+        removeMembers(members);
```
rajshekhar-nexthop: Instead of removing, is it good to dynamically set the member's weight to 0, or update an admin-state/active attribute, preserving the object so NhgOrch state remains intact and the ASIC still has a reference to the object. Let they be removed by pathd later which will be a cleaner approach

doc/bfd/BFD_SAI_Triggered_NHG_Fast_Switchover_HLD.md
yuezhoujk: Thanks for raising this. The current implementation already handles BFD UP through the control plane path (pathd → zebra → fpmsyncd → NhgOrch) — there is no autonomous restore in the SAI SDK. Section 4.3.4 has been updated to clarify this.

doc/bfd/BFD_SAI_Triggered_NHG_Fast_Switchover_HLD.md
yuezhoujk: Yes, the current implementation already takes this approach — on BFD DOWN we use weight reduction and mark the member inactive, preserving the SAI object. Updated the HLD to reflect this more clearly in Section 7.2.

PR #2376, opened 2026-06-08
[pon] Add HLD for Dual Homed PON Gateway active/standby solution

anuragprakash:
Introduce a Dual Homed Gateway design using XGS-PON, with OLT-interface active/standby protection adapted from SONiC DualToR: ponprotd (composite FSM), dcpon-syncd (SDK bridge), PonOrch (route/IPinIP programming), failover and TSA flows, and new SAI APIs.


rajesh-kumar-raju (commented):
What are the exact criteria  by which an OLT interface becomes ACTIVE or remains STANDBY? The HLD states that the peer OLT SDK independently detects the vacancy and promotes the standby OLT, but the following details are unclear:

How is the initial active OLT selected when both gateways start simultaneously?
Is selection based on configured priority, gateway/OLT ID, optical detection, or another mechanism?

anuragprakash:
> What are the exact criteria by which an OLT interface becomes ACTIVE or remains STANDBY? The HLD states that the peer OLT SDK independently detects the vacancy and promotes the standby OLT, but the following details are unclear:
> 
> How is the initial active OLT selected when both gateways start simultaneously? Is selection based on configured priority, gateway/OLT ID, optical detection, or another mechanism?

An OLT interface becomes ACTIVE only when the local OLT SDK reports that it owns the active PON role for the protected ONUs and the associated Ethernet link is up. Evidence may include ONU registration/ranging completion, OMCI reachability, DBA scheduling with alloc-IDs, upstream burst reception, and service-flow counters. It remains STANDBY when the local OLT is deliberately held out of service, or the SDK determines that the peer-side session owns the service. UNKNOWN is used for incomplete, conflicting, or failed observations and is typically handled with tunnel forwarding; the “pon-protd" combines this local OLT role, link state, and any TSA drain state into the forwarding intent, while the "orch-agent" only programs the resulting direct or tunnel routes.

The HLD does not define an ITU-T-standard mechanism for selecting the initial active OLT when both gateways start simultaneously; ITU-T PON still defines ONU activation, OMCI, GEM transport, and upstream DBA behaviour for an OLT, but not redundant dual-OLT election. In that case, initial selection must come from the vendor OLT protection implementation, such as configured priority, deterministic OLT or gateway ID, explicit hardware protection-group policy, or vendor-specific optical/session arbitration. Once the SDK reports the selected local role, SONiC consumes it as ACTIVE, STANDBY, or UNKNOWN; SONiC design does not independently elect the active peer.

PR #2296, opened 2026-04-16
[HLD]: adds the High-Level Design document for Sampled Port Mirroring with Truncation 

Janetxxx:
#### Description
   
This PR adds the High-Level Design document for **Sampled Port Mirroring with Truncation** in SONiC.
   
#### Motivation
   
Current SONiC ERSPAN port mirroring mirrors every packet in full. At high line rates (e.g., 51 Tbps), this generates excessive mirror traffic (~1 Gbps per switch without truncation). 
Sampled mirroring with truncation reduces per-switch mirror bandwidth to ~52 Mbps, enabling scalable traffic monitoring across thousands of switches.
   
#### Design Summary
   
The feature leverages SAI `SamplePacket` objects with `TYPE=MIRROR_SESSION` to enable per-port hardware-based sampled mirroring. 

 **Implementation PRs:**
 - [YANG] Add sample_rate and truncate_size to sonic-mirror-session: https://github.com/sonic-net/sonic-buildimage/pull/26756
 - [CLI] Add sample_rate and truncate_size to mirror session CLI: https://github.com/sonic-net/sonic-utilities/pull/4459
 - [SwitchOrch/MirrorOrch] Add sampled port mirror and truncation capability discovery: https://github.com/sonic-net/sonic-swss/pull/4502

PR #2378, opened 2026-06-10
[HLD][Port Mirror] Add direction support, remove in-place update, and YANG model updates

Janetxxx:
#### Why I did it
 
 Follow-up to #2296, this PR updates the design with refinements made as the feature was implemented:
 
 - **Direction support (RX/TX/BOTH):** extend from ingress-only to per-direction sampling; add the `EGRESS_SAMPLEPACKET_ENABLE` / `EGRESS_SAMPLE_MIRROR_SESSION` port attributes (RX = ingress pair, TX = egress pair, BOTH = both) and update the SAI call sequence.
 - **Stricter configuration semantics:** a SET on an existing session is rejected as a duplicate instead of updated in place (reconfigure = explicit delete + re-add), and a sampled session must specify a direction.
 - **Fail-closed on unsupported platforms:** reject the session when the required capability is missing, instead of silently downgrading to full mirroring.
 - **Per-direction conflict detection (5.2.4):** describe the symmetric, per-direction SamplePacket conflict between sampled mirroring and sFlow.
 - **YANG:** relax the `sample_rate` range to `2..max`.
 - **Flow diagram (Section 4):** simplify to the create/delete workflow with the rejection path and per-direction bind/unbind steps.
 - **Test plan (9.3.1 / 9.3.2):** expand and renumber VS and system test cases (TX/BOTH/LAG, duplicate rejection, sFlow conflict, inactive/no-route, route withdraw–reactivate, truncate-without-rate default)

### DIFF
--- a/doc/port-mirroring/SONiC_Port_Mirroring_HLD.md
+++ b/doc/port-mirroring/SONiC_Port_Mirroring_HLD.md
@@ -187,7 +187,7 @@ flowchart LR
 ### 3.1.2 Data Format
 
 The mirrored packet goes through the following processing in the ASIC hardware pipeline:
-1. Sampling: For each ingress packet on the monitored port, a hardware counter is incremented. When the counter reaches N (the configured sample rate), the packet is selected for mirroring.
+1. Sampling: For each packet on the monitored port in the configured direction (RX, TX or BOTH), a hardware counter is incremented. When the counter reaches N (the configured sample rate), the packet is selected for mirroring.
 2. Truncation: The selected packet is truncated to the configured truncate size (e.g., 128 bytes). Only the packet headers are retained.
 3. GRE encapsulation: The truncated packet is encapsulated with outer headers for tunneling to the collector.
 4. Forwarding: The encapsulated mirror packet is forwarded to the monitor port and routed to the collector. The original packet continues normal forwarding unaffected.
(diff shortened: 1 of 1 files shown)


doc/port-mirroring/SONiC_Port_Mirroring_HLD.md
```diff
@@ -187,7 +187,7 @@ flowchart LR
 ### 3.1.2 Data Format
 
 The mirrored packet goes through the following processing in the ASIC hardware pipeline:
-1. Sampling: For each ingress packet on the monitored port, a hardware counter is incremented. When the counter reaches N (the configured sample rate), the packet is selected for mirroring.
+1. Sampling: For each packet on the monitored port in the configured direction (RX, TX or BOTH), a hardware counter is incremented. When the counter reaches N (the configured sample rate), the packet is selected for mirroring.
```
tirupatihemanth, line 190: This seems like one way to do sampling - is this platform specific? Sampling can be probabilistic and each packet could be selected with a probability of (1/sampling_rate)?

doc/port-mirroring/SONiC_Port_Mirroring_HLD.md
```diff
@@ -296,12 +296,19 @@ Mirror Orchestration agent is modified to support this feature:
...
+See Section 3.1.1 for the architecture overview and Section 4 for the detailed workflow.
+
+**Configuration semantics:**
+
+- **Duplicate session names are rejected; no in-place update.** When a SET arrives for a session name that already exists, `createEntry()` returns `task_duplicated` and the existing session is left unchanged — the new attributes are not applied. Reconfiguring any field of a sampled mirror session, including `sample_rate` or `truncate_size`, therefore requires deleting the session and re-adding it with the new values.
+- **Unsupported platforms are rejected, not silently downgraded.** If the required per-direction sampled mirroring capability (or samplepacket truncation) is not supported by the ASIC, MirrorOrch rejects the session (`task_invalid_entry`).
+- **Sampled mirroring requires an explicit direction.** A session that configures `sample_rate` must also specify a `direction` (RX, TX or BOTH); a sampled session without a direction is rejected.
```
tirupatihemanth, line 311: CMIIW: Are we just aligning HLD to the implementation we already have? The PRs that were already merged for this feature already has this behavior implemented right? We reject duplicate session name and not do in-place update. Also I think we don't silently downgrade but reject on unsupported platforms.

tirupatihemanth (approved):

doc/port-mirroring/SONiC_Port_Mirroring_HLD.md
Janetxxx: Yes correct, this updated HLD is just aligning with the implementation we're already had.

doc/port-mirroring/SONiC_Port_Mirroring_HLD.md
Janetxxx: Good catch, the counter-based description was just one possible implementation; the design only requires "on average 1 in N" and doesn't mandate a specific algorithm.

PR #2379, opened 2026-06-10
[doc] High frequency telemetry support of MIXED tam tel type mode HLD

DavidZagury:
## Summary
Extends the existing High frequency telemetry HLD (https://github.com/sonic-net/SONiC/blob/master/doc/high-frequency-telemetry/high-frequency-telemetry-hld.md) to add support for SAI_TAM_TEL_TYPE_MODE_MIXED_TYPE. Some vendor SAI implementations advertise only MIXED_TYPE, on which SONiC HFT cannot be enabled today because the orchagent hardcodes SAI_TAM_TEL_TYPE_MODE_SINGLE_TYPE.

The change is additive and orchagent-internal:

- HFTelOrch auto-detects which mode(s) the SAI advertises via sai_query_attribute_enum_values_capability(SAI_TAM_TEL_TYPE_ATTR_MODE) at init and stores the chosen mode for the lifetime of the orchagent.
- Selection rule: MIXED_TYPE when advertised (the default via HFTelOrch::DEFAULT_TEL_TYPE_MODE); SINGLE_TYPE otherwise; HFT disabled with a clear log when neither is advertised.

No changes to CONFIG_DB, YANG, CLI, STATE_DB schema, IPFIX wire format, or the OpenTelemetry exporter. The mode is selected once at orchagent init and fixed for its lifetime — no runtime mode switching in this phase.

## Related implementation PR

| Repo | PR title | state |
|------|----------|-------|
| [sonic-swss](https://github.com/sonic-net/sonic-swss/) | [[HFT][countersyncd]: Support SAI_TAM_TEL_TYPE_MODE_MIXED_TYPE](https://github.com/sonic-net/sonic-swss/pull/4732) | GitHub issue/pull request detail |
| [sonic-sairedis](https://github.com/sonic-net/sonic-sairedis/) | [[vslib][HFT] Advertise SAI_TAM_TEL_TYPE_ATTR_MODE capability](https://github.com/sonic-net/sonic-sairedis/pull/1969) | GitHub issue/pull request detail |
| [sonic-utilities](https://github.com/sonic-net/sonic-utilities/) | [[hft] Target per-key path in JSON Patch ops to avoid wiping the table](https://github.com/sonic-net/sonic-utilities/pull/4663) | GitHub issue/pull request detail |

### DIFF
--- a/doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
+++ b/doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
@@ -15,6 +15,7 @@
   - [7.4. Collapsing the tam\_report object](#74-collapsing-the-tam_report-object)
   - [7.5. State machine](#75-state-machine)
   - [7.6. Config-ready notification and templates](#76-config-ready-notification-and-templates)
+    - [7.6.1. CounterSyncd label resolution in MIXED](#761-countersyncd-label-resolution-in-mixed)
   - [7.7. STATE\_DB](#77-state_db)
   - [7.8. Work flow](#78-work-flow)
 - [8. SAI API](#8-sai-api)
@@ -38,7 +39,7 @@
 
 ## 2. Scope
 
-This document extends the existing [High frequency telemetry high level design](high-frequency-telemetry-hld.md) to add support for the `SAI_TAM_TEL_TYPE_MODE_MIXED_TYPE` mode of `sai_tam_tel_type`. It covers only orchagent-internal changes. The CONFIG_DB schema, YANG model, CLI, STATE_DB schema, IPFIX wire format, CounterSyncd, and OpenTelemetry exporter are unchanged.
+This document extends the existing [High frequency telemetry high level design](high-frequency-telemetry-hld.md) to add support for the `SAI_TAM_TEL_TYPE_MODE_MIXED_TYPE` mode of `sai_tam_tel_type`. It covers only orchagent-internal changes plus one contained internal extension to CounterSyncd's label-resolution path (§7.6.1). The CONFIG_DB schema, YANG model, CLI, STATE_DB schema, IPFIX wire format, the CounterSyncd public interface, and the OpenTelemetry exporter are unchanged.
 
 ## 3. Definitions/Abbreviations
 
@@ -69,12 +70,12 @@ Requirements specific to MIXED_TYPE support, in addition to those listed in the 
 Non-requirements:
 
 - No CONFIG_DB, YANG, CLI, or STATE_DB schema changes.
-- No CounterSyncd changes.
+- No CounterSyncd schema or public-interface changes. One internal extension to the label-resolution path is required so that the per-group sessions sharing a template_id in MIXED resolve correctly (see §7.6.1).
 - No runtime mode switching - the chosen mode is fixed for the lifetime of the orchagent process.
 
 ## 6. Architecture Design
 
-The architecture diagram from the [base HLD §6](high-frequency-telemetry-hld.md#6-architecture-design) is unchanged. The change is internal to `Orchagent → High frequency telemetry Orch`; the SAI/syncd boundary, CounterSyncd, OpenTelemetry container, and Redis databases are unaffected.
+The architecture diagram from the [base HLD §6](high-frequency-telemetry-hld.md#6-architecture-design) is unchanged. The bulk of the change is internal to `Orchagent → High frequency telemetry Orch`; the SAI/syncd boundary, the CounterSyncd public interface, the OpenTelemetry container, and the Redis databases are unaffected. CounterSyncd's internal label-resolution path is extended to handle the MIXED-mode case where multiple per-group sessions share a template_id (§7.6.1).
 
 ## 7. High-Level Design
 
@@ -145,7 +146,19 @@ The `SAI_TAM_TEL_TYPE_ATTR_STATE` state machine (`STOP_STREAM` ↔ `CREATE_CONFI
 
 `HFTelProfile::getObjectType(tam_tel_type_obj)` returns the singleton in MIXED mode. `updateTemplates(tam_tel_type_obj)` queries `SAI_TAM_TEL_TYPE_ATTR_IPFIX_TEMPLATES` and stores the resulting buffer under the singleton key. `getTemplates(object_type)` returns that same buffer for every `object_type` requested by `HFTelOrch` when populating per-group STATE_DB session entries.
 
-This means the combined IPFIX template returned by the single tel_type is **replicated** into each `HIGH_FREQUENCY_TELEMETRY_SESSION|profile|group_name` entry. CounterSyncd's existing template-set parser handles both single-template and multi-template-set buffers, so no CounterSyncd change is needed.
+This means the combined IPFIX template returned by the single tel_type is **replicated** into each `HIGH_FREQUENCY_TELEMETRY_SESSION|profile|group_name` entry. CounterSyncd's existing template-set parser handles both single-template and multi-template-set buffers, so no parser change is needed. The replication does require an extension to CounterSyncd's label-resolution path, since multiple per-group sessions now share a template_id; see §7.6.1.
+
+### 7.6.1. CounterSyncd label resolution in MIXED
+
+CounterSyncd resolves IPFIX data-record fields to SAI counter identities by looking up each field's IPFIX element ID (the per-object label assigned by `HFTelProfile`) against the `object_names` list of the STATE_DB session that owns the template. In SINGLE mode each `sai_tam_tel_type` carries its own template_id, so there is exactly one session per template_id and the lookup is unambiguous.
+
+In MIXED mode the orchagent replicates the combined IPFIX template into every per-group `HIGH_FREQUENCY_TELEMETRY_SESSION_TABLE` entry (§7.6), so all per-group sessions of a profile share a template_id. A resolution that consults only one session per template_id would correctly resolve labels owned by that session, but labels owned by sibling sessions (for example a PORT label seen when the QUEUE session won CounterSyncd's internal `template_id → session` race) would fall back to `unknown_<label>`.
+
+CounterSyncd handles this with a per-session `session_template_ids` set that records every template_id a session registered. At per-record lookup time it unions the `object_id_name_map` entries of all sessions sharing the record's template_id and resolves the label against that union.
+
+The aggregation is unambiguous because in MIXED mode the orchagent allocates labels from a per-profile monotonic counter (`HFTelProfile::m_next_label`, see §12), so two sessions within the same profile never share a label. In SINGLE mode each template_id has exactly one registering session, the aggregation collapses to that session's lookup, and behavior is identical to the pre-MIXED resolution path.
+
+The CounterSyncd public interface — `HIGH_FREQUENCY_TELEMETRY_SESSION_TABLE` schema, IPFIX wire format, OpenTelemetry export — is unchanged. The extension is internal to CounterSyncd's IPFIX data-record processing path.
 
 ### 7.7. STATE_DB
 
@@ -329,7 +342,7 @@ Limitations introduced by this design:
 - The TAM tel_type mode is chosen at orchagent init from SAI capability and cannot be changed at runtime.
 - In MIXED mode all three `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_*_STATS` flags supported by SONiC HFT (PORT, MMU, OUTPUT_QUEUE) are enabled on the single tel_type even if a given profile only references a subset of object types. The streamed counters remain constrained by `sai_tam_counter_subscription` objects.
 - If the vendor SAI advertises neither `SINGLE_TYPE` nor `MIXED_TYPE` for `SAI_TAM_TEL_TYPE_ATTR_MODE`, HFT is disabled at orchagent init and a notice is logged.
-- In MIXED mode the per-profile IPFIX label allocator (`HFTelProfile::m_next_label`) is monotonic and never reuses values. Because the label field is 16-bit (`sai_uint16_t`), at most 65 535 distinct objects may be subscribed to a single profile over its lifetime; profiles approaching this limit must be deleted and recreated to reset the counter. The limit is per profile and per orchagent lifetime, so warm-restart or orchagent restart implicitly resets it.
+- In MIXED mode the per-profile IPFIX label allocator (`HFTelProfile::m_next_label`) is monotonic and never reuses values. Because the label field is 16-bit (`sai_uint16_t`), at most 65 535 distinct objects may be subscribed to a single profile over its lifetime; profiles approaching this limit must be deleted and recreated to reset the counter. The limit is per profile and per orchagent lifetime, so warm-restart or orchagent restart implicitly resets it. This monotonic allocation is what guarantees label uniqueness across all per-group sessions of a profile, on which CounterSyncd's label-resolution path in §7.6.1 depends.
 
 Vendor-specific limitations inherited from the underlying SAI implementation. These are not introduced by this design but are surfaced by it on platforms that only support MIXED_TYPE:
 
--- a/doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
+++ b/doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
@@ -340,7 +340,7 @@ In MIXED mode the orchagent holds one `sai_tam_tel_type` and one `sai_tam_report
 Limitations introduced by this design:
 
 - The TAM tel_type mode is chosen at orchagent init from SAI capability and cannot be changed at runtime.
-- In MIXED mode all three `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_*_STATS` flags supported by SONiC HFT (PORT, MMU, OUTPUT_QUEUE) are enabled on the single tel_type even if a given profile only references a subset of object types. The streamed counters remain constrained by `sai_tam_counter_subscription` objects.
+- In MIXED mode all three `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_*_STATS` flags (PORT, MMU, OUTPUT_QUEUE) are set on the single tel_type, even if the profile only uses a subset. Only counters with a matching `sai_tam_counter_subscription` are actually streamed.
 - If the vendor SAI advertises neither `SINGLE_TYPE` nor `MIXED_TYPE` for `SAI_TAM_TEL_TYPE_ATTR_MODE`, HFT is disabled at orchagent init and a notice is logged.
 - In MIXED mode the per-profile IPFIX label allocator (`HFTelProfile::m_next_label`) is monotonic and never reuses values. Because the label field is 16-bit (`sai_uint16_t`), at most 65 535 distinct objects may be subscribed to a single profile over its lifetime; profiles approaching this limit must be deleted and recreated to reset the counter. The limit is per profile and per orchagent lifetime, so warm-restart or orchagent restart implicitly resets it. This monotonic allocation is what guarantees label uniqueness across all per-group sessions of a profile, on which CounterSyncd's label-resolution path in §7.6.1 depends.
 
--- a/doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
+++ b/doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
@@ -75,7 +75,7 @@ Non-requirements:
 
 ## 6. Architecture Design
 
-The architecture diagram from the [base HLD §6](high-frequency-telemetry-hld.md#6-architecture-design) is unchanged. The bulk of the change is internal to `Orchagent → High frequency telemetry Orch`; the SAI/syncd boundary, the CounterSyncd public interface, the OpenTelemetry container, and the Redis databases are unaffected. CounterSyncd's label-resolution path is extended for the MIXED case where one template is shared across per-group sessions (see §7.6.1).
+The architecture diagram from the [base HLD §6](high-frequency-telemetry-hld.md#6-architecture-design) is unchanged. The bulk of the change is internal to `Orchagent → High frequency telemetry Orch`; the SAI/syncd boundary, the CounterSyncd public interface, the OpenTelemetry container, and the Redis databases are unaffected. CounterSyncd's label-resolution path is extended for the MIXED case where one set of templates is shared across per-group sessions (see §7.6.1).
 
 ## 7. High-Level Design
 
--- a/doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
+++ b/doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
@@ -232,9 +232,9 @@ sequenceDiagram
     end
     box OpenTelemetry container
         participant otel as OpenTelemetry Collector
-        participant counter as counter syncd
     end
     box SWSS container
+        participant counter as counter syncd
         participant port_orch as Port Orch
         participant hft_orch as High Frequency Telemetry Orch
     end
--- a/doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
+++ b/doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
@@ -1,0 +1,365 @@
+# High frequency telemetry - TAM telemetry type MIXED mode support <!-- omit in toc -->
+
+## Table of Content <!-- omit in toc -->
+
+- [1. Revision](#1-revision)
+- [2. Scope](#2-scope)
+- [3. Definitions/Abbreviations](#3-definitionsabbreviations)
+- [4. Overview](#4-overview)
+- [5. Requirements](#5-requirements)
+- [6. Architecture Design](#6-architecture-design)
+- [7. High-Level Design](#7-high-level-design)
+  - [7.1. Mode auto-detection in HFTelOrch](#71-mode-auto-detection-in-hftelorch)
+  - [7.2. HFTelProfile data structures](#72-hftelprofile-data-structures)
+  - [7.3. Creating the tam\_tel\_type object](#73-creating-the-tam_tel_type-object)
+  - [7.4. Collapsing the tam\_report object](#74-collapsing-the-tam_report-object)
+  - [7.5. State machine](#75-state-machine)
+  - [7.6. Config-ready notification and templates](#76-config-ready-notification-and-templates)
+  - [7.7. STATE\_DB](#77-state_db)
+  - [7.8. Work flow](#78-work-flow)
+- [8. SAI API](#8-sai-api)
+- [9. Configuration and management](#9-configuration-and-management)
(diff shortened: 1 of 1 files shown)


doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,378 @@
...
+
+| Abbreviation | Description                                                                                              |
+| ------------ | -------------------------------------------------------------------------------------------------------- |
+| SINGLE_TYPE  | `SAI_TAM_TEL_TYPE_MODE_SINGLE_TYPE`: one `sai_tam_tel_type` object binds counters of exactly one type.   |
+| MIXED_TYPE   | `SAI_TAM_TEL_TYPE_MODE_MIXED_TYPE`: one `sai_tam_tel_type` object binds counters of multiple types.      |
+| HFT          | High frequency telemetry.                                                                                |
+
+## 4. Overview
+
+The current SONiC HFT implementation hardcodes `SAI_TAM_TEL_TYPE_ATTR_MODE = SAI_TAM_TEL_TYPE_MODE_SINGLE_TYPE` when creating `sai_tam_tel_type` objects. Because a SINGLE_TYPE tel_type can bind only one counter category, the orchagent creates one tel_type per SAI object type (`SAI_OBJECT_TYPE_PORT`, `SAI_OBJECT_TYPE_BUFFER_POOL`, `SAI_OBJECT_TYPE_QUEUE`, `SAI_OBJECT_TYPE_INGRESS_PRIORITY_GROUP`) per profile.
```
Pterosaur: I think it's not true here. We allow multiple tel_type per profile under the single_type mode.
The key difference between single mode and mixed mode is whether multiple SAI object type can be bound to ONE tel_type object.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,378 @@
...
+
+Non-requirements:
+
+- No CONFIG_DB, YANG, CLI, or STATE_DB schema changes.
+- No CounterSyncd schema or public-interface changes. One internal extension to the label-resolution path is required so that the per-group sessions sharing a template_id in MIXED resolve correctly (see §7.6.1).
+- No runtime mode switching - the chosen mode is fixed for the lifetime of the orchagent process.
+
+## 6. Architecture Design
+
+The architecture diagram from the [base HLD §6](high-frequency-telemetry-hld.md#6-architecture-design) is unchanged. The bulk of the change is internal to `Orchagent → High frequency telemetry Orch`; the SAI/syncd boundary, the CounterSyncd public interface, the OpenTelemetry container, and the Redis databases are unaffected. CounterSyncd's internal label-resolution path is extended to handle the MIXED-mode case where multiple per-group sessions share a template_id (§7.6.1).
```
Pterosaur: why multiple per-group sessions share a template_id ?

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,378 @@
...
+Selection rules:
+
+| Advertised values                | Selected mode | HFT enabled |
+| -------------------------------- | ------------- | ----------- |
+| `SINGLE_TYPE`                    | SINGLE_TYPE   | yes         |
+| `SINGLE_TYPE` and `MIXED_TYPE`   | SINGLE_TYPE   | yes         |
+| `MIXED_TYPE` only                | MIXED_TYPE    | yes         |
+| neither                          | -             | no (logged) |
+
+SINGLE_TYPE is preferred when both are advertised so that the behavior of all existing platforms is unchanged. This is consistent with the SAI specification, which declares `SAI_TAM_TEL_TYPE_MODE_SINGLE_TYPE` as the default value of `SAI_TAM_TEL_TYPE_ATTR_MODE`.
```
Pterosaur, line 95: IMHO, MIXED_TYPE can be the default behavior in the Orchagent, since mixed_type mode is more efficient and uses fewer states and SAI objects.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,378 @@
...
+```mermaid
+sequenceDiagram
+    autonumber
+    box Redis
+        participant config_db as CONFIG_DB
+        participant state_db as STATE_DB
+    end
+    box OpenTelemetry container
+        participant otel as OpenTelemetry Collector
+        participant counter as counter syncd
```
Pterosaur: counter syncd is in the swss container, not open telemetry container.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,378 @@
...
+## 11. Memory Consumption
+
+In MIXED mode the orchagent holds one `sai_tam_tel_type` and one `sai_tam_report` per profile instead of up to four of each. The number of `sai_tam_counter_subscription` objects, the IPFIX template buffer size (per profile total), and CounterSyncd buffering are unchanged. The dynamic memory formula in the base HLD §8.5 still applies.
+
+## 12. Restrictions/Limitations
+
+Limitations introduced by this design:
+
+- The TAM tel_type mode is chosen at orchagent init from SAI capability and cannot be changed at runtime.
+- In MIXED mode all three `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_*_STATS` flags supported by SONiC HFT (PORT, MMU, OUTPUT_QUEUE) are enabled on the single tel_type even if a given profile only references a subset of object types. The streamed counters remain constrained by `sai_tam_counter_subscription` objects.
```
Pterosaur: What do you mean by “The streamed counters remain constrained by `sai_tam_counter_subscription` objects” ？

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,378 @@
...
+
+### 7.6.1. CounterSyncd label resolution in MIXED
+
+CounterSyncd resolves IPFIX data-record fields to SAI counter identities by looking up each field's IPFIX element ID (the per-object label assigned by `HFTelProfile`) against the `object_names` list of the STATE_DB session that owns the template. In SINGLE mode each `sai_tam_tel_type` carries its own template_id, so there is exactly one session per template_id and the lookup is unambiguous.
+
+In MIXED mode the orchagent replicates the combined IPFIX template into every per-group `HIGH_FREQUENCY_TELEMETRY_SESSION_TABLE` entry (§7.6), so all per-group sessions of a profile share a template_id. A resolution that consults only one session per template_id would correctly resolve labels owned by that session, but labels owned by sibling sessions (for example a PORT label seen when the QUEUE session won CounterSyncd's internal `template_id → session` race) would fall back to `unknown_<label>`.
```
Pterosaur, line 155:
```suggestion
In MIXED mode the orchagent replicates the combined IPFIX template into every per-group `HIGH_FREQUENCY_TELEMETRY_SESSION_TABLE` entry (§7.6), so all per-group sessions of a profile share a set of template_id. A resolution that consults only one session per template_id would correctly resolve labels owned by that session, but labels owned by sibling sessions (for example a PORT label seen when the QUEUE session won CounterSyncd's internal `template_id → session` race) would fall back to `unknown_<label>`.
```
 A session may correspond to multiple IPFIX templates.

Also, we need to support dynamic updates, so a session may temporarily have two sets of templates during the transition period. This is similar to a two-step commit: the old template ID set will be deleted only after the data plane, meaning the telemetry data, starts using the new template ID set.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,378 @@
...
+
+CounterSyncd handles this with a per-session `session_template_ids` set that records every template_id a session registered. At per-record lookup time it unions the `object_id_name_map` entries of all sessions sharing the record's template_id and resolves the label against that union.
+
+The aggregation is unambiguous because in MIXED mode the orchagent allocates labels from a per-profile monotonic counter (`HFTelProfile::m_next_label`, see §12), so two sessions within the same profile never share a label. In SINGLE mode each template_id has exactly one registering session, the aggregation collapses to that session's lookup, and behavior is identical to the pre-MIXED resolution path.
```
Pterosaur: Could you please clarify what m_next_label is and how we should use it? I don’t fully understand it.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: I have no objection to using MIXED_TYPE as default. I was just not sure there will be no issues with other vendors with changing that behavior. If you think there shouldn't be an issue with changing the behavior I will agree.

Pterosaur: I think it's not an issue. I can discuss it with other vendors.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: It just means that we are setting all as true in HFTelProfile::getTAMTelTypeObjID unlike in SINGLE mode.
But this sentence is confusing, and maybe just unnecessary

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: Multiple per-group SESSION entries advertise the same template_id because in MIXED mode the orchagent creates a single tel_type per profile, queries its IPFIX template once, and then writes the same template into every HIGH_FREQUENCY_TELEMETRY_SESSION_TABLE|<profile>|<group_name> entry (PORT, QUEUE, …).

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: Fixed

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: Multiple per-group SESSION entries advertise the same template_id because in MIXED mode the orchagent creates a single tel_type per profile, queries its IPFIX template once, and then writes the same template into every HIGH_FREQUENCY_TELEMETRY_SESSION_TABLE|<profile>|<group_name> entry (PORT, QUEUE, …). The mechanism is §7.6; the resulting label-resolution wrinkle and its fix are §7.6.1.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: dynamic updates is a general countersyncd improvement, not something that supported today and is not specific to MIXED. 
The current replace behavior is unchanged by this design, I am not sure it should be part of this design

Pterosaur:
Hi @DavidZagury ,  It looks like the local commit hasn’t been pushed yet.

DavidZagury:
> Hi @DavidZagury , It looks like the local commit hasn’t been pushed yet.

@Pterosaur pushed

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
Pterosaur: My comment is about two things:

1. We need to use multiple IPFIX templates instead of a single template_id, so this description must be corrected.

3. I agree that dynamic update do not need to be included in this PR. However, our design should not prevent it in the future. I’m wondering whether we should duplicate templates for each type entry, because in mixed mode, dynamic updates may require an atomic update to the template set of a session.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
Pterosaur: Ok, Got it. Thanks.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
Pterosaur: Could you please update the default mode to mixed_type in this document?

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
Pterosaur: Could we emit only one combined session into State DB in mixed mode, such as HIGH_FREQUENCY_TELEMETRY_SESSION_TABLE|profile? Do you have any concerns or suggestions? I think this would make everything simpler.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,382 @@
...
+
+Non-requirements:
+
+- No CONFIG_DB, YANG, CLI, or STATE_DB schema changes.
+- No CounterSyncd schema or public-interface changes. One internal extension to the label-resolution path is required so that the per-group sessions sharing a template_id in MIXED resolve correctly (see §7.6.1).
+- No runtime mode switching - the chosen mode is fixed for the lifetime of the orchagent process.
+
+## 6. Architecture Design
+
+The architecture diagram from the [base HLD §6](high-frequency-telemetry-hld.md#6-architecture-design) is unchanged. The bulk of the change is internal to `Orchagent → High frequency telemetry Orch`; the SAI/syncd boundary, the CounterSyncd public interface, the OpenTelemetry container, and the Redis databases are unaffected. CounterSyncd's label-resolution path is extended for the MIXED case where one template is shared across per-group sessions (see §7.6.1).
```
Pterosaur:
```suggestion
The architecture diagram from the [base HLD §6](high-frequency-telemetry-hld.md#6-architecture-design) is unchanged. The bulk of the change is internal to `Orchagent → High frequency telemetry Orch`; the SAI/syncd boundary, the CounterSyncd public interface, the OpenTelemetry container, and the Redis databases are unaffected. CounterSyncd's label-resolution path is extended for the MIXED case where one set of templates is shared across per-group sessions (see §7.6.1).
```

 Same question: how about combining all per-group sessions into one profile session?

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: While there are upsides to use a combined STATE_DB profile key, but the original thought was that keeping the same STATE_DB key shape in SINGLE and MIXED keeps CounterSyncd, the inspect CLI, and any external consumers mode-agnostic — no per-read branching on m_tel_type_mode, which makes its changes simpler.

Creating the two options might also create a need to duplicate the VS tests or UT to cover both branches.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
r12f: I would also vote for mixed_type, because it is much better. If mixed_type is not supported, we can always fallback to single type anyway.

abelamit (dismissed):

liat-grozovik:
@r12f @Pterosaur please let us know if something is still required or we can go a head and merge.
please also review code PRs.

Pterosaur:
Hi @liat-grozovik , This comment still needs to be addressed.: https://github.com/sonic-net/SONiC/pull/2379/changes#r3396236077
Other than that, the PR looks good to me.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: @Pterosaur updated

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,383 @@
...
+Selection rules:
+
+| Advertised values                | Selected mode | HFT enabled |
+| -------------------------------- | ------------- | ----------- |
+| `SINGLE_TYPE`                    | SINGLE_TYPE   | yes         |
+| `SINGLE_TYPE` and `MIXED_TYPE`   | MIXED_TYPE    | yes         |
+| `MIXED_TYPE` only                | MIXED_TYPE    | yes         |
+| neither                          | -             | no (logged) |
+
+MIXED_TYPE is preferred whenever it is advertised, including on platforms that also advertise SINGLE_TYPE. In MIXED mode the orchagent holds one `sai_tam_tel_type` and one `sai_tam_report` per profile instead of up to four of each (§11), so preferring MIXED whenever the vendor supports it minimizes SAI object count and memory footprint even on platforms where SINGLE_TYPE would also work. This departs from the SAI specification's default of `SAI_TAM_TEL_TYPE_MODE_SINGLE_TYPE` for `SAI_TAM_TEL_TYPE_ATTR_MODE`; the default only applies when a platform does not explicitly select a mode, and the orchagent always makes an explicit choice via `SAI_TAM_TEL_TYPE_ATTR_MODE` at create time.
```
prvattem, line 96: This contradicts with the overview section 
`SINGLE_TYPE remains the default whenever the vendor advertises it, preserving today's behavior for existing platforms; MIXED_TYPE is chosen only when SINGLE_TYPE is not advertised.`

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,383 @@
...
+
+### 13.2. System Test cases
+
+Implemented in `sonic-swss/tests/` using DVS:
+
+- Run the existing HFT DVS tests against a virtual switch that simulates a MIXED-only SAI capability. Verify:
+  - Counters reach CounterSyncd through the genetlink path.
+  - STATE_DB session entries are populated for each configured group with non-empty `session_config`, `object_ids`, and `object_names`.
+  - The inspect-stream CLI returns the expected counters per group without modification.
+- Run the same tests against a virtual switch advertising both modes. Verify that SINGLE_TYPE is selected and existing behavior is preserved.
```
prvattem, line 379: This again contradicts with section 7.1 - in this test, you should test that MIXED_TYPE is selected?

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,383 @@
...
+
+- **MIXED mode**: on the first call, create a single tel_type with:
+  - `SAI_TAM_TEL_TYPE_ATTR_TAM_TELEMETRY_TYPE = SAI_TAM_TELEMETRY_TYPE_COUNTER_SUBSCRIPTION`
+  - `SAI_TAM_TEL_TYPE_ATTR_MODE = SAI_TAM_TEL_TYPE_MODE_MIXED_TYPE`
+  - **All** applicable enable attributes set to `true`: `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_PORT_STATS`, `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_MMU_STATS`, `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_OUTPUT_QUEUE_STATS`.
+  - `SAI_TAM_TEL_TYPE_ATTR_REPORT_ID` pointing to the single report object created in §7.4.
+
+  Subsequent calls return the cached tel_type oid regardless of the `object_type` argument. The tel_type is added to `SAI_TAM_TELEMETRY_ATTR_TAM_TYPE_LIST` exactly once.
+
+Setting all enable attributes (rather than only those required by the current profile) is acceptable because the actual counters streamed are constrained by the `sai_tam_counter_subscription` objects bound to the tel_type. Enabling categories that no subscription references has no functional effect.
```
prvattem, line 130: Since MIXED mode uses a single shared tel_type, please define how add/update/delete operations affect existing streaming. Is the profile rebuilt? Is streaming interrupted? Are templates regenerated?

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,383 @@
...
+
+`getTAMTelTypeObjID(object_type)` becomes mode-aware:
+
+- **SINGLE mode** (unchanged): create one tel_type per object type, with exactly one of `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_PORT_STATS`, `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_MMU_STATS`, or `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_OUTPUT_QUEUE_STATS` set to `true` according to the object type. `SAI_TAM_TEL_TYPE_ATTR_MODE = SAI_TAM_TEL_TYPE_MODE_SINGLE_TYPE`.
+
+- **MIXED mode**: on the first call, create a single tel_type with:
+  - `SAI_TAM_TEL_TYPE_ATTR_TAM_TELEMETRY_TYPE = SAI_TAM_TELEMETRY_TYPE_COUNTER_SUBSCRIPTION`
+  - `SAI_TAM_TEL_TYPE_ATTR_MODE = SAI_TAM_TEL_TYPE_MODE_MIXED_TYPE`
+  - **All** applicable enable attributes set to `true`: `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_PORT_STATS`, `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_MMU_STATS`, `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_OUTPUT_QUEUE_STATS`.
+  - `SAI_TAM_TEL_TYPE_ATTR_REPORT_ID` pointing to the single report object created in §7.4.
```
prvattem, line 126: Does support for MIXED_TYPE imply support for every SWITCH_ENABLE_* attribute SONiC enables? If not, capability validation may be required.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
```diff
@@ -0,0 +1,383 @@
...
+
+Implemented in `sonic-swss/tests/mock_tests/`:
+
+- Mode selection: mock `sai_query_attribute_enum_values_capability` to return (a) SINGLE only, (b) MIXED only, (c) both, (d) neither. Assert that `HFTelOrch` selects SINGLE in (a), MIXED in (b) and (c), and disables HFT in (d).
+- MIXED mode SAI calls: assert exactly one `create_tam_tel_type` and one `create_tam_report` call per profile, with `SAI_TAM_TEL_TYPE_ATTR_MODE = SAI_TAM_TEL_TYPE_MODE_MIXED_TYPE` and all three `SAI_TAM_TEL_TYPE_ATTR_SWITCH_ENABLE_*_STATS` set to `true`.
+- MIXED mode state machine: assert that `STOP_STREAM → CREATE_CONFIG` is not issued until every configured object type is ready, and is issued exactly once when the last one becomes ready.
+- MIXED mode template propagation: with two groups configured (e.g. PORT and QUEUE), assert that the same combined IPFIX template buffer is written to both `HIGH_FREQUENCY_TELEMETRY_SESSION|profile|PORT` and `HIGH_FREQUENCY_TELEMETRY_SESSION|profile|QUEUE` entries.
+- SINGLE mode regression: the existing tests continue to pass without modification.
+
+### 13.2. System Test cases
```
prvattem, line 371: Please add tests for add/remove/update of groups and objects while streaming is active.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: Fixed - §4 now matches §7.1's selection rule: MIXED_TYPE is preferred whenever advertised, SINGLE_TYPE only when MIXED_TYPE isn't. Pushed.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: Fixed - this test now expects MIXED_TYPE to be selected, matching §7.1. Pushed.

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: Added §7.5 'Impact of configuration changes during streaming': any add/remove/update to a group stops the affected tel_type via setStreamState(object_type, STOP_STREAM) before applying the change. In SINGLE mode mapKey() is identity, so only the mutated object type's tel_type stops. In MIXED mode mapKey() collapses every object type onto the shared tel_type, so the whole profile stops - not just the mutated group. Resuming re-walks STOP_STREAM -> CREATE_CONFIG -> START_STREAM, gated on every group being ready, and re-queries a fresh combined template. Pushed, with matching mock-test coverage (see reply below on the testing comment).

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: Good catch - no, it didn't imply that. Added validation: HFTelOrch::querySupportedTelTypeModes now probes each SWITCH_ENABLE_PORT_STATS/MMU_STATS/OUTPUT_QUEUE_STATS attribute independently via sai_query_attribute_capability, in both SINGLE and MIXED (not just MIXED - a vendor could support SINGLE_TYPE mode but still be missing one of these attributes). getTAMTelTypeObjID only enables supported categories on the shared MIXED tel_type, and groupTableSet rejects (task_failed, logged) a group whose category isn't supported, in either mode, before any counter subscription/tel_type-creation is attempted. If a vendor supports none of the three, HFT is disabled entirely rather than silently selecting an unusable mode. HLD updated (§5/§7.1/§7.3/§12/§8/§13.1) and code+mock tests pushed on the implementation PR (sonic-swss#4732).

doc/high-frequency-telemetry/high-frequency-telemetry-mixed-mode-hld.md
DavidZagury: Added: hftelprofile_ut.cpp now has a StreamingMutationTest fixture covering exactly this - with a profile actively streaming (START_STREAM), updating a group's stats, adding an object, and removing an object each get asserted to stop streaming for the whole profile in MIXED mode (shared tel_type) vs. only the mutated group's own tel_type in SINGLE mode. §13.1 updated to describe this. Pushed on sonic-swss#4732.

michaeloso:
link to recording - https://zoom.us/rec/share/xfVWJpdwF6_y9ZjC4XqCliAdl8nK9tRE1FQ-zVKgcoBIk6ADrqbXOEqhj6WmhYF_.Nd-UyDxc8W2_CKNI?startTime=1788879363000

PR #2387, opened 2026-06-12
Add vpp vlan-bvi HLD

yue-fred-gao:
### why
VLAN BVI is not supported today and an issue has been created: https://github.com/sonic-net/sonic-buildimage/issues/26936

### what this PR does
Create an HLD to document how we plan to support ARP, DHCP and unicast packet to VLAN.

### DIFF
--- a/doc/vpp/vlan-bvi-hld.md
+++ b/doc/vpp/vlan-bvi-hld.md
@@ -1,0 +1,864 @@
+# VLAN Bridge Domain with BVI — High-Level Design
+
+## Revision History
+
+| Rev | Date | Author | Summary |
+|-----|------|--------|---------|
+| 1.0 | 2026-04 | yuega2 | Initial design.  ARP/DHCP punted via the `l2-input-classify` clone-on-hit VPP patch (0010); IP unicast and ARP redirected via sonic_ext plugin features on the `ip4-punt`/`ip6-punt`/`arp` arcs; per-buffer state held in a global `orig_rx_by_bi[]` sidecar keyed by buffer index. |
+| 1.1 | 2026-06 | yuega2 | Re-architected control-plane punt around the `sonic_ext` out-of-tree VPP plugin. Removed all dependence on classifier clone-on-hit for ARP/DHCP. Per-buffer state moved into the `vnet_buffer2(b)->unused` overlay (`sonic_ext_buffer_opaque_t`) so it survives `vlib_buffer_clone()` automatically with no atomic ops or worker handoff concerns. ARP and DHCP now reach the kernel via the `l2-flood -> linux-cp-punt-xc` path with a BVI-side feature node, then through `linux-cp-punt -> aggr-tap-redirect` on the BVI host tap, which restores the original member tap and re-pushes the wire VLAN tag from the cookie. The classifier now handles only LLDP / LACP (always redirect-punt). |
+| 1.2 | TBD | — | Update §5 packet-flow walkthroughs to match v1.1 architecture. |
+
+---
+
+## 1. Problem Statement
+
+SONiC on VPP needs to support L2 bridging with L3 SVI (VLAN interfaces) for both
+tagged and untagged VLAN members. The standard SONiC data model creates:
+
+- A VLAN (e.g., Vlan10)
+- An SVI with an IP address (e.g., 10.0.0.1/24 on Vlan10)
+- Member ports with tagging mode (tagged or untagged)
+
+VPP's bridge-domain (BD) and BVI (Bridge Virtual Interface) constructs map to this
+model, but require explicit wiring for two distinct traffic classes:
+
+1. **Control-plane protocols (ARP, LLDP, LACP, DHCP)** — must be delivered to
+   Linux on the **member interface tap** (the LCP host tap of the wire phy /
+   sub-if that received the frame), so per-port services (LLDP/LACP agents,
+   DHCP relay) see frames with the right ifindex.
+
+2. **L3 traffic over the VLAN SVI IP** — IP unicast destined to the SVI address
+   (e.g., ping, SSH, routing protocols) is also delivered to the Linux
+   control plane via the **member interface tap**, not via a BVI tap. The BVI
+   exists only to give the BD an L3 endpoint inside VPP for routing/ARP
+   resolution; no packets are punted through the BVI to Linux.
+
+In summary, SONiC doesn't expect NPU to punt packets to virtual netdev, such as Vlan,
+Port Channel directly. The packets should be punted via member interface then delivered
+to the virtual netdev by linux kernel.
+
+### Challenges
+
+1. **No BVI → Linux punt path**: VPP's BVI is purely an internal L3 endpoint
+   for the BD. The control plane must instead see all punted traffic on the
+   originating member interface, so Linux observes ports the same way it
+   would on physical hardware.
+
+2. **Per-port punt for control protocols**: ARP, LLDP, LACP, and DHCP must be
+   punted directly to the LCP host tap of the wire ingress port so Linux
+   daemons (lldpd, teamd, dhcrelay) receive frames on the correct netdev.
+
+3. **Original wire ingress is lost by L2 bridging**: ARP and DHCP broadcasts
+   reach the BVI via `l2-flood`, by which point `vnet_buffer(b)->sw_if_index[VLIB_RX]`
+   has been overwritten with the BVI sw_if. The redirect path needs to know
+   which member the frame entered VPP on. We carry this in a per-buffer
+   cookie stamped at `device-input` (see §3.7).
+
+4. **Wire VLAN tag is consumed by the sub-interface**: a tagged frame's
+   802.1Q header is stripped by `ethernet-input` (sub-if dispatch) or by L2
+   VTR pop-1, before any redirect node runs.  The kernel still needs to see
+   the original wire frame on the member tap, so the tag must be re-pushed
+   on egress (see §3.7).
+
+5. **Tagged member dispatch**: VPP must create dot1q sub-interfaces for tagged
+   members, strip the tag on ingress to the BD, and push it back on egress.
+
+6. **Untagged member handling**: The physical interface itself joins the BD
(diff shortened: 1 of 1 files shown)


doc/vpp/vlan-bvi-hld.md
```diff
@@ -0,0 +1,864 @@
...
+  → bobm1 → ethernet-input → bobm1
+    → l2-input (BD 10)
+      → l2-input-classify: etype=0x88CC → "other" table
+        → untag_other: etype=0x88CC → HIT, action=NONE
+          → punt (consume): → linux-cp-punt → Ethernet4 tap
+            → kernel lldpd/teamd processes the frame
+```
+
+
+### 5.4 DHCP Broadcast from a BD Member
```
lolyu: **Major — state the relay-vs-direct-L2 DHCP model explicitly.** This trace shows a DISCOVER being *both* flooded to sibling BD members (`bobm1-tx`, `bobm0-tx`) **and** punted to Linux on the originating member tap. That means if a DHCP **server** is attached to another member of the same VLAN, the client gets two service paths at once: the directly-flooded L2 frame to the in-VLAN server, **and** whatever `dhcrelay` generates from the punted copy. The HLD should say explicitly which model SONiC-VPP targets — pure-L2 (server in the same VLAN/broadcast domain) vs. relay (server L3-remote) — and whether simultaneous double-servicing can occur. This is the single most important clarification for anyone implementing or testing DHCP on a VLAN BVI.

lolyu (commented):
## VLAN BVI HLD — Review (Comment)

Strong design overall. The v1.0->v1.1 re-architecture is a real improvement: moving per-buffer state into the `vnet_buffer2(b)->unused` overlay so it survives `vlib_buffer_clone()` is elegant and removes the v1.0 atomic-swap recycled-buffer race. The "no upstream VPP source modified" out-of-tree `sonic_ext` plugin approach is the right long-term call, and the per-protocol punt table (S3.1) is an excellent reference artifact.

Not blocking — posting as a **Comment**. The substantive items are clarifications I'd like resolved before this becomes the reference doc people implement against. Details inline; summary:

**Major (clarify):**
1. **S5.4 — relay vs. direct-L2-server model is never stated.** A DHCP DISCOVER is both flooded to sibling members *and* punted to Linux. If a DHCP server lives on another member of the same VLAN, the client can be serviced via two paths (direct L2 flood + relay on the punted member tap). The HLD should state explicitly whether intra-VLAN DHCP is expected to work via pure L2 (in-VLAN server) or always via relay (L3-remote server), and whether double-servicing is possible.
2. **DHCP snooping interaction is unaddressed.** With snooping on, the member port facing a DHCP server must be trusted or OFFER/ACK are dropped. Worth a paragraph.
3. **DHCPv6 is silent.** ARP/DHCPv4 get full treatment, IPv6 ND is deferred to S7, but DHCPv6 isn't mentioned. It's link-scoped multicast (`ff02::1:2`), so the `dst==255.255.255.255` gate in `bcast-redirect` won't catch it — the design silently doesn't handle it today. State it as out-of-scope explicitly.

**Minor:**
4. Rev 1.2 is "TBD" — S5 walkthroughs flagged as still pending v1.1 reconciliation; clarify what's authoritative.
5. S6 `show lcp` example contradicts S3.6 (`lcp-auto-subint` disabled => no `Ethernet0.10` netdev).
6. `bcast-redirect` punts *any* limited broadcast with no L4/port match — note rate-limit/policer expectations (CPU-DoS surface).

Full reasoning on each is attached to the relevant lines.

doc/vpp/vlan-bvi-hld.md
```diff
@@ -0,0 +1,864 @@
...
+                          aggr-tap 25 orig-rx 3 member-tap 19
+                          vlan-pushed vid 10 tpid 0x8100 → REDIRECTED
+                        → member-host-tap (tap4103) interface-output
+                          → tap4103-tx  (DHCP, 802.1q vlan 10 restored)
+                            → kernel sees DHCP on Ethernet0.10
+```
+
+Key observations:
+
+- **Member flood copies are not punted** — they just go out the wire on
```
lolyu, line 762: **Major — DHCP snooping interaction is unaddressed.** If DHCP snooping is enabled on the VLAN, the member port facing a DHCP server must be marked *trusted*, or server-originated OFFER/ACK arriving on an untrusted member port will be dropped. Since OFFER/ACK return either as broadcast or unicast back through the BD, please document how snooping trust state composes with the punt-via-member redirect path (does the redirect run before or after snooping enforcement?).

doc/vpp/vlan-bvi-hld.md
```diff
@@ -0,0 +1,864 @@
...
+  bcast punts     : <counter>
+
+# Promiscuous mode
+vppctl show interface bobm0
+  flags: ... promisc ...
+```
+
+---
+
+## 7. IPv6 Neighbor Discovery (Future Work)
```
lolyu, line 979: **Major — DHCPv6 is silent (not just ND).** Section 7 defers IPv6 *Neighbor Discovery*, but DHCPv6 is never mentioned. DHCPv6 relay/forward uses link-scoped multicast `ff02::1:2`, not a limited IPv4 broadcast, so the `dst==255.255.255.255` gate in `sonic-ext-bcast-redirect` (S3.7.2) will never match it — the current design silently does not punt DHCPv6. Please state DHCPv6 as explicitly out-of-scope here (alongside ND), or describe the intended multicast-punt mechanism, so readers don't assume v4 coverage implies v6.

doc/vpp/vlan-bvi-hld.md
```diff
@@ -0,0 +1,864 @@
...
+
+## Revision History
+
+| Rev | Date | Author | Summary |
+|-----|------|--------|---------|
+| 1.0 | 2026-04 | yuega2 | Initial design.  ARP/DHCP punted via the `l2-input-classify` clone-on-hit VPP patch (0010); IP unicast and ARP redirected via sonic_ext plugin features on the `ip4-punt`/`ip6-punt`/`arp` arcs; per-buffer state held in a global `orig_rx_by_bi[]` sidecar keyed by buffer index. |
+| 1.1 | 2026-06 | yuega2 | Re-architected control-plane punt around the `sonic_ext` out-of-tree VPP plugin. Removed all dependence on classifier clone-on-hit for ARP/DHCP. Per-buffer state moved into the `vnet_buffer2(b)->unused` overlay (`sonic_ext_buffer_opaque_t`) so it survives `vlib_buffer_clone()` automatically with no atomic ops or worker handoff concerns. ARP and DHCP now reach the kernel via the `l2-flood -> linux-cp-punt-xc` path with a BVI-side feature node, then through `linux-cp-punt -> aggr-tap-redirect` on the BVI host tap, which restores the original member tap and re-pushes the wire VLAN tag from the cookie. The classifier now handles only LLDP / LACP (always redirect-punt). |
+| 1.2 | TBD | — | Update §5 packet-flow walkthroughs to match v1.1 architecture. |
```
lolyu: **Minor — Rev 1.2 is `TBD` and flags S5 as still needing v1.1 reconciliation.** Some S5 traces already look v1.1-correct (e.g. S5.5 references `ip4-punt-redirect -> ip4-dvr-dpo`), but this row tells reviewers the S5 walkthroughs aren't fully updated. Either complete the S5 pass and promote 1.2, or drop the placeholder so it's clear what's authoritative.

doc/vpp/vlan-bvi-hld.md
```diff
@@ -0,0 +1,864 @@
...
+  BD-ID 10, flood, learn
+    bvi10        (BVI, sw_if_index 25)
+    bobm0.10     (tagged member, sw_if_index 23, vtr pop-1)
+    bobm1        (untagged member, sw_if_index 3)
+
+# LCP pairs
+vppctl show lcp
+  bvi10     → tap_Vlan10
+  bobm0     → Ethernet0      (physical)
+  bobm0.10  → Ethernet0.10   (auto-subint)
```
lolyu: **Minor — this example contradicts S3.6.** S3.6 mandates `lcp-auto-subint` **off** and states no Linux `Ethernet0.10` netdev is created. But this `show lcp` output lists `bobm0.10 -> Ethernet0.10   (auto-subint)`. One of them is wrong — fix this example to reflect the disabled-auto-subint reality (parent `Ethernet0` only), otherwise it misleads implementers about what netdevs exist.

doc/vpp/vlan-bvi-hld.md
```diff
@@ -0,0 +1,864 @@
...
+| `sonic-ext-host-xc` | `device-input` | runs before `ethernet-input`, enabled on host taps | When a Linux reply re-enters VPP on a host tap, jump straight to the paired phy's `interface-output` (skip `ethernet-input` re-parse / sub-if dispatch). |
+| `sonic-ext-bcast-redirect` | `ip4-unicast` | runs before `ip4-lookup`, enabled on every BVI that has an LCP pair (i.e. every L3 SVI BVI).  BVIs created purely as the L2 endpoint of a VXLAN tunnel termination have no LCP pair and are therefore **not** instrumented by this node — limited broadcasts on those BVIs continue down the regular L2/VXLAN forwarding path and can be encapsulated to the remote VTEP. | Gate on cookie-magic + `dst==255.255.255.255`.  No protocol or port match: any limited-broadcast IPv4 frame carrying a valid cookie is dispatched to `linux-cp-punt` instead of `ip4-lookup`.  Today this is the entry point for DHCP broadcasts that have been L2-flooded into the BVI; the same node will serve future non-BVI broadcast redirect use cases without code changes. |
```
lolyu: **Minor — unbounded broadcast punt is a CPU-DoS surface.** `sonic-ext-bcast-redirect` matches on cookie + `dst==255.255.255.255` with *no protocol or L4/port filter*, so every limited-broadcast IPv4 frame on the BVI is punted to `linux-cp-punt`. The doc frames this generality as a feature (NetBIOS/custom UDP), but punting all 255.255.255.255 traffic to the control plane invites a flood-to-CPU. Add a sentence on the expected policer/rate-limit on the punt path.

dypet (approved):

lolyu (approved):
LGTM

PR #2388, opened 2026-06-14
Add HLD for Console Mirror Feature

V2lsbGlhbVpoYW5n:
The console mirror feature enables on-demand recording of serial console traffic for a specific console line. It mirrors RX and TX data streams observed by the per-line console proxy to secure local recording files and can optionally display the mirrored stream live to the mirror user. This helps diagnose serial console issues without taking over or interrupting the active console session.

 | Description | Repo | PR title | State |
|-------------|------|----------|-------|
| Implement design in sonic-host-services | sonic-host-services | [Add console mirror recording and control support](https://github.com/sonic-net/sonic-host-services/pull/414) | Merged |
| Support CLI commands in sonic-utilities | sonic-utilities | [Add console mirror CLI commands](https://github.com/sonic-net/sonic-utilities/pull/4697) | Merged |
| Publish test plan in sonic-mgmt | sonic-mgmt | [Add Console Mirror test plan](https://github.com/sonic-net/sonic-mgmt/pull/26629) | Merged |
| Implement test in sonic-mgmt | sonic-mgmt | [Add Console Mirror test cases](https://github.com/sonic-net/sonic-mgmt/pull/27003) | Merged |

### DIFF
--- a/doc/console/Console-Mirror-High-Level-Design.md
+++ b/doc/console/Console-Mirror-High-Level-Design.md
@@ -111,38 +111,44 @@ The design extends the existing console monitor DCE architecture. The per-line p
 
 ```mermaid
 flowchart LR
-    userA["User A<br/>consutil connect 1"]
-    pts["PTS<br/>/dev/ttyUSB1-PTS"]
-    ptm["PTM<br/>/dev/ttyUSB1-PTM"]
-    proxy["Per-line proxy<br/>console-monitor-proxy@1"]
-    tty["Physical TTY<br/>/dev/ttyUSB1"]
     target["Managed device<br/>serial console"]
+    tty["Physical TTY<br/>/dev/ttyUSB1"]
 
-    userA <-- "interactive console" --> pts
-    pts <-- "PTY bridge" --> ptm
-    ptm <-- "normal TX/RX stream" --> proxy
-    proxy <-- "serial stream" --> tty
-    tty <-- "physical serial" --> target
-
-    subgraph MirrorPath["Mirror side path inside DCE"]
+    subgraph proxyBox["Per-line proxy process (console-monitor-proxy@1)"]
+        core["Existing proxy forwarding logic"]
         mirror["MirrorManager"]
         writer["RecordingWriter"]
-        log["Secure text logs<br/>/var/log/sonic/console-mirror/..."]
         state["STATE_DB<br/>CONSOLE_MIRROR|1"]
+        control["MirrorControlServer<br/>/run/console-monitor/mirror/line1.sock"]
+        display["DisplaySink"]
+        log["Text logs<br/>/var/log/sonic/console-mirror/..."]
     end
 
-    proxy -. "best-effort RX/TX copy" .-> mirror
+    ptm["PTM<br/>/dev/ttyUSB1-PTM"]
+    pts["PTS<br/>/dev/ttyUSB1-PTS"]
+    userA["User A<br/>consutil connect 1"]
+
+    
+
+    target <-- "physical serial" --> tty
+    tty <-- "serial stream" --> core
+
+    core <-- "normal TX/RX stream" --> ptm
+    ptm <-- "PTY bridge" --> pts
+    pts <-- "interactive console" --> userA
+
+    core -->|best-effort RX/TX copy| mirror
+
     mirror --> writer
     writer --> log
     mirror --> state
+    mirror --> display
 
-    userB["User B<br/>sudo consutil mirror start 1"]
-    uds["MirrorControlServer<br/>/run/console-monitor/mirror/line1.sock"]
+    control -->|start / stop / status / attach-display| mirror
 
-    userB <-- "start / stop / status" --> uds
-    uds --> mirror
-    mirror -. "optional escaped live display" .-> uds
-    uds -. "display stream" .-> userB
+    userB["User B<br/>sudo consutil mirror start 1 --display"]
+    userB -. "UDS control messages" .-> control
+    display -. "UDS display stream" .-> userB
 ```
 
 ### 2.2 Design Principles
@@ -222,7 +228,7 @@ For accepted records, `MirrorManager` creates a mirror record containing timesta
 * `RecordingWriter`, if recording is active.
 * `DisplaySink`, if live display is attached.
 
-Writer submission is best effort. If the writer queue is full, `MirrorManager` increments `pending_writer_drop_count`, writes a syslog warning, and does not block the forwarding loop. When the writer queue later accepts records again, `MirrorManager` emits one `EVENT` record with `reason=writer_queue_full` and the accumulated `dropped_records` value before recording subsequent data records.
+Writer submission is best effort. If the writer queue is full, `MirrorManager` increments `pending_writer_drop_count`, writes a syslog warning, and does not block the forwarding loop. When the writer queue later accepts records again, `MirrorManager` emits one `EVENT` record with `reason=writer_queue_full` and the accumulated `pending_writer_drop_count` value before recording subsequent data records.
 
 Display submission is also best effort. If the display queue is full, the display record is dropped and a syslog warning is written.
 
@@ -280,7 +286,7 @@ Only one `DisplaySink` is allowed per line. If a display sink is already attache
 
 | Resource | Description |
 |----------|-------------|
-| UDS connection | Connection back to the `consutil mirror start --display` process |
+| UDS connection | Accepted client connection from the `MirrorControlServer` back to the `consutil mirror start --display` process |
 | Display queue | Bounded queue receiving display records from `MirrorManager` |
 | Display task/thread | Background execution context that sends records to the CLI |
 
--- a/doc/console/Console-Mirror-High-Level-Design.md
+++ b/doc/console/Console-Mirror-High-Level-Design.md
@@ -26,7 +26,8 @@
     - [3.1 Proxy Changes](#31-proxy-changes)
       - [3.1.1 MirrorManager](#311-mirrormanager)
       - [3.1.2 RecordingWriter](#312-recordingwriter)
-      - [3.1.3 MirrorControlServer](#313-mirrorcontrolserver)
+      - [3.1.3 RecordingArchiver](#313-recordingarchiver)
+      - [3.1.4 MirrorControlServer](#314-mirrorcontrolserver)
     - [3.2 Mirror Control Message Protocol](#32-mirror-control-message-protocol)
     - [3.3 RX Mirroring](#33-rx-mirroring)
     - [3.4 TX Mirroring](#34-tx-mirroring)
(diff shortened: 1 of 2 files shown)


doc/console/Console-Mirror-High-Level-Design.md
```diff
@@ -0,0 +1,920 @@
...
+
+```bash
+sudo consutil mirror start <target> [OPTIONS]
+```
+
+Options:
+
+| Option | Description |
+|--------|-------------|
+| `--devicename`, `-d` | Interpret target as remote device name instead of line number |
```
lizhijianrd: Since all the CLI design within the existing `consutil` command, let's keep current `-d` param to make it consistency.

doc/console/Console-Mirror-High-Level-Design.md
```diff
@@ -0,0 +1,920 @@
...
+If a mirror session is already active for the target line, the command fails:
+
+```text
+Cannot start mirror: line [1] already has an active mirror session
+```
+
+### 5.2 Stop Mirroring
+
+```bash
+sudo consutil mirror stop <target> [--devicename]
```
lizhijianrd: Could you please add a `-a, --archive` param for the `stop` command?
If user use `a` param, then we archive the mirror log files into zip, otherwise (by default), we keep the log files in the original position on the disk.

doc/console/Console-Mirror-High-Level-Design.md
```diff
@@ -0,0 +1,920 @@
...
+sudo consutil mirror show [target] [--devicename]
+```
+
+Example output:
+
+```text
+Line  State   Direction  Timeout  Stop Time             File
+----  ------  ---------  -------  --------------------  ----
+1     active  both       24h      2026-06-14T14:12:00Z  /var/log/sonic/console-mirror/line1/console-mirror-line1-both-20260613T141200Z-part0001.log
+2     idle    -          -        -                     -
```
lizhijianrd: 1. Can we add a column `Start Time` to show the mirror start time of the line?
2. Can we change the `Stop Time` column to `Remaining`, indicating the remaining mirror session time before timeout? e.g., Timeout: 24h, Remaining: 12h15m
3. Can we add a command `sudo consutil mirror timeout` command to allow user update the timeout of a running session? e.g., Timeout = 1h and Remaining = 5m, if user issue `consutil mirror timeout 1 2h`, then the timeout is reset to 2h and remaining is also 2h.

doc/console/Console-Mirror-High-Level-Design.md
```diff
@@ -0,0 +1,920 @@
...
+Stopped mirror on line [1]; packaging recording
+Expected archive: /var/log/sonic/console-mirror/line1/console-mirror-line1-both-20260613T141200Z.zip
+Waiting for packaging to complete...
+
+Recording archive: /var/log/sonic/console-mirror/line1/console-mirror-line1-both-20260613T141200Z.zip
+```
+
+### 6.2 Automatic Stop
+
+User B starts recording with the default 24-hour timeout:
```
lizhijianrd, line 995: For manual stop, if user add `-a` in command, then generate archive file. Otherwise (default), don't generate archive file.
For auto stop, is usually happens when user would like to mirror for a long time and analysis later, or user forget the mirror session after start. In such situation, we generate the archived zip file and cleanup the original log to save the disk space.

doc/console/Console-Mirror-High-Level-Design.md
```diff
@@ -0,0 +1,920 @@
...
+| Key Format | Field | Value | Description |
+|------------|-------|-------|-------------|
+| `CONSOLE_MIRROR\|<line>` | `state` | `idle` / `active` / `error` | Mirror state |
+| `CONSOLE_MIRROR\|<line>` | `owner_pid` | PID | CLI process that started the session |
+| `CONSOLE_MIRROR\|<line>` | `started_by` | username | User that started mirroring |
+| `CONSOLE_MIRROR\|<line>` | `start_time` | Unix timestamp | Mirror start time |
+| `CONSOLE_MIRROR\|<line>` | `timeout` | positive integer seconds | Configured auto-stop timeout |
+| `CONSOLE_MIRROR\|<line>` | `file_path` | path | Current active recording part; cleared when recording stops |
+| `CONSOLE_MIRROR\|<line>` | `direction` | `rx` / `tx` / `both` | Mirrored direction |
+| `CONSOLE_MIRROR\|<line>` | `resolution` | `us` / `ms` | Timestamp resolution |
```
lizhijianrd: Band rate 9600 transfers 0.96 Bytes/ms.
Band rate 115200 transfers 11.52 Bytes/ms.
I think we use `ms` as resolution is enough, no need to provide `us` option to user.

doc/console/Console-Mirror-High-Level-Design.md
V2lsbGlhbVpoYW5n: `--devicename` param is now aligned with other commands.

doc/console/Console-Mirror-High-Level-Design.md
V2lsbGlhbVpoYW5n: `mirror stop` command now support explicitly using `-a` param to create a zip file in [
6ed3dd3](https://github.com/sonic-net/SONiC/pull/2388/commits/6ed3dd39323611bd00a792ac43a445b439dad590)

doc/console/Console-Mirror-High-Level-Design.md
V2lsbGlhbVpoYW5n: - Changed status command output.
- New command added (5.3 Update Mirror Timeout).

[6ed3dd3](https://github.com/sonic-net/SONiC/pull/2388/commits/6ed3dd39323611bd00a792ac43a445b439dad590)

doc/console/Console-Mirror-High-Level-Design.md
V2lsbGlhbVpoYW5n: The HLD now uses fixed millisecond resolution and removes the microsecond option.

[6ed3dd3](https://github.com/sonic-net/SONiC/pull/2388/commits/6ed3dd39323611bd00a792ac43a445b439dad590)

yxieca (dismissed):
## AI-assisted review — Console Mirror HLD

Strong, well-structured HLD. The architecture (passive RX/TX tap inside the existing per-line proxy → non-interference), the SCM-Text v1 record format, and the security posture are sound, and the concrete examples check out (verified: `--devicename` matches the existing `consutil` convention; all SCM-Text `length` examples are byte-exact — 14/5/10 for data, 28/35/41/41 for event JSON; the `+000000000877ms` delta arithmetic is correct; the referenced Console-Monitor HLD exists in-repo). Using STATE_DB (not CONFIG_DB) for reboot-transient runtime state is the idiomatic SONiC choice.

No blocking design flaw. The following are consistency/completeness items worth resolving before merge:

**Consistency / correctness**
- **State-enum mismatch.** §3.1.1 (MirrorManager) defines `state` as `idle` / `active` / `stopping`, but §4.1.1 (STATE_DB `CONSOLE_MIRROR`) lists `idle` / `active` / `error`. These are never reconciled, and §3.6 (fatal writer error / disk full "update STATE_DB state") vs §3.7 ("mark state idle after restart") add ambiguity. Please provide one authoritative state machine (idle↔active→stopping→idle, plus active→error) and state how `stopping`/`error` map into STATE_DB and what clears `error`.
- **`max_file_size` default/minimum unspecified.** Rotation only occurs "if configured" (§3.1.2), and the `--max-file-size` option table (§5.1) gives no default and no floor. Please state the default-when-omitted (a 24h default session with no rotation implies one unbounded `part0001.log`) and reject/floor pathological small values.
- **Control-socket authorization is permission-only.** The UDS is "created with restrictive permissions so only Admin/root" (§3.1.4/§7), but there is no peer-credential (`SO_PEERCRED`) validation mentioned. For a channel that records credentials, please state explicitly that filesystem ownership/permissions are the trust boundary (anyone able to open the socket is trusted), or add peer-cred checking.
- **Recording retention / `/var/log` sizing.** Recording parts and ZIP archives persist indefinitely under `/var/log/sonic/console-mirror/`; auto-stop archives but never purges, and there is no logrotate/retention policy. On many SONiC platforms the `/var/log` partition is deliberately small, so large sensitive recordings can starve syslog. Please specify a retention/logrotate integration, a max-total-size cap, or a data partition (e.g. under `/host/…`) instead of `/var/log`.
- **Wall-clock `delta` vs monotonic deadline.** The auto-stop deadline correctly uses a monotonic clock (§3.1.1), but record `timestamp`/`delta` are wall-clock UTC (§3.5.4), so a backward clock adjustment can make `delta` non-monotonic or negative. Please note that `seq` (not `delta`) is the monotonic ordering key.

**Completeness**
- **Missing Test Plan section.** Please outline the unit/integration test approach — in particular how non-interference (mirroring does not alter the byte stream / line ownership) and RX/TX capture are validated.
- **Missing Scalability note.** Max concurrent mirror sessions across lines and per-session memory/disk cost.
- **Impacted modules / schema home.** The doc references `sonic-utilities` for the CLI but does not name the console-monitor daemon's repo/package, nor where the new STATE_DB `CONSOLE_MIRROR` schema is registered (swss schema). SONiC HLDs normally enumerate impacted modules + the schema location.
- **Explicit warm-reboot / `config reload` behavior.** §3.6 uses generic "Proxy restart" language; please add an explicit row for warm reboot / config reload (recording ends, parts preserved, STATE_DB reset).
- **UTF-8 escaping classification.** §3.5.6 mixes "UTF-8 character" and "byte" rows. Please clarify whether classification is per-decoded-codepoint or per-byte, and how valid-but-non-printable codepoints (NEL, zero-width, BiDi overrides) are escaped — these are the terminal-spoofing vectors the escaping is meant to neutralize.
- **`stop --archive` blocking wait.** The CLI blocks on the socket until the terminal response (§3.2/§5.2). Please document a server-side keepalive/max-wait, or explicitly state the CLI blocks indefinitely until the terminal response.
- **Minor:** clarify the `RX`/`TX` (record tokens) vs `rx`/`tx`/`both` (path/DB/CLI) casing split; note `timeout` is stored as integer seconds in STATE_DB but as duration strings in the CLI/protocol; mark `owner_pid` as informational (it can point to an exited CLI for most of a session, and auto-stop has no CLI); and state explicitly that `direction=both` produces a single interleaved file.

Overall this is a solid v0.1 design; approving to unblock, with the above items recommended before implementation lands.

_Reviewed by AI agent on behalf of Ying._

lizhijianrd (dismissed):

doc/console/Console-Mirror-High-Level-Design.md
```diff
@@ -0,0 +1,1057 @@
...
+  "state": "active",
+  "line": "1",
+  "start_time": "2026-06-13T14:12:00Z",
+  "direction": "both",
+  "timeout": "24h",
+  "remaining": "12h15m",
+  "file_path": "/var/log/sonic/console-mirror/line1/console-mirror-line1-both-20260613T141200Z-part0001.log"
+}
+```
+
```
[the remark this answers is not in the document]
V2lsbGlhbVpoYW5n: It is designed to that way.

The actual implementation is:
```python
# STATE_DB
("start_time", str(int(self.start_time)))

# Control protocol
"start_time": _rfc3339_ms(self.start_time)

# CLI: passes the control value through
status.get("start_time") or "-"
```

lizhijianrd (approved):

PR #2396, opened 2026-06-18
docs: Add HLD for Bazel migration

blorente:
Add HLD describing a potential migration to Bazel, including the changes necessary in the current build so that the migration is as un-disruptive as possible.

### DIFF
--- a/doc/bazel/bazel_migration_hld.md
+++ b/doc/bazel/bazel_migration_hld.md
@@ -30,12 +30,9 @@
 8. [SAI API](#8-sai-api)
 9. [Configuration and management](#9-configuration-and-management)
 10. [Warmboot and Fastboot Design Impact](#10-warmboot-and-fastboot-design-impact)
-    - [Warmboot and Fastboot Performance Impact](#warmboot-and-fastboot-performance-impact)
 11. [Memory Consumption](#11-memory-consumption)
 12. [Restrictions/Limitations](#12-restrictionslimitations)
 13. [Testing Requirements/Design](#13-testing-requirementsdesign)
-    - [13.1. Unit Test cases](#131-unit-test-cases)
-    - [13.2. System Test cases](#132-system-test-cases)
 14. [Open/Action items - if any](#14-openaction-items---if-any)
 
 ### 1. Revision
--- a/doc/bazel/bazel_migration_hld.md
+++ b/doc/bazel/bazel_migration_hld.md
@@ -112,9 +112,9 @@ All components should still be buildable with the regular Make-based flow.
 
 When a component is built with Bazel, this will entail:
 
-- Eliminating system dependencies (e.g. from `apt` and `whl`), as well as other sonic-make injected deps. All dependencies will go through the hermetic Bazel build graph.
-- Removing `deb` packages as an intermediate format, instead relying on the Bazel dependency graph.
-- Removing `docker` as a dependency, instead building containers directly with Bazel.
+- Eliminating system dependencies for that component (e.g. from `apt` and `whl`), as well as other sonic-make injected deps. All dependencies will go through the hermetic Bazel build graph.
+- Removing `deb` packages as an intermediate format for that component, instead relying on the Bazel dependency graph.
+- Removing `docker` as a dependency for that component, instead building containers directly with Bazel.
 
 However, we will not attempt to:
 
@@ -141,13 +141,13 @@ This will signal to the community that we do intend to adopt Bazel, and that the
 
 By turning `BUILD_WITH_BAZEL_WHEN_AVAILABLE` on, we'll also be making the Bazel build **blocking in CI**. As a consequence, any changes that break the Bazel build will have to be fixed before they're merged.
 
-It is possible that, at this point, the need arises for a per-target toggle. This is left as an [open question](#14-openaction-items---if-any).
+It is possible that, at this point, the need arises for a per-component Bazel toggle. This is left as an [open question](#14-openaction-items---if-any).
 
 We expect to increase coverage of targets that build with Bazel. Specifically, we'll transition to building the base layers (`docker-base-*`, `docker-config-engine-*`, and `docker-swss-layer-*`) with Bazel by default.
 
 By the end of this phase, we expect most users to be able to build their components entirely in Bazel, without the need of a slave container.
 
-**This opt-out window will only exist for 1 release. After that, we will move into Phase 3.**
+**This opt-out window will only exist for 1 release after the Bazel build for a component has been introduced. After that, we will move into Phase 3.**
 
 ##### Phase 3: Bazel-only
 
@@ -161,8 +161,8 @@ One open question is when we will establish a blocking CI pipeline for Bazel bui
 
 We propose the following structure:
 
-- Phase 1: A nightly, post-merge job that tests the Bazel builds only. There are no expectations to keep this job green.
-- Phase 2: At the start of Phase 2, the components that are migrated to Bazell will now be blocking pre-merge, as we flip the default of `BUILD_WITH_BAZEL_WHEN_AVAILABLE`.
+- Phase 1: A nightly, post-merge job that tests the Bazel builds only. There are no expectations to keep this job green, but it will be useful in spotting regressions.
+- Phase 2: Bazel builds are blocking pre-submit. When we flip the default of `BUILD_WITH_BAZEL_WHEN_AVAILABLE`, the components that are migrated to Bazell will now be blocking the pre-submit checks.
 
 This phased approach allows us to establish critical infrastructure and let the build mature before we make it required for anyone.
 
--- a/doc/bazel/bazel_migration_hld.md
+++ b/doc/bazel/bazel_migration_hld.md
@@ -8,6 +8,8 @@
 
 This document covers the migration of the SONiC build system into a modern, Bazel-based system. It proposes an end state of the build system, as well as a migration path to get us there.
 
+Please note that this design covers only the migration of individual components. Migrating the full image assembly process is out of the scope of this document.
+
 ### 3. Definitions/Abbreviations 
 
 - Bazel: A build system open-sourced by Google. It specializes in polyglot builds, and promises fast, reproducible builds through hermeticity. [Documentation](https://bazel.build/docs)
@@ -43,12 +45,60 @@ Early results demonstrate that we can produce cold builds similar to those in th
 
 // TODO BL: Should we include a table here?
 
-#### 4.b Migration Overview
+#### 4.b Migration
 
-Bazel migrations are expensive, and famously disruptive. To alleviate this issue as much as possible, we propose a gradual rollout. We will migrate each component in turn, in a manner such that we:
+Bazel migrations are expensive, and famously disruptive. To alleviate this issue as much as possible, we propose a gradual, non-mandatory rollout.
 
-- Keep compatibility with the old build system. A component built with Bazel can be depended on from a component built with Make.
-- Can roll back if needed. The old build system will remain in place, even for migrated components, to ensure users can roll back quickly if there are issues.
+We will migrate each component in turn, starting from the components that are depended on the least.
+For each component, we will introduce the Bazel build for its container, while retaining the ability to build the container with the current system.
+Users can control whether they want to build with Bazel or GNU make with a command line flag:
+
+```
+$ make build // TODO BL: Current command
+$ BUILD_WITH_BAZEL_WHEN_AVAILABLE=true make build  // TODO BL: Current command
+```
+
+This flag will start off by default. The implementation of this flag's semantics is defined in [TODO BL:](section 7.h).
+
+To minimize disruption, we propose the following phases to the migration:
+
+##### Phase 1: Opt-in trial period
+
+Phase 1 will introduce Bazel as an optional build system for some components. For this phase, `BUILD_WITH_BAZEL_WHEN_AVAILABLE` will be turned off by default.
+We will start with small components such as `sflow`, `teamd`, and `database`, and continue onto progressively larger leaf components until we have sufficient coverage to be representative of the Bazel build.
+We expect this phase to last until the November release.
+
+The community can use this period to experiment with Bazel, adopt it into their own builds, and generally gather information on whether this is a net benefit.
+
+After the November release, the community will face a decision point: Should we adopt Bazel fully?
+
+If we decide to move forward, we will continue onto Phase 2.
+
+##### Phase 2: Opt-out migration period
+
+At the start of this phase, we will flip `BUILD_WITH_BAZEL_WHEN_AVAILABLE` to be on by default.
+This will signal to the community that we do intend to adopt Bazel, and that they should start adopting it into their internal forks if they haven't already.
+
+We expect to massively increase coverage of targets that build with Bazel, specifically extending to building the base layers (`docker-base-*`, `docker-config-engine-*`, and `docker-swss-layer-*`) with Bazel by default.
+
+By the end of this phase, we expect most users to be able to build their components entirely in Bazel, without the need of a slave container.
+
+##### Phase 3: Bazel-only
+
+When every component can be built with Bazel, and most users have had a reasonable opportunity to migratie, we expect to be able to remove the current build system, and transition to using exclusively Bazel.
+
+We cannot give estimations of when this will be, as it will depend on community support and involvement.
+
+##### CI considerations
+
+One open question is when we will establish a blocking CI pipeline for Bazel builds.
+We propose doing this early, as part of Phase 1.
+
+After the first container build merges, we will stand up a non-blocking CI pipeline that will test the Bazel build.
+That pipeline will run on every build, but failures in it won't block contributors from merging code.
+
+When that pipeline has been proven to be stable for a reasonable period, we propose to make it blocking, so that contributions that would break the Bazel build cannot be merged.
+This will avoid code drift between the two builds, helping tremendously with migration speed.
 
 ### 5. Requirements
 
@@ -56,10 +106,7 @@ TODO BL: I need to read more what they expect here.
 
 ### 6. Architecture Design 
 
-This section covers the changes that are required in the SONiC architecture. In general, it is expected that the current architecture is not changed.
-This section should explain how the new feature/enhancement (module/sub-module) fits in the existing architecture. 
-
-If this feature is a SONiC Application Extension mention which changes (if any) needed in the Application Extension infrastructure to support new feature.
+There are no expected changes to the SONiC architecture.
 
 ### 7. High-Level Design 
 
@@ -226,6 +273,16 @@ Every component repository will need to be migrated to Bazel. `sonic-buildimage`
 We expect the resulting containers to be comparable (if not equal) to those produced by the old build system, both in terms of size and runtime performance and memory footprint.
 
 We aim for build times to be reduced dramatically, especially for incremental builds. We expect noticeable second-order effects on developer ergonomics, as a more reliable build means fewer cold builds overall. We expect on-disk build caches to be significantly larger than those of the old build system.
+
+#### 7.h Bazel/make interoperability
+
+During the migration, Bazel and make will have to interoperate. We propose a model where we introduce top-level make targets for Bazel-compatible containers.
+
+// TODO BL: Fill out with code examples
+
+Some other considerations:
+- Bazel should consume layers built by make.
+- Bazel would execute inside the slave container for now. There are still hermeticity issues in the Bazel build, so it does end up depending on the host system. Therefore, to ensure interoperability with the make-built layers, we'll build in the slave container. These hermeticity issues should be treated as bugs, but in the interest of getting the build in the hands of users we choose to solve them later in the migration.
 
 ### 8. SAI API 
 
--- a/doc/bazel/bazel_migration_hld.md
+++ b/doc/bazel/bazel_migration_hld.md
@@ -16,16 +16,16 @@
 5. [Requirements](#5-requirements)
 6. [Architecture Design](#6-architecture-design)
 7. [High-Level Design](#7-high-level-design)
-   - [7.a Groundwork](#7a-groundwork)
-   - [7.b Managing Patched External Dependencies](#7b-managing-patched-external-dependencies)
-   - [7.c Building Component Containers](#7c-building-component-containers)
-   - [7.d Platform & Device Support](#7d-platform--device-support)
-   - [7.e Debuggability](#7e-debuggability)
-   - [7.f Affected Systems](#7f-affected-systems)
+   - [7.a Bazel/make interoperability](#7a-bazelmake-interoperability)
+   - [7.b Groundwork](#7b-groundwork)
+   - [7.c Managing Patched External Dependencies](#7c-managing-patched-external-dependencies)
+   - [7.d Building Component Containers](#7d-building-component-containers)
+   - [7.e Platform & Device Support](#7e-platform--device-support)
+   - [7.f Debuggability](#7f-debuggability)
+   - [7.g Affected Systems](#7g-affected-systems)
      - [Dependency Changes](#dependency-changes)
      - [Changed Repositories](#changed-repositories)
-   - [7.g Performance Summary](#7g-performance-summary)
-   - [7.h Bazel/make interoperability](#7h-bazelmake-interoperability)
+   - [7.h Performance Summary](#7h-performance-summary)
 8. [SAI API](#8-sai-api)
 9. [Configuration and management](#9-configuration-and-management)
 10. [Warmboot and Fastboot Design Impact](#10-warmboot-and-fastboot-design-impact)
@@ -70,6 +70,10 @@ Please see these publications about our work:
 
 ### 4. Overview 
 
+> [!tip]
+> A full example of migrating a container can be found in [sonic-buildimage#28005](https://github.com/sonic-net/sonic-buildimage/pull/28005).
+> We will explore the different moving pieces in that PR in [Section 7](7-high-level-design).
+
 #### 4.a Motivation
 
 The current SONiC build system, while being ergonomic and accommodating the vast matrix of platforms and vendors that are required, has some shortcomings, that the community has been feeling for some time.
@@ -78,14 +82,13 @@ After two decades of SONiC development, we now have a fuller picture of what SON
 
 Early results demonstrate that we can produce cold builds similar to those in the current system, but **sub-minute builds** for some classes of incremental changes, all while keeping hermeticity and reproducibility, so that the same build on the same commit will produce the same results a year from now.
 
-// TODO BL: Should we include a table here?
-
 #### 4.b Migration
 
 Bazel migrations are expensive, and famously disruptive. To alleviate this issue as much as possible, we propose a gradual, non-mandatory rollout.
 
-We will migrate each component in turn, starting from the components that are depended on the least.
-For each component, we will introduce the Bazel build for its container, while retaining the ability to build the container with the current system.
+We will extend the existing build system with a new target type that can build docker containers in Bazel.
+The mechanics of this new target are discussed in [7.a](#7a-bazelmake-interoperability).
+
 Users can control whether they want to build with Bazel or GNU make with a command line flag:
 
 ```
@@ -96,7 +99,7 @@ $ BUILD_WITH_BAZEL_WHEN_AVAILABLE=true make target/docker-sysmgr.gz
 # make runs Bazel to build the target
 ```
 
-This flag will start off by default. The implementation of this flag's semantics is defined in [section 7.h](#7h-bazelmake-interoperability).
+This flag will start off by default. The implementation of this flag's semantics is defined in [section 7.a](#7a-bazelmake-interoperability).
 
 To minimize disruption, we propose the following phases to the migration:
 
@@ -168,21 +171,87 @@ There are no expected changes to the SONiC architecture.
 
 ### 7. High-Level Design 
 
-This section specifies how different parts of the build will work under Bazel. Everything explained here has already been implemented in a proof of concept, in [thesayyn/sonic-buildimage](https://github.com/thesayyn/sonic-buildimage).
+This section specifies how different parts of the build will work under Bazel.
+Everything explained here has already been implemented in a proof of concept migrating `docker-sysmgr`, in [sonic-buildimage#28005](https://github.com/sonic-net/sonic-buildimage/pull/28005).
+The following sections will explain different parts of that PR.
 
-#### 7.a Groundwork
+#### 7.a Changes to Existing Build System (Bazel/make Interoperability)
+
+During the migration, Bazel and make will have to interoperate. This section explains the changes needed in the current build system to make that happen.
+
+We propose to **extend the existing build system to add the ability to build some targets with Bazel**. For example, with [docker-sysmgr](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/rules/docker-sysmgr.mk#L21-L30):
+
+```make
+# From sonic-buildimage#28005/rules/docker-sysmgr.mk
+
+...
+ifeq ($(BUILD_WITH_BAZEL_WHEN_AVAILABLE),n)
+
+SONIC_DOCKER_IMAGES += $(DOCKER_SYSMGR)
+SONIC_INSTALL_DOCKER_IMAGES += $(DOCKER_SYSMGR)
+
+else
+
+# When BUILD_WITH_BAZEL_WHEN_AVAILABLE is enabled, build this docker with Bazel.
+# Bazel only supports bookworm today, which is enforced at the root Makefile.
+$(DOCKER_SYSMGR)_BAZEL_BASE += $(DOCKER_CONFIG_ENGINE_BOOKWORM)
+SONIC_BAZEL_DOCKER_IMAGES += $(DOCKER_SYSMGR)
+SONIC_BOOKWORM_DOCKERS += $(DOCKER_SYSMGR)
+
+endif
+...
+```
+
+As shown in the example, the `BUILD_WITH_BAZEL_WHEN_AVAILABLE` configuration flag will toggle the entire Bazel behaviour. When switched off, the system will use the regular make-based build:
+
+```make
+# From sonic-buildimage#28005/rules/config
+
+...
+# Build eligible dockers (those registered in SONIC_BAZEL_DOCKER_IMAGES) with
+# Bazel instead of the legacy `docker build` flow.
+# When disabled, fall back to the normal Make docker build.
+BUILD_WITH_BAZEL_WHEN_AVAILABLE ?= n
+```
+
+This flag is defined in [rules/config](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/rules/config#L164-L177), along with another flag to control where the Bazel cache directory goes, `SONIC_BAZEL_CACHE_SOURCE`.
+
+We modify the rules execution engine to create targets for this new target type. In [`slave.mk`](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/slave.mk#L1413-L1420):
+
+```make
+# From sonic-buildimage#28005/slave.mk
+
+...
+# Targets for building docker images with Bazel.
+$(addprefix $(TARGET_PATH)/, $(SONIC_BAZEL_DOCKER_IMAGES)) : $(TARGET_PATH)/%.gz : .platform \
+		$$(addprefix $(TARGET_PATH)/,$$($$*.gz_BAZEL_BASE))
+	$(HEADER)
+	bazel run --config=slave //dockers/$*:write_$*.gz $(LOG)
+	$(FOOTER)
+
+SONIC_TARGET_LIST += $(addprefix $(TARGET_PATH)/, $(SONIC_BAZEL_DOCKER_IMAGES))
+```
+
+Please note that these new targets depend on make-built base images, through the `$*_BAZEL_BASE` condition.
+We have written [some tooling to import make-built base images into Bazel](https://github.com/sonic-net/sonic-buildimage/tree/e09be005b19c3521c674e4415d08a25648fc15f4/tools/bazel/oci), but they are out of scope of this section.
+
+This ensures that Bazel-built dockers are built exactly any other Docker, in the slave container, while maintaining Bazel's benefits like hermeticity and a more granular cache.
+It also ensures that **there should be no change to the workflow of someone using Bazel**. The way they call make is the same, and the produced artifacts should be interchangeable.
+
+Our goal is that there can be ICs that have switched to Bazel without realizing it.
+
+#### 7.b Groundwork
 
 The Bazel ethos is that **every input to a build must be known, down to the checksum, before the build starts**. The current build system has a few instances where we cannot deterministically predict these inputs.
 
-- Define a hermetic gcc toolchain, so that we always use the same version for every build. [Source](https://github.com/thesayyn/sonic-build-infra/tree/master/toolchains/gcc).
-- Fetch Debian packages deterministically, instead of relying in `apt install`. We do that by using `rules_distroless` to fetch from a Debian snapshot. [Source](https://github.com/thesayyn/sonic-buildimage/blob/master/dockers/docker-base-bookworm/base_bookworm.MODULE.bazel).
-- For components that have Python dependencies, we created `pyproject.toml` files to be able to generate deterministic `uv` lockfiles. [Example](https://github.com/thesayyn/sonic-utilities/blob/master/pyproject.toml).
+- Define a hermetic gcc toolchain, so that we always use the same version for every build. [Source](https://github.com/blorente/sonic-build-infra/tree/master/toolchains/gcc).
+- Fetch Debian packages deterministically, instead of relying in `apt install`. We do that by using `rules_distroless` to fetch from a Debian snapshot. [Source](TODO LInk to the PR).
 
-#### 7.b Managing Patched External Dependencies
+#### 7.c Managing Patched External Dependencies
 
-The SONiC build relies on patching several third party dependencies (e.g. [libnl3](https://github.com/thesayyn/sonic-buildimage/tree/master/src/libnl3)). Furthermore, it relies on the Debian suite of tools ([debhelper](https://man7.org/linux/man-pages/man1/dh.1.html) and associated tools), which is not compatible with Bazel. Hence, another solution is required.
+The SONiC build relies on patching several third party dependencies (e.g. [libnl3](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/src/libnl3/BUILD.bazel)). Furthermore, it relies on the Debian suite of tools ([debhelper](https://man7.org/linux/man-pages/man1/dh.1.html) and associated tools), which is not compatible with Bazel. Hence, another solution is required.
 
-We propose four approaches to tackle this problem. They are detailed in [this PR to `thesayyn/sonic-buidimage`](https://github.com/thesayyn/sonic-buildimage/pull/15), but we will list their summaries here for ease of reference.
+We propose four approaches to tackle this problem. They are detailed in [this documentation](https://github.com/thesayyn/sonic-buildimage/blob/master/tools/bazel/docs/import-external-projects.md), but we will list their summaries here for ease of reference.
 
 In order of most preferred to least preferred, they are:
 
@@ -191,14 +260,16 @@ In order of most preferred to least preferred, they are:
 3. Migrate said dependency to Bazel. The specific method will depend on the dependency, but tips and tricks have been listed in the document above.
 4. Patch and build the dependency out of band, and then import the built artifacts into the Bazel build.
 
-#### 7.c Building Component Containers
+An early draft of documentation explaining these can be found [here](https://github.com/thesayyn/sonic-buildimage/blob/master/tools/bazel/docs/import-external-projects.md).
+
+#### 7.d Building Component Containers
 
 Currently, most components in SONiC are bundled into `.deb` archives before being installed in the containers to be deployed.
 
 We propose a similar model except, instead of `.deb` archives, we will bundle the components into tar archives for ease of consumption. Using `tar.bzl` and `mtree`, we can replicate the rootfs structure of `.deb` packages:
 
 ```starlark
-# Example from: https://github.com/thesayyn/sonic-buildimage/blob/master/src/sonic-sysmgr/BUILD.bazel#L4-L11
+# Example from: https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/src/sonic-sysmgr/BUILD.bazel#L4-L11
 # Note how we're able to place a file in `/debian/sysmgr/rebootbackend`.
 tar(
     name = "sysmgr_pkg",
@@ -213,31 +284,28 @@ tar(
 Then, we can skip the `apt install` step of building containers and just treat each tar as an OCI container layer with `rules_oci`:
 
 ```starlark
-# Example from: https://github.com/thesayyn/sonic-buildimage/blob/master/dockers/docker-base-bookworm/BUILD.bazel#L123-L139
+# Example from: https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/dockers/docker-sysmgr/BUILD.bazel#L59-L72
 oci_image(
-    name = "docker-base-bookworm",
-    architecture = "amd64",
-    entrypoint = ["/usr/bin/supervisord"],
+    name = "docker-sysmgr",
+    base = ":config_engine_base_layout",
+    entrypoint = ["/usr/local/bin/supervisord"],
     env = {
         "DEBIAN_FRONTEND": "noninteractive",
     },
-    os = "linux",
-    tars = [ # Each of these is a tar() target.
-        ":rootfs",
-        ":rootdirslinks",
-        ":flat",
-        ":site-packages",
-        "@sonic_supervisord_utilities//:dist",
+    tars = [
+        ":apt_deps",
+        ":rdeps",
+        ":source_files",
     ],
     visibility = ["//visibility:public"],
 )
 ```
 
-#### 7.d Platform & Device Support
+#### 7.e Platform & Device Support
 
 The current SONiC build supports a rich matrix of vendors, platforms, and devices. This can be replicated with Bazel's expressive configuration language.
 
-We define command line flags to allow users to specify which device and ASIC manufacturer they're targeting ([link](https://github.com/thesayyn/sonic-build-infra/blob/master/config/BUILD.bazel)):
+We define command line flags to allow users to specify which device and ASIC manufacturer they're targeting ([link](https://github.com/blorente/sonic-build-infra/blob/master/config/BUILD.bazel)):
 
 ```starlark
 # SONiC platform, which maps to ASIC manufacturers
@@ -287,13 +355,13 @@ For ease of use, we propose bundling default configurations for well-known combi
 $ bazel build <target> --config=broadcom # Equivalent to the above.
 ```
 
-#### 7.e Debuggability
+#### 7.f Debuggability
 
 We will implmement automatic debug container generation, mirroring the current system.
 
 We will write a Bazel rule that will crawl the dependency tree of an image, capture the debug symbols of its binaries, and bundle them with well-known debugging tools such as `gdb` to create debug containers.
 
-For instance, following the example in [7.d](#7d-platform--device-support):
+For instance, following the example in [7.e](#7e-platform--device-support):
 
 ```starlark
 oci_image(
@@ -309,7 +377,7 @@ debug_container(
 
 Users will then be able to load these containers into switches normally for debugging.
 
-#### 7.f Affected Systems
+#### 7.g Affected Systems
 
 ##### Dependency Changes
 
@@ -326,21 +394,11 @@ And we gain a few dependencies:
 
 Every component repository will need to be migrated to Bazel. `sonic-buildimage` will also need to change, but it will be migrated piecemeal.
 
-#### 7.g Performance Summary
+#### 7.h Performance Summary
 
 We expect the resulting containers to be comparable (if not equal) to those produced by the old build system, both in terms of size and runtime performance and memory footprint.
 
 We aim for build times to be reduced dramatically, especially for incremental builds. We expect noticeable second-order effects on developer ergonomics, as a more reliable build means fewer cold builds overall. We expect on-disk build caches to be significantly larger than those of the old build system.
-
-#### 7.h Bazel/make interoperability
-
-During the migration, Bazel and make will have to interoperate. We propose a model where we introduce top-level make targets for Bazel-compatible containers.
-
-// TODO BL: Fill out with code examples
-
-Some other considerations:
-- Bazel should consume layers built by make.
-- Bazel would execute inside the slave container for now. There are still hermeticity issues in the Bazel build, so it does end up depending on the host system. Therefore, to ensure interoperability with the make-built layers, we'll build in the slave container. These hermeticity issues should be treated as bugs, but in the interest of getting the build in the hands of users we choose to solve them later in the migration.
 
 ### 8. SAI API 
 
--- a/doc/bazel/bazel_migration_hld.md
+++ b/doc/bazel/bazel_migration_hld.md
@@ -250,7 +250,7 @@ This section describes how individual components are built with Bazel, focusing 
 The Bazel ethos is that **every input to a build must be known, down to the checksum, before the build starts**. The current build system has a few instances where we cannot deterministically predict these inputs.
 
 - Define a hermetic gcc toolchain, so that we always use the same version for every build. [Source](https://github.com/blorente/sonic-build-infra/tree/master/toolchains/gcc).
-- Fetch Debian packages deterministically, instead of relying on `apt install`. We do that by using `rules_distroless` to fetch from a Debian snapshot. [Source](TODO LInk to the PR).
+- Fetch Debian packages deterministically, instead of relying on `apt install`. We do that by using `rules_distroless` to fetch from a Debian snapshot. [Source](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/MODULE.bazel#L23-L56).
 
 ##### 7.b.2 Managing Patched External Dependencies
 
--- a/doc/bazel/bazel_migration_hld.md
+++ b/doc/bazel/bazel_migration_hld.md
@@ -89,8 +89,11 @@ For each component, we will introduce the Bazel build for its container, while r
 Users can control whether they want to build with Bazel or GNU make with a command line flag:
 
 ```
-$ make build // TODO BL: Current command
-$ BUILD_WITH_BAZEL_WHEN_AVAILABLE=true make build  // TODO BL: Current command
+$ make target/docker-sysmgr.gz
+# Builds make-based target as usual
+
+$ BUILD_WITH_BAZEL_WHEN_AVAILABLE=true make target/docker-sysmgr.gz
+# make runs Bazel to build the target
 ```
 
 This flag will start off by default. The implementation of this flag's semantics is defined in [section 7.h](#7h-bazelmake-interoperability).
--- a/doc/bazel/bazel_migration_hld.md
+++ b/doc/bazel/bazel_migration_hld.md
@@ -1,6 +1,6 @@
-# SONiC Build System Migration #
+# SONiC Build System Migration
 
-## Table of Content 
+## Table of Content
 
 1. [Revision](#1-revision)
 2. [Scope](#2-scope)
@@ -38,15 +38,15 @@
     - [13.2. System Test cases](#132-system-test-cases)
 14. [Open/Action items - if any](#14-openaction-items---if-any)
 
-### 1. Revision  
+### 1. Revision
 
-### 2. Scope  
+### 2. Scope
 
 This document covers the migration of the SONiC build system into a modern, Bazel-based system. It proposes an end state of the build system, as well as a migration path to get us there.
 
 Please note that this design covers only the migration of individual components. Migrating the full image assembly process is out of the scope of this document.
 
-### 3. Definitions/Abbreviations 
+### 3. Definitions/Abbreviations
 
 - Bazel: A build system open-sourced by Google. It specializes in polyglot builds, and promises fast, reproducible builds through hermeticity. [Documentation](https://bazel.build/docs)
 - Bazel rules, also called rulesets: Extensions to Bazel to provide additional capabilities, usually to integrate Bazel with a new programming language. Usually named `rules_<lang>`. For instance, `rules_go` extends Bazel to be able to build and test Go sources. [Documentation](https://bazel-docs-staging.netlify.app/versions/master/skylark/rules.html)
@@ -69,7 +69,7 @@ Please see these publications about our work:
 - Talk from OCP EMEA Summit 2026: https://www.youtube.com/watch?v=uSKCNDWuXjc
 - Blog article about our work:  https://aspect.build/blog/bazel-for-sonic
 
-### 4. Overview 
+### 4. Overview
 
 > [!tip]
 > A full example of migrating a container can be found in [sonic-buildimage#28005](https://github.com/sonic-net/sonic-buildimage/pull/28005).
@@ -87,7 +87,7 @@ Early results demonstrate that we can produce cold builds similar to those in th
 
 Bazel migrations are expensive, and famously disruptive. To alleviate this issue as much as possible, we propose a gradual, non-mandatory rollout.
 
-We will extend the existing build system with a new target type that can build docker containers in Bazel.
+We will **extend the existing build system** with a new target type that can build docker containers in Bazel.
 The mechanics of this new target are discussed in [7.a](#7a-changes-to-existing-build-system-bazelmake-interoperability).
 
 Users can control whether they want to build with Bazel or GNU make with a command line flag:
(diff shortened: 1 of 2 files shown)


doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,387 @@
...
+
+The current SONiC build system, while being ergonomic and accommodating the vast matrix of platforms and vendors that are required, has some shortcomings, that the community has been feeling for some time.
+
+After two decades of SONiC development, we now have a fuller picture of what SONiC needs in a build. With this information, we believe we can build a faster, more reliable build system on top of Bazel.
+
+Early results demonstrate that we can produce cold builds similar to those in the current system, but **sub-minute builds** for some classes of incremental changes, all while keeping hermeticity and reproducibility, so that the same build on the same commit will produce the same results a year from now.
+
+// TODO BL: Should we include a table here?
+
+#### 4.b Migration
```
bocon13, line 96: include an example of changes required to *mk to opt in

blorente: @bocon13 I have included a more thorough example of the changes in section 7.a (formerly 7.h, but I moved it up because I thought it would be more important than anything else in section 7). Please let me know what you think!

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+Non-Bazel Definitions
+- debhelper, or `dh`: A suite of Debian tools to rebuild and repack Debian packages. In SONiC, they are used to apply patches to external dependencies. [Documentation](https://man7.org/linux/man-pages/man1/dh.1.html)
+- Open Container Initiative (OCI): An industry standard to specify container formats and runtimes. [Documentation](https://opencontainers.org/).
+
+#### 3.a Prior Art
+
+We have already implemented large parts of this document as proofs of concept.
+
+Please see these publications about our work:
+- Talk from OCP EMEA Summit 2026: https://www.youtube.com/watch?v=uSKCNDWuXjc
```
bocon13, line 79: add a few more links

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+```
+
+This flag will start off disabled by default. The implementation of this flag's semantics is defined in [section 7.a](#7a-changes-to-existing-build-system-bazelmake-interoperability).
+
+To minimize disruption, we propose the following phases to the migration:
+
+##### Phase 1: Opt-in trial period
+
+Phase 1 will introduce Bazel as an optional build system for some components. For this phase, `BUILD_WITH_BAZEL_WHEN_AVAILABLE` will be turned off by default.
+We will start with small components such as `sflow`, `teamd`, and `database`, and continue onto progressively larger leaf components until we have sufficient coverage to be representative of the Bazel build.
```
bocon13: sysmgr, p4rt, watchdogs as first components.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+
+$ BUILD_WITH_BAZEL_WHEN_AVAILABLE=y make target/docker-sysmgr.gz
+# make runs Bazel to build the target
+```
+
+This flag will start off disabled by default. The implementation of this flag's semantics is defined in [section 7.a](#7a-changes-to-existing-build-system-bazelmake-interoperability).
+
+To minimize disruption, we propose the following phases to the migration:
+
+##### Phase 1: Opt-in trial period
```
bocon13: this phase is about:

- eliminating sonic make injected deps
- eliminating system deps (apt/whl)
- removing deb package targets for leaf components
- building containers directly with Bazel

excluded:
- building natively with bazel outside of sonic make infra
- bazel still depends on sonic make base OCI layers (base, config-engine, possibly swss?)

Caveat: there is still some dependency (runtime / build) defined by sonic make and bazel

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+
+Phase 1 will introduce Bazel as an optional build system for some components. For this phase, `BUILD_WITH_BAZEL_WHEN_AVAILABLE` will be turned off by default.
+We will start with small components such as `sflow`, `teamd`, and `database`, and continue onto progressively larger leaf components until we have sufficient coverage to be representative of the Bazel build.
+We expect this phase to last until the November release.
+
+The community can use this period to experiment with Bazel, adopt it into their own builds, and generally gather information on whether this is a net benefit.
+
+After the November release, the community will face a decision point: Should we adopt Bazel fully?
```
bocon13: what does this mean?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+
+If we decide to move forward, we will continue onto Phase 2.
+
+##### Phase 2: Opt-out migration period
+
+At the start of this phase, we will flip `BUILD_WITH_BAZEL_WHEN_AVAILABLE` to be on by default.
```
bocon13, line 162: If there is a regression in one component, do we have to roll back all components?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+This will signal to the community that we do intend to adopt Bazel, and that they should start adopting it into their internal forks if they haven't already.
+
+We expect to massively increase coverage of targets that build with Bazel, specifically extending to building the base layers (`docker-base-*`, `docker-config-engine-*`, and `docker-swss-layer-*`) with Bazel by default.
+
+By the end of this phase, we expect most users to be able to build their components entirely in Bazel, without the need of a slave container.
+
+##### Phase 3: Bazel-only
+
+When every component can be built with Bazel, and most users have had a reasonable opportunity to migrate, we expect to be able to remove the current build system, and transition to using exclusively Bazel.
```
bocon13: How long does each component owner have to support BUILD_WITH_BAZEL....? Devs want to have to minimize the time supporting multiple build systems.

doc/bazel/bazel_migration_hld.md
bocon13: Should Bazel become a first class build system?

What does that mean?

- bazel may be the only option for some components (already true today for components like p4rt)
- bazel will handle more of dependency graph for these components (sonic deps might also need to be converted to bazel e.g. sonic-utils, py-common)
- reasonable build performance requires basic bazel infra (i.e. local cache, remote cache)

bocon13: reframe phase 1 around trial of build and bazel infra

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+We will start with small components such as `sflow`, `teamd`, and `database`, and continue onto progressively larger leaf components until we have sufficient coverage to be representative of the Bazel build.
+We expect this phase to last until the November release.
+
+The community can use this period to experiment with Bazel, adopt it into their own builds, and generally gather information on whether this is a net benefit.
+
+After the November release, the community will face a decision point: Should we adopt Bazel fully?
+
+If we decide to move forward, we will continue onto Phase 2.
+
+##### Phase 2: Opt-out migration period
```
bocon13: expectation is that users and CI should leverage bazel cache (and maybe more) and that parts of the system will be built with Bazel by default. Opt-out window will exist for some reasonable period (1 release).

Possibility to have opt-out window on component/container basis.

doc/bazel/bazel_migration_hld.md
bocon13: Component owners can remove support for make-based build.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+##### Phase 3: Bazel-only
+
+When every component can be built with Bazel, and most users have had a reasonable opportunity to migrate, we expect to be able to remove the current build system, and transition to using exclusively Bazel.
+
+We cannot give estimations of when this will be, as it will depend on community support and involvement.
+
+##### CI considerations
+
+One open question is when we will establish a blocking CI pipeline for Bazel builds.
+We propose doing this early, as part of Phase 1.
```
bocon13: non-blocking during phase 1. blocking during phase 2

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+    * A CI pipeline that validates the Bazel build to prevent drift between the two systems
+5. Platform and device coverage
+    * Present patterns to migrate the existing matrix of vendors, platforms, devices, and ASIC manufacturers
+    * Allow users to select the target platform and SAI implementation at build time from the command line
+6. Maintained developer experience
+    * Automatic generation of debug containers, mirroring the current system
+
+### 6. Architecture Design 
+
+There are no expected changes to the SONiC architecture.
```
bocon13: From build system architecture, we are extending the build system to support a new bazel target type for containers.

bocon13: some of the dependency graph is moving to bazel at the binary/library level

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+
+Our goal is that a contributor could switch to Bazel without even realizing it.
+
+#### 7.b Bazel Build
+
+This section describes how individual components are built with Bazel, focusing on the Bazel build itself rather than its interoperability with the existing make-based system.
+
+##### 7.b.1 Groundwork
+
+The Bazel ethos is that **every input to a build must be known, down to the checksum, before the build starts**. The current build system has a few instances where we cannot deterministically predict these inputs.
```
bocon13, line 314: during transition. bazel determines "system" dependency versions via rules_distroless at build time. docker-base++ Dockerfile determiens runtime dependency versions.

Vs. Today --- slave Dockerfile for build and base for runtime.

Not a regressions, per se, but one more place to keep in sync during transition.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+
+There are no changes to the SAI API.
+
+### 9. Configuration and management 
+
+There are no changes to the configuration.
+
+### 10. Warmboot and Fastboot Design Impact  
+
+TODO BL: I probably need help from Brian for this
```
bocon13: no impact

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,450 @@
...
+
+### 12. Restrictions/Limitations  
+
+We should not depend on the host system at all. This means that we must find hermetic alternatives to historically unhermetic software, such as gcc. For instance, we should never rely on gcc being installed on the system.
+
+This means that, sometimes, we'll need to migrated dependencies to Bazel, which incurs a non-trivial cost. A large portion of these dependencies have already been migrated, and the maintenance cost is estimated to be low.
+
+### 13. Testing Requirements/Design  
+
+// TODO BL: I probably need Brian's help for this, since his team has been donig the validation
```
bocon13: Note: no new sonic functional tests required --- rely on existing testing framework

but,
adding CI jobs, validating bazel in these environments, required bazel infra for build performance.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,478 @@
...
+### 2. Scope
+
+This document covers the migration of the SONiC build system into a modern, Bazel-based system. It proposes an end state of the build system, as well as a migration path to get us there.
+
+Please note that this design covers only the migration of individual components. Migrating the full image assembly process is out of the scope of this document.
+
+### 3. Definitions/Abbreviations
+
+- Bazel: A build system open-sourced by Google. It specializes in polyglot builds, and promises fast, reproducible builds through hermeticity. [Documentation](https://bazel.build/docs)
+- Bazel rules, also called rulesets: Extensions to Bazel to provide additional capabilities, usually to integrate Bazel with a new programming language. Usually named `rules_<lang>`. For instance, `rules_go` extends Bazel to be able to build and test Go sources. [Documentation](https://bazel-docs-staging.netlify.app/versions/master/skylark/rules.html)
```
judsonwilson-nvidia: Would be helpful to describe that users define "targets" by invoking a rule.

Also, I think that even the basic capabilities are defined by rules?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,478 @@
...
+- BUILD files: Files where the Bazel build is defined. Usually named `BUILD.bazel`, and written by hand in [Starlark](https://bazel.build/rules/language). [Documentation](https://bazel.build/concepts/build-files)
+- Bazel registry: A specially-shaped directory that hosts bazel rules and their versions. [Documentation](https://bazel.build/external/registry)
```
judsonwilson-nvidia, line 62: This seems to be omitting several details, which I found confusing. 

If I understand correctly:
- A registry can be a local directory or an HTTP server.
- Registries host different kinds of bazel modules, including bazel rules, as well as common C, C++ and Java libraries and other projects.

doc/bazel/bazel_migration_hld.md
judsonwilson-nvidia: I see the "target" definition below. I would put these definitions next to eachother and describe the relationship.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,478 @@
...
+
+After two decades of SONiC development, we now have a fuller picture of what SONiC needs in a build. With this information, we believe we can build a faster, more reliable build system on top of Bazel.
+
+Early results demonstrate that we can produce cold builds similar to those in the current system, but **sub-minute builds** for some classes of incremental changes, all while keeping hermeticity and reproducibility, so that the same build on the same commit will produce the same results a year from now.
+
+#### 4.b Migration
+
+Bazel migrations are expensive, and famously disruptive. To alleviate this issue as much as possible, we propose a gradual, non-mandatory rollout.
+
+We will **extend the existing build system** with a new target type that can build docker containers in Bazel.
```
judsonwilson-nvidia:
```suggestion
We will **extend the existing build system** with a new makefile target type that can build docker containers in Bazel.
```

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,478 @@
...
+# make runs Bazel to build the target
+```
+
+This flag will start off disabled by default. The implementation of this flag's semantics is defined in [section 7.a](#7a-changes-to-existing-build-system-bazelmake-interoperability).
+
+To minimize disruption, we propose the following phases to the migration:
+
+##### Phase 1: Trial of Bazel Infrastructure
+
+Phase 1 will introduce Bazel as an optional build system for some components. The goal is to validate whether and how SONiC could most benefit from Bazel.
```
judsonwilson-nvidia: It's not clear to me what is the unit of granularity for a "component". Will this likely be an entire container?

Does the entire "component" need to have full Bazel support to build it with Bazel, or can components themselves be hybrid? I assume we can add BUILD files incrementally within the component but won't be able to make an image build using bazel until the full component can be built with Bazel?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,478 @@
...
+# When BUILD_WITH_BAZEL_WHEN_AVAILABLE is enabled, build this docker with Bazel.
+# Bazel only supports bookworm today, which is enforced at the root Makefile.
+$(DOCKER_SYSMGR)_BAZEL_BASE += $(DOCKER_CONFIG_ENGINE_BOOKWORM)
+SONIC_BAZEL_DOCKER_IMAGES += $(DOCKER_SYSMGR)
+SONIC_BOOKWORM_DOCKERS += $(DOCKER_SYSMGR)
+
+endif
+...
+```
+
```
judsonwilson-nvidia, line 268: I think it's easier to see what is going on if the change to the makefile is shown as a diff/patch rather than the final block:

```suggestion

```diff
# From sonic-buildimage#28005/rules/docker-sysmgr.mk

...
+ifeq ($(BUILD_WITH_BAZEL_WHEN_AVAILABLE),n)
+
 SONIC_DOCKER_IMAGES += $(DOCKER_SYSMGR) 
 SONIC_INSTALL_DOCKER_IMAGES += $(DOCKER_SYSMGR)
+
+else
+
+# When BUILD_WITH_BAZEL_WHEN_AVAILABLE is enabled, build this docker with Bazel.
+# Bazel only supports bookworm today, which is enforced at the root Makefile.
+$(DOCKER_SYSMGR)_BAZEL_BASE += $(DOCKER_CONFIG_ENGINE_BOOKWORM)
+SONIC_BAZEL_DOCKER_IMAGES += $(DOCKER_SYSMGR)
+SONIC_BOOKWORM_DOCKERS += $(DOCKER_SYSMGR)
+
+endif
...
```

```

File: doc/bazel/bazel_migration_hld.md [hunk 19]
```diff
@@ -0,0 +1,478 @@
...
+...
+ifeq ($(BUILD_WITH_BAZEL_WHEN_AVAILABLE),n)
+
+SONIC_DOCKER_IMAGES += $(DOCKER_SYSMGR)
+SONIC_INSTALL_DOCKER_IMAGES += $(DOCKER_SYSMGR)
+
+else
+
+# When BUILD_WITH_BAZEL_WHEN_AVAILABLE is enabled, build this docker with Bazel.
+# Bazel only supports bookworm today, which is enforced at the root Makefile.
```
judsonwilson-nvidia: SONiC is now trixie-only, so I would scrub out the bookworm stuff to reduce friction.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,478 @@
...
+#### 7.b Bazel Build
+
+This section describes how individual components are built with Bazel, focusing on the Bazel build itself rather than its interoperability with the existing make-based system.
+
+##### 7.b.1 Groundwork
+
+The Bazel ethos is that **every input to a build must be known, down to the checksum, before the build starts**. The current build system has a few instances where we cannot deterministically predict these inputs.
+
+- Define a hermetic gcc toolchain, so that we always use the same version for every build. [Source](https://github.com/blorente/sonic-build-infra/tree/master/toolchains/gcc).
+- Fetch Debian packages deterministically, instead of relying on `apt install`. We do that by using `rules_distroless` to fetch from a Debian snapshot. [Source](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/MODULE.bazel#L23-L56).
```
judsonwilson-nvidia: Are we going to be changing over to distroless containers?  It might be good to clarify this.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,478 @@
...
+##### 7.b.2 Managing Patched External Dependencies
+
+The SONiC build relies on patching several third party dependencies (e.g. [libnl3](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/src/libnl3/BUILD.bazel)). Furthermore, it relies on the Debian suite of tools ([debhelper](https://man7.org/linux/man-pages/man1/dh.1.html) and associated tools), which is not compatible with Bazel. Hence, another solution is required.
+
+We propose four approaches to tackle this problem. They are detailed in [this documentation](https://github.com/thesayyn/sonic-buildimage/blob/master/tools/bazel/docs/import-external-projects.md), but we will list their summaries here for ease of reference.
+
+In order of most preferred to least preferred, they are:
+
+1. Find the module we want to patch in the BCR. Please see the PR for instructions on how to patch it.
+2. Pull the dependency as a Debian package with `rules_distroless`. This method does not allow patching.
```
judsonwilson-nvidia: Could rules_distroless be extended to handle patching?

judsonwilson-nvidia:
I think exit-criteria for Phase 1 will be a hot topic. Can we call it out, but punt on it and say that developing the criteria will be part of Phase 1?  E.g. we are going to need to ensure certain CUJs for devolopers, CI, release builds, etc. We can gather a checklist during Phase 1 and declare beginning of Phase 2 once X happens.

doc/bazel/bazel_migration_hld.md
blorente: I see it so clearly now, the targets should come _before_ rules. Moved up.
Re: Basic capabilities, it depends on what we mean by that (e.g. the scheduling engine is a basic capability, which will never be rule-ified), but I think I see your point. I've added a reference to the starlarkification of rules to this line, to make the distinction clear.

doc/bazel/bazel_migration_hld.md
blorente: Added some clarifying lines, this is indeed confusing and the BCR UI doesn't do a good job of explaining this. Please let me know if I should expand further.

doc/bazel/bazel_migration_hld.md
blorente: Done, thanks :D Good catch, target is overloaded in this document.

doc/bazel/bazel_migration_hld.md
blorente: > It's not clear to me what is the unit of granularity for a "component". Will this likely be an entire container?

Absolutely fair point. I have added a definition of "component" at the very top, and I've disambiguated this section replacing "component" with "container" whenever I meant "contianer".

> Does the entire "component" need to have full Bazel support to build it with Bazel, or can components themselves be hybrid? 

Components themselves _can_ be hybrid temporarily, though it's probably going to result in bad outcomes if we stick with that in the long term. Bazel thrives in everything being Bazelified. If we do need to do that, I think the right granularity is what I describe in the definition for "components": The right abstraction layer is the `.deb`, `.whl`, or container layer. We can consume Make-built debs from Bazel and vice versa if we need to, but it'd be a bad idea to consume e.g. Make-built `.so` libraries from a Bazel built binary. In general, we should aim to migrate "everything under a `rules/*.mk` file" at once.

> I assume we can add BUILD files incrementally within the component but won't be able to make an image build using bazel until the full component can be built with Bazel?

By "image build", I'm assuming you mean "a build of a container's image", not a full SONiC image build.
If so, this is not necessarily true. We can install a make-built `.deb` into a Bazel-built component image. It would probably lead to worse results than Bazelifying everything (e.g. we lose control of the GCC toolchain), but we can technically do it.

doc/bazel/bazel_migration_hld.md
blorente: This is a very good point, thanks. I've added a warning at the top of Section 7, and scrubbed the examples.

doc/bazel/bazel_migration_hld.md
blorente: Great suggestion, thanks. Applied manually.

doc/bazel/bazel_migration_hld.md
blorente: Good point. This has nothing to do with distroless containers, so I've made it very clear in the prose.

doc/bazel/bazel_migration_hld.md
blorente: Good question. It cannot, because it doesn't build the dependencies from source (and using `dh` et al would essentially undo all the benefits we get from Bazel). Added a note to that effect.

blorente (commented):
Thank you so much @judsonwilson-nvidia and @bocon13  for taking a look :D 
Hopefully I've addressed your comments, but looking forward to discussing this further :D 

@judsonwilson-nvidia , regarding Phase 1 exit criteria, I've added some words near the end of Phase 1. I agree that punting and gathering information during the phase is the right move. To be completely honest, I would be guessing blind without putting this in the hands of _some_ users at least.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+$ BUILD_WITH_BAZEL_WHEN_AVAILABLE=y make target/docker-sysmgr.gz
+# make runs Bazel to build the target
+```
+
+This flag will start off disabled by default. The implementation of this flag's semantics is defined in [section 7.a](#7a-changes-to-existing-build-system-bazelmake-interoperability).
+
+To minimize disruption, we propose the following phases to the migration:
+
+##### Phase 1: Trial of Bazel Infrastructure
+
```
jusherma, line 118: During this phase, could the build behavior/outputs of a given component diverge over time depending on whether the build is being done via the traditional `make` approach or with the `BUILD_WITH_BAZEL_WHEN_AVAILABLE` flag?

If so, is that something we're ok with or something we need to mitigate?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+For this phase, `BUILD_WITH_BAZEL_WHEN_AVAILABLE` will be turned off by default.
+We will aim to migrate entire containers at once, though it's possible that a container can have _some_ Bazel-built parts and some Make-built parts.
+We will start adding Bazel builds for leaf, small containers such as `sysmgr`, `p4rt`, and the small watchdog binaries, and continue onto progressively larger leaf containers until we have sufficient coverage to be representative of the benefits and challenges that Bazel poses.
+
+All components (containers and individual `.deb`s) should still be buildable with the regular Make-based flow.
+
+When a container is built with Bazel, this will entail:
+
+- Eliminating system dependencies for that container (e.g. from `apt` and `whl`), as well as other sonic-make injected deps. All dependencies will go through the hermetic Bazel build graph.
+- Removing `deb` packages as an intermediate format for that container, instead relying on the Bazel dependency graph. We will retain the ability to create `deb` packages with Bazel, it will just not be required.
```
jusherma: Does this refer to debian packages which are built solely to install into a docker image, where the latter is the "real" build output and the `.deb` is discarded after install, or am I misunderstanding?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+**We expect this phase to last until the November release.**
+
+After the November release, the community will face a decision point: Should Bazel be a first-class build system in SONiC?
+
+Specifically:
+
+- Is the development experience in Bazel supported well enough to be used in day-to-day operations for migrated components and containers?
+    - Are there any gaps left to fill before this is true?
+- Should Bazel be allowed to be the _only option_ for certain containers (like it is today for `p4rt`)?
+- Should Bazel be allowed to handle most of the dependency graph for these containers? As a consequence, some SONiC dependencies will have to be converted to Bazel (e.g. `sonic-utilities`).
```
jusherma: > Should Bazel be allowed to handle most of the dependency graph

What would be handling the rest of the dependency graph?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+If we decide there are still gaps, we will work to correct them before proceeding.
+
+If we decide to move forward, we will continue onto Phase 2.
+
+##### Phase 2: Migration Period
+
+At the start of this phase, we will flip `BUILD_WITH_BAZEL_WHEN_AVAILABLE` to be on by default.
```
jusherma, line 162: If build behavior diverges between the make and bazel flows during phase 1, that will also need to be reconciled in the bazel files before phase 2 starts

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+This will signal to the community that we do intend to adopt Bazel, and that they should start adopting it into their internal forks if they haven't already.
+
+By turning `BUILD_WITH_BAZEL_WHEN_AVAILABLE` on, we'll also be making the Bazel build **blocking in CI**. As a consequence, any changes that break the Bazel build will have to be fixed before they're merged.
+
+It is possible that, at this point, the need arises for a per-component Bazel toggle. This is left as an [open question](#14-openaction-items---if-any).
+
+We expect to increase coverage of targets that build with Bazel. Specifically, we'll transition to building the base layers (`docker-base-*`, `docker-config-engine-*`, and `docker-swss-layer-*`) with Bazel by default.
+
+By the end of this phase, we expect most users to be able to build their components entirely in Bazel, without the need of a slave container.
```
jusherma: > without the need of a slave container.

I'm excited for this milestone. Will there be any additional changes needed in phase 2 to support this? I was under the impression the only code change for phase 2 was making `BUILD_WITH_BAZEL_WHEN_AVAILABLE` on by default. Wouldn't that imply makefile/docker-slave wrappers still?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+This can happen piecemeal, and be up to the discretion of each component owner.
+
+##### CI considerations
+
+One open question is when we will establish a blocking CI pipeline for Bazel builds.
+
+We propose the following structure:
+
+- Phase 1: A nightly, post-merge job that tests the Bazel builds only. There are no expectations to keep this job green, but it will be useful in spotting regressions.
```
jusherma: I would recommend naming these something different to avoid confusion with the larger phases 1/2/3.

Also, I think creating a formal nightly job to test bazel builds only should be done at the beginning of "Phase 1: Trial of Bazel Infrastructure" (also with no expectation of keeping it green), so we immediately have continuous monitoring on regressions and build time.

It also can give confidence to the community that the bazel infra is robust

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+- Phase 2: Bazel builds are blocking pre-submit. When we flip the default of `BUILD_WITH_BAZEL_WHEN_AVAILABLE`, the components that are migrated to Bazel will now be blocking the pre-submit checks.
+
+This phased approach allows us to establish critical infrastructure and let the build mature before we make it required for anyone.
+
+We expect to be able to leverage the remote caching mechanism with Bazel to make these pipelines significantly faster than the Make-based version.
```
jusherma: Where will the remote cache be stored and will it be publicly readable? Presumably it won't be publicly _writable_ for security?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+2. Hermetic and reproducible builds
+    * No dependency on the host system (e.g. host-installed `gcc`, `python`, or `docker`)
+    * The same build on the same commit produces the same artifacts
+3. Equivalent build output
+    * Resulting containers are comparable to those of the current system in size, runtime performance, and memory footprint
+    * No regression in functionality of the produced images
+4. Gradual migration
+    * Bazel and GNU make coexist, with per-component opt-in and opt-out controlled by a single flag
+    * Components are migrated one at a time, starting from the least depended-on
+    * A CI pipeline that validates the Bazel build to prevent drift between the two systems
```
jusherma, line 217: This is critical. But will this pipeline be able to validate the build outputs between bazel and make-based builds are functionally identical, or just that both build?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+SONIC_TARGET_LIST += $(addprefix $(TARGET_PATH)/, $(SONIC_BAZEL_DOCKER_IMAGES))
+```
+
+Please note that these new targets depend on make-built base images, through the `$*_BAZEL_BASE` condition.
+We have written [some tooling to import make-built base images into Bazel](https://github.com/sonic-net/sonic-buildimage/tree/e09be005b19c3521c674e4415d08a25648fc15f4/tools/bazel/oci), but they are out of scope of this section.
+
+This ensures that Bazel-built dockers are built exactly like any other Docker, in the slave container, while maintaining Bazel's benefits like hermeticity and a more granular cache.
+It also ensures that **there should be no change to the workflow of someone using Bazel**. The way they call make is the same, and the produced artifacts should be interchangeable.
+
```
jusherma, line 305: Are there any "gotchas" in terms of managing bazel's own version when using it inside the sonic-slave container?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+#### 7.b Bazel Build
+
+This section describes how individual components are built with Bazel, focusing on the Bazel build itself rather than its interoperability with the existing make-based system.
+
+##### 7.b.1 Groundwork
+
+The Bazel ethos is that **every input to a build must be known, down to the checksum, before the build starts**. The current build system has a few instances where we cannot deterministically predict these inputs.
+
+- Define a hermetic gcc toolchain, so that we always use the same version for every build. [Source](https://github.com/blorente/sonic-build-infra/tree/master/toolchains/gcc).
```
jusherma: This is obviously a major step in the right direction for reproducible builds, but during phase 1, could there be cases where the gcc toolchain used by bazel differs from that installed directly in the sonic-slave container? If so, any concerns about inter-target compatibility?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+##### 7.b.2 Managing Patched External Dependencies
+
+The SONiC build relies on patching several third party dependencies (e.g. [libnl3](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/src/libnl3/BUILD.bazel)). Furthermore, it relies on the Debian suite of tools ([debhelper](https://man7.org/linux/man-pages/man1/dh.1.html) and associated tools), which is not compatible with Bazel. Hence, another solution is required.
+
+We propose four approaches to tackle this problem. They are detailed in [this documentation](https://github.com/thesayyn/sonic-buildimage/blob/master/tools/bazel/docs/import-external-projects.md), but we will list their summaries here for ease of reference.
+
+In order of most preferred to least preferred, they are:
+
+1. Find the module we want to patch in the BCR. Please see the PR for instructions on how to patch it.
```
jusherma: If we use this approach, can the data fetched from the BCR be stored in an internal cache as well? I want to make sure we're not trading dependency on one external registry (apt) for another (BCR)

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+2. Pull the dependency as a Debian package with `rules_distroless`. This method does not build the dependencies from source, and thus **does not allow patching**.
+3. Migrate said dependency to Bazel. The specific method will depend on the dependency, but tips and tricks have been listed in the document above.
+4. Patch and build the dependency out of band, and then import the built artifacts into the Bazel build.
+
+An early draft of documentation explaining these can be found [here](https://github.com/thesayyn/sonic-buildimage/blob/master/tools/bazel/docs/import-external-projects.md).
+
+##### 7.b.3 Building Component Containers
+
+Currently, most components in SONiC are bundled into `.deb` archives before being installed in the containers to be deployed.
+
```
jusherma, line 347: Do any of these intermediate debs have install scripts?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+```shell
+$ bazel build <target> --config=broadcom # Equivalent to the above.
+```
+
+##### 7.b.5 Debuggability
+
+We will implement automatic debug container generation, mirroring the current system.
+
+We will write a Bazel rule that will crawl the dependency tree of an image, capture the debug symbols of its binaries, and bundle them with well-known debugging tools such as `gdb` to create debug containers.
```
jusherma: Will the dependency tree crawling approach be robust and maintainable? I generally associate crawling with hacky systems

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+Users will then be able to load these containers into switches normally for debugging.
+
+#### 7.c Affected Systems
+
+##### Dependency Changes
+
+Once components build entirely in Bazel, we lose a few dependencies:
+- We no longer need a sonic-slave container. Note that, during the migration, Bazel still runs inside the slave container (see [7.a](#7a-changes-to-existing-build-system-bazelmake-interoperability)); this dependency only goes away once a component builds entirely in Bazel.
+- We no longer depend on developer tools such as `gcc` and `python` being installed in the build machine.
+- We no longer depend on Docker at build time.
```
jusherma, line 511: This would be fantastic

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+And we gain a few dependencies:
+- We depend on Google servers to download and install Bazel at build time.
```
jusherma, line 514: Does this need to be done for each build? Can't we install bazel on a build host and use it for all builds?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+##### Changed Repositories
+
+Every component repository will need to be migrated to Bazel. `sonic-buildimage` will also need to change, but it will be migrated piecemeal.
+
+#### 7.d Performance Summary
+
+We expect the resulting containers to be comparable (if not equal) to those produced by the old build system, both in terms of size and runtime performance and memory footprint.
+
+We aim for build times to be reduced dramatically, especially for incremental builds. We expect noticeable second-order effects on developer ergonomics, as a more reliable build means fewer cold builds overall. We expect on-disk build caches to be significantly larger than those of the old build system.
```
jusherma, line 525: > We expect on-disk build caches to be significantly larger than those of the old build system.

Any estimates or benchmarks on the size?

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+### 11. Memory Consumption
+
+We expect the resulting containers to be comparable (if not equal) to those produced by the old build system, both in terms of size and runtime performance and memory footprint.
+
+### 12. Restrictions/Limitations  
+
+We should not depend on the host system at all. This means that we must find hermetic alternatives to historically unhermetic software, such as gcc. For instance, we should never rely on gcc being installed on the system.
+
+This means that, sometimes, we'll need to migrate dependencies to Bazel, which incurs a non-trivial cost. A large portion of these dependencies have already been migrated, and the maintenance cost is estimated to be low.
```
jusherma, line 547: To ensure the bazel infra is maintainable long-term, I think it will be critical that engineers from multiple organizations develop expertise in it. This should be a conscious effort

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+However, we will be adding CI jobs to validate the Bazel builds in the required environments, as outlined in the [CI Considerations](#ci-considerations). We expect to be able to leverage remote caching to make these builds significantly more performant than the alternative.
+
+#### 13.1. Unit Test cases  
+
+#### 13.2. System Test cases
+
+### 14. Open/Action items - if any 
+
+- **Should we add a per-component toggle during the migration?**
+  - If necessary, it's entirely possible to add a per-target flag, so that individual components can be toggled without affecting the rest of the build. We're not confident that the feature is needed, and the effort to implement it is relatively low, therefore we've left it out of this document.
```
jusherma: My vote is also to omit this, as it may discourage timely migration to bazel

bocon13 (dismissed):

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+#### 7.b Bazel Build
+
+This section describes how individual components are built with Bazel, focusing on the Bazel build itself rather than its interoperability with the existing make-based system.
+
+##### 7.b.1 Groundwork
+
+The Bazel ethos is that **every input to a build must be known, down to the checksum, before the build starts**. The current build system has a few instances where we cannot deterministically predict these inputs.
+
+- Define a hermetic gcc toolchain, so that we always use the same version for every build. [Source](https://github.com/blorente/sonic-build-infra/tree/master/toolchains/gcc).
+- Fetch Debian packages deterministically, instead of relying on `apt install`. We do that by using `rules_distroless` to fetch from a Debian snapshot. [Source](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/MODULE.bazel#L23-L56). Please note that this only refers to the Bazel ruleset, and is **unrelated to distroless containers** as a concept. Switching to distroless containers is out of the scope of this document.
```
prasoon-nexthop, line 319: **Reproducibility toolchain / mirroring.** The §4.a goal ("the same build on the same commit will produce the same results a year from now") would be stronger if the pinning + mirroring strategy were stated here. The PoC already pins a lot — `.bazelversion`, a timestamp-pinned Debian snapshot, and `apt.lock`/`packages.lock` — which is great. Two gaps worth addressing in the doc: (1) `MODULE.bazel.lock` doesn't appear to be committed, so the bzlmod dependency graph isn't locked or CI-verified (consider committing it + `--lockfile_mode=error` in CI); (2) the guarantee still depends on the live availability of the BCR, `snapshot.debian.org`, and Google's Bazel download servers (see §7.c) — for a true "a year from now" guarantee, consider mirroring the BCR and the snapshot. Also worth noting the guarantee holds *fully* only once base layers are Bazel-built (Phase 2), since until then Bazel consumes Make-built base layers (per the warning just below).

prasoon-nexthop (commented):
Added some inline comments, please check.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+If we decide there are still gaps, we will work to correct them before proceeding.
+
+If we decide to move forward, we will continue onto Phase 2.
+
+##### Phase 2: Migration Period
+
+At the start of this phase, we will flip `BUILD_WITH_BAZEL_WHEN_AVAILABLE` to be on by default.
+This will signal to the community that we do intend to adopt Bazel, and that they should start adopting it into their internal forks if they haven't already.
+
+By turning `BUILD_WITH_BAZEL_WHEN_AVAILABLE` on, we'll also be making the Bazel build **blocking in CI**. As a consequence, any changes that break the Bazel build will have to be fixed before they're merged.
```
prasoon-nexthop: **Global flag = large blast radius in Phase 2.** Flipping the single global `BUILD_WITH_BAZEL_WHEN_AVAILABLE` default to on makes every migrated component blocking in presubmit at the same moment; one broken Bazel build then blocks all contributors. The per-component toggle is listed as an open item (§14), but it looks like the mechanism that makes Phase 2 genuinely incremental rather than all-or-nothing. Consider promoting it into the design (flip per component as each is validated) instead of leaving it optional.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+### 12. Restrictions/Limitations  
+
+We should not depend on the host system at all. This means that we must find hermetic alternatives to historically unhermetic software, such as gcc. For instance, we should never rely on gcc being installed on the system.
+
+This means that, sometimes, we'll need to migrate dependencies to Bazel, which incurs a non-trivial cost. A large portion of these dependencies have already been migrated, and the maintenance cost is estimated to be low.
+
+### 13. Testing Requirements/Design  
+
+There are no new SONiC functional, unit, or system tests required.
```
prasoon-nexthop, line 552: **No artifact-equivalence validation for Requirement 3.** Req 3 (comparable size/runtime/memory, no functional regression) has no validation mechanism, and this section says no new tests are required. During the coexistence window you can cheaply diff the Make-built vs Bazel-built artifacts — installed package set, file tree, sizes, layer digests — and gate CI on it. That equivalence check is arguably the single most valuable safety net for the migration; without it, drift between the two systems only surfaces after it ships. Suggest adding it to the CI plan.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+We propose four approaches to tackle this problem. They are detailed in [this documentation](https://github.com/thesayyn/sonic-buildimage/blob/master/tools/bazel/docs/import-external-projects.md), but we will list their summaries here for ease of reference.
+
+In order of most preferred to least preferred, they are:
+
+1. Find the module we want to patch in the BCR. Please see the PR for instructions on how to patch it.
+2. Pull the dependency as a Debian package with `rules_distroless`. This method does not build the dependencies from source, and thus **does not allow patching**.
+3. Migrate said dependency to Bazel. The specific method will depend on the dependency, but tips and tricks have been listed in the document above.
+4. Patch and build the dependency out of band, and then import the built artifacts into the Bazel build.
```
prasoon-nexthop: **Patched-dep plan may be optimistic + a hermeticity note.** For SONiC's patched deps (libnl3, etc.), approach 1 (module already in BCR) will rarely apply and approach 2 explicitly can't patch — so most patched deps likely land on 3 (migrate to Bazel) or 4 (build out-of-band, import). Worth adding: (a) a rough bucketing of today's patched deps across these approaches + the maintenance cost, and (b) a note that approach 4 imports artifacts *not* built by Bazel, which is in tension with the hermeticity requirement (§5.2) — fine as a transitional escape hatch, but call it out as non-hermetic.

doc/bazel/bazel_migration_hld.md
prasoon-nexthop: **Question: hermetic gcc/glibc vs vendor SAI ABI.** Which gcc/glibc version does the hermetic toolchain pin to, and has ABI compatibility with the prebuilt vendor SAI blobs (e.g. Broadcom `.deb`s) been validated? Vendor SAI is typically compiled against a specific Debian toolchain/glibc; a divergent hermetic gcc could cause link/runtime ABI issues in `syncd` even though the SAI *API* is unchanged (§8). A sentence on how the toolchain stays ABI-compatible with vendor artifacts would help.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+By turning `BUILD_WITH_BAZEL_WHEN_AVAILABLE` on, we'll also be making the Bazel build **blocking in CI**. As a consequence, any changes that break the Bazel build will have to be fixed before they're merged.
+
+It is possible that, at this point, the need arises for a per-component Bazel toggle. This is left as an [open question](#14-openaction-items---if-any).
+
+We expect to increase coverage of targets that build with Bazel. Specifically, we'll transition to building the base layers (`docker-base-*`, `docker-config-engine-*`, and `docker-swss-layer-*`) with Bazel by default.
+
+By the end of this phase, we expect most users to be able to build their components entirely in Bazel, without the need of a slave container.
+
+**This opt-out window will only exist for 1 release after the Bazel build for a component has been introduced. After that, we will move into Phase 3.**
```
prasoon-nexthop: **Phase 2→3 is a one-way door.** Two concerns: (1) a 1-release opt-out window may be short for downstream vendor forks carrying heavy patches — consider making it longer/configurable; (2) once component owners remove Make support (Phase 3), there's no rollback path described. Suggest explicit rollback criteria and keeping Make removal reversible for a grace period after Phase 3 begins.

doc/bazel/bazel_migration_hld.md
prasoon-nexthop: **Remote-cache trust model.** The perf argument leans on remote caching, but for a public repo the trust model matters: who hosts the cache, who has write access, how is cache poisoning from untrusted contributors prevented, and is it shared across forks? A short section on the remote-cache infrastructure + its security model would back the CI speed claims.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+	$(FOOTER)
+
+SONIC_TARGET_LIST += $(addprefix $(TARGET_PATH)/, $(SONIC_BAZEL_DOCKER_IMAGES))
+```
+
+Please note that these new targets depend on make-built base images, through the `$*_BAZEL_BASE` condition.
+We have written [some tooling to import make-built base images into Bazel](https://github.com/sonic-net/sonic-buildimage/tree/e09be005b19c3521c674e4415d08a25648fc15f4/tools/bazel/oci), but they are out of scope of this section.
+
+This ensures that Bazel-built dockers are built exactly like any other Docker, in the slave container, while maintaining Bazel's benefits like hermeticity and a more granular cache.
+It also ensures that **there should be no change to the workflow of someone using Bazel**. The way they call make is the same, and the produced artifacts should be interchangeable.
```
prasoon-nexthop: **Confirm `.gz` format parity.** Since downstream image assembly (out of scope here) consumes these `.gz` outputs, worth stating explicitly that the Bazel-produced `.gz` is byte/format-interchangeable with the Make `docker save` output (same tar layout, loads identically). "Should be interchangeable" is the crux of the transparent-swap promise — good to confirm it's verified, not assumed.

doc/bazel/bazel_migration_hld.md
prasoon-nexthop: **Debug-container rule is underspecified.** This rule crawls the dep tree, captures debug symbols, matches them to the stripped release binaries (build-ids), and bundles gdb. Debug-symbol handling in Bazel (split DWARF / fission, strip vs keep, `.dwo`) is historically fiddly, and debug containers are a required DX feature (§5.6). A bit more detail on how symbols are captured and matched to the release binaries would de-risk this.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+
+Our goal is that a contributor could switch to Bazel without even realizing it.
```
prasoon-nexthop: **First-build UX vs "without even realizing it."** Great goal, but the first Bazel build pulls the gcc toolchain + BCR modules (potentially a large download) and grows a sizable on-disk cache (acknowledged in §7.d). New contributors will notice the first build — worth setting that expectation somewhere.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,502 @@
...
+    * The same build on the same commit produces the same artifacts
+3. Equivalent build output
+    * Resulting containers are comparable to those of the current system in size, runtime performance, and memory footprint
+    * No regression in functionality of the produced images
+4. Gradual migration
+    * Bazel and GNU make coexist, with per-component opt-in and opt-out controlled by a single flag
+    * Components are migrated one at a time, starting from the least depended-on
+    * A CI pipeline that validates the Bazel build to prevent drift between the two systems
+5. Platform and device coverage
+    * Present patterns to migrate the existing matrix of vendors, platforms, devices, and ASIC manufacturers
```
Bojun-Feng: CPU architecture (amd64/arm64/armhf) is another axis of the platform matrix not mentioned in this section. P4RT is already disabled on arm64/armhf because pre-built Bazel binaries aren't available for those architectures (See [slave.mk](https://github.com/sonic-net/sonic-buildimage/blob/9e8dd55d195993e7d7ea5b7c3a3015183b77f901/slave.mk#L230-L246))

This is not a Phase 1 concern, but a brief acknowledgment here would help ARM platform maintainers understand what to expect.

Bojun-Feng (approved):

doc/bazel/bazel_migration_hld.md
blorente: I see two failure modes here: 
- One is that the Bazel build becomes incorrect, and it stops being built. This should be caught by the CI job explained below.
- The other one is that the Make and Bazel builds both complete successfully, but they produce substantially different artifacts. This is very difficult to guard against, because to my knowledge there are no automated, end-to-end integration tests publicly available. Efforts like Alpine usage for these tests are too far away to be practical preconditions. So, I think we have to rely on mitigating this by (1) staying in Phase 1 for as short of a time as possible, so that we converge on one build system for each container as soon as possible, and (2) Relying on community validation of containers. This is not ideal, but I also think that the risk is very small. When it goes wrong, it's usually easy to catch and fix.

blorente: Added some words to that effect.

doc/bazel/bazel_migration_hld.md
blorente: Your understanding is correct. I've added an example to illustrate the point.

doc/bazel/bazel_migration_hld.md
blorente: Make, for now. This is specifically for the cases where the base layer of a container has not been migrated yet, but it will in the future.

But, to be completely honest, we _will_ also have to answer the question of "should Bazel handle the _whole_ dependency graph" eventually, so I think I'll just reword the question to remove "most of"

doc/bazel/bazel_migration_hld.md
blorente: Good point. I've explicitly noted that we'll have to make an effort to reconcile before flipping.

doc/bazel/bazel_migration_hld.md
blorente: Yep, good point. Made this an explicit point.

doc/bazel/bazel_migration_hld.md
blorente: If we gate merges on CI, as explained below, then a broken build should not make it into the repository, right?

Taking a step back, it is unclear to me whether you're advocating to promote it, or just asking us to consider it. Could you elaborate on how strongly you feel about this?

doc/bazel/bazel_migration_hld.md
blorente: Not really. The only hard link to the slave container right now is the fact that we need to consume base layers from Make. If we migrate the base layers, we don't need the slave container. The PoC in https://github.com/thesayyn/sonic-buildimage doesn't need the slave container to create functioning images.

I've added this to the previous paragraph to make the link a little clearer :)

doc/bazel/bazel_migration_hld.md
blorente: I think we should prefer a longer rollout grace period than a rollback mechanism. Added some wording to relax the period, please let me know what you think!

doc/bazel/bazel_migration_hld.md
blorente: Oh, I meant to indicate that these steps should happen at the beginning of the indicated phases. I'll make that clear.

doc/bazel/bazel_migration_hld.md
blorente: This is a really good point :) 
It's is still an open question (I've added it down there), but it's a well-understood model with a lot of prior art (at Aspect, this is a big part of what we do as a business), so I'm sure we'll arrive at a good conclusion.

blorente: Please see this response: https://github.com/sonic-net/SONiC/pull/2396/changes#r3637847060

doc/bazel/bazel_migration_hld.md
blorente: Now that you mention them being identical, I think there's space for a job that takes the Make outputs from a previous job, and compares them against Bazel. I don't think we should go for "functionally identical", because of the aforementioned reasons: As far as I understand it, there's no automated way today to test a container end to end in public CI. However, tests like "the containers have exactly the same file list with the same permissions" would be easy to implement, and would go a long way towards catching drift. 

How does that sound?
Also, how much of a blocker do you think it is to figure this out?

doc/bazel/bazel_migration_hld.md
blorente: Actually, we should explore this further in this thread: https://github.com/sonic-net/SONiC/pull/2396/changes#r3637864817 
Looking forward to your thoughts!

doc/bazel/bazel_migration_hld.md
blorente: Well spotted :D We already have an aarch64 toolchain, and this was definitely one of the guiding research questions of the PoC, so I should make an explicit note :D

doc/bazel/bazel_migration_hld.md
blorente: This is not a claim we're willing to make. In particular, the layout of the container will have an OCI layout , which is different than the docker image layout. I will relax the language, please let me know if this is a problem.

doc/bazel/bazel_migration_hld.md
blorente: Not really. Because the source directory will be mounted directly into the slave container, the `.bazelversion` file will be picked up, as if we were outside of the slave.

doc/bazel/bazel_migration_hld.md
blorente: Good point, added a note on that.

doc/bazel/bazel_migration_hld.md
blorente: Good point. The short story is that we chose a gcc that was compatible with bookworm packages for the PoC, but we may have to upgrade when switching to trixie. We can generate gcc builds more or less at will with https://github.com/blorente/gcc-builds, so this should not be an issue. Same for prebuilt SAIs from vendors: Ostensibly there is a gcc version that fits all use cases (because this is true today for SONiC). If it's a different one than the one we picked, we will discover that quickly and switch easily.

I've added this to the text, please let me know what you think.

blorente: Please see this comment: https://github.com/sonic-net/SONiC/pull/2396/changes#r3638792655 . 
The toolchain we've chosen has been validated with one Broadcom `.deb` (as well as the rest of the container), but needs to be re-validated when we switch to Trixie.

doc/bazel/bazel_migration_hld.md
blorente: I'm not sure if this is a blocker for the HLD or a suggestion on wording. Could you please clarify?

I've added some words at the end of this section, please let me know what you think.

doc/bazel/bazel_migration_hld.md
blorente: Good point. The BCR is backed by a github repository, so while this is a legitimate concern lately, the risk profile is exactly the same as it is for `sonic-buildimage` itself. To be truly resilient, we'd have to either:
- Stand up a community instance of a service like Artifactory to mirror the BCR _and_ the rulesets (e.g. `rules_cc`. Remember, the BCR doesn't contain the contents of the dependency itself, just metadata on how to use the dependency).
- Stand up a mirror of the BCR in a more-resilient-than-public-GitHub service (e.g. a GH Enterprise instance with SLAs).

I'm not sure of the practicalities of either option, but they're both doable. This would be a great topic to discuss in a WG meeting. Would it be okay to punt on this until then? I've added a note on the text.

doc/bazel/bazel_migration_hld.md
blorente: I've added some words for the above, please let me know what you think.

doc/bazel/bazel_migration_hld.md
blorente: Oh, this is a great point. Some of them do, and the solution is to replicate that in Bazel before creating the tar. Added that to the text.

doc/bazel/bazel_migration_hld.md
blorente: Yes, but it's under-specified in the doc. We use a Bazel aspect, which is a first class mechanism to traverse up the graph in a structured way: https://bazel.build/extending/aspects

Let me update the code samples to show what we landed on, please let me know what you think.

blorente: This is covered by the `force_debug_build` parameter in `sonic_deploy_tar`. Please see the updated document and let me know what you think.

doc/bazel/bazel_migration_hld.md
blorente: I agree :D

doc/bazel/bazel_migration_hld.md
blorente: > Can't we install bazel on a build host and use it for all builds?

Yes, the Bazel downloads (and most other things) will be cached across builds in the same host. So a given CI host will only have to download Bazel and the rulesets once (and, ideally, every other build artifact as well).

For more context on Bazel specifically: We will be using Bazel through `bazelisk`, which manages Bazel versions and caches downloads: https://github.com/bazelbuild/bazelisk
Other deps will be cached by bazel itself on the host.

doc/bazel/bazel_migration_hld.md
blorente: Only loose estimates, but all 12 containers of the PoC could comfortably fit in around 60GB of disk space. This seems like a lot, but it has three caveats:
- This was through _a lot_ of heavy churn because I was experimenting with pretty core parts of the build, such as the toolchain and global gcc flags. These can blow up the cache size, because the entire subtree will need to be rebuilt. I expec this to be a rare case.
- Bazel itself is improving to manage this cache and garbage collection better. We're around two major releases behind, for good, but not unsurmountable reasons. This problem should get smaller as we upgrade Bazel versions.
- If and when we enable remote caching, the local on-disk caches will be reduced drastically, because Bazel can avoid downloading artifacts it doesn't need (e.g. if it already knows the result of the test because it's cached, it doesn't need to download the test).

doc/bazel/bazel_migration_hld.md
blorente: I've added some words to the effect. In practice, given that we already have several organizations interested and trying to submit contributions, I expect this to happen with or without our help.

doc/bazel/bazel_migration_hld.md
blorente: This is a really good point. Added to the plan.

doc/bazel/bazel_migration_hld.md
blorente: I agree. I don't think we should add it until we run against that particular wall. I trust that CI will catch most of the build breakages early enough anyway.

doc/bazel/bazel_migration_hld.md
blorente: @bocon13 Any thoughts on this before merging?

doc/bazel/bazel_migration_hld.md
blorente: @bocon13 Do you think this description is rigorous enough before merging this document, or should we dive deeper?

doc/bazel/bazel_migration_hld.md
jusherma: Perfect!

doc/bazel/bazel_migration_hld.md
jusherma: Excellent

doc/bazel/bazel_migration_hld.md
jusherma: Yeah, I think "same files with same permissions" is a pretty good heuristic that should catch most divergence. If we want more certainty, comparing hashes of files in the containers might be worthwhile too, but not sure that would be worth running regularly. Perhaps just once prior to phase 2.

I think drift during phase 1 is impossible to avoid completely, as it's less of a technical problem than a process one. Any changes people commit to the Makefile targets during phase 1 could introduce gaps between what make and bazel are actually building, even if both build successfully.

At this juncture, it may be difficult  to convince the whole community to keep bazel files in sync with all makefile changes they commit, the two general process approaches I see are:

1. During phase 1, the build working group keeps an eye on makefile changes that require corresponding bazel changes to keep parity and address them as they come
2. Just prior to phase 2, we reconcile all drift between the make and bazel systems

My vote is for option 1, as much as possible. These are not mutually exclusive options, of course.

I don't think this issue a blocker, per se, but it would be good to have a plan in mind, as option 2 could leave a lot of technical debt to address, depending on how much makefile churn there is during phase 1.

As you mentioned in another thread, limiting the time in phase 1 could help here, as long as we don't rush it and ensure everything is stable before making bazel the default.

How does this all sound?

doc/bazel/bazel_migration_hld.md
jusherma: That doesn't sound bad to me. Our plan internally will probably be to allocate a large, fixed amount of space for the cache, to get as much speed benefit as possible. So the rate that the cache fills up is—at least in our case—more about how much history we can save to speed up as many builds as possible.

doc/bazel/bazel_migration_hld.md
```diff
@@ -303,14 +306,18 @@ This section describes how individual components are built with Bazel, focusing
...
 - Fetch Debian packages deterministically, instead of relying on `apt install`. We do that by using `rules_distroless` to fetch from a Debian snapshot. [Source](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/MODULE.bazel#L23-L56). Please note that this only refers to the Bazel ruleset, and is **unrelated to distroless containers** as a concept. Switching to distroless containers is out of the scope of this document.
 
 > [!warning]
 > During the transition, Bazel will consume base layers from the Make-based build. This can lead to discrepancies in runtime dependencies if we don't keep the `rules_distroless` dependencies up to date.
 >
 > This is not a regression, but it is one more synchronization point that we'll need to keep up to date temporarily.
 
+Please note that this groundwork alone is not sufficient to guarantee full reproducibility. For instance, we'll still rely on `snapshot.debian.org` being available. However, this brings us close enough that we can start working on the Bazel build.
```
jusherma, line 327: Is removing dependencies on domains like snapshot.debian.org (and/or caching the items we download from them) in the future roadmap?

doc/bazel/bazel_migration_hld.md
jusherma: Gotcha

doc/bazel/bazel_migration_hld.md
jusherma: Oh that's excellent news!

doc/bazel/bazel_migration_hld.md
jusherma: Perfect, thanks!

doc/bazel/bazel_migration_hld.md
jusherma: Yes, I think it's ok to punt to a WG meeting as long as we have it tracked

doc/bazel/bazel_migration_hld.md
jusherma: Gotcha. As long as it's a bazel feature I don't have any specific concerns. I was mostly worried about it being (potentially brittle) code maintained in the sonic build infra

jusherma (approved):

doc/bazel/bazel_migration_hld.md
blorente: This sounds great!

> comparing hashes of files in the containers might be worthwhile too, but not sure that would be worth running regularly

Comparing hashes of files will be tricky, because the Make-based system is not hermetic to begin with. So, depending on _when_ you perform a Make build, the Bazel build may or may not produce the same hashes.

> My vote is for option 1, as much as possible. These are not mutually exclusive options, of course.

I agree. This is something I was planning on doing anyway, but you're absolutely right that it's nowhere in the document. Formalized it in the "CI Considerations" section.

doc/bazel/bazel_migration_hld.md
blorente: To clarify, this is for CI machines, right? If so, this is a good idea, and Bazel should be fairly good about GCing that shared cache. 

The point that may be a bit more problematic is developer workstations. They will _also_ have fairly big caches. There are ways to mitigate this, and a very reasonable thing is to `rm -rf` the whole cache and start over, but just note that some developers may notice an increase in disk usage. This is tweakable via Bazel flags. I'm more than happy to provide guidance, but the specifics will of course depend on how your developers like to work.

doc/bazel/bazel_migration_hld.md
blorente: It is not _yet_, but I'd like it to be. My main limitation is that I'm not familiar with how SONiC as a project wants to do shared community hosted services, but we should discuss that in the WG. I've added an Open question to that effect.

blorente: @bocon13 in case you have any thoughts on this.

bocon13: For internal dev / CI, we should absolutely be able to remove/rewrite/replace external registries (i.e. Debian, PyPI, GitHub), and Bazel supports this in a number of ways (e.g. downloader config, source cache).

For public builds, the primary benefit of this is to prevent build failures due to outages, removed artifacts, etc. We can certainly put it on the roadmap, but it needs to fit into the overall community CI infra strategy.

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,565 @@
...
+- Fetch Debian packages deterministically, instead of relying on `apt install`. We do that by using `rules_distroless` to fetch from a Debian snapshot. [Source](https://github.com/sonic-net/sonic-buildimage/blob/e09be005b19c3521c674e4415d08a25648fc15f4/MODULE.bazel#L23-L56). Please note that this only refers to the Bazel ruleset, and is **unrelated to distroless containers** as a concept. Switching to distroless containers is out of the scope of this document.
+
+> [!warning]
+> During the transition, Bazel will consume base layers from the Make-based build. This can lead to discrepancies in runtime dependencies if we don't keep the `rules_distroless` dependencies up to date.
+>
+> This is not a regression, but it is one more synchronization point that we'll need to keep up to date temporarily.
```
lunyue-ms, line 325: Currently, package versions are centrally managed under files/build/versions-public and updated by the UpgradeVersion pipeline. With Bazel, the version pins seem to be spread across MODULE.bazel, BUILD.bazel, direct download checksums, Debian snapshots, and lockfiles.
Will we have similar central version files or tooling to update and validate both Make and Bazel pins together, especially during Phases 1 and 2?

blorente: This took me a while to think through, great question. Here is my proposed plan:

Right now:
- `apt` and Python deps will live in `sonic-build-infra`, and users can check them by looking at `@sonic-build-infra//deps/{apt.bzl,python.txt}`.  This will be a centralized list, so no component may use an `apt` or Python dependency not listed here.
- Bazel modules we depend on can be inspected via `bazel mod graph`. This can be centralized further (more on that later), but since there is no parity to be kept, I don't think we should do it right now.
- Every version of every dependency that goes into the build (even raw http fetches) are be recorded into `MODULE.bazel.lock`. This is not intended to be human-readable. We can use it to tell that _something_ changed, but not to figure out what changed.

Before Phase 1:
- We can modify `sonic_deb` so that they report their versions in the same format as the current `files/build/versions` files.
- We can write a small script to use `bazel cquery` and extract the versions of `apt` deps from `@sonic-build-infra//deps/apt.bzl` (not by parsing that file, but rather by inspecting the Bazel graph in a principled way).

During Phase 1:
- We can add a new CI check that:
	- Collects the versions of `apt` and Python dependencies, compares them to `files/build/versions-public/build/build-sonic-slave-trixie`, and fails if they differ.
	- Fails if there are any changes in any of:
		-  `$ bazel mod graph` 
		- The files in `@sonic_build_infra//deps`.
		- `MODULE.bazel.lock`.
- This new CI check could be bypassed by a comment in the PR.
	- This guarantees that all changes are only performed explicitly, never accidentally.

In the future:
- If we want stronger guarantees, we could import all the relevant versions of Bazel modules into the local registry, and remove the BCR fallback.

I've tried to reflect this (with less detail) in this commit: https://github.com/sonic-net/SONiC/pull/2396/commits/5cb3c982cae690f6551656d4edaf05d4d831c375

Please let me know what you think!

blorente: Also, for SBOMs specifically, Bazel has quite a few tools that make this process easy. I'll make sure that the example PR has an example of SBOM generation before it lands.

lunyue-ms: Thanks for the detailed proposal. It makes sense to me. It looks like we have several ways to detect and control dependency/version drift during the migration.

lunyue-ms (dismissed):

doc/bazel/bazel_migration_hld.md
```diff
@@ -0,0 +1,569 @@
...
+Specifically:
+
+- Is the development experience in Bazel supported well enough to be used in day-to-day operations for migrated components and containers?
+    - Are there any gaps left to fill before this is true?
+- Should Bazel be allowed to be the _only option_ for certain containers (like it is today for `p4rt`)?
+- Should Bazel be allowed to handle the dependency graph for these containers? As a consequence, some SONiC dependencies will have to be converted to Bazel (e.g. `sonic-utilities`).
+
+If we decide there are still gaps, we will work to correct them before proceeding.
+
+If we decide to move forward, we will continue onto Phase 2.
```
yijingyan2, line 158: do we plan to define some performance metrics that we'll collect during phase 1 to validate that Bazel meets the requirements in section 5? maybe this could help inform phase 2 decision?

doc/bazel/bazel_migration_hld.md
bocon13:
```suggestion
- Talk from OCP SONiC Workshop 2025: https://www.youtube.com/watch?v=1YNmuzMqV24
- Talk from OCP EMEA Summit 2026: https://www.youtube.com/watch?v=uSKCNDWuXjc
```

doc/bazel/bazel_migration_hld.md
blorente: That is part of what I meant at the end of Phase 1: We should develop the criteria for Phase 2 as Phase 1 proceeds and developers have a chance to try it out. Once this is in the hands of users, we'll have a much better base to guess what types of checks we need.

bocon13 (approved):

lunyue-ms (approved):

PR #2398, opened 2026-06-18
Add VPP SAI Unit Test Framework HLD

yue-fred-gao:
### why
vpp SAI layer doesn't have unit test. There is no protection on the existing code. We can use sonic-mgmt to achieve some sanity check and get some code coverage. But it is a long chain of operations and hard to get higher code coverage.

### what this PR does
Propose a new unit test framework that leverage existing infrastructure as much as possible. The solution uses sai test infra (saiserver, ptf tester etc) in SAI but plugin vpp and vpp sai layer as dataplane. The existing test cases in sai test can be reused. It also lays out plans that go beyond the sai test, include notification verification, verification of punt/inject to control plane and data driven test framework

### DIFF
--- a/doc/vpp/saivpp-unittest-hld.md
+++ b/doc/vpp/saivpp-unittest-hld.md
@@ -1,0 +1,702 @@
+# HLD: VPP SAI Unit Test Framework
+Rev v0.1
+
+## Table of Contents
+
+1. [Objectives](#1-objectives)
+2. [Scope](#2-scope)
+3. [Overview](#3-overview)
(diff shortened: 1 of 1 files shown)


yue-fred-gao:
hi @vitor-arista, we decided to use SONiC as central location for sonic-vpp HLDs, same as other HLD. Can you please approve this PR? There is no change comparing to the approved version in sonic-platform-vpp repo. Thanks!

nicholasching (approved):
Looks good to me!

vitor-arista:
LGTM too.

PR #2409, opened 2026-06-24
docs: detach member instead of admin-down for egress-disable HLD

cyw233:
## What

Updates the SONiC-VPP LAG Member Egress Disable HLD to **v0.3**: model `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE` by **detaching the member from the VPP bond** (while keeping its link administratively up), replacing the v0.2 approach that set the member port **admin state down**.

### Architecture

In SONiC-VPP:
- The VPP bond (`BondEthernet`) is created in **XOR mode**; VPP itself does **not** run LACP.
- **LACP is run by teamd in Linux**, over each LAG member's LCP tap (e.g. `Ethernet0`).
- LACP control PDUs are punted/injected through the member's VPP hardware-interface, and teamd's per-member link-watch (ethtool) follows that hardware-interface's link state.

So a member hardware-interface is **simultaneously** (1) the LACP control-plane path for that member and (2) the bond's egress member. Anything that takes the member link down also stops LACP for that member.

### The failure

Setting the member admin state **down** to represent egress-disable also drops the LACP control path. This is fatal at bring-up, because SONiC creates **every** LAG member with `EGRESS_DISABLE=true` *before* LACP has selected it (the initial unselected state, where teamd has not yet enabled collection/distribution):

1. Member created egress-disabled → its link is forced down.
2. Link down → no LACP PDUs → teamd can never select the member.
3. Never selected → SONiC never sets `EGRESS_DISABLE=false`.
4. Member stuck down permanently → the PortChannel and any BGP session over it never come up.

The same deadlock blocks any later LACP re-selection, since the link must be up to renegotiate.

### Empirical validation (vlab-vpp-01)

With an admin-down build loaded, all uplink (T2) LACP PortChannels stayed down and their BGP sessions never established, while single-port (T0) BGP was unaffected:

- `show interfaces portchannel` → all `LACP(A)(Dw)`, members deselected.
- `teamdctl PortChannel105` → the admin-down member showed `link: down, selected: no, state: disabled`.
- Bringing the members **up** → LACP converged, PortChannel `Up`, BGP came up (6370 prefixes).
- **Detaching** a member while keeping its link up → teamd kept it `Selected` and BGP stayed up.

### Why detach is the only viable VPP mechanism

- VPP exposes **no per-member egress-only disable** on an XOR bond.
- Member `weight` only applies to active-backup mode; VPP rejects it in XOR: `set interface bond: Weight valid for active-backup only` (verified on the DUT).
- **Detach** removes the member from egress distribution while keeping its link up, so the teamd-run LACP control plane keeps working. This mirrors how a hardware LAG implements egress-disable: the member link stays up and LACP-trapped, while the member is removed from the egress distribution set.

## v0.2 concerns about detach are addressed

The reasons v0.2 gave for rejecting detach are handled by this design:

- **Re-add must not re-evaluate LCP/tap** — re-attach calls only `create_bond_member()` and never re-enters `vpp_ensure_lag_lcp()`, so LCP/tap and the bond MAC are untouched.
- **Member attributes on re-add** — the bond index and member port are looked up from stored SAI object state, not reconstructed.
- **Remove-while-already-detached** — handled explicitly by tracking egress-disabled members and skipping the redundant detach on remove.

### DIFF
--- a/doc/vpp/vpp-lag-member-egress-disable-HLD.md
+++ b/doc/vpp/vpp-lag-member-egress-disable-HLD.md
@@ -67,6 +67,23 @@ The existing VPP LAG design creates the VPP LAG bond first, then adds member por
 
 One important detail is that LCP/tap creation is intentionally delayed until the first enabled member is added. Creating the tap too early, while the VPP bond has no active member, can leave the tap with a stale generated MAC address instead of the first member's MAC address. When the first member is added, the VPP bond adopts that member's MAC, so delaying tap creation until then ensures the tap, bond, and first-member MAC addresses all match.
 
+### Egress disable in the data plane
+
+`SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE` controls only **data-plane egress distribution** for a LAG member. It is the egress half of the IEEE 802.1AX *collecting / distributing* model that SONiC uses to represent a member's LACP selection state, and it has an ingress counterpart:
+
+| SAI attribute | 802.1AX role | Direction | Effect when set to `true` |
+|---|---|---|---|
+| `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE` | Distributing off | Egress (Tx data frames) | Member is removed from the LAG's egress distribution (hash) set |
+| `SAI_LAG_MEMBER_ATTR_INGRESS_DISABLE` | Collecting off | Ingress (Rx data frames) | Member stops accepting data frames into the LAG |
+
+SONiC drives these attributes from teamd's LACP state: a member is added with both disabled while it is still unselected, and Orchagent clears them once teamd reports the member as collecting and distributing. SONiC's expectation for `EGRESS_DISABLE=true` is therefore deliberately narrow:
+
+- **Only data frames are affected.** The member is taken out of the port channel's egress traffic distribution, so the forwarding plane no longer hashes data traffic onto it. `EGRESS_DISABLE` is the egress counterpart of `INGRESS_DISABLE`; it is **not** a port shutdown and **not** a "drop all egress" rule.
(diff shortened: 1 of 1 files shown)


doc/vpp/vpp-lag-member-egress-disable-HLD.md
yue-fred-gao, line 18: Can you add a section to describe SONiC's expectation on egress disable in data plane? LACP continues to run over the member interface. The interface is taken out of traffic distribution of the port channel. egress disable sounds like dropping all egress packets. This is not true otherwise LACP will stop working, which is still required based on your description.

cyw233: I've added a new "Egress disable in the data plane" subsection to make SONiC's expectation explicit.

In short, `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE` is the egress (Distributing) half of the 802.1AX collecting/distributing model, paired with `SAI_LAG_MEMBER_ATTR_INGRESS_DISABLE` (Collecting). Setting it true:

- Only affects data frames - the member is removed from the LAG's egress distribution (hash) set. It is not a port shutdown and not a "drop all egress" rule.
- Keeps the link up and LACP running - LACPDUs are control-plane traffic that keeps flowing on the member regardless of distribution state, so teamd can still select/deselect it. If egress-disable dropped all egress, LACPDUs couldn't be sent and the member could never be (re)selected.
- Is a normal transient state - SONiC adds every member with `EGRESS_DISABLE=true` (and `INGRESS_DISABLE=true`) before LACP selects it, then teamd clears both once the member is collecting + distributing.

This is exactly the contract the detach approach satisfies (remove from egress distribution while keeping the link up and LACP alive), which is why admin-down doesn't work here.

yue-fred-gao: thanks

yue-fred-gao (approved):

PR #2413, opened 2026-06-25
Resync leak detection docs with each other

fraserg-arista:
The two documents that describe liquid cooling and leak detection have diverged from each other. This change brings them back into alignment with themselves and with the code that implements them.

Both documentations have been updated to reflect that they are living design documents and not proposals to make a change.

The style of the two documents has been unified to make them easier to read together (e.g the API spec is now in table form in both instead of a mix with pseudocode).

The leak detection HLD defers to the PMON doc for describing policy and the larger context of thermalctld/bmcctld but it remains the source of truth for the liquid cooling APIs.


fraserg-arista:
My intention is to make sure the docs and code are consistent with each other. There shouldn't be any behaviour changes.

The diff for leakage_detection_hld.md is pretty hard to read, unfortunately. The original content should all still be there but with the extra info that was added and tense change (present rather than a future proposal), a lot of the lines were touched.

@yuazhe could you take a look and tag anyone else at Nvidia who should review this?

doc/bmc/leakage_detection_hld.md
```diff
@@ -1,156 +1,267 @@
...
 ```
 
-### the GNMI event model
-EVENTS_PUBLISHER_SOURCE = "sonic-events-host"
+### 6.3 Aggregate System Leak State
 
-EVENTS_PUBLISHER_TAG = "liquid-cooling-leak"
+The system leak result is stored in `STATE_DB` and consumed by `bmcctld`.
 
-## 7. CLI management
```
fraserg-arista, line 125: The CLI spelling change has been brought up as a breaking change for some implementers.

Our docs and code are inconsistent on 'leak' vs 'leakage' for the show command; since the code currently uses 'leak' as the keyword, that's what I've used in this doc. Maybe we could add 'leakage' as an alias for compatibility?

fraserg-arista:
@judyjoseph fyi

doc/bmc/leakage_detection_hld.md
```diff
@@ -1,156 +1,267 @@
-# Liquid Cooling leakage detection in SONiC
```
judyjoseph, line 1: A general comment: I am not sure if we would like to move this doc as applicable to SONiC BMC .. This was a generic implementation where BMC could be running openBMC and these API's were for Sonic running in Switch-host

doc/bmc/leakage_detection_hld.md
```diff
@@ -1,156 +1,267 @@
...
 
-```
-Defines a logical structure for liquid cooling devices, with keys for various sensors.
+`LeakageSensorBase` provides the per-sensor leak interface used by `thermalctld`.
+
+| Method | Description |
+|--------|-------------|
+| `get_name()` | Returns the leak sensor name |
+| `is_leak()` | Returns whether a stable leak condition is detected |
+| `is_leak_sensor_ok()` | Returns whether the sensor is healthy or faulty |
```
judyjoseph, line 45: these data (and a few more below) is probably duplicated - not sure if this needs to be repeated again. We could just give reference to https://github.com/sonic-net/SONiC/blob/master/doc/bmc/sonicBMC/pmon-bmc-design.md#224-platform-apis ?

PR #2415, opened 2026-06-26
Added doc that describes dual U-Boot environment synchronization

luka-micas:
1st draft,added doc that describes dual U-Boot environment synchronization.


doc/bmc/sonicBMC/dual-uboot-bmc-design.md
```diff
@@ -0,0 +1,431 @@
...
+- `fit_name`
+- `sonic_version_1`
+- `linuxargs`
+- `bootcmd`
+
+Verify that the alternate environment is synchronized immediately after the primary update flow succeeds.
+
+### Test 7: Flash Switchover
+
+After a successful update, switch to the alternate SPI flash.
```
adamcalabrigo: What are you expecting regarding SONiC's mechanism for switching flash boot source? Does this proposal expect AST2700's ABR functionality to be enabled? If so, do you propose that SONiC include a generic mechanism to handle the boot watchdog timer that Aspeed uses to implement the boot recovery?

luka-micas: Thanks for the comment. The flash boot source switching mechanism (including ABR/watchdog behavior) is a platform boot flow before SONiC-BMC starts and is outside the scope of this HLD.
This design only focuses on synchronizing the dual U-Boot environment partitions to keep boot metadata consistent after a possible boot source switch.

PR #2417, opened 2026-06-27
Update vlan_bvi_hld with new design of punting DHCP packet

yue-fred-gao:
### why
when I ran dhcp_relay sonic-mgmt tests, I found sonic expects DHCP broadcast packets like discovery are punted to SONIC only. It has a specific test case to check if dhcp broadcast packets are not flooded in the BD: test_dhcp_broadcast_not_flooded. 

### what this PR does
Update the HLD to use l2-input-classify to match on dhcp packets (udp to port 68) tagged and untagged and punt the packets via lcp. Also, the classifier only matches on dport because sport can be random after NAT.


AkeelAli (approved):
LGTM. PR description says "udp to port 68", should be 67.

PR #2423, opened 2026-06-30
doc/acl: document the POLICER_ACTION ACL rule action

AnantKishorSharma:
Documents the `POLICER_ACTION` ACL rule action, which binds a CONFIG_DB `POLICER` entry to an ACL rule and is realized as `SAI_ACL_ENTRY_ATTR_ACTION_SET_POLICER`.

Updates `doc/acl/ACL-High-Level-Design.md`:
- Adds `policer_action` to the ACL_RULE actions table.
- Adds a design note covering the SAI mapping, policer reference-counting (held for the rule lifetime — acquired before the SAI ACL entry and released on rule removal or entry-creation failure), and deferred rule creation until the referenced policer exists.
- Adds a revision entry (Rev 1.2).

Related PRs:

| Repo | PR title | State |
| ---- | -------- | ----- |
| sonic-swss | [orchagent: compose POLICER_ACTION into the ACL rule packet action](https://github.com/sonic-net/sonic-swss/pull/4666) | state |
| sonic-buildimage | [sonic-yang-models: POLICER_ACTION leafref](https://github.com/sonic-net/sonic-buildimage/pull/27859) | state |
| sonic-mgmt | [tests: POLICER_ACTION ACL action](https://github.com/sonic-net/sonic-mgmt/pull/25896) | state |

Feature request: sonic-net/sonic-buildimage#27857.

### DIFF
--- a/doc/acl/ACL-High-Level-Design.md
+++ b/doc/acl/ACL-High-Level-Design.md
@@ -1,6 +1,6 @@
 # ACL in SONiC
 # High Level Design Document
-### Rev 1.1
+### Rev 1.2
 
 # Table of Contents
   * [List of Tables](#list-of-tables)
@@ -88,6 +88,7 @@
 | 0.3 | 10-Nov-2016 | Andriy Moroz       | Updated according to the comments   |
 | 0.4 | 20-Dec-2016 | Oleksandr Ivantsiv | Update data structures              |
 | 1.1 | 08-Apr-2025 | Anish Narsian      | VXLAN inner src mac rewrite support |
+| 1.2 | 30-Jun-2026 | Anant Kishor Sharma | POLICER_ACTION (policer/rate-limit) ACL rule action |
 # About this Manual
 This document provides general information about the ACL feature implementation in SONiC.
 # Scope
@@ -568,6 +569,14 @@ to class AclRuleMirror to indicate the stage the ACL mirror rule, according to t
 action, "SAI_ACL_ENTRY_ATTR_ACTION_MIRROR_INGRESS" for ingress ACL rule, "SAI_ACL_ENTRY_ATTR_ACTION_MIRROR_EGRESS"
 for egress ACL rule.  
 Add possibility to receive updates about mirror sessions state change and perform mirroring rules state change accordingly.
+
+To meter/rate-limit ACL-matched traffic, the policer action (keyword "POLICER_ACTION") is realized by class "AclRulePolicer",
+which resolves the referenced POLICER table entry to its SAI policer object and programs "SAI_ACL_ENTRY_ATTR_ACTION_SET_POLICER"
+on the ACL entry. "AclRulePolicer" holds a reference on the bound policer for the lifetime of the rule: the reference is
+acquired before the SAI ACL entry is created and released when the rule is removed (or if entry creation fails), so a
+POLICER cannot be deleted while an ACL rule binds it. If the referenced policer does not exist yet, rule creation is deferred
+and retried (via the Consumer m_toSync mechanism) until the policer appears, mirroring the existing "wait for ACL table" behavior.
+
 # 4 Flows
 ## 4.1 Creating of ACL Objects
 ![](acl_create.png)
@@ -629,6 +638,7 @@ Ansible + PTF
 |packet_action | ACL Rule property. Packet actions "forward" or "drop". Valid for rules in "L3" tables only
 |mirror_action | Action "mirror". Valid for rules in "mirror" tables only
 |inner_src_mac_rewrite_action | Action to rewrite the inner src mac rewrite field.
+|policer_action | Action to meter/rate-limit matching packets via a referenced POLICER table entry. Maps to SAI_ACL_ENTRY_ATTR_ACTION_SET_POLICER
 *Keywords derived from the SAI ACL attributes.*
 # Appendix B: Sample input json file
 ```
--- a/doc/acl/ACL-High-Level-Design.md
+++ b/doc/acl/ACL-High-Level-Design.md
@@ -570,9 +570,10 @@ action, "SAI_ACL_ENTRY_ATTR_ACTION_MIRROR_INGRESS" for ingress ACL rule, "SAI_AC
 for egress ACL rule.  
 Add possibility to receive updates about mirror sessions state change and perform mirroring rules state change accordingly.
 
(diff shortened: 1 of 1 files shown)


prsunny (dismissed):
lgtm,

rdhegde:
With reference to https://github.com/sonic-net/sonic-swss/pull/4666
In AclRulePacket::createRule()

Should you not increment the policer reference count only once the createRule() is successfull
What is the need for "if ( !m_policerRefHeld )" before incrementing the policer reference count ? On the contrary if m_policerRefHeld is true in this routine then it should be considered as an error.

rdhegde:
With reference to https://github.com/sonic-net/sonic-swss/pull/4666
Instead of accessing gPolicerOrch, why not store it in AclOrch like how the others such as m_mirrorOrch, m_neighOrch ... are stored

rdhegde:
With reference to https://github.com/sonic-net/sonic-swss/pull/4666
In AclOrch::doAclRuleTask(), we should wait for policer creation only when all the other rule attributes are valid else we will end up waiting for a policer creation when the rule itself is invalid

rdhegde:
Is there a plan to add CLI and yang support for the policer creation ?

AnantKishorSharma:
> With reference to [sonic-net/sonic-swss#4666](https://github.com/sonic-net/sonic-swss/pull/4666) In AclRulePacket::createRule()
> 
> Should you not increment the policer reference count only once the createRule() is successfull What is the need for "if ( !m_policerRefHeld )" before incrementing the policer reference count ? On the contrary if m_policerRefHeld is true in this routine then it should be considered as an error.

The reference is taken before `AclRule::createRule()` and released in the failure path, so the count stays balanced: if `increaseRefCount()` fails (only reachable if the policer disappeared after `validate()`) we return before any SAI entry is created — nothing to unwind — and if entry creation itself fails we release the reference. Incrementing only on success would also be correct; I took it before-create so a refcount failure never leaves a half-created entry. Both are leak-free. On the guard: `!m_policerRefHeld` keeps the count balanced if `createRule()` is re-entered without an intervening `removeRule()` (which clears the flag). If we confirm that path never occurs, treating a held reference as an error would be reasonable.

AnantKishorSharma:
> With reference to [sonic-net/sonic-swss#4666](https://github.com/sonic-net/sonic-swss/pull/4666) Instead of accessing gPolicerOrch, why not store it in AclOrch like how the others such as m_mirrorOrch, m_neighOrch ... are stored

An `m_policerOrch` member would match `m_mirrorOrch`/`m_neighOrch`. `gPolicerOrch` is a process-global singleton like the other `g*Orch` handles, so it's equivalent and correct here. No functional issue.

AnantKishorSharma:
> With reference to [sonic-net/sonic-swss#4666](https://github.com/sonic-net/sonic-swss/pull/4666) In AclOrch::doAclRuleTask(), we should wait for policer creation only when all the other rule attributes are valid else we will end up waiting for a policer creation when the rule itself is invalid

The policer deferral runs before the rule is validated, so an invalid rule that also references a missing policer is deferred instead of reported immediately. Not a mis-programming risk, it's rejected once the policer appears (and any rule with a missing policer is deferred regardless), so only the error is delayed unless you see a failure mode I'm missing.

AnantKishorSharma:
> Is there a plan to add CLI and yang support for the policer creation ?

YANG support for policer creation is already in place: the `POLICER` table is modeled by `sonic-policer.yang`, with YANG model tests, so `POLICER` config is YANG-validated today. 

On the CLI side there's no dedicated `config policer` create command, a `config policer` convenience command could be a follow-up if there's interest.

rdhegde:
> > With reference to [sonic-net/sonic-swss#4666](https://github.com/sonic-net/sonic-swss/pull/4666) In AclRulePacket::createRule()
> > Should you not increment the policer reference count only once the createRule() is successfull What is the need for "if ( !m_policerRefHeld )" before incrementing the policer reference count ? On the contrary if m_policerRefHeld is true in this routine then it should be considered as an error.
> 
> The reference is taken before `AclRule::createRule()` and released in the failure path, so the count stays balanced: if `increaseRefCount()` fails (only reachable if the policer disappeared after `validate()`) we return before any SAI entry is created — nothing to unwind — and if entry creation itself fails we release the reference. Incrementing only on success would also be correct; I took it before-create so a refcount failure never leaves a half-created entry. Both are leak-free. On the guard: `!m_policerRefHeld` keeps the count balanced if `createRule()` is re-entered without an intervening `removeRule()` (which clears the flag). If we confirm that path never occurs, treating a held reference as an error would be reasonable.

The first part of the comment was to ensure that you take reference only when everything is done rather than take it and then clear it. Also, taking the reference is not needed for createRule() so why take it then. Agree there is no functional issue, just looks redundant. 
Regarding the second part the question comes back to why `createRule()` is re-entered without an intervening `removeRule()` I do not see any reason for that to happen. Actually this should never happen. So adding an error will highlight such case. The current approach hides it.

rdhegde:
> > With reference to [sonic-net/sonic-swss#4666](https://github.com/sonic-net/sonic-swss/pull/4666) Instead of accessing gPolicerOrch, why not store it in AclOrch like how the others such as m_mirrorOrch, m_neighOrch ... are stored
> 
> An `m_policerOrch` member would match `m_mirrorOrch`/`m_neighOrch`. `gPolicerOrch` is a process-global singleton like the other `g*Orch` handles, so it's equivalent and correct here. No functional issue.

There is no functional issue but isn't it better to avoid exposing globals when there is a choice ? Also, there is already a precedent for this in terms of how the other dependent orch's do it, so I thought we should just follow the same.

rdhegde:
> > With reference to [sonic-net/sonic-swss#4666](https://github.com/sonic-net/sonic-swss/pull/4666) In AclOrch::doAclRuleTask(), we should wait for policer creation only when all the other rule attributes are valid else we will end up waiting for a policer creation when the rule itself is invalid
> 
> The policer deferral runs before the rule is validated, so an invalid rule that also references a missing policer is deferred instead of reported immediately. Not a mis-programming risk, it's rejected once the policer appears (and any rule with a missing policer is deferred regardless), so only the error is delayed unless you see a failure mode I'm missing.

This is a miss in terms that an invalid rule is in the queue and constantly retrying create until the policer appears. I feel it would be better to just move the deferral code below, so if there is an error then we flag it immediately.

PR #2425, opened 2026-06-30
Initial draft of CONFIG DB Data Encryption Master Key Management

fredxia-nexthop:
This document describes the design of a Master Key Encryption Infrastructure for SONiC that enables at-rest encryption of sensitive configuration fields in ConfigDB (Redis). It defines the general infrastructure — master key management, an AES-GCM encryption library, a code-defined encryption registry, and a transparent ConfigDB interception layer — and then describes how BGP MD5 peer authentication passwords are the first application built on top of this infrastructure.

### DIFF
--- a/doc/master-key-encryption/master-key-encryption.md
+++ b/doc/master-key-encryption/master-key-encryption.md
@@ -35,6 +35,7 @@
 |-----|------------|-----------|--------------------|
 | 0.1 | 2026-06-16 | Fred Xia  | Initial draft      |
 | 0.2 | 2026-08-06 | Fred Xia  | Second draft       |
+| 0.3 | 2026-08-06 | Fred Xia  | Third draft        |
 
 ---
 
@@ -121,7 +122,7 @@ Our implementation follows the Type-6 (reversible AES) model, using AES-256-GCM 
  │  │    master-key-manager CLI         │   │   Encryption Library          │  │
  │  │                                   │   │ (master_key_manager.py)       │  │
  │  │  set --master-key-file | --name   │   │  AES-256-GCM (Type-6)         │  │
- │  │  encrypt --table / decrypt --table│   │  Nonce + CT + AAD             │  │
+ │  │  encrypt/decrypt --file|--string  │   │  Nonce + CT + AAD             │  │
  │  │  activate/deactivate --name|--all │   │  base64-encoded output        │  │
  │  │  status / list                    │   └────────────────┬──────────────┘  │
  │  └──────────────┬────────────────────┘                    │                 │
@@ -435,14 +436,21 @@ Commands:
         master-key-manager set --master-key-file /etc/sonic/bgp_master_key mySecretKey
         master-key-manager set --name BGP mySecretKey
 
-  encrypt --table <TABLE> [-o <OUTPUT>]
-      Read all entries of <TABLE> from ConfigDB (or --config-file),
-      encrypt every registered secret field, and print the result as JSON
-      (or write to -o).  Table must appear in the encryption registry.
+  encrypt (--file <JSON> | --name <NAME> --string <STRING>) [-o <OUTPUT>]
+      Two mutually exclusive forms:
+      --file <JSON>
+          Read a CONFIG DB JSON file, encrypt every registered secret field
+          (all registry entries are applied; entries whose tables are absent
+          from the file are skipped), and print the resulting JSON (or write
+          it to -o).  The input file is not modified.  Every registry entry
+          with a table present in the file must have a provisioned master key.
+      --name <NAME> --string <STRING>
+          Encrypt a single string with the master key of the registry entry
+          named <NAME> and print the resulting blob (or write it to -o).
 
-  decrypt --table <TABLE> [-o <OUTPUT>]
-      Read all entries of <TABLE>, decrypt every registered secret field,
-      and print the result as JSON (or write to -o).
+  decrypt (--file <JSON> | --name <NAME> --string <STRING>) [-o <OUTPUT>]
+      Same two forms, decrypting instead: a CONFIG DB JSON file back to
+      plaintext, or a single encrypted blob back to its plaintext string.
 
   activate (--name <NAME> | --all)
       Activate encryption for the registry entry selected by --name, or for
@@ -699,7 +707,7 @@ A separate proposal (`sonic-py-common/sonic_py_common/security_cipher.py`) was s
 |-----------|-------------|
 | `tests/master_key_encryption_test.py` | AES-GCM deterministic test vectors; `MasterKeyManager` key update, encrypt, decrypt, history; max 8 keys; file permission enforcement |
 | `tests/config_db_encryptor_test.py` | `ConfigDBEncryptor` interception: `set_entry`, `mod_entry`, `mod_config` hooks; encryption activated/deactivated; multiple field names |
-| `tests/master_key_manager_test.py` | All CLI subcommands: `set` (`--master-key-file`/`--name`), `encrypt --table`, `decrypt --table`, `activate`/`deactivate` (`--name`/`--all`), `status`, `list`; custom registry via `--registry-config` |
+| `tests/master_key_manager_test.py` | All CLI subcommands: `set` (`--master-key-file`/`--name`), `encrypt`/`decrypt` (`--file` or `--name` + `--string`), `activate`/`deactivate` (`--name`/`--all`), `status`, `list`; custom registry via `--registry-config` |
 
 ### System Tests (Testbed)
 
--- a/doc/master-key-encryption/master-key-encryption.md
+++ b/doc/master-key-encryption/master-key-encryption.md
@@ -1,0 +1,709 @@
+# Master Key Encryption Infrastructure for SONiC
+
+## Table of Contents
+
+- [Revision History](#revision-history)
+- [Scope](#scope)
+- [Definitions and Abbreviations](#definitions-and-abbreviations)
+- [Overview](#overview)
+- [Motivation](#motivation)
+- [Requirements](#requirements)
+- [Architecture](#architecture)
+- [Master Key Manager Infrastructure](#master-key-manager-infrastructure)
+  - [Master Key File](#master-key-file)
+  - [Master Key Generation and Provisioning](#master-key-generation-and-provisioning)
+  - [Key Rotation](#key-rotation)
+  - [Encryption Library](#encryption-library)
+  - [ConfigDB Interception Layer](#configdb-interception-layer)
+  - [Encryption Registry: Static Configuration](#encryption-registry-static-configuration)
+- [CLI: master-key-manager](#cli-master-key-manager)
+- [Application: BGP MD5 Password Encryption](#application-bgp-md5-password-encryption)
+  - [Registration](#registration)
+  - [Operational Workflow](#operational-workflow)
+  - [FRR Integration](#frr-integration)
+- [Comparison With Prior Proposal](#comparison-with-prior-proposal)
+- [Warmboot and Fastboot Design Impact](#warmboot-and-fastboot-design-impact)
+- [Restrictions and Limitations](#restrictions-and-limitations)
+- [Test Plan](#test-plan)
+- [Future Work](#future-work)
+
+---
+
+## Revision History
+
+| Rev | Date       | Author    | Change Description |
+|-----|------------|-----------|--------------------|
+| 0.1 | 2026-06-16 | Fred Xia  | Initial draft      |
+
+---
+
+## Scope
+
+This document describes the design of a **Master Key Encryption Infrastructure** for SONiC that enables at-rest encryption of sensitive configuration fields in ConfigDB (Redis). It defines the general infrastructure — master key management, an AES-GCM encryption library, a code-defined encryption registry, and a transparent ConfigDB interception layer — and then describes how BGP MD5 peer authentication passwords are the first application built on top of this infrastructure.
+
+---
+
+## Definitions and Abbreviations
+
+| Term               | Meaning                                                              |
+|--------------------|----------------------------------------------------------------------|
+| AES-GCM  | Advanced Encryption Standard in Galois/Counter Mode       |
+| AAD      | Additional Authenticated Data (used in AES-GCM)          |
+| Type-6   | Cisco-compatible password encryption type: reversible AES |
+| ConfigDB | SONiC configuration database (Redis)                      |
+
+---
+
+## Overview
+
+SONiC stores all configuration — including protocol passwords such as BGP MD5 authentication passwords, TACACS shared secrets, and RADIUS keys — as cleartext in Redis (ConfigDB) and in the `config_db.json` startup file. Any local account can dump these values with standard Redis tools. This design proposes a general-purpose Master Key Encryption Infrastructure to protect such fields at rest, with BGP MD5 password encryption as its first application.
+
+The infrastructure provides:
+
+1. **A code-defined encryption registry** (`master_key_manager/master_key_encryption_config.py`) that statically declares which tables and fields are subject to automatic encryption. Adding a new encrypted field requires a code change (not a YANG model edit or CLI registration), which is deliberate: any new use of encryption must also involve feature-specific handling code (e.g., decryption in `frrcfgd` before rendering to FRR).
+2. **A master key management library and CLI tool** (`master-key-manager`) to provision master keys, activate/deactivate encryption, and inspect status.
+3. **An AES-256-GCM encryption library** (`master_key_manager/master_key_encryption.py`) offering authenticated symmetric encryption (Type-6, reversible).
+4. **A transparent ConfigDB interception layer** so that standard SONiC tools (`config`, `sonic-cfggen`) automatically encrypt sensitive fields on write without requiring application-level changes.
+5. **Extensibility** so that additional features (TACACS, RADIUS, LDAP) can be supported by adding an entry to the registry and the corresponding decrypt/render code in the feature daemon.
+
+---
+
+## Motivation
+
+### At-Rest Security
+
+Community SONiC writes all configuration values to Redis plaintext. An operator or compromised process with read access to the local Redis socket can trivially extract BGP MD5 passwords, TACACS shared secrets, RADIUS keys, and similar credentials. FIPS-compliant deployments and customer security policies require these values to be stored encrypted.
+
+### Industry Precedent
+
+Proprietary vendors address this with a "Type-6" (or equivalent) reversible AES encryption tied to a device-local master key:
+
+- **Cisco IOS**: `password encryption aes` (Type-6). Once enabled, all service passwords and BGP neighbor passwords are stored AES-encrypted.
+- **Arista EOS**: `password encryption reversible aes-256-gcm`. Uses a 32-byte master key and AES-256-GCM for at-rest encryption.
+- **Dell Enterprise SONiC**: Provides similar master key encryption for BGP and AAA passwords.
+
+Our implementation follows the Type-6 (reversible AES) model, using AES-256-GCM for both confidentiality and integrity.
+
+---
+
+## Requirements
+
+### Functional Requirements
+
+1. **FR-1**: ConfigDB fields designated as sensitive shall be stored encrypted using AES-256-GCM with a device-local master key.
+2. **FR-2**: Encryption and decryption shall be transparent to standard SONiC configuration tools (`config load`, `config reload`, `config apply`).
+3. **FR-3**: The system shall support master keys supplied by a central controller, enabling enterprise key management integration.
+4. **FR-4**: A CLI tool (`master-key-manager`) shall allow the operator to set master keys and activate/deactivate encryption.
+5. **FR-5**: When encryption is activated, all plaintext secret fields in registered tables shall be re-encrypted automatically; when deactivated, they shall be decrypted.
+6. **FR-6**: Encryption shall apply only to ConfigDB. FRR running configuration continues to use cleartext passwords.
+7. **FR-7**: The master key file shall be accessible only to root (mode 0600), protected by filesystem permissions.
+8. **FR-8**: The infrastructure shall retain up to 8 historical master keys to enable decryption of values encrypted under a previous key.
+9. **FR-9**: BGP MD5 passwords shall be the first feature protected by this infrastructure.
+10. **FR-10**: The catalog of tables and fields subject to encryption shall be defined in code (`master_key_encryption_config.py`). Adding encryption to a new table requires a code change, not a YANG model edit or CLI command, because any new encrypted field also requires feature-specific decrypt/render code in the consuming daemon.
+11. **FR-11**: A single table may declare multiple encrypted fields (e.g., both `auth_password` and `md5_key`).
+
+### Non-Functional Requirements
+
+1. The master key file format shall be a versioned, human-readable JSON structure for auditability and disaster recovery.
+2. Encryption/decryption shall use the Python `cryptography` library — no shell subprocesses.
+3. One `MasterKeyManager` instance per master key file, shared across all tables that point to the same file.
+
+---
+
+## Architecture
+
+```
+ ┌─────────────────────────────────────────────────────────────────────────────┐
+ │                    Master Key Encryption Infrastructure                     │
+ │                                                                             │
+ │  ┌───────────────────────────────────┐   ┌───────────────────────────────┐  │
+ │  │    master-key-manager CLI         │   │   Encryption Library          │  │
+ │  │                                   │   │ (master_key_manager.py)       │  │
+ │  │  set  --master-key-file  <key>    │   │  AES-256-GCM (Type-6)         │  │
+ │  │  encrypt --table / decrypt --table│   │  Nonce + CT + AAD             │  │
+ │  │  activate / deactivate            │   │  base64-encoded output        │  │
+ │  │  status / list                    │   └────────────────┬──────────────┘  │
+ │  └──────────────┬────────────────────┘                    │                 │
+ │                 │ reads                  ┌────────────────▼──────────────┐  │
+ │                 ▼                        │   MasterKeyManager            │  │
+ │  ┌───────────────────────────────────┐   │   /etc/<feature>_master_key   │  │
+ │  │  master_key_encryption_config.py  │   │   mode 0600, JSON             │  │
+ │  │  (code-defined, static registry)  │   │   up to 8 historical keys     │  │
+ │  │  • name  (e.g. "BGP")             │   │   enabled flag (on/off state) │  │
+ │  │  • master_key_file                │   │   one instance per file       │  │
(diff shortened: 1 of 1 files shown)


nmoray:
@fredxia-nexthop I guess we already have similar sort of infra in the SONIC. Please refer to the following PRs.

https://github.com/sonic-net/sonic-buildimage/pull/22711
https://github.com/sonic-net/sonic-buildimage/pull/17201

Anders from Nexthop already knew about this feature. :-)

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+
+---
+
+## Definitions and Abbreviations
+
+| Term               | Meaning                                                              |
+|--------------------|----------------------------------------------------------------------|
+| AES-GCM  | Advanced Encryption Standard in Galois/Counter Mode       |
+| AAD      | Additional Authenticated Data (used in AES-GCM)          |
+| Type-6   | Cisco-compatible password encryption type: reversible AES |
```
bhouse-nexthop: I'd recommend dropping any mention of "Type-6" it really doesn't have any relation to SONiC.

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+
+The infrastructure provides:
+
+1. **A code-defined encryption registry** (`master_key_manager/master_key_encryption_config.py`) that statically declares which tables and fields are subject to automatic encryption. Adding a new encrypted field requires a code change (not a YANG model edit or CLI registration), which is deliberate: any new use of encryption must also involve feature-specific handling code (e.g., decryption in `frrcfgd` before rendering to FRR).
+2. **A master key management library and CLI tool** (`master-key-manager`) to provision master keys, activate/deactivate encryption, and inspect status.
+3. **An AES-256-GCM encryption library** (`master_key_manager/master_key_manager.py`) offering authenticated symmetric encryption (Type-6, reversible).
+4. **A transparent ConfigDB interception layer** so that standard SONiC tools (`config`, `sonic-cfggen`) automatically encrypt sensitive fields on write without requiring application-level changes.
+5. **Extensibility** so that additional features (TACACS, RADIUS, LDAP) can be supported by adding an entry to the registry and the corresponding decrypt/render code in the feature daemon.
```
bhouse-nexthop, line 69: Remove "and the corresponding decrypt/render code in the feature daemon", the decrypt code should be automatic if its in the registry, remove "

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+
+---
+
+## Motivation
+
+### At-Rest Security
+
+Community SONiC writes all configuration values to Redis plaintext. An operator or compromised process with read access to the local Redis socket can trivially extract BGP MD5 passwords, TACACS shared secrets, RADIUS keys, and similar credentials. FIPS-compliant deployments and customer security policies require these values to be stored encrypted.
+
+### Industry Precedent
```
bhouse-nexthop, line 79: I would remove this section all-together, we really shouldn't call out other NOS's

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+
+---
+
+## Requirements
+
+### Functional Requirements
+
+1. **FR-1**: ConfigDB fields designated as sensitive shall be stored encrypted using AES-256-GCM with a device-local master key.
+2. **FR-2**: Encryption and decryption shall be transparent to standard SONiC configuration tools (`config load`, `config reload`, `config apply`).
+3. **FR-3**: The system shall support master keys supplied by a central controller, enabling enterprise key management integration.
```
bhouse-nexthop, line 97: I think this is misleading as it indicates that this is also providing some form of central key management controller which it is not.  Just specifying that the user is in control of the keys themselves is sufficient which gives deployment flexibility.

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+4. **FR-4**: A CLI tool (`master-key-manager`) shall allow the operator to set master keys (by key file path or registry entry name) and activate/deactivate encryption per registry entry (`--name`) or for all registry entries (`--all`).
```
bhouse-nexthop: Get rid of the concept of master keys by registry.  There should be a single master key only that can be rotated (or disabled).  Segmentation adds maintenance burdens, additional code, and little to no value.

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+5. **FR-5**: When encryption is activated, all plaintext secret fields in registered tables shall be re-encrypted automatically; when deactivated, they shall be decrypted.
```
bhouse-nexthop, line 99: remove "in registered tables"

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+6. **FR-6**: Encryption shall apply only to ConfigDB. FRR running configuration continues to use cleartext passwords.
+7. **FR-7**: The master key file shall be accessible only to root (mode 0600), protected by filesystem permissions.
+8. **FR-8**: The infrastructure shall retain up to 8 historical master keys. The retained history preserves older keys so they remain available for explicit-key decryption (via `decrypt_string(..., key_idx=N)`) and for future automated rolling-rotation support; automatic decryption of old ciphertexts with the historical keys is called out as Future Work.
```
bhouse-nexthop: I would remove this FR-8, it doesn't serve any real purpose.  Most security standards require destruction of historic keys immediately.

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+9. **FR-9**: BGP MD5 passwords shall be the first feature protected by this infrastructure.
```
bhouse-nexthop, line 103: since this implementation should be generic and _automatic_, we should just up front enumerate all consumers and add them to the registry so they are automatically encrypted and decrypted.

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+
+---
+
+## Overview
+
+SONiC stores all configuration — including protocol passwords such as BGP MD5 authentication passwords, TACACS shared secrets, and RADIUS keys — as cleartext in Redis (ConfigDB) and in the `config_db.json` startup file. Any local account can dump these values with standard Redis tools. This design proposes a general-purpose Master Key Encryption Infrastructure to protect such fields at rest, with BGP MD5 password encryption as its first application.
+
+The infrastructure provides:
+
+1. **A code-defined encryption registry** (`master_key_manager/master_key_encryption_config.py`) that statically declares which tables and fields are subject to automatic encryption. Adding a new encrypted field requires a code change (not a YANG model edit or CLI registration), which is deliberate: any new use of encryption must also involve feature-specific handling code (e.g., decryption in `frrcfgd` before rendering to FRR).
```
bhouse-nexthop: likely this needs to be defined in a C++ file for portability reasons.  

Honestly, I'd remove any references to explicit files that any code belongs to in this entire HLD as that won't age well over time.  Also you're not indicating which repository said files are in anyhow so its already only half-referenced.

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+5. **FR-5**: When encryption is activated, all plaintext secret fields in registered tables shall be re-encrypted automatically; when deactivated, they shall be decrypted.
+6. **FR-6**: Encryption shall apply only to ConfigDB. FRR running configuration continues to use cleartext passwords.
+7. **FR-7**: The master key file shall be accessible only to root (mode 0600), protected by filesystem permissions.
+8. **FR-8**: The infrastructure shall retain up to 8 historical master keys. The retained history preserves older keys so they remain available for explicit-key decryption (via `decrypt_string(..., key_idx=N)`) and for future automated rolling-rotation support; automatic decryption of old ciphertexts with the historical keys is called out as Future Work.
+9. **FR-9**: BGP MD5 passwords shall be the first feature protected by this infrastructure.
+10. **FR-10**: The catalog of tables and fields subject to encryption shall be defined in code (`master_key_encryption_config.py`). Adding encryption to a new table requires a code change, not a YANG model edit or CLI command, because any new encrypted field also requires feature-specific decrypt/render code in the consuming daemon.
```
bhouse-nexthop: Remove ", because any new encrypted field also requires feature-specific decrypt/render code in the consuming daemon."

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+11. **FR-11**: A single table may declare multiple encrypted fields (e.g., both `auth_password` and `md5_key`).
+
+### Non-Functional Requirements
+
+1. The master key file format shall be a versioned, human-readable JSON structure for auditability and disaster recovery.
+2. Encryption/decryption shall use the Python `cryptography` library — no shell subprocesses.
```
bhouse-nexthop, line 110: should not define the library being used, also this should be in C++ for the most part, except maybe the part that does the encryption enablement/disablement that might need to bulk encrypt/decrypt if that is in python.

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+3. One `MasterKeyManager` instance per master key file, shared across all tables that point to the same file.
```
bhouse-nexthop, line 111: There should only be a single key file at any given time system-wide.

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+ │  │    master-key-manager CLI         │   │   Encryption Library          │  │
+ │  │                                   │   │ (master_key_manager.py)       │  │
+ │  │  set --master-key-file | --name   │   │  AES-256-GCM (Type-6)         │  │
+ │  │  encrypt --table / decrypt --table│   │  Nonce + CT + AAD             │  │
+ │  │  activate/deactivate --name|--all │   │  base64-encoded output        │  │
+ │  │  status / list                    │   └────────────────┬──────────────┘  │
+ │  └──────────────┬────────────────────┘                    │                 │
+ │                 │ reads                  ┌────────────────▼──────────────┐  │
+ │                 ▼                        │   MasterKeyManager            │  │
+ │  ┌───────────────────────────────────┐   │   /etc/<feature>_master_key   │  │
```
bhouse-nexthop, line 131: no `<feature>`

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+ │  │  master_key_encryption_config.py  │   │   mode 0600, JSON             │  │
+ │  │  (code-defined, static registry)  │   │   up to 8 historical keys     │  │
+ │  │  • name  (e.g. "BGP")             │   │   enabled flag (on/off state) │  │
+ │  │  • master_key_file                │   │   one instance per file       │  │
+ │  │  • fields: {TABLE: [field, ...]}  │   └───────────────────────────────┘  │
+ │  └──────────────┬────────────────────┘                                      │
+ │                 │ read at startup                                           │
+ │                 ▼                                                           │
+ │  ┌──────────────────────────────────────────────────────────────────────┐   │
+ │  │          ConfigDB Interception Layer (Python)                        │   │
```
bhouse-nexthop, line 141: This needs to be lifted to C++ in the sonic-swss-common submodule.  All redis reads/writes for configdb flow through that, including from python and rust as it provides interop.  Nothing reads/writes directly to redis, they all flow through that path.

By making it python-only, you just lost anything written in any other language which defeats the purpose of this HLD.

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+ └───────────────────────────────┘      └──────────────────────────────────┘
+```
+
+---
+
+## Master Key Manager Infrastructure
+
+### Master Key File
+
+The master key file is a JSON document (e.g., `/etc/sonic/bgp_master_key`) with mode 0600 (readable only by root). It holds a named key set and retains up to **8 historical master keys** in descending timestamp order. The retained history preserves older keys so they remain available for explicit-key decryption (via `decrypt_string(..., key_idx=N)`) and for future automated rolling-rotation support; automatic decryption of values encrypted under an older key using the history is Future Work (see [Future Work](#future-work)).
```
bhouse-nexthop: fixup as per prior comments

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+The `set` command accepts any string, pads or truncates it to 32 bytes for AES-256, and prepends a new `MasterKey` entry to the list. The previous key is retained in the history list; it remains available for explicit-key decryption via `decrypt_string(..., key_idx=N)` and for future automated rolling-rotation support. Automatic decryption of old ciphertexts using the historical keys is Future Work.
+
+This is the primary model for enterprise deployments where a controller pushes a deterministic key to a fleet of devices.
+
+#### 2. Local Random Key (Ad-hoc)
+
+For standalone devices without a central key management system, the operator can supply any high-entropy string:
+
+```
+# master-key-manager set --master-key-file /etc/sonic/bgp_master_key $(openssl rand -hex 32)
```
bhouse-nexthop: you should have a built-in generator so someone doesn't need to know they need to shell out to openssl

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+Saved master key to /etc/sonic/bgp_master_key
+```
+
+### Key Rotation
+
+When a new master key is set with `set`, the old key is retained in the history list. To re-encrypt all ConfigDB values with the new key:
+
+```
+# master-key-manager deactivate --name BGP
```
bhouse-nexthop: shouldn't need a name field since there's only 1 key

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+# master-key-manager activate --name BGP
+```
+
+Deactivation decrypts all fields back to plaintext, then activation re-encrypts them with the current (newest) master key. Because FRR reads decrypted passwords via the interception layer, no BGP flap occurs during rotation.
+
+> **Note:** The safe rotation procedure above re-encrypts *all* values with the new key by first decrypting to plaintext (`deactivate`) and then re-encrypting (`activate`), so every value ends up under the current key. Automatic decryption of values that were encrypted under an older key — i.e. a rolling rotation where `set` is called without a preceding `deactivate` — is **not** supported in this version: `decrypt_string` and `is_encrypted` only consult the current key (index 0) by default. Performing a rolling `set` without `deactivate` first will cause `is_encrypted` to misclassify old-key ciphertexts as cleartext, leading to double-encryption. Operators must follow the `deactivate` → `activate` sequence. Automatic historical-key decryption during rolling rotation is tracked as Future Work.
+
+### Encryption Library
+
```
bhouse-nexthop, line 260: This section needs to be rewritten due to needing to be in C++

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+        Does not write to the database — caller is responsible for the write."""
+
+    def encrypt_config(self, data) -> dict:
+        """Return data (or a deep copy if anything was changed) with all registered
+        plaintext secret fields encrypted.  Returns original dict unchanged if no
+        encryption is needed (lazy copy — avoids deep-copying large route tables)."""
+```
+
+#### ConfigDBConnector Interception (both py-swsssdk and swss-common)
+
```
bhouse-nexthop, line 372: for clarification, this needs to be raised to C++

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+    "name": "BGP",
+    "master_key_file": BGP_MASTER_KEY_FILE,
+    "fields": {
+        "BGP_NEIGHBOR":   ["auth_password"],
+        "BGP_PEER_GROUP": ["auth_password"],
+    },
+}
+
+_MasterKeyEncryptionRegistries = [ _BGP_REGISTRY ]
+```
```
bhouse-nexthop, line 429: This would be in C++, and just be a list of Table to field names.  No need to reference the master key file name nor the "NAME" as we only need support for 1 master key.

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+
+Both tables share the same key file. `master-key-manager activate/deactivate --name BGP` bulk-encrypts or decrypts all entries of the BGP registry entry's tables; `--all` targets every registry entry.
+
+The decrypt-and-render step for FRR is handled explicitly in `frrcfgd`.
+
+```python
+def bgp_neighbor_handler(self, table, key, data):
```
bhouse-nexthop, line 582: none of this should need to change at all.  This should be handled *transparently* in the C++ code.  You already know what fields should be stored encrypted or not, and its fairly easy to detect if a field is decryptable or not (to handle cases where a plaintext value was passed when encrypted was expected)

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+
+
+#### Step 4: Key Rotation
+
+```
+# master-key-manager set --name BGP newKey2026
+Saved master key to /etc/sonic/bgp_master_key
+
+# master-key-manager deactivate --name BGP
+# master-key-manager activate --name BGP
```
bhouse-nexthop: no "rotate" feature?

doc/master-key-encryption/master-key-encryption.md
```diff
@@ -0,0 +1,731 @@
...
+
+- The master key file (`/etc/sonic/bgp_master_key`) is a host filesystem file. It persists across warmboot/fastboot cycles; no special handling is required.
+- During warmboot, `bgpcfgd`/`frrcfgd` start and read ConfigDB. The interception layer decrypts passwords transparently. No warmboot-specific code path is needed.
+- If the master key file is missing or corrupted during boot, decryption will fail and `frrcfgd` will use the raw (encrypted) string as the BGP password, resulting in BGP authentication failures. Operators should include the master key file in backup/restore procedures.
+
+---
+
+## Restrictions and Limitations
+
+1. **Python-only interception**: Encryption is transparent only for Python callers using `ConfigDBConnector`. Tools that bypass Python and write directly to Redis (`redis-cli`, `sonic-db-cli`, native C++ or Rust applications) will not have passwords encrypted automatically. From an operational standpoint this is acceptable because network admins configure the switch via `config` or `sonic-cfggen`. C++ and Rust interception is noted as future work.
```
bhouse-nexthop, line 686: i think this needs to be a day1 lift

bhouse-nexthop (changes requested):

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: Type-6, while is specific to a vendor, is a widely accepted term in network switching industry. It is also recommended by NSA/CISA. The actual implementation does not have reference to "Type-6" though. Will remove

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: I can remove reference to specific files. However this proposal is for Python based implementation. Implementation in C++ as well as Rust is mentioned in Future Works as discussed in previous meetings

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: OK

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: I think the sentence is quite clear in that it makes the point that this feature is friendly to centralized key distribution, which is common practice in many deployments

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: The registry idea was agreed during previous meetings. Using a single master key makes the system hard to extend and scale. It is also not friendly to fault containment. I a community-driven ecosystem it also helps to reduce the risk of errors during system migration.

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: It is a registration mechanism.

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: Maybe not 8. But a previous master key will make it possible to recover the system config if a backup config is available.

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: The framework needs to be extensible. One cannot foresee or implement all features that may require encryption. For example OSPF and ISIS may be added later on. So it won't be possible to enumerate all consumers at this time

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: What I mean that when adding a new table/field for encryption, there is likely to be new application-specific logic added somewhere. For example, adding BGP MD5 password means additional code needs to be added to decrypt the password in CONFIG DB before passing it to FRR. It is not about adding decrypt/render code in the master key management library or tools

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: I can remove this line. The point is that for security reason it won't use a subprocess. But again, C++ or Rust implementation is listed in Future Works

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: Again I think it is better (and by consultation with AI), that per-subsystem master key, distributed by a controller, is preferrable in large system deployment. Of course this framework does not prevent one to have just one master key for all systems if one chooses to do so

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: It's not python-only. C++ and/or Rust implementation is Future Work if the need arises

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: openssl is built-in on SONIC switch, and its random number generation logic is compliant with many industry and governmental requirements such as FIPS. It will take a lot of effort to qualify a random password generator for the same requirements

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: I think it should be one key per sub system.

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: C++ is for future works. Currently we don't have the need

doc/master-key-encryption/master-key-encryption.md
fredxia-nexthop: Again this does not need to be in C++. In fact some of the SONIC tools are python only. The python code directly save data to CONFIG DB without C++

doc/master-key-encryption/master-key-encryption.md
bhouse-nexthop: I completely disagree here, and I don't recall any consensus on this being accepted previously. There is no "fault containment".  Either your file system is corrupt or its not, trying to segment the key to try to limit a blast radius if your filesystem corrupts is meaningless.

doc/master-key-encryption/master-key-encryption.md
bhouse-nexthop: then say that explicitly.

doc/master-key-encryption/master-key-encryption.md
bhouse-nexthop: Then say 'all registered plaintext secret fields', specifically calling out a table as registered implies all fields in a table are encrypted.

doc/master-key-encryption/master-key-encryption.md
bhouse-nexthop: Under [NIST SP 800-53 Rev. 5](https://csrc.nist.gov/publications/detail/sp/800-53/rev-5/final) control SC-12 (Cryptographic Key Establishment and Management), the destruction of previous encryption keys requires removing all physical and electronic traces to prevent data recovery once keys reach the end of their cryptoperiod or rotation cycle.

Again, remove this concept.

doc/master-key-encryption/master-key-encryption.md
bhouse-nexthop: I should have said "enumerate all *existing* consumers", we obviously can't enumerate future ones.

doc/master-key-encryption/master-key-encryption.md
bhouse-nexthop: Python-only is fine for a prototype as a proving ground, but the final implementation can't be python-only.

doc/master-key-encryption/master-key-encryption.md
bhouse-nexthop: This should, explicitly, NOT be required.  This should be *transparent* to the consumer.  You have a registry of fields that are encrypted, when reading from Redis you simply decrypt all those fields in memory if you are root and have access to the keys.  There is absolutely no reason to make an edit to every single consumer when you already have centralized code in sonic-swss-common that everyone uses to read the configdb.

doc/master-key-encryption/master-key-encryption.md
bhouse-nexthop: Can you share references for justification of this.  We need a write-up as to why this exists and what the actual use-case is for this.  How people would consume this.  Adding additional key-management overhead on operators is not something we should take lightly.  Key management is often the hardest part of the security of any cryptographic subsystem.

doc/master-key-encryption/master-key-encryption.md
bhouse-nexthop: I'm not saying to re-invent the wheel here.  Use openssl from within python via pyOpenSSL and call RAND_bytes.  Or heck, even call subprocess.run() and shell out to openssl yourself.  Just don't make the operator explicitly type `$(openssl rand -hex 32)` on the command line.  Also that would leak a hex cryptographic key into the "ps" list that could be visible to unprivileged processes.

doc/master-key-encryption/master-key-encryption.md
bhouse-nexthop: SONiC is actively rewriting a lot of python tooling in RUST.  It is an initiative inside of SONiC.

PR #2440, opened 2026-07-02
vpp: portchannel punt via member interface

yue-fred-gao:
### why 
The current port-channel implementation punts control-plane packets directly to the `PortChannel<id>` (bond) netdev. 
This does not conform to the SONiC data-plane model. On real hardware the NPU punts control traffic to the CPU on the physical/member port it was received on. 
The most visible symptom is **DHCP relay**. `sonic-dhcp-relay` / `dhcrelay`
(the SONiC relay agent) binds and accepts uplink packets **only on port-channel member interfaces**. Because today's solution delivers the frame on the `PortChannel<id>` netdev instead of the member, the relay agent never sees the
uplink packet and DHCP relay over a port-channel uplink is broken.

### what this PR does
Redesign the punt path for packets to port channel and port channel sub interface using sonic-ext plugin that was created for VLAN/BVI.

### DIFF
--- a/doc/vpp/vpp_portchannel_redesign_hld.md
+++ b/doc/vpp/vpp_portchannel_redesign_hld.md
@@ -1,0 +1,335 @@
+# Port Channel Punt Redesign — High-Level Design
+
+## Revision History
+
+| Rev | Date       | Author  | Summary |
+|-----|------------|---------|---------|
+| 1.0 | 2026-06-30 | Yue Gao | Initial design. Redesign of port-channel control-plane punt so that traffic received on a bonded member is delivered to Linux **on the originating member netdev** (and let the kernel bond/8021q stack carry it up to `PortChannel<id>[.<vlan>]`), instead of being punted directly onto the `PortChannel<id>` netdev via the `be<id>` host tap + TC redirect. Reuses the existing `sonic_ext` capture + aggregate-tap-redirect machinery (originally built for VLAN/BVI) by teaching `sonic_ext_phy_is_aggregate()` to recognize bond masters and bond sub-interfaces. |
+
+---
+
+## 1. Problem Statement
+
+The current port-channel implementation punts control-plane packets **directly to
+the `PortChannel<id>` (bond) netdev**:
+
+- `sonic-sairedis` creates the bond as `BondEthernet<id>` and, on the first member
+  enslave, wires a linux-cp pair `BondEthernet<id> ↔ be<id>` plus a TC filter
+  redirect `be<id> → PortChannel<id>`
+  (`SwitchVppFdb.cpp`, `vpp_create_lag_member`).
+- A packet received on a bonded member and punted by VPP is dispatched by
+  `linux-cp-punt-xc` to the `be<id>` host tap, and the TC redirect moves it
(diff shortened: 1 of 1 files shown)


doc/vpp/vpp_portchannel_redesign_hld.md
```diff
@@ -0,0 +1,335 @@
...
+the packet and redirect it to the original member tap interface.
+
+The design divides port-channel traffic into classes and defines, for each,
+**where the punted copy must land in Linux** and **how VPP delivers it there**.
+
+| Traffic class | Received on | Must appear in Linux on | VPP punt mechanism |
+|---------------|-------------|--------------------------|--------------------|
+| **ARP** (main interface) | member phy | member netdev → kernel bond → `PortChannel10` | `linux-cp-punt-xc` → `be10` tap → **`sonic-ext-aggr-tap-redirect` → member tap** |
+| **IPv4 unicast for-us** (main interface) | member phy | member netdev → kernel bond → `PortChannel10` | same as ARP |
+| **ARP** (sub-interface) | member phy (tagged) | member netdev → bond → 8021q → `PortChannel10.1` | `linux-cp-punt-xc` → `be10.1` tap → **`sonic-ext-aggr-tap-redirect` → member tap (re-push .1Q)** |
```
lunyue-ms, line 110: Hi @yue-fred-gao, please correct me if my understanding is not correct: the packet delivered to be10.1 tap should have a vlan tag(vid1). In member tap (re-push .1Q), the packet will be re-push with a vlan tag. Can the re-pushed vid can be different from the vid in be10.1` tap in some cases? 
Section6.3 notes mentioned "No tag strip/re-push is required".  It seems that (re-push .1Q) needs to be changed?

lunyue-ms (dismissed):
Thanks, Fred, this HLD is very clear and constructed. From the log, it seems that you have already implemented the main part.

doc/vpp/vpp_portchannel_redesign_hld.md
```diff
@@ -0,0 +1,335 @@
...
+|-------------------------|--------------------------|------------------------------|
+| `PortChannel10`         | `BondEthernet10`         | `be10` (LCP tap) → `PortChannel10` (team) |
+| `PortChannel10` members | `BondEthernet10` members | `Ethernet0`, `Ethernet4` member taps |
+| `PortChannel10.1`       | `BondEthernet10.1`       | `PortChannel10.1` (8021q on the team netdev) |
+
+---
+
+## 3. Design Principles (Punt Strategy Summary)
+
+The design utilizes the punt-via-member infra in sonic-exe plugin introduced for VLAN/BVI. It uses
```
AkeelAli: typo: sonic-ext plugin

AkeelAli (dismissed):
Great design.

doc/vpp/vpp_portchannel_redesign_hld.md
yue-fred-gao: Yes. The vlan is retained in L3 processing flow. sonic-ext-aggr-tap-redirect re-pushes VLAN only if the packet doesn't have VLAN tag. Let me remove "repush .1Q"

lunyue-ms (approved):

PR #2441, opened 2026-07-02
Add API design for leak sensor test bits

fraserg-arista:
The intention of this API is to allow platforms to expose leak test functionality in their liquid cooling implementation. The state is written to the state DB so that other platform components can distinguish between real and test leaks (e.g. by logging that they would trigger a power cycle, without actually carrying out the action).


fraserg-arista:
Note: this is written on top of https://github.com/sonic-net/SONiC/pull/2413

@judyjoseph @yuazhe 

The intent here is that the API is simple enough that it can be implemented in software or hardware, as appropriate for each implementer's leak detection design, so let me know if this is a practical API for you to implement.

judyjoseph:
@fraserg-arista would be great if we have this as a separate PR and get it merged first -- This is useful for community, Let me know your thoughts

judyjoseph:
@chinmoy-nexthop f.y.i

chinmoy-nexthop:
> @chinmoy-nexthop f.y.i

Thank you Judy. will take a look.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -499,22 +504,26 @@ Reference doc:
...
+| get_leak_severity() | Get the per-sensor leak severity as `LeakSeverity.CRITICAL`, `LeakSeverity.MINOR`, or `None` |
+| get_leak_profile() | Return the `LeakSensorProfileBase` associated with this leak sensor type |
+| set_test_leak(enable) | Enable or disable a per-sensor test bit that forces the sensor to report a leak |
+| is_test_leak_enabled() | Return whether the per-sensor forced-leak test bit is enabled |
 
 **Note**
 
 The get_leak_severity() API call has to determine whether any condition indicates a severe leak detected from a leak sensor.This decision could be based on:
 
+When the test bit is enabled, the effective leak state for that sensor is forced to `True` even if the hardware leak input is clear. This affects only leak presence. The physical health returned by `is_leak_sensor_ok()` is unchanged, and `get_leak_severity()` continues to return the platform-defined severity for a leak on that sensor. The test bit is intended for validation and is cleared by default at boot.
```
chinmoy-nexthop: a privileged caller that enables a test leak may be able to trigger the same destructive path as a real leak.  yes, we talked about distinction between test leaks from real leaks, but I do not see that reflect in the design.

chinmoy-nexthop: i mean , the documented defaults are system_leak_policy=enabled and system_critical_leak_action=power_off here i see.

doc/bmc/leakage_detection_hld.md
```diff
@@ -1,156 +1,299 @@
...
-2. For platform that doesn't support liquid cooling at all, there should be no further performance overheading.
 
-## 3. Detection and alarm flow
-The leak alarm process is straightforward. The platform API first acquires the status of the leak detection sensors. Then, thermalctld has a thread that calling the API and in turn notifies system health monitor, who ultimately sends out a gNMI event.
+1. Monitor leak sensors only on liquid-cooled or hybrid-cooled platforms.
+2. Detect per-sensor leak state and per-sensor leak severity.
+3. Aggregate per-sensor results into a system leak severity.
+4. Expose per-sensor and system leak status through Redis DB for BMC policy enforcement and CLI.
+5. Avoid additional PMON overhead on platforms that are not liquid cooled.
+6. Support a per-sensor test API that can force a leak indication without requiring physical liquid in the chassis.
```
chinmoy-nexthop: I think the test bit intent is good, but we need to add the the design of  “test vs real leak” with explicit flow  diagram and boundaries . can you please add.

doc/bmc/leakage_detection_hld.md
```diff
@@ -1,156 +1,299 @@
...
+|--------|-------------|
+| `get_name()` | Returns the leak sensor name |
+| `is_leak()` | Returns whether a stable leak condition is detected |
+| `is_leak_sensor_ok()` | Returns whether the sensor is healthy or faulty |
+| `get_leak_sensor_type()` | Returns the sensor type, for example `rope`, `spot`, or `flex_pcb` |
+| `get_leak_sensor_location()` | Returns the physical sensor location |
+| `get_leak_severity()` | Returns the per-sensor leak severity as `LeakSeverity.CRITICAL`, `LeakSeverity.MINOR`, or `None` |
+| `get_leak_profile()` | Returns the `LeakSensorProfileBase` associated with this sensor type |
+| `set_test_leak(enable)` | Enables or disables a per-sensor test bit that forces the sensor to report a leak |
+| `is_test_leak_enabled()` | Returns whether the per-sensor forced-leak test bit is enabled |
```
chinmoy-nexthop: is there a way to disable these two APIs in production? Since they appear to be intended only for testing, we may not need them in production. Leaving them enabled could introduce a potential misuse risk.

doc/bmc/leakage_detection_hld.md
```diff
@@ -1,156 +1,299 @@
...
+`bmcctld` subscribes to `SYSTEM_LEAK_STATUS|system` and applies the configured action from `LEAK_CONTROL_POLICY`.
+
+- `system_critical_leak_action` defaults to `power_off`
+- `system_minor_leak_action` defaults to `syslog_only`
+
+Supported actions are:
+
+- `syslog_only`
+- `graceful_shutdown`
+- `power_off`
```
chinmoy-nexthop: Would it make sense to limit the graceful_shutdown/power-off action to cases where the leak source is a test injection? What do you think?

chinmoy-nexthop (commented):
Nice addition.  added few suggestions.

also, please add tests that prove forced test leaks cannot power off the host even when they aggregate to CRITICAL.

doc/bmc/leakage_detection_hld.md
fraserg-arista: An interesting question! Do you know if we have any models in Sonic for this kind of behaviour? (e.g. something that needs a configuration setting to enable a test/debug mode)

doc/bmc/sonicBMC/pmon-bmc-design.md
fraserg-arista: Yeah, I think you're right here: the fact of this being a test is passed through to thermalctld but it only calculates the overall state of the system for bmcctld to act on, but bmcctld doesn't know that this is a test so it will apply the system policy.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -499,22 +504,26 @@ Reference doc:
...
+|---------|----------|
+| get_name() | Get leak sensor name |
+| is_leak() | Is there a leak detected? **Only stable leak conditions** are asserted or cleared. This could be done by debounce logic in <vendor>platform/firmware |
+| is_leak_sensor_ok() | Is the leak sensor OK or faulty? |
+| get_leak_sensor_type() | Get the sensor type, for example rope, flex_pcb, or spot |
+| get_leak_sensor_location() | Get the sensor location |
+| get_leak_severity() | Get the per-sensor leak severity as `LeakSeverity.CRITICAL`, `LeakSeverity.MINOR`, or `None` |
+| get_leak_profile() | Return the `LeakSensorProfileBase` associated with this leak sensor type |
+| set_test_leak(enable) | Enable or disable a per-sensor test bit that forces the sensor to report a leak |
+| is_test_leak_enabled() | Return whether the per-sensor forced-leak test bit is enabled |
```
judyjoseph: @fraserg-arista Can we have a separate LeakTest class eg: **LeakageSensorTestBase** where we can define a set of APIs which every platform owner can implement in their platform modules ? This way we can simulate any of the above platofrm API's in the sonic-mgmt test infra

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -699,6 +711,38 @@ config liquid-cool leak-action [system|rack_mgr] [critical|minor]  [syslog_only|
    - power_off        : Immediately power off Switch-Host via platform API module->set_admin_state(DOWN).
 ```
 
+* **config liquid-cool leak-test sensor**
```
judyjoseph: @fraserg-arista let me know your thought .. I was thinking we really don't need a config command .. but instead using the platform_modules which each platform owner can implement and use it in the sonic_mgmt leak tests to simulate leaks.

We can discuss ideas in the coming workgroup meeting

doc/bmc/leakage_detection_hld.md
```diff
@@ -7,6 +7,7 @@ Due to the excessive heat generated by the equipment, traditional air-cooling me
 ## 2. Requirements
 1. Monitoring the liquid colling leakge detection sensor, and alarm accordingly.
 2. For platform that doesn't support liquid cooling at all, there should be no further performance overheading.
+3. Support controlled per-sensor leak-test injection for platform validation without allowing a test-only leak to trigger host power control.
```
nikamirrr, line 10: This limits the e2e testing for the BMC pmon.

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -518,6 +522,19 @@ This base class is already defined in sonic-platform-common, additional new plat
...
 The get_leak_severity() API call has to determine whether any condition indicates a severe leak detected from a leak sensor.This decision could be based on:
 
+#### LeakageSensorTestBase
+
+`LeakageSensorTestBase` is an optional platform test interface used by `sonic-mgmt` and platform validation tooling. It is not exposed through the normal production CLI.
+
+| Method | Description |
+|---------|----------|
+| is_leak_test_supported() | Return whether this platform supports controlled leak-test injection |
+| set_test_leak(sensor_name, enable) | Enable or disable test injection for one sensor |
```
nikamirrr: Do we need a severity param to toggle the leak state from None, MINOR, CRITICAL.

How about simulating a sensor fault?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -518,6 +522,19 @@ This base class is already defined in sonic-platform-common, additional new plat
...
+| is_test_leak_enabled(sensor_name) | Return whether test injection is enabled for one sensor |
```
judyjoseph: should we have test API's to set the severity and other attributes of the test leak sensor?

doc/bmc/sonicBMC/pmon-bmc-design.md
```diff
@@ -518,6 +522,19 @@ This base class is already defined in sonic-platform-common, additional new plat
...
+| clear_test_leaks() | Clear test injection on every sensor |
+
+Platforms that do not implement this interface do not support test injection. Test state is cleared at boot and interface use is audited. On platforms with hardware test bits, enabling a test may cause `is_leak()` to return `True` because the hardware leak indication is forced. `thermalctld` uses `is_test_leak_enabled()` to classify that indication as test-injected and excludes it from the physical result used for policy enforcement. If the underlying physical state is not independently observable while a hardware test bit is enabled, test injection is restricted to controlled validation where a concurrent real leak can be ruled out.
```
judyjoseph: do we need to disable the policy enforcement in case of test leak ?

judyjoseph: @fraserg-arista There is the leak policy may be we can use this in tests to prevent enforcement. This is up to testcase to decide
```
config liquid-cool leak-action [system|rack_mgr] [critical|minor]  [syslog_only|graceful_shutdown|power_off]

   - syslog_only      : Log the event; no Switch-Host power action taken.
   - graceful_shutdown: Issue a graceful GNOI shutdown to Switch-Host; force power-off after graceful_shutdown_timeout (CHASSIS_MODULE|SWITCH-HOST) if unresponsive.
   - power_off        : Immediately power off Switch-Host via platform API module->set_admin_state(DOWN).
DB schema
  "LEAK_CONTROL_POLICY": {                                      ; In CONFIG_DB
      "system_leak_policy"            : "enabled | disabled",   ; enabled by default
      "system_critical_leak_action"   : "power_off",            ; default is power_off   
      "system_minor_leak_action"      : "syslog_only",          ; default is syslog_only
      "rack_mgr_leak_policy"          : "enabled | disabled",   ; enabled by default
      "rack_mgr_critical_alert_action": "syslog_only",          ; default is syslog_only
      "rack_mgr_minor_alert_action"   : "syslog_only"           ; default is syslog_only
  }

```

judyjoseph:
@fraserg-arista please could you update this HLD based on review comments during community discussion last week.

judyjoseph (approved):

judyjoseph:
@chinmoy-nexthop @nikamirrr could you check any open comments you still have -- and add it to this PR. if there are any we can get them merged in a follow up PR

PR #2444, opened 2026-07-07
HLD: CPO support in `xcvrd`

bgallagher-nexthop:
HLD that describes how `xcvrd` can add support for hardware platforms with co-packaged optics (CPO).

### DIFF
--- a/doc/pmon/xcvrd_cpo.md
+++ b/doc/pmon/xcvrd_cpo.md
@@ -55,7 +55,7 @@ The scope of this HLD is to define how `xcvrd` will use the CPO platform API obj
 | CMIS | Common Management Interface Specification |
 | CPO | Co-packaged Optics |
 | OE | Optical Engine |
-| ELSFP | External Laser Small Factor Pluggable |
+| ELSFP | External Laser Small Form Factor Pluggable |
 | DOM | Digital Optical Monitoring |
 
 ### 4. Overview 
@@ -148,7 +148,7 @@ power mode, since low power mode is a module-wide state. This requires fetching 
 disabling / deinitializing their lanes. See the below code snippet outlining `CpoManagerTask` for how this could be implemented simply.
 
 Later on, if further customization of state machine logic is required to aid in debugging or manage extra steps in the state machine for CPO,
-the `CmisManagerTask` can be refactored and/or methods can be overriden in `CmisManagerTask` by `CpoManagerTask`.
+the `CmisManagerTask` can be refactored and/or methods can be overridden in `CmisManagerTask` by `CpoManagerTask`.
 
 ```python3
 class CpoManagerTask(CmisManagerTask):
--- a/doc/pmon/xcvrd_cpo.md
+++ b/doc/pmon/xcvrd_cpo.md
@@ -1,0 +1,265 @@
+# xcvrd CPO Support #
+
+## Table of Content 
+
+- [1. Revision](#1-revision)
+- [2. Scope](#2-scope)
+- [3. Definitions/Abbreviations](#3-definitionsabbreviations)
+- [4. Overview](#4-overview)
+- [5. Requirements](#5-requirements)
+- [6. Architecture Design](#6-architecture-design)
+- [7. High-Level Design](#7-high-level-design)
+- [8. SAI API](#8-sai-api)
+- [9. Configuration and management](#9-configuration-and-management)
+- [10. Warmboot and Fastboot Design Impact](#10-warmboot-and-fastboot-design-impact)
+- [11. Memory Consumption](#11-memory-consumption)
+- [12. Restrictions/Limitations](#12-restrictionslimitations)
+- [13. Testing Requirements/Design](#13-testing-requirementsdesign)
+  - [13.1. Unit Test cases](#131-unit-test-cases)
+  - [13.2. System Test cases](#132-system-test-cases)
+- [14. Open/Action items - if any](#14-openaction-items---if-any)
+
+### 1. Revision  
+| Rev |     Date     |       Author       | Change Description                                                               |
+|:---:|:------------:|:------------------:|----------------------------------------------------------------------------------|
+| 0.1 |  07/02/2026  |  Brian Gallagher   | Initial version                                                                  |
+
+### 2. Scope  
+
+The scope of this HLD is to define how `xcvrd` will use the CPO platform API objects introduced in the [Port Mapping for CPO Platforms](https://github.com/sonic-net/SONiC/pull/2211) HLD. This includes:
+- DOM telemetry (DomInfoUpdateTask)
+- CMIS state machine management (CmisManagerTask)
+- Transceiver presence management (SfpStateUpdateTask)
+
+### 3. Definitions/Abbreviations 
+
+| Term | Definition |
+|------|------------|
+| CMIS | Common Management Interface Specification |
+| CPO | Co-packaged Optics |
+| OE | Optical Engine |
+| ELSFP | External Laser Small Factor Pluggable |
+| DOM | Digital Optical Monitoring |
+
+### 4. Overview 
+
+This HLD covers:
+- the changes required to refactor `xcvrd` to support the usage of platform APIs that are not traditional SfpBase-derived objects while maximizing code re-use
+- how CPO support for DOM telemetry, CMIS state machine management and physical port device presence management will be added to `xcvrd`
+
+### 5. Requirements
+
+This section list out all the requirements for the HLD coverage and exemptions (not supported) if any for this design.
+
+### 6. Architecture Design 
+
+This HLD does not propose any architectural changes to SONiC.
+
+### 7. High-Level Design 
+
+The primary problem posed by the `xcvrd` integration with CPO platform APIs is that they are not guaranteed to
+expose the same interface that traditional SfpBase-derived objects do, which makes code re-use difficult.
+This HLD proposes to solve this problem by committing to the following shared interfaces:
+- CPO objects inherit from DeviceBase, so hardware interaction like presence detection is a common interface between SfpBase and CPO objects.
+- CPO objects will expose a CmisApi, so any CMIS related logic can leverage the API instance to interact with the hardware irrespective of whether it is a SfpBase or CPO object.
+
+For each of the tasks in `xcvrd` (`CmisManagerTask`, `SfpStateUpdateTask`, `DomInfoUpdateTask`), a CPO-specific subclass will be introduced. Each subclass
+will aim to re-use as much code as possible from the original task code, overriding methods where necessary to introduce any required custom logic for CPO hardware.
+
+| Task | CPO Subclass |
+|------|--------------|
+| CmisManagerTask | CpoManagerTask |
+| DomInfoUpdateTask | CpoDomInfoUpdateTask |
+| SfpStateUpdateTask | CpoStateUpdateTask |
+
+#### 7.1 CMIS State Machine Management (CmisManagerTask)
+
+##### 7.1.1 Interface Gap Analysis
+
+| Call | Present on CPO objects |
+|------|:----------------------:|
+| `sfp.get_presence()` | yes, via DeviceBase |
+| `sfp.get_xcvr_api()` | yes, via CpoBase |
+
+The philosophy of leveraging DeviceBase and the CmisApi exposed via `get_xcvr_api()` will work for this task, allowing code to be re-used without changes.
+
+##### 7.1.2 Changes Required
+
+###### 7.1.2.1 CmisManagerTask Refactor
+
+`CmisManagerTask` currently iterates over every logical interface, and internally decides whether it should process a given port. When introducing `CpoManagerTask`, 
+we will need to ensure that `CmisManagerTask` does not attempt to touch any CPO ports. This will be addressed by refactoring CmisManagerTask to accept a dictionary
+of ports it is responsible for. Internally, it will then decide to skip any port that is not present in the dictionary passed to it.
+```python3
+        # Start the CMIS manager
+        # self.sfp_obj_dict contains all traditional pluggable ports
+        cmis_manager = None
+        if not self.skip_cmis_mgr:
+            cmis_manager = CmisManagerTask(self.namespaces, port_mapping_data, self.sfp_obj_dict, self.stop_event, skip_cmis_mgr=self.skip_cmis_mgr)
+            cmis_manager.start()
+            self.threads.append(cmis_manager)
+
+        # Start the CPO manager
+        # self.cpo_obj_dict contains all CPO ports
+        cpo_manager = None
+        if self.cpo_obj_dict:
+            cpo_manager = CpoManagerTask(self.namespaces, port_mapping_data, self.cpo_obj_dict, self.stop_event)
+            cpo_manager.start()
+            self.threads.append(cpo_manager)
+```
+
+This will allow both a `CmisManagerTask` and `CpoManagerTask` to run in parallel and manage their respective ports.
+
+###### 7.1.2.2 CpoManagerTask
+
+A `CpoManagerTask` will be introduced to handle CPO ports. Initially, this will be a minimal skeleton that re-uses the `CmisManagerTask` logic as-is.
+Later on, if customization of state machine logic is required to aid in debugging or manage extra steps in the state machine for CPO, the `CmisManagerTask`
+can be refactored and/or methods can be overriden in `CmisManagerTask` by `CpoManagerTask`.
+
+```python3
+class CpoManagerTask(CmisManagerTask):
+    def __init__(self, namespaces, port_mapping, sfp_obj_dict, main_thread_stop_event, skip_cpo_mgr=False):
+        super().__init__(namespaces, port_mapping, sfp_obj_dict, main_thread_stop_event,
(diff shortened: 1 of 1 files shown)


doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,265 @@
...
+thin wrappers around `CmisApi` functions, providing basic normalization. Most just check whether `self.xcvr_api` is None before calling the API
+method in order to avoid an exception being thrown.
+
+`xcvrd` should be refactored to just directly call the `CmisApi` functions and perform normalization in `xcvrd` helper methods. These thin
+wrapper functions can then be removed from `SfpOptoeBase` afterwards. That will allow us to rely on the `xcvr_api` for both SfpBase and CPO
+objects in the DOM task.
+
+###### 7.2.2.2 Refactor of DomInfoUpdateTask Database Publishing Utilities
+
+The `DomInfoUpdateTask` currently uses many utilities to collect information and publish it to the database. Many of these functions follow
```
tshalvi: How are the DomInfoUpdateTask functions that do not follow the “read from EEPROM, then publish to DB” pattern expected to be handled by this refactoring? Could you also provide an example of such a function and explain how the proposed design would apply to it?

doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,265 @@
...
+        if not self.skip_cmis_mgr:
+            cmis_manager = CmisManagerTask(self.namespaces, port_mapping_data, self.sfp_obj_dict, self.stop_event, skip_cmis_mgr=self.skip_cmis_mgr)
+            cmis_manager.start()
+            self.threads.append(cmis_manager)
+
+        # Start the CPO manager
+        # self.cpo_obj_dict contains all CPO ports
+        cpo_manager = None
+        if self.cpo_obj_dict:
+            cpo_manager = CpoManagerTask(self.namespaces, port_mapping_data, self.cpo_obj_dict, self.stop_event)
```
tshalvi, line 123: I think we should add a `skip_cpo_mgr` flag for CpoManagerTask, similar to the existing `skip_cmis_mgr` flag for CmisManagerTask. This would provide the option to delegate CPO module configuration to the firmware instead of handling it through xcvrd on platforms that support this.

doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,265 @@
...
+
+`CmisManagerTask` currently iterates over every logical interface, and internally decides whether it should process a given port. When introducing `CpoManagerTask`, 
+we will need to ensure that `CmisManagerTask` does not attempt to touch any CPO ports. This will be addressed by refactoring CmisManagerTask to accept a dictionary
+of ports it is responsible for. Internally, it will then decide to skip any port that is not present in the dictionary passed to it.
+```python3
+        # Start the CMIS manager
+        # self.sfp_obj_dict contains all traditional pluggable ports
+        cmis_manager = None
+        if not self.skip_cmis_mgr:
+            cmis_manager = CmisManagerTask(self.namespaces, port_mapping_data, self.sfp_obj_dict, self.stop_event, skip_cmis_mgr=self.skip_cmis_mgr)
```
micas-net, line 115: do we need to run normal manager if all ports are cpo related?

prgeor (commented):
@bgallagher-nexthop anything pending? I see the PR in draft.

doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,265 @@
...
+the logic that publishes to Redis into its own function that can be overriden in the CPO subclass to publish
+both OE and ELSFP information.
+
+#### 7.4 Port Device Access
+
+TODO: How `xcvrd` accesses CPO objects on the Chassis object and differentiates between what is a CPO port
+and what is a traditional pluggable port.
+
+Two options:
+1. Define new `get_cpo()` accessors on ChassisBase and store CPO ports on in a `self.cpo_list` instance variable separate to `self.sfp_list`
```
prgeor: @bgallagher-nexthop can we take this approach. cpo_list and sfp_list separate

doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,265 @@
...
+the form of:
+1. Read information from transceiver EEPROM
+2. Publish to DB
+
+These functions will be refactored to allow customization of step 1, so that CPO can read information from both the optical engine and
+the ELSFP before publishing both to the database. This will maximize code re-use while allow CPO to diverge in its high-level control flow.
+
+###### 7.2.2.3 CpoDomInfoUpdateTask
+
+A subclass of `DomInfoUpdateTask` will be introduced, called `CpoDomInfoUpdateTask`. This task will re-implement the high-level DOM logic
```
micas-net: Please share more detials or examples about it, for example when the cpodominforupdatetask will be run, etc.

doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,265 @@
...
+| get_transceiver_vdm_thresholds | no, but is a thin wrapper around a CmisApi function of the same name |
+
+The philosophy of leveraging DeviceBase and the CmisApi should also work for this task. We will require
+the same refactoring to call the CmisApi instead of the thin wrapper functions on SfpOptoeBase that we
+need to do for the DOM telemetry above.
+
+##### 7.3.2 Changes Required
+
+A CPO specific `SfpStateUpdateTask` subclass will be introduced. All of the existing logic can be re-used, but
+the only difference will be the logic to publish transceiver info to Redis. For CPO, we will need to publish
```
micas-net: Please list the information of ELS which will be present in DATABASE.

doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,265 @@
...
+| get_transceiver_threshold_info | no, but is a thin wrapper around a CmisApi function of the same name |
+| get_transceiver_vdm_thresholds | no, but is a thin wrapper around a CmisApi function of the same name |
+
+The philosophy of leveraging DeviceBase and the CmisApi should also work for this task. We will require
+the same refactoring to call the CmisApi instead of the thin wrapper functions on SfpOptoeBase that we
+need to do for the DOM telemetry above.
+
+##### 7.3.2 Changes Required
+
+A CPO specific `SfpStateUpdateTask` subclass will be introduced. All of the existing logic can be re-used, but
```
KroosMicas: Are all the information of OE/ELS stored in the TRANSCEIVER_INFO table? Is it necessary to create a new database table?

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: You are correct, we can skip creation of this when there are no non-CPO ports. I have updated the code snippet to reflect that.

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: Updated code snippet to support skipping creation of the CPO manager task, same as the existing `CmisManagerTask`.

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: Added more details to this section about the CPO DOM task, including a snippet showing how & when it will be created.

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: Done, committed to this approach in the HLD. I will also update the Port Mapping HLD to explain that `ChassisBase` will provide these methods, since that is probably the better place for this information than this HLD.

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: I have updated this part of the HLD to explain that there will be two tables:
- OE information will go into TRANSCEIVER_INFO. orchagent can continue to respond to creation of this table.
- ELSFP information will go into TRANSCEIVER_ELS_INFO.

Both tables will be created on ELSFP plug-in events and deleted on ELSFP removal.

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: I have updated this section with more in-depth details including a code snippet showing the high-level control flow of the CPO task. I think this likely answers your question -- since the CPO task will differ in its high-level control flow, it will no longer adhere to this same approach of having utility functions that perform the "read from EEPROM and publish to DB" logic internally and is instead free to define its own approach.

Please let me know if the latest updates do not fully answer your question and I can provide more detail.

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: https://gist.github.com/bgallagher-nexthop/5c6f23595a394b3799fad2c367869bb9

This contains a list of the ELSFP fields that are intended to be initially published

I am hestitant to add it to the HLD though, as it is quite extensive and is quite likely to become stale or not reflect reality as time goes on.

bgallagher-nexthop: Scratch that, I have included it in the HLD (with a disclaimer).

micas-net (dismissed):
LGTM!

abelamit (dismissed):

doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,612 @@
...
+        super().log_notice("CPO: {}".format(message))
+
+    def log_error(self, message):
+        super().log_error("CPO: {}".format(message))
+
+    def handle_cmis_dp_deinit_state(self, lport):
+        # Fetch all sibling physical ports that share this optical engine
+        # and disable and deinitialize their lanes.
+        api = self.port_dict[lport].get('api')
+        if self.check_module_state(api, ['ModuleLowPwr']):
```
prgeor, line 173: @bgallagher-nexthop I am not quite sure about necessity of sibling physical ports. Isn't it quite same as the case of breakout ports which is sharing the same physical port? We have all the mapping information in cpo.json to map the lanes for each logical port, physical port and the OE.

doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,612 @@
...
+| `voltage_laneN` | FLOAT | Per-lane supply voltage (V); N = 1..lane_count |
+| `laser_bias_current_laneN` | FLOAT | Per-lane laser bias current (mA); N = 1..lane_count |
+| `optical_power_laneN` | FLOAT | Per-lane output optical power (dBm); N = 1..lane_count |
+| `icc_monitor` | FLOAT | Integrated coherent controller monitor value |
+| `tec_current` | FLOAT | TEC drive current (mA); optional, present only if vendor supports aux monitor |
+| `laser_temperature` | FLOAT | Laser junction temperature (°C); optional, present only if vendor supports aux monitor |
+
+**TRANSCEIVER_ELS_DOM_THRESHOLD**
+
+The following per-lane threshold fields are added for ELSFP entries:
```
prgeor, line 295: @bgallagher-nexthop please confirm if the information is duplicated for each logical port which share the same ELSP?

doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,612 @@
...
+| `module_state` | STRING | CMIS module state |
+| `module_fault_cause` | STRING | Fault cause if module_state is Fault |
+| `enable_laneN` | BOOLEAN | Whether lane N is enabled |
+| `state_laneN` | STRING | Per-lane state: "Lane Output off" \| "Lane Output ramping" \| "Lane Output on" |
+| `output_fiber_checked_laneN` | BOOLEAN | Whether output fiber check has completed for lane N |
+
+Fields that are specific to datapath-capable modules (e.g., `dpinit_pending_hostlane1-8`, TX/RX output status) are **not** written for ELSFP entries.
+
+**TRANSCEIVER_ELS_STATUS_FLAG**
+
```
prgeor, line 352: @bgallagher-nexthop can we make sure that only lanes corresponding to the logical port 
`TRANSCEIVER_ELS_STATUS_FLAG|EthernetXX` is captured in this table? In pluggable optics the number of lanes was limited to 8, but for ELSP there could be more than 8 lanes and doing mental mapping of all these lanes for each logical port for the user is not that easy while debugging any issue.

doc/pmon/xcvrd_cpo.md
```diff
@@ -0,0 +1,612 @@
...
+        return False
+    try:
+        return platform_chassis.get_cpo(physical_port) is not None
+    except (NotImplementedError, AttributeError, IndexError):
+        return False
+```
+This will allow `xcvrd` to make a decision on whether a given port should be handled by the existing set of
+tasks or the new set of CPO-specific tasks described in this HLD.
+
+In the above sections, `xcvrd` also requires the ability to fetch which other physical ports share an
```
prgeor, line 597: @bgallagher-nexthop I am bit lost there as all the mapping information is captured in cpo.json so each cpo object should be independently managed without impacting other cpo objects sharing the OE/ELSP.

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: > Isn't it quite same as the case of breakout ports which is sharing the same physical port?

Correct, it is the exact same problem that is solved for breakout interfaces [here](https://github.com/sonic-net/sonic-platform-daemons/blob/6acdb85aa4d860d2bf174ddb33e65be148b351c6/sonic-xcvrd/xcvrd/cmis/cmis_manager_task.py#L1038-L1041): the entire module is in low power mode, so all lanes module-wide should be disabled/deinitialized.

That logic is not re-usable for CPO though, since the optical engine or ELSFP can be in low power mode but span multiple physical ports so it's no longer as simple as just setting the lane mask to `0xFF`. We now need to fetch the sibling ports and deinitialize/disable their lanes.

> We have all the mapping information in cpo.json to map the lanes for each logical port, physical port and the OE.

That is true, the implementation of `get_oe_sibling_pports` and `get_elsfp_sibling_pports` will leverage `cpo.json` to understand the mapping.

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: Yeah, there will be duplication:

1. non-banked information will be duplicated for each logical interface. We will not read this information multiple times from the hardware, but we will publish it to Redis for each interface since the cost of space usage is so low.
2. banked information (anything per-lane for instance) will not be duplicated. We will only publish per-lane information for the relevant logical interface table.
3. the existing database schema strategy for breakout ports will be used for CPO (DOM info published to first breakout port).

I would be particularly interested in your thoughts on 3 -- it seems nicer to me to have each subport have its own table containing information about the lanes it is responsible for instead of adhering to the old approach used for traditional pluggable transceivers. It comes at the cost of breaking convention (which we are already doing for the CPO task, so this seems like a non-issue) and extra memory usage (more duplicate information in Redis) though that will be a very small amount of extra memory in reality (kilobytes).

I have added a section explaining this to the HLD in "7.2.2.1 Database Schema for CPO".

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: I agree. Any per-lane information published for an interface should only contain the lanes associated with that interface. The only exception could be breakout interfaces (see above comment). I would prefer to also extend this policy to cover breakout interfaces.

Let me know your thoughts.

doc/pmon/xcvrd_cpo.md
bgallagher-nexthop: These functions will leverage the data contained in `cpo.json` to understand which ports share the same underlying OE or ELSFP.

There are two places where this information is required:
- in the CMIS state machine, to ensure we deinit + disable all lanes of an OE when in low power mode
- in the CPO DOM task, to make sure we do not read the same non-banked information from the OE/ELSFP multiple times in one polling cycle. We should read static module-wide information once, and reuse that information for each interface sharing the device. We need to understand what interfaces share the same device in order to deduplicate these reads.

doc/pmon/xcvrd_cpo.md
prgeor: @bgallagher-nexthop  Aren't [2] and [3] conflicting? I thought we want to publish dom info related to that specific logical port? Currently the breakout port dom info is empty for non-primary subport.

bgallagher-nexthop: [2] would just apply in non-breakout case (since most physical ports are 8 lanes on current hardware, banking and interfaces will be 1:1), [3] would only apply if breakout ports are present.

I think we should just merge them both into "Any per-lane, banked information will only be published to the logical interface's table associated with those lanes." and cover both:
- when physical ports are sharing a bank
- when breakout ports are sharing a physical port

I have updated the HLD to reflect this. This means we will no longer follow the approach of publishing DOM info for only the first subport on CPO hardware. We will publish for every subport, but only publish information about the lanes associated with that logical interface.

doc/pmon/xcvrd_cpo.md
prgeor: @bgallagher-nexthop lets do that.

prgeor (approved):

micas-net (approved):
LGTM!!

abelamit (approved):

PR #2453, opened 2026-07-09
Update BMC HLD with Redis communication for SONiC BMC

benle7:
|Repo | PR title                                                                  | state                                                                                                                  | 
|-----| ------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- | 
|[sonic-buildimage](https://github.com/sonic-net/sonic-buildimage/)|[[SONiC BMC] Create BMC OS configuration value for choosing the communication path](https://github.com/sonic-net/sonic-buildimage/pull/29206) | GitHub issue/pull request detail |
|[sonic-platform-common](https://github.com/sonic-net/sonic-platform-common/)|[[SONiC BMC] Get the BMC eeprom from Redis](https://github.com/sonic-net/sonic-platform-common/pull/747) | GitHub issue/pull request detail |
|[sonic-utilities](https://github.com/sonic-net/sonic-utilities/)| [[SONiC BMC] Align BMC CLIs to use the new BMC OS configuration value](https://github.com/sonic-net/sonic-utilities/pull/4807) | GitHub issue/pull request detail |

This PR updates the existing BMC HLD, which was originally designed for communication with OpenBMC via Redfish.
The document has been extended to describe the required changes for supporting communication with a SONiC BMC through the BMC's Redis database.

### DIFF
--- a/doc/bmc/bmc_hld.md
+++ b/doc/bmc/bmc_hld.md
@@ -4,12 +4,29 @@
 Board Management Controller (BMC) is a specialized microcontroller embedded on a motherboard. It manages the interface between system management software and hardware. BMC provides out-of-band management capabilities, allowing administrators to monitor and manage hardware remotely.
 OpenBMC is an open-source project that provides a Linux-based firmware stack for Board Management Controllers (BMCs). It implements the Redfish standard, allowing for standardized and secure remote management of server hardware. In essence, OpenBMC serves as the software that runs on BMC hardware, utilizing the Redfish API to facilitate efficient hardware management.
 Redfish is a standard for managing and interacting with hardware in a datacenter, designed to be simple, secure, and scalable. It works with BMC to provide a RESTful API for remote management of servers. Together, Redfish and BMC enable efficient and standardized hardware management.
-In summary, NOS will deal with BMC through the redfish RESTful API.
+
+In summary, when the BMC runs OpenBMC, the Switch-Host NOS interacts with the BMC through the Redfish RESTful API.
+
+When the BMC runs SONiC OS, the Switch-Host NOS interacts with the BMC through Redis over the host–BMC link (`usb0`). The host connects to the Redis instance on the BMC and reads platform data.
+
+The host selects the communication path at runtime based on `DEVICE_METADATA|bmc.os`.
 
 
 ## 2. BMC flows in SONiC
-The implementation is straightforward: SONiC will incorporate a Redfish client as the underlying infrastructure to support the BMC action. This Redfish client object is implemented and initialized at runtime by SONiC itself. 
+
+SONiC supports two BMC operating-system cases on the Switch-Host side:
+
+| BMC OS | Transport | Client on Switch-Host |
+|--------|-----------|------------------------|
+| **OpenBMC** | Redfish over `usb0` / `bmc_addr` | `RedfishClient` (`redfish_client.py`) |
+| **SONiC** (default) | Redis over `usb0` / `bmc_addr` | Remote STATE_DB connector (`db_connect_remote`) |
+
+For **OpenBMC**, SONiC incorporates a Redfish client as the underlying infrastructure to support BMC actions.
+
 ![general flow](https://github.com/yuazhe/SONiC/blob/90aa83c07c4ae9502a78bd33ce5dc8b8e41b8b7e/images/bmc/bmc_overall_flow.png)
+
+For **SONiC BMC**, the host opens a TCP connection to the BMC Redis instance (STATE_DB, db 6), using the same `bmc_addr` from `bmc.json` / `DEVICE_METADATA|bmc`.
+On Switch-BMC systems, Redis is bound to the BMC link IP at startup (see `dockers/docker-database/docker-database-init.sh` and `dockers/docker-database/supervisord.conf.j2`).
 
 ## 3. BMC ip address initialization
 This is the flow of the bmc ip address configuration: 
@@ -27,7 +44,7 @@ iface usb0 inet static
 ![ip address init flow](https://github.com/yuazhe/SONiC/blob/c3912566a589767e43f12d822dc3611734ae84dc/images/bmc/bmc_ip_set_flow.png)
 
 
-## 3. BMC firmware upgrade flow
+## 3. BMC firmware upgrade flow (OpenBMC only)
 
 It requires a new ComponenetBMC object to be added to the component.py
 
@@ -35,8 +52,11 @@ It requires a new ComponenetBMC object to be added to the component.py
 
 ## 4. Sonic-platform-common support for bmc
 
-### 4.1 BMC redfish client
-The redfish_client.py module provides the RedfishClient class, which facilitates BMC access via cURL requests to Redfish APIs. This class serves as a cURL wrapper for executing various Redfish commands. The class utilizes callback functions to obtain user credentials securely and supports asynchronous task monitoring to handle long-running operations like firmware updates and log dump.
+### 4.1 BMC Redfish client (OpenBMC only)
+
+The `redfish_client.py` module provides the `RedfishClient` class, which facilitates BMC access via cURL requests to Redfish APIs. This class serves as a cURL wrapper for executing various Redfish commands. The class utilizes callback functions to obtain user credentials securely and supports asynchronous task monitoring to handle long-running operations like firmware updates and log dump.
+
+Used when `DEVICE_METADATA|bmc.os` is `openbmc`.
 
 Key functionalities:
 1.	Session Management: Handles login and logout operations, ensuring secure sessions with the BMC. It manages tokens and session IDs, and automates re-login if tokens expire.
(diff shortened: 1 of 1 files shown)


oleksandrivantsiv (dismissed):

benle7:
Hi @judyjoseph 

This PR includes the design updates I presented in the "SONiC BMC WG Meeting".
Can you please review/approve?

Thanks!

doc/bmc/bmc_hld.md
```diff
@@ -4,13 +4,30 @@
...
 OpenBMC is an open-source project that provides a Linux-based firmware stack for Board Management Controllers (BMCs). It implements the Redfish standard, allowing for standardized and secure remote management of server hardware. In essence, OpenBMC serves as the software that runs on BMC hardware, utilizing the Redfish API to facilitate efficient hardware management.
 Redfish is a standard for managing and interacting with hardware in a datacenter, designed to be simple, secure, and scalable. It works with BMC to provide a RESTful API for remote management of servers. Together, Redfish and BMC enable efficient and standardized hardware management.
-In summary, NOS will deal with BMC through the redfish RESTful API.
+
+In summary, when the BMC runs OpenBMC, the Switch-Host NOS interacts with the BMC through the Redfish RESTful API.
+
+When the BMC runs SONiC OS, the Switch-Host NOS interacts with the BMC through Redis over the host–BMC link (`usb0`). The host connects to the Redis instance on the BMC and reads platform data.
```
chander-nexthop, line 10: Why can't we implement the API's in SONiC Redfish too (barring the dump)? Should be fairly straightforward to implement. This way we keep the API layer clean and do away with the bmc os config too?

benle7: As part of the SONiC BMC project architecture, the agreed-upon design was that communication between the host side and the BMC side would be via Redis. Dedicated Redfish endpoints will be added for specific use cases, but those are outside the scope of this PR.

Most of the existing BMC APIs (bmc_base.py) are OpenBMC-specific and are therefore not relevant for SONiC BMC. Examples include session management, resetting the root password, and firmware updates, which in our case will be handled directly on the BMC side.

However, the host still needs access to the BMC EEPROM data so that we can continue supporting the existing `show platform bmc eeprom` CLI. For that reason, the EEPROM information is retrieved from the SONiC BMC Redis database.

chinmoy-nexthop, line 10: I think this design needs another look. It exposes the BMCs internal OS details to every hostside consumer instead of encapsulating them behind a stable API abstraction. this way sharing implementation details across layers and tightly couples consumers(of BMC) to the BMC OS internals, which could make maintenance and future changes significantly harder. please think about it

doc/bmc/bmc_hld.md
```diff
@@ -90,22 +116,57 @@ get_bmc_debug_log_dump(task_id, filename, path)
...
-show platform bmc summary 
+show platform bmc os
+---------------------------
+openbmc
+
+config bmc os sonic
+config bmc os openbmc
+```
+
+`config bmc os` writes `DEVICE_METADATA|bmc.os`. Subsequent BMC CLI/API calls on the Switch-Host use Redfish or Redis according to this setting.
```
chinmoy-nexthop, line 157: why an operator need this API at all ? exposing `config bmc os` as a configurable option suggests the BMC OS is a runtime choice, which it is not.

I would recommend removing it, or at most exposing it as read only platform metadata. a get is ok but not set . please think about it.

doc/bmc/bmc_hld.md
```diff
@@ -146,7 +214,7 @@ do not block or interrupt the standard dump flow.
 ![show techsupport flow](https://github.com/sonic-net/SONiC/blob/30d7b3524e1e1f25abb4679f7ffa777eabe9f499/images/bmc/show_techsupport_flow.png)
 
 ### 6.3 Errors Handling: 
-- generate_dump check whether BMC is suppported (via bmc.json file). If not, BMC logic is skipped. 
+- `generate_dump` checks whether BMC Redfish dump is supported: Switch-Host, `bmc.json` present, and `get_bmc_os() == openbmc`. If not, host-side BMC dump logic is skipped.
```
chinmoy-nexthop, line 217: > `get_bmc_os() == openbmc`  is not looking good . 

ideally , this check should be be in only one place something like "ChassisBase.get_bmc()" ( from bmc.json) , all implementation of subclass creation  like vendor specific custom code and transport strategy can grow independently

doc/bmc/bmc_hld.md
```diff
@@ -45,35 +65,41 @@ Key functionalities:
...
+
+| API | OpenBMC | SONiC BMC |
+|-----|---------|-----------|
+| `get_name()` | Yes | Yes |
+| `get_presence()` | Yes | Yes |
+| `get_model()` | Yes | Yes |
+| `get_serial()` | Yes | Yes |
+| `get_revision()` | Yes | Yes |
+| `get_status()` | Yes | Yes |
+| `is_replaceable()` | Yes | Yes |
```
chinmoy-nexthop, line 86: OpenBMC and SONiC BMC already provide almost all of the capabilities we need (EEPROM, presence, model/serial, status, version, etc.). For those common capabilities, we should expose a single interface that hides the underlying BMC OS.

The current design seems to push the BMC OS decision into the caller and we need to hard code like this 
> if (bmc.os == OpenBMC)
>     use OpenBMC-specific API
> else
>     use SONiC BMC API

That means every consumer now needs to know which BMC OS is running, which creates unnecessary tight coupling.

My proposal : 
 - I think we should separate this by capability, not by BMC OS. Common functionality  should always be accessed through Redfish (or another common API), regardless of whether the platform is running OpenBMC or SONiC BMC. 
 - only genuinely platform specific features should require a platform extension.

The underlying BMC OS should remain an implementation detail. hostside consumers should nt need to know or care about which BMC OS is running.

chinmoy-nexthop:
@benle7  thanks for this proposal, appreciate the effort here.

@oleksandrivantsiv   @benle7  , when you get a chance, could you take another look at some of the points above? I think considering them would help make the design cleaner and more maintainable in future .

doc/bmc/bmc_hld.md
oleksandrivantsiv: @chander-nexthop,

Thanks for the comment.

This HLD extension was done based on a request we received from @judyjoseph. According to the conversation we had, the Redfish API on SONiC BMC has to stay minimal, and the main communication interface between the Host and the BMC should be the Redis DB.

The extension was implemented based on those requirements.

doc/bmc/bmc_hld.md
oleksandrivantsiv: For NVIDIA platforms, the BMC OS *is* a runtime choice — the same hardware can run either OpenBMC or SONiC BMC.

This should not be an issue in practice: the initial configuration is deployed after image installation, or migrated from the previous version, so the value is set as part of the normal configuration flow rather than being flipped arbitrarily at runtime.

doc/bmc/bmc_hld.md
```diff
@@ -90,22 +116,57 @@ get_bmc_debug_log_dump(task_id, filename, path)
...
 ## 5. CLI commands
+
+### 5.1 BMC OS selection
+
+The Switch-Host stores the configured BMC operating system in CONFIG_DB:
+
+- **Table/key**: `DEVICE_METADATA|bmc`
+- **Field**: `os`
+- **Values**: `openbmc` | `sonic`
+- **Default**: `openbmc`
```
judyjoseph: Can we make "sonic" as default in switch host ? Else we need to add change in the internal tooling to add this golden config,

benle7: Hi @judyjoseph 
I updated the default BMC OS to "sonic" as we discussed in the last meeting.
Thanks

judyjoseph: Sure thanks for that update. Btw wanted to confirm that today we don't use this config -"DEVICE_METADATA|bmc". Also since sonic is default value of os field - I assume the code logic will take care of this case when DEVICE_METADATA|bmc is not present in configDB and is a no-op in Switch-Host side of sonic device.

benle7: Correct. If DEVICE_METADATA|bmc is not present in ConfigDB, we’ll treat it as the default case (sonic).

oleksandrivantsiv (approved):

doc/bmc/bmc_hld.md
```diff
@@ -90,22 +116,57 @@ get_bmc_debug_log_dump(task_id, filename, path)
...
+
+Runtime API: `sonic_py_common.device_info.get_bmc_os()`.
+
 ```
-show platform bmc summary 
+show platform bmc os
+---------------------------
+sonic
+
+config bmc os sonic
```
judyjoseph, line 153: I am not sure if we need to have a config command to select bmc OS. We need to find other ways to get the BMC os from sku or some other platform parameter. @benle7, there was a similar comment during community presentation also.  Can you take care of this

doc/bmc/bmc_hld.md
```diff
@@ -90,22 +116,57 @@ get_bmc_debug_log_dump(task_id, filename, path)
...
+| `get_bmc_debug_log_dump()` | Yes | **No** | Redfish only |
+| `update_firmware(fw_image)` | Yes | **No** | Redfish only |
 
 ## 5. CLI commands
+
+### 5.1 BMC OS selection
+
+The Switch-Host stores the configured BMC operating system in CONFIG_DB:
+
+- **Table/key**: `DEVICE_METADATA|bmc`
```
judyjoseph, line 139: Again this OS info could be populated in STATE_DB and shown via a show command. I doub;t if this should be in config db. Usually we don't change OS dynamically.

PR #2461, opened 2026-07-13
HLD for Redfish Certificate Provisioning Integration

shreyansh-nexthop:


doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
```diff
@@ -0,0 +1,798 @@
...
+
+- **Automatic certificate consumption.** A staging helper (`stage-credentials.sh`) inside the redfish container reshapes the provisioned files from `/etc/sonic/credentials/` (`server.crt`, `server.key`, `server_ca.crt`) into what bmcweb expects: a combined `server.pem`, a CA truststore with its `<hash>.0` symlink. bmcweb's code is untouched.
+- **Automatic rotation.** A watcher picks up installs and rotations (inotify, plus a 60 second reconcile as a safety net) and restarts bmcweb, since bmcweb reads certificates only at startup. A cert/key pair guard ensures a mid-rotation mismatch is never staged.
+- **mTLS enforcement.** `TLSStrict` and CN-to-local-user mapping (`MTLSCommonNameParseMode 2`) are enabled once, at first provisioning. Clients must present a certificate signed by the staged CA whose CN names a local BMC user.
+- **Secure mode (fail closed).** A CONFIG_DB flag, `DEVICE_METADATA|localhost` field `redfish_secure_mode`, is auto-set at first provisioning. Once on, every bmcweb start is gated: with no valid staged certificate, bmcweb does not start at all, instead of falling back to a self-signed cert. Recovery is automatic when valid certs reappear.
```
judyjoseph: suggest to keep it out of DEVICE_METADATA .. we can keep it as a separate table for REDFISH -- something like this - is this ok ?

  "REDFISH": {
    "AUTHENTICATION_MODE": {
      "secure_mode": "true"
    }
  }

shreyansh-nexthop: Sure, I've moved the flag out of DEVICE_METADATA into a dedicated REDFISH table as you proposed:

```json
"REDFISH": {
    "AUTHENTICATION_MODE": {
        "secure_mode": "true"
    }
}
```

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
```diff
@@ -0,0 +1,798 @@
+# Redfish Certificate Provisioning Integration
+
+## At a glance
+
+What this design delivers, in one list. Each item has a full section below.
+
+- **Automatic certificate consumption.** A staging helper (`stage-credentials.sh`) inside the redfish container reshapes the provisioned files from `/etc/sonic/credentials/` (`server.crt`, `server.key`, `server_ca.crt`) into what bmcweb expects: a combined `server.pem`, a CA truststore with its `<hash>.0` symlink. bmcweb's code is untouched.
+- **Automatic rotation.** A watcher picks up installs and rotations (inotify, plus a 60 second reconcile as a safety net) and restarts bmcweb, since bmcweb reads certificates only at startup. A cert/key pair guard ensures a mid-rotation mismatch is never staged.
```
judyjoseph: Please add details on what happens to ongoing redfish requests ? So here can we assume that the request is either accepted and already in process by bmcctld or the request itself couldn't go through

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
```diff
@@ -0,0 +1,798 @@
...
+- **mTLS enforcement.** `TLSStrict` and CN-to-local-user mapping (`MTLSCommonNameParseMode 2`) are enabled once, at first provisioning. Clients must present a certificate signed by the staged CA whose CN names a local BMC user.
```
judyjoseph: As dicussed we would not need a local BMC user

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
```diff
@@ -0,0 +1,798 @@
...
+- **Secure mode (fail closed).** A CONFIG_DB flag, `REDFISH|AUTHENTICATION_MODE` field `secure_mode`, is auto-set at first provisioning. Once on, every bmcweb start is gated: with no valid staged certificate, bmcweb does not start at all, instead of falling back to a self-signed cert. Recovery is automatic when valid certs reappear.
```
judyjoseph: Please update back if we really need a config flag for "secure_mode". Before the CA certificate is present, we could use the self-signed certificate and support https requests like /redfish/v1. 

But once we have the CA certificate and mTLS handshake is complete - can we store this status somewhere -- may be in redis STATE_DB or in a file  ? This could be used to prevent fallback to self-signed certificate any time in the switch lifecycle.

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
```diff
@@ -0,0 +1,798 @@
...
+- The condition is immediately visible: STATE_DB reports `in_sync false` with a "credentials not ready" reason.
+- If bmcweb restarts while the source is gone, the guard decides: the staged certificate is still present and valid, so bmcweb starts normally even in secure mode. Only if the staged certificate is also gone does secure mode fail closed.
+
+Revocation is done by rotating the CA (clients trusting the old CA stop being accepted), not by deleting files.
+
+## Observability
+
+A rotation that silently fails to land is a security flaw: the endpoint keeps answering with the old certificate and nobody notices. The design makes cert state observable in two places.
+
+**STATE_DB status.** On every cycle the watcher publishes to STATE_DB key `REDFISH_CERT_STATUS|global`:
```
judyjoseph, line 291: So will this STATE_DB table/key share the status of certificate provision ? Can we add a similar one to show that the mTLS handshake is complete?

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
```diff
@@ -0,0 +1,798 @@
...
+-v /var/run/redis:/var/run/redis:rw      # redis unix socket
+-p 0.0.0.0:443:18080                     # bmcweb listens on 18080, published on host :443
+```
+
+What each mount means for this design:
+
+- **`/etc/sonic:ro`**: the provisioning agent installs certs under `/etc/sonic/credentials/`, which lives under `/etc/sonic`, so they are already visible inside the container. The mount is read-only, and the staged files live elsewhere regardless: `/etc/sonic/credentials` is the provisioning directory, and the reshaped copies bmcweb reads are kept separate from it.
+- **`/var/lib/bmcweb:rw`**: the fingerprint stamp lives here (`/var/lib/bmcweb/.staged-credentials.stamp`). Unlike files in the container's writable layer, it survives container recreation, which the boot and restart behavior relies on.
+- **`/var/run/redis:rw`**: the container reads the secure-mode flag from CONFIG_DB and publishes certificate status to STATE_DB through this mounted redis socket.
+- **`-p 443:18080`**: client requests hit `:443` while bmcweb internally binds `18080`.
```
judyjoseph: Can we make this bmcweb port open to clients -- configurable ? and keep it default as 443

judyjoseph:
@qiluo-msft @yxieca for review

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
shreyansh-nexthop: Yes, have updated the doc

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
shreyansh-nexthop: Have added

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
shreyansh-nexthop: Have removed it.

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
shreyansh-nexthop: Will store this info in /var/lib/bmcweb/provisioned in the host. No need of redis here.

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
shreyansh-nexthop: Yes, the key `REDFISH_CERT_STATUS|global` will carry the info. Have added the new field `mtls_enforced`. It will be true when TLSStrict is enabled and valid CA-signed certificate is staged.

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
```diff
@@ -0,0 +1,864 @@
...
+
+Staging the certificates makes mTLS possible but does not by itself make bmcweb request or require a client certificate. Enforcement is controlled by bmcweb's persistent state file `/bmcweb_persistent_data.json`, and enabling it (`enable_mtls`) is a required part of first-time provisioning.
+
+The keys that matter in its `auth_config` block:
+
+| Key | Value | Effect |
+|---|---|---|
+| `TLS` | `true` | allow client-certificate authentication |
+| `TLSStrict` | `true` | require a verified client certificate (verify_peer + fail_if_no_peer_cert) |
+| `MTLSCommonNameParseMode` | `2` | take the client cert's CommonName as the session identity |
```
judyjoseph, line 223: @shreyansh-nexthop @chinmoy-nexthop, We need to add support for enforcement with the CommonName to make sure we accept the valid CN's (these we configure explicitly in the CONFIG_DB). The common Names could be wildcard entries too -- I can share a sample of wildcard CNs later.

Another similar wildcard CN was added in restapi server where client autentication is done : https://github.com/sonic-net/sonic-restapi/pull/168. 

We need a similar support in redfish/bmcweb as well when we process the CN

@qiluo-msft cc .. please comment further

judyjoseph:
@shreyansh-nexthop you could refer to this config https://github.com/sonic-net/sonic-mgmt/tree/master/ansible/roles/testbed/nut/templates/config_patch/common .. we could have redfish config similar to how it is done for other services like gnmi/restapi/telemetry

Sample config as below, so with this we can use the attributes server_crt/server_key/ca_crt etc and the other configs to set the port as configurable, client_auth = cert, where CN validation is enforced OR none.
```
  // Similar to restapi/telemetry certs
 "REDFISH": {
    "config": {
      "client_auth": "cert|none",
      "port": "443"
    },
    "certs": {
      "server_crt": "/etc/sonic/redfish/redfishserver.cer",
      "server_key": "/etc/sonic/redfish/redfishserver.key",
      "ca_crt": "/etc/sonic/credentials/ROOT_CERTIFICATE.pem"
    }
 }
```

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
shreyansh-nexthop: @judyjoseph, this is doable. We can modify the sonic-dbus-bridge code to first verify if the CN is acceptable or not. Will update the HLD with the info.

shreyansh-nexthop:
> @shreyansh-nexthop you could refer to this config https://github.com/sonic-net/sonic-mgmt/tree/master/ansible/roles/testbed/nut/templates/config_patch/common .. we could have redfish config similar to how it is done for other services like gnmi/restapi/telemetry
> 
> Sample config as below, so with this we can use the attributes server_crt/server_key/ca_crt etc and the other configs to set the port as configurable, client_auth = cert, where CN validation is enforced OR none.
> 
> ```
>   // Similar to restapi/telemetry certs
>  "REDFISH": {
>     "config": {
>       "client_auth": "cert|none",
>       "port": "443"
>     },
>     "certs": {
>       "server_crt": "/etc/sonic/redfish/redfishserver.cer",
>       "server_key": "/etc/sonic/redfish/redfishserver.key",
>       "ca_crt": "/etc/sonic/credentials/ROOT_CERTIFICATE.pem"
>     }
>  }
> ```

Sure, this can be done, will update the HLD

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
```diff
@@ -0,0 +1,966 @@
...
+- **Deletion behavior.** Deleting the source certificates does not tear down the running service (bmcweb keeps serving the last staged certificate), and it is not a way to revoke access: the running bmcweb keeps serving the already-loaded certificate and keeps trusting the same CA. To actually revoke, rotate the CA (clients whose certs were issued by the old CA then fail the mTLS handshake).
+- **Observability.** The watcher publishes sync status to STATE_DB (`REDFISH_CERT_STATUS|global`: `in_sync`, `last_error`, fingerprints, served serial, plus `mtls_enforced` for the enforcement state) every cycle, and every decision is logged to syslog. A rotation that fails to land is visible, not silent; monitoring must watch `in_sync`.
+- **Boot resilience preserved.** Before certificates are ever provisioned, the BMC still boots with bmcweb's self-signed certificate and the API is reachable.
```
judyjoseph, line 13: So assuming the following is true 

```
By default when BMC boots up for first time, the certificate bmcweb serves is not CA-signed. 
It is a self-signed certificate that bmcweb generates itself at first startup -- There is no CA, no truststore, and mTLS is off. In that default state /redfish/v1 works. 
Because no client certificate is required, you can reach it with curl -k https:///redfish/v1                                                                                                                                                                                                                                                                        
```

shreyansh-nexthop: Yes, this is true

doc/sonic-redfish/Redfish_Certificate_Provisioning_HLD.md
```diff
@@ -0,0 +1,966 @@
...
+- **Automatic rotation.** A watcher picks up installs and rotations (inotify, plus a 60 second reconcile as a safety net) and restarts bmcweb, since bmcweb reads certificates only at startup. A cert/key pair guard ensures a mid-rotation mismatch is never staged. Requests in flight at the restart fail at the connection level; see "Applying a change" for the exact semantics.
+- **mTLS enforcement, without local users.** With `client_auth` set to `cert`, `TLSStrict` requires clients to present a certificate signed by the staged CA and valid for client authentication. The certificate's common name is then checked against the trusted names configured in CONFIG_DB (`REDFISH|certs` field `client_crt_cname`), which accepts exact entries and wildcards such as `*.example.com`. There are no local users or passwords in the container: an accepted client is authorized as an administrator, and the common name is also what appears in sessions and logs.
```
judyjoseph, line 9: @shreyansh-nexthop from the redfish spec https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.23.0.pdf

```
13.3.2 Authentication requirements
13.3.2.1 Resource and operation authentication requirements
Services shall authenticate all write requests to Redfish resources. For example:
       • POST, except to the
       Sessions resource collection for authentication
        • PUT
        • PATCH
        • DELETE
Redfish resources shall not be available as unauthenticated, except for:
        • The service root to identify the device and service locations.
```
The redfish/v1 service root call unauthenticated and needs to be available allways. With TLS strict mode we cannot achive as we need to present the client certificate which server needs to autenticate. Can we use "TLS true, TLSStrict false." I think bmcweb support this ? @qiluo-msft  cc

shreyansh-nexthop: We can achieve this @judyjoseph. As you mentioned, "TLS as true and TLSStrict as false", should fulfil this requirement. With this, all the authenticated endpoints will need certificates and un-authenticated ones works without any authentication.

I will do a testing with this change and then update the HLD. Thanks!

PR #2466, opened 2026-07-20
Event driven switchover in DualToR

vivekverma-arista:
This HLD proposes an event driven switchover for DualToR deployments in case of link failures. Today the switchover is partially poll driven.

With an end-to-end event driven switchover, the convergence times are no longer a function of configured poll interval on the devices and reduce significantly.


vivekverma-arista:
@lolyu @StormLiangMS @judyjoseph @lguohan

PR #1187, opened 2022-12-15
Updating Contributors

ixodie:
I was trying to add a new community contributor and noticed that the website uses a single image which has not been updated in years.  The source of the image cannot be found.

I implemented a dynamic responsive display of the contibutors.json file using js and added all of the images to a new directory and put that data into the json.

I could not find 3 companies (I cannot read Chinese), but I left the original images in case someone can determine who they are.  They are noted in the json file as "unknown".

Several companies have been acquired, I did not include any acquired companies provided that the new company is already listed (ex. Apstra/Juniper).

PR #2481, opened 2026-07-28
[fpmsyncd] HLD: Producer-side ZMQ route delivery resiliency and heap release

deepak-singhal0408:
## Summary
- Adds a new HLD under `doc/fpmsyncd/` describing a producer-side redesign of the **fpmsyncd → orchagent ZMQ route delivery** path.
- The design decouples FPM/netlink ingest from ZMQ delivery via a **coalescing map** (last-writer-wins) + a **dedicated send thread**, drains it in **bounded, byte-capped chunks** (one wire message each, under the transport's single-message ceiling), and replaces the inline per-prefix retry ladder with a two-level **inner blip-absorber + outer retain-and-retry**. A prolonged unrecoverable stall becomes a bounded, observable failure (sticky STATE_DB assert record + process exit → restart RIB replay), never a silent drop.
- Adds an **event-driven, chunked, abortable heap release** on the send thread's busy→idle edge (fpmsyncd links **tcmalloc**), returning burst-inflated free memory to the OS without a dedicated thread and without delaying route delivery.
- Documents the STATE_DB telemetry contract (route-stat + heap-stat tables, ABNF schema) and the internal tuning knobs (no CONFIG_DB/CLI/YANG change).
- Diagrams are checked in as external PNGs under `doc/fpmsyncd/images/fpmsyncd_zmq_route_delivery/` (no inline base64).

## Why
Tracked under [sonic-net/sonic-buildimage#28369](https://github.com/sonic-net/sonic-buildimage/issues/28369): under a large FIB burst fpmsyncd's inline, one-prefix-per-message send path spins a bounded retry loop on a full ZMQ socket and then **silently drops** the route — no back-pressure, no telemetry, no crash — so the ASIC FIB diverges from the RIB with no signal. Stacked issue [sonic-net/sonic-buildimage#28245](https://github.com/sonic-net/sonic-buildimage/issues/28245): fpmsyncd's RSS ratchets up after each burst and never comes back down (heap free-but-not-returned). This producer-side design is complementary to, and independent of, the orchagent consumer redesign in [sonic-net/SONiC#2328](https://github.com/sonic-net/SONiC/pull/2328).

## Test plan
- [ ] HLD reviewed against the [SONiC HLD template](https://github.com/sonic-net/SONiC/blob/master/doc/guidelines/hld_template.md).
- [ ] Diagrams render correctly on GitHub via the relative `images/fpmsyncd_zmq_route_delivery/*.png` links.
- [ ] Doc-only change; no build/test impact.


doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
```diff
@@ -0,0 +1,829 @@
...
+## 5. Requirements
+
+RFC-2119 keywords. A requirement states an **outcome**; the mechanism that achieves it is a design
+choice (§7) guarded by a test (§13).
+
+### 5.1 Functional
+
+- **R1.** fpmsyncd **shall not** silently drop a route update in any recoverable regime;
+  `routes_lost_total` **shall** remain `0` except across a deliberate assert+restart, which the
+  restart RIB replay then repairs.
```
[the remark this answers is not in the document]
deepak-singhal0408: Reworded R1 to remove the "except across a deliberate assert+restart" phrasing that implied the counter could go non-zero. It now states routes_lost_total remains 0 in all regimes; an assert+restart interrupts in-flight delivery but the restart RIB replay re-drives the affected routes, so the counter still does not increment — consistent with the §9.3.1 schema ("0 always").

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
```diff
@@ -0,0 +1,829 @@
...
+key                        = FPMSYNCD_HEAP_STAT_TABLE|global
+heap_rss_bytes             = 1*DIGIT      ; RSS symptom (/proc/self/statm)
+heap_inuse_bytes           = 1*DIGIT      ; genuinely-live bytes
+heap_free_held_bytes       = 1*DIGIT      ; free-but-not-returned to the OS = the retained bytes
+retention_ratio            = 1*DIGIT "." 1*DIGIT ; derived free_held/rss — one-glance health (<0.3)
+allocator                  = "tcmalloc"  ; linked allocator (fpmsyncd links tcmalloc)
+release_runs_total         = 1*DIGIT      ; releases executed
+release_skipped_busy_total = 1*DIGIT      ; releases aborted/skipped because busy (starvation signal)
+released_bytes_total       = 1*DIGIT      ; cumulative bytes returned to the OS
+release_last_ts            = ; "YYYY-MM-DD HH:MM:SS" UTC — NOT epoch
```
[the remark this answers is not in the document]
deepak-singhal0408: Fixed the ABNF: release_last_ts now has an explicit value production (4DIGIT "-" 2DIGIT "-" 2DIGIT SP 2DIGIT ":" 2DIGIT ":" 2DIGIT) matching the "YYYY-MM-DD HH:MM:SS" example, instead of an empty right-hand side.

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
```diff
@@ -0,0 +1,829 @@
...
+**Backward compatibility:** STATE_DB *outputs* only — like the schemas above, they add no existing
+table or configuration change (§9.3), and consumers that ignore them are unaffected.
+
+---
+
+## 10. Warmboot and Fastboot Design Impact
+
+Warmboot/fastboot behavior is **out of scope** for this design and unchanged from the baseline —
+this change touches only fpmsyncd's steady-state producer path, not the boot/reboot sequence, and
+adds nothing to the boot critical chain.
```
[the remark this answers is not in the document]
deepak-singhal0408: another subsection would be redundant here.

deepak-singhal0408:
tag @prsunny @qiluo-msft @prabhataravind @abdosi for viz/review..
 
also cc: @rlhui @lguohan for viz.

venkit-nexthop (approved):
Looks good to me @deepak-singhal0408

kenneth-arista:
@rejithomas-arista

rejithomas-arista:
Hi Deepak,

I couldn't find the comparison findings you mentioned in the #28245 thread. I may have overlooked them, so just wanted to check.
Since #28245 is the issue I had raised around glibc arena-top pinning and freed memory not being returned to the OS, wanted to share a few observations on the proposed fix.

**On HeapReleaseManager / #28245**

I believe the tcmalloc library referenced here is the gperftools implementation you mentioned in the issue thread (0110-lib-build-support-gperftools-tcmalloc-release-free-memory-to-host-OS.patch). In that context, HeapReleaseManager appears to add application-level machinery around allocator release behavior, including chunked 4 MiB releases, an abort predicate, a 500 ms debounce, send-thread state transitions, and additional telemetry.

My understanding is that this is needed because the chosen tcmalloc implementation does not independently release memory during idle periods unless explicitly driven. Since the release runs from the same thread that delivers routes, the abort predicate is also needed to avoid delaying route delivery while reclaim is in progress.

Advantage with using jemalloc is that memory release happens independently through its decay/background purge mechanism, so fpmsyncd does not need route-send-thread involvement or application-level release scheduling. It also avoids the glibc brk/arena-top fragmentation behavior by using extent-based allocation.  
If tcmalloc is required for broader SONiC consistency, google/tcmalloc  may also be worth considering, as it includes a background release thread and would likely avoid the need for HeapReleaseManager. 

**On end-to-end back-pressure and the route delivery design**

The producer-side coalescing and retry design looks like a good fix for the silent drop problem in #28369. One observation I had  was around end-to-end back-pressure.   
Today, orchagent's ZMQ receive path  drain the ZMQ socket into an internal queue independently of SAI progress. The poll thread can continue receiving and enqueueing while RouteOrch/syncd/SAI are still behind. As a result, fpmsyncd may not see sustained ZMQ back-pressure even when the real bottleneck is downstream. It continues reading from the FPM TCP socket, so zebra's FPM output buffer never fills, the dplane queue never reaches its limit, and zebra's RIB processing never slows down. Instead, the unbounded queue in orchagent ends up absorbing the backlog. 
With the proposed producer-side design, fpmsyncd also decouples FPM ingest from ZMQ delivery, the FPM thread writes into a coalescing map and returns immediately. That improves fpmsyncd responsiveness, but under sustained downstream overload it also means zebra may still not receive meaningful back-pressure.
The backlog can therefore exist in two places: the fpmsyncd coalescing map and the orchagent ZMQ consumer queue. If overload continues, the producer-side map is bounded by mMax and eventually exits/restarts, which is observable and certainly better than silent loss, but it is still not true end-to-end back-pressure back to zebra.

One question I had while reading the HLD was about the coalescing map. I couldn't find any details on the underlying data structure, so I wasn't sure how ordering across different keys is preserved during chunk extraction. Since FRR sends updates in dependency order (for example, NHGs before the routes that reference them etc), I was wondering whether insertion order is tracked, or whether ordering is preserved in some other way if an unordered container is used.
 
 To me, the producer-side work and the consumer-side work appear complementary, as you mentioned. I hope issue #28576 will remain open to track the consumer-side queue growth and dedup work. 
 
Just wanted to share a few observations after reading through the current proposal.

deepak-singhal0408:
Thanks a lot for the detailed review and sharp analysis @rejithomas-arista. My response inline.


> Hi Deepak,
> 
> I couldn't find the comparison findings you mentioned in the #28245 thread. I may have overlooked them, so just wanted to check. Since #28245 is the issue I had raised around glibc arena-top pinning and freed memory not being returned to the OS, wanted to share a few observations on the proposed fix.
[DS]: No, you haven't missed it. I haven't posted the comparison data yet.
> 
> **On HeapReleaseManager / #28245**
> 
> I believe the tcmalloc library referenced here is the gperftools implementation you mentioned in the issue thread (0110-lib-build-support-gperftools-tcmalloc-release-free-memory-to-host-OS.patch). In that context, HeapReleaseManager appears to add application-level machinery around allocator release behavior, including chunked 4 MiB releases, an abort predicate, a 500 ms debounce, send-thread state transitions, and additional telemetry.
> 
> My understanding is that this is needed because the chosen tcmalloc implementation does not independently release memory during idle periods unless explicitly driven. Since the release runs from the same thread that delivers routes, the abort predicate is also needed to avoid delaying route delivery while reclaim is in progress.
> 
> Advantage with using jemalloc is that memory release happens independently through its decay/background purge mechanism, so fpmsyncd does not need route-send-thread involvement or application-level release scheduling. It also avoids the glibc brk/arena-top fragmentation behavior by using extent-based allocation.   If tcmalloc is required for broader SONiC consistency, google/tcmalloc may also be worth considering, as it includes a background release thread and would likely avoid the need for HeapReleaseManager. 

[DS]: First, onto your original point: jemalloc's extent-based allocation does avoid the glibc brk/arena-top pinning you reported. That isn't in dispute.

Where I start from is consistency rather than a claim that tcmalloc is the better allocator. FRR is already built with gperftools tcmalloc by default in master, and fpmsyncd runs in that same bgp container alongside zebra. Introducing a second allocator there means two sets of tuning knobs and two debugging surfaces in one container, so I'd rather not diverge without a concrete reason.

You anticipated exactly that and offered google/tcmalloc as the consistency-preserving option. I went through both at source level, and my read is that neither removes the need for the release hook.

**jemalloc 5.3.0** — at defaults (background_thread:false), decay purge is driven by a probabilistic tick on malloc/free (arena_decay_ticks, arena_inlines_b.h) and runs on whichever thread trips the tick. In fpmsyncd that can land on the route send thread, and it means no purge during idle (no allocations, no ticks) and purge during bursts. With background_thread:true it moves off-thread, but the wake is a deadline timer, and the only load coupling is arena_should_decay_early(), which wakes the purger earlier on an influx of dirty pages. The purge loop itself (pac_decay_stashed(), pac.c) has no cancellation point.

**google/tcmalloc** — the rate-limited release is genuinely better bounded than jemalloc's decay stash; bytes = rate × elapsed per tick is close to the chunking we want. But background_release_rate defaults to 0, and ProcessBackgroundActions() is a loop the application has to spawn a thread for — the library doesn't create one. So the application-level machinery doesn't go away, it moves, and it has no abort. Separately, skip_subrelease (on by default, 60 s / 300 s) declines to release below recent demand peaks; on a bursty route workload peaks recur inside those windows, so it would tend to hold RSS near the peak rather than return it between bursts.

So HeapReleaseManager isn't there to work around gperftools. It's there because this workload needs release that starts on an idle edge and stops when routes arrive. Under all three allocators that hook still has to exist — only the backend primitive changes (mallctl("arena.<i>.decay") / ReleaseMemoryToSystem() / ReleaseFreeMemory()).

**On the comparison data:**  Sorry, my bad — I didn't post an update on the issue.
Having gone through both allocators' release paths, the allocator choice isn't what decides this — the idle-edge hook is needed under all three, so a jemalloc-vs-tcmalloc bake-off wouldn't change the design. 
I'd rather retire that explicitly (have done some simple absolute case benchmarking with jemalloc as well, and perf is no better than tcmalloc). Your RSS field data stands on its own. 
With tcmalloc + HeapReleaseManager, the release-cost side is already measured on the shipping allocator in Appendix A.2 of this HLD: 129 MiB returned in one pass at ~108 µs/MiB, handed back in 4 MiB slices with the abort predicate checked between slices, and with bursts arriving every ~2 ms drain throughput is unchanged within ±1.5% and typical latency flat (p99 up a few percent).   Hope this one is settled with this data. Let me know if you think otherwise.
> 
> **On end-to-end back-pressure and the route delivery design**
> 
> The producer-side coalescing and retry design looks like a good fix for the silent drop problem in #28369. One observation I had was around end-to-end back-pressure.    Today, orchagent's ZMQ receive path drain the ZMQ socket into an internal queue independently of SAI progress. The poll thread can continue receiving and enqueueing while RouteOrch/syncd/SAI are still behind. As a result, fpmsyncd may not see sustained ZMQ back-pressure even when the real bottleneck is downstream. It continues reading from the FPM TCP socket, so zebra's FPM output buffer never fills, the dplane queue never reaches its limit, and zebra's RIB processing never slows down. Instead, the unbounded queue in orchagent ends up absorbing the backlog.  With the proposed producer-side design, fpmsyncd also decouples FPM ingest from ZMQ delivery, the FPM thread writes into a coalescing map and returns immediately. That improves fpmsyncd responsiveness, but under sustained downstream overload it also means zebra may still not receive meaningful back-pressure. The backlog can therefore exist in two places: the fpmsyncd coalescing map and the orchagent ZMQ consumer queue. If overload continues, the producer-side map is bounded by mMax and eventually exits/restarts, which is observable and certainly better than silent loss, but it is still not true end-to-end back-pressure back to zebra.
> 
[DS]: Agree with you. The producer-side change does decouple FPM ingest from ZMQ delivery, and backlog can sit in both the coalescing map and the orchagent consumer queue. mMax converts silent loss into a bounded, observable failure, but that's containment, not back-pressure to zebra.

And yes — [28576](https://github.com/sonic-net/sonic-buildimage/issues/28576) will stay open until all the consumer-side PRs are in; agreed the producer and consumer halves are complementary.

> One question I had while reading the HLD was about the coalescing map. I couldn't find any details on the underlying data structure, so I wasn't sure how ordering across different keys is preserved during chunk extraction. Since FRR sends updates in dependency order (for example, NHGs before the routes that reference them etc), I was wondering whether insertion order is tracked, or whether ordering is preserved in some other way if an unordered container is used.    To me, the producer-side work and the consumer-side work appear complementary, as you mentioned. I hope issue #28576 will remain open to track the consumer-side queue growth and dedup work.    Just wanted to share a few observations after reading through the current proposal.
[DS]: 

**NHG-before-route:**  fpm use-next-hop-groups is off by default in SONiC (dockers/docker-fpm-frr/frr/zebra/zebra.conf.j2), so zebra includes nexthop information in the route messages themselves. Each APP_ROUTE_TABLE entry is self-contained and there's no NHG→route reference to order. When it is enabled, next-hop groups go to APP_NEXTHOP_GROUP_TABLE over Redis while routes go over ZMQ — those two were never ordered relative to each other, before or after this change — and RouteOrch retries a route until its group resolves. There are some future enhancements planned to make NHG support enabled by default. We could explore further how to streamline both.

**The container is std::map<std::string, KeyOpFieldsValuesTuple>.** It drains in prefix order — sorted by key, not by arrival — so on its own it gives no FIFO or dependency guarantee.

Across different prefixes the map delivers in key order, not arrival order, and that's deliberate. orchagent's m_toSync is itself a std::multimap keyed by key, and only P4Orch opts into the ordered-queue path (setOrderedQueue()), so route delivery is already key-ordered rather than FIFO downstream. Per-key ordering is strict: a key holds one pending entry, last-writer-wins, and the op is part of the coalesced value, so SET/DEL can't invert. On a transient send failure the chunk is re-merged without clobbering newer ingest. Because entries are self-contained, collapsing a prefix to its latest state is lossless.

rejithomas-arista:
Thanks Deepak for the thorough responses and the work.

**On HeapReleaseManager / jemalloc:** I have left a follow up in the jemalloc thread asking whether background_thread:true was tested, since the observed 9-10 ms tail latency appears to reflect foreground purging rather than the configuration I had in mind. I won't repeat the details here—just mentioning it for completeness. Thanks for putting the data together.

**On back-pressure:** I agree that bounding the queue with mMax is a clear improvement over silently dropping updates. Since work is happening on both the producer and consumer sides, I was wondering whether end-to-end back-pressure across the full pipeline (syncd -> orchagent -> fpmsyncd -> zebra) has been considered or is being tracked as future work. Zebra's dplane already has the machinery to apply back-pressure, but it never really comes into play if both orchagent's poll thread and fpmsyncd's ingest thread immediately drain their sockets into in-memory buffers. I was not following the Routing WG discussions, so if there is already work in progress in this area, I would appreciate a pointer to understand what's happening..

**On ordering:** Fair point that NHGs are disabled by default today and routes are effectively self-contained. I am also not convinced there is a practical benefit to preserving arrival order in the current SONiC routing architecture. I only wanted to raise it as something worth keeping in mind since I am not uptodate with work happening on the infra side..

venkit-nexthop:
> 
> **On back-pressure:** I agree that bounding the queue with mMax is a clear improvement over silently dropping updates. Since work is happening on both the producer and consumer sides, I was wondering whether end-to-end back-pressure across the full pipeline (syncd -> orchagent -> fpmsyncd -> zebra) has been considered or is being tracked as future work. Zebra's dplane already has the machinery to apply back-pressure, but it never really comes into play if both orchagent's poll thread and fpmsyncd's ingest thread immediately drain their sockets into in-memory buffers. I was not following the Routing WG discussions, so if there is already work in progress in this area, I would appreciate a pointer to understand what's happening..

@rejithomas-arista I was wondering what would zebra do when it receives the backpressure? Dont we have to implement a similar coalescing datastructure there (which would look similar to the one in fpmsyncd)?
We cannot backpressure the protocol that produces the route (example BGP), because it may still be capable of advertising the routes to peers.

rejithomas-arista:
Hi Venkat,
As I see zebra already has same-prefix coalescing. When dataplane processing backs up, zebra’s meta queue  stops draining  and route changes for the same prefix continue to converge. If fpmsyncd slows its socket reads, pressure can eventually propagate back through zebra's FPM output path and dplane provider queues. Once that reaches zebra's RIB/meta-queue processing, the existing RIB-level coalescing can help unless I am missing something in my reading of code. 
Also there is no protocol-level back-pressure due to this. zebra continues reading from BGP ie the back-pressure stays between zebra and orchagent's m_toSync…
The reason I raised the back-pressure question is that I see the ZMQ route path as a fast pipe between two existing  convergence points: zebra's RIB processing on the producer side and orchagent's  m_toSync (which coalesces via addToSync) processing on the consumer side removing the earlier Redis round-trip overhead.  fpmsyncd was mostly acting as a bridge: read FPM and forward inline. orchagent's poll thread is a tight loop — recv, push through. The pipe seemed as intended to be fast and transparent.
The current proposed changes introduces additional processing and buffering  within this pipe. I am not raising this as a blocker. Based on my reading of the code, I arrived at a slightly different mental model of where back-pressure would naturally converge. Since I am still building context on the FRR/SONIC routing side, wanted to understand whether there's additional context or design rationale..
One related thing I noticed while looking at the zebra FPM dplane code: if the FPM output path cannot make write progress for 15 seconds, the FPM plugin resets/reconnects which schedules RIB replay...

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
```diff
@@ -0,0 +1,834 @@
...
+```
+
+A dedicated `show` wrapper is a possible future enhancement (§14); it is intentionally not part of
+this change to keep the surface minimal and backward compatible.
+
+### 9.3 Config DB and tuning knobs
+
+**No CONFIG_DB tables are added or changed** — hence the design is trivially **backward
+compatible**: upgrading to (or downgrading from) an image with this change does not alter any
+operator-visible configuration or default behavior. The correctness fix is **always on** (it has no
```
anamehra: Current code enables the ZMQ route path only when `SYSTEM_DEFAULTS|swss_zmq` is enabled, otherwise it falls back to `ProducerStateTable`. Should the HLD say this fix applies only when route ZMQ is enabled, or is the intent to make this path unconditional?

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
```diff
@@ -0,0 +1,834 @@
...
+health.
+
+### 7.11 Repository / component change map
+
+| Repo | Component | Change |
+|---|---|---|
+| `sonic-swss-common` | `ZmqClient` / `ZmqProducerStateTable` | Generic send-path overflow guard + a clamped retry-config setter (max attempts and backoff ceiling) + send-path counters (raw blips, blips absorbed, max backoff). **No route-specific logic.** Version-bumped. |
+| `sonic-swss` | `fpmsyncd` — new `RouteSendCoalescer` | Coalescing map + dedicated send thread + chunked drain + inner blip-absorber wiring + outer retry + episode hysteresis + assert/liveness + route & heap STATE_DB telemetry. Exposes an **idle-hook seam**. |
+| `sonic-swss` | `fpmsyncd` — new `HeapReleaseManager` | Chunked, abortable heap release on the busy→idle edge; wires the seam above. |
+| `sonic-swss` | `fpmsyncd` — `RouteSync` | Funnels the steady-state route/label-route write path through the coalescer; installs the heap-release idle hook. |
```
anamehra: `ZmqClient::sendMsg(dbName, tableName, kcos)` sends one DB/table per message, while `RouteSync` has both route and label-route tables. Should the coalescing key include `(db, table, key)`, or should route and label-route use separate coalescers?

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
```diff
@@ -0,0 +1,834 @@
...
+## 7. High-Level Design
+
+### 7.1 Design overview
+
+The producer path is redesigned around a single principle: **decouple FPM ingest from ZMQ delivery,
+and never let the delivery side drop a route.** Concretely, the inline "serialize-and-send each
+prefix on the FPM thread" path is replaced by the following set of mechanisms, listed here in
+dataflow order; each is detailed in the subsection noted.
+
+1. **Asynchronous send thread** — the FPM thread hands each update off in microseconds and returns
```
anamehra, line 287: Current `RouteSync` can send the FPM offload reply before the route update is delivered to APP_DB/ZMQ when `suppress-fib-pending` is disabled. With async queueing, should the reply happen after enqueue, after successful ZMQ send, or remain as-is?

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
```diff
@@ -0,0 +1,834 @@
...
+
+- **Inner blip absorber:** on a transient "would block", the send path retries a small number of
+  times with a short backoff (a few milliseconds total). A momentary HWM blip clears here and
+  **never escalates**; absorbed blips are counted separately from real congestion (§7.10).
+- **Outer retry:** if the blip absorber is exhausted, the whole chunk is **re-merged into the live
+  map** (§7.3) and the drain aborts to a backoff; the next drain retries. A congestion **episode**
+  opens under hysteresis, so isolated blips do not inflate the episode count.
+- **Liveness bound:** if the time since the last successful send exceeds `tFailMs`, or the map depth
+  exceeds `mMax`, the producer writes a **sticky STATE_DB assert record** (which survives the
+  process exit) and exits for recovery. The restart triggers **RIB replay**, which
```
anamehra: The recovery path depends on fpmsyncd restart causing FRR RIB replay. Can we add evidence or a test showing that routes queued only in the coalescer are recovered after fpmsyncd exits?

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
```diff
@@ -0,0 +1,834 @@
...
+producer-side send path has two independent defects that surface under **burst** (tens of
+thousands of route updates in a short window) — a scale point SONiC routinely hits on session
+flap, peer bounce, or a full-table refresh.
+
+### 4.1 Problem A — silent route drop under burst (#28369)
+
+In the baseline (pre-fix) implementation, fpmsyncd sends **one prefix per ZMQ message**, **inline
+on its single main thread**. When a
+burst fills the ZMQ send buffer, the send returns a would-block (`EAGAIN`); the code spins a **bounded**
+inline retry ladder and, once the ladder exhausts, **silently drops** the route — no
```
anamehra: Current `ZmqClient` appears to throw on oversized messages or retry exhaustion, and `fpmsyncd` does not catch those ZMQ exceptions. Should the baseline be updated from "silent drop" to "process exits/restarts without retaining pending work", or is this HLD based on an older code baseline?

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
```diff
@@ -0,0 +1,834 @@
...
+FPM netlink stream (north) and produces route updates to orchagent over ZMQ (south); "producer"
+refers to that ZMQ egress channel, the subject of this document. It specifies:
+
+- **(a)** a non-blocking, coalescing, chunked send path that removes a class of **silent route drops**
+  under burst ([sonic-buildimage #28369](https://github.com/sonic-net/sonic-buildimage/issues/28369));
+- **(b)** an event-driven **heap release** that returns burst-inflated memory to the OS
+  ([sonic-buildimage #28245](https://github.com/sonic-net/sonic-buildimage/issues/28245)); and
+- **(c)** the **STATE_DB telemetry** that makes both observable.
+
+**Out of scope:** the *consumer* side (orchagent `ZmqRouteConsumerStateTable` / `RouteOrch`
```
anamehra: The HLD mentions `ZmqRouteConsumerStateTable`, but current `ZmqRouteOrch` appears to construct generic `ZmqConsumerStateTable`. Is this just background context, or does this design depend on the specialized route consumer being wired in?

anamehra (commented):
Adding a few clarification questions after checking the HLD against current code paths.

venkit-nexthop:
> Hi Venkat, As I see zebra already has same-prefix coalescing. When dataplane processing backs up, zebra’s meta queue stops draining and route changes for the same prefix continue to converge. If fpmsyncd slows its socket reads, pressure can eventually propagate back through zebra's FPM output path and dplane provider queues. Once that reaches zebra's RIB/meta-queue processing, the existing RIB-level coalescing can help unless I am missing something in my reading of code. Also there is no protocol-level back-pressure due to this. zebra continues reading from BGP ie the back-pressure stays between zebra and orchagent's m_toSync… 

I see..

> The reason I raised the back-pressure question is that I see the ZMQ route path as a fast pipe between two existing convergence points: zebra's RIB processing on the producer side and orchagent's m_toSync (which coalesces via addToSync) processing on the consumer side removing the earlier Redis round-trip overhead. fpmsyncd was mostly acting as a bridge: read FPM and forward inline. orchagent's poll thread is a tight loop — recv, push through. The pipe seemed as intended to be fast and transparent. 

Yes, you are right about the ZMQ path just being a fast pipe between fpmsyncd and orchagent.

> The current proposed changes introduces additional processing and buffering within this pipe. 

It is not within the pipe. It is still on the fpmsyncd side. The pipe should ideally be very shallow and transient. Since it is non-coalescing, it should not hold a lot of items at any given time and should drain as quickly as possible. This proposal simply adds a coalescing structure before the pipe.

> I am not raising this as a blocker. Based on my reading of the code, I arrived at a slightly different mental model of where back-pressure would naturally converge. Since I am still building context on the FRR/SONIC routing side, wanted to understand whether there's additional context or design rationale.. One related thing I noticed while looking at the zebra FPM dplane code: if the FPM output path cannot make write progress for 15 seconds, the FPM plugin resets/reconnects which schedules RIB replay...

Your instincts are correct. Normally we would like to apply back-pressure to the RIB (which is implemented inside Zebra), when the dataplane is backed up. But, there is a HLD for FIB design and PRs have started coming in. The fpmsyncd process is going to house the FIB. The FIB is going to maintain forwarding chains and it is meant to help with back walks when certain convergence events happen. This helps better convergence in a prefix independent manner. To aid that, I think it may be better to not apply back pressure all the way to the RIB and implement coalescing inside fpmsyncd.

deepak-singhal0408:
Great discussions on the end to end back-pressure topic.
JFYI, I will be oof next week. So my responses on this thread will be a bit delayed.

Will address the open questions/comments upon my return. 

On Heapreleasemanager/jemalloc topic:
Thought a bit more on this, and I plan to go deeper. Next benchmarking will focus on a few metrics (fpmsyncd route download performance, CPU usage, page faults, preemptive ness (may be moot with jemalloc + bg=true)) during mem release under various incoming load patterns (sustained burst, burst--idle(long)--burst--, burst--idle(short)--burst) + prefix combinations(all unique, set of prefixes with repeated updated) to simulate this more realistically, I plan to use a lightweight kvm setup (t0 topo). the focus will be only at fpmsyncd route delivery.. If a need arises to slow down orchagent to build pressure on fpmsyncd will do so as well (same for both).. Any kvm related noise will be same for both benchmarking. If anything comes up that skews the results, will be transparent about it.

deepak-singhal0408:
> Great discussions on the end to end back-pressure topic. JFYI, I will be oof next week. So my responses on this thread will be a bit delayed.
> 
> Will address the open questions/comments upon my return.
> 
> On Heapreleasemanager/jemalloc topic: Thought a bit more on this, and I plan to go deeper. Next benchmarking will focus on a few metrics (fpmsyncd route download performance, CPU usage, page faults, preemptive ness (may be moot with jemalloc + bg=true)) during mem release under various incoming load patterns (sustained burst, burst--idle(long)--burst--, burst--idle(short)--burst) + prefix combinations(all unique, set of prefixes with repeated updated) to simulate this more realistically, I plan to use a lightweight kvm setup (t0 topo). the focus will be only at fpmsyncd route delivery.. If a need arises to slow down orchagent to build pressure on fpmsyncd will do so as well (same for both).. Any kvm related noise will be same for both benchmarking. If anything comes up that skews the results, will be transparent about it.

@rejithomas-arista, have posted my new benchamrking data here.. https://github.com/sonic-net/sonic-buildimage/issues/28245#issuecomment-5352712074
And my conclusion is actually inverted and leaning towards your suggestion. Please have a look and let me know if any other thoughts before I revise the HLD.

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
deepak-singhal0408: @anamehra, yes, the coalascing + send in a separate thread fix only applies to swss_zmq path. Traditional Redis path stays untouched. I will update the HLD to make this point clear.

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
deepak-singhal0408: @anamehra the effective key is `(table, key)`, matching what Redis already does: `ProducerStateTable` dedups via a per-table `<TABLE>_KEY_SET` (SADD) and overwrites the per-key `_<TABLE>:<key>` hash (HSET), so a key updated twice before the consumer drains is delivered once with the latest value. ZMQ has no server-side dedup, so the coalescer restores that parity rather than adding a new optimization — I'll say this explicitly in the HLD.

The one place we don't match Redis is fairness: all tables share a single FIFO `RedisPipeline` today, so none can starve another, whereas the coalescer drains route to empty before reaching label-route. I'll fix that (single ordered pass covering both) and add test coverage.

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
deepak-singhal0408: @anamehra with `suppress-fib-pending` disabled the reply is informational — it sets the offload flag in zebra's RIB and nothing gates on it — so placement doesn't change behavior. The semantics are also unchanged from today: the reply already precedes the write, so it has never meant "programmed in ASIC" in this mode. Async doesn't change delivery latency either — ingest no longer stalls behind the send and coalescing removes work from the wire, so a burst clears sooner end-to-end; what changes is that the reply and the send are no longer the same synchronous call.

Plan is to move it after the producer hand-off anyway, as tidiness — the reply then corresponds to work fpmsyncd has accepted (in the coalescing map, so R1/R2 apply) rather than just read off netlink. Same thread, no plumbing.

Not doing it after the ZMQ send: that needs a completion callback from the send thread plus per-key ack tracking through LWW coalescing, and re-couples the FPM thread to wire completion.

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
deepak-singhal0408: @anamehra unit tests cover the producer half — routes stay in the map when the send fails, and I'll add one for the kill case: routes resident in the map, coalescer destroyed, nothing dispatched.

They can't cover recovery itself, since the replay would just be the test re-feeding the routes. It also needs real zebra to be meaningful: those routes are ones zebra already considers delivered, so recovery depends on the reconnect doing an unconditional RIB re-drive rather than resending only unacked routes.

So I'll add a sonic-mgmt test that stalls the send path until routes are parked in the map, kills fpmsyncd before they're dispatched, and verifies they land after recovery. Will cover it in the test plan.

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
deepak-singhal0408: @anamehra it throws, but `Select::poll_descriptors` in swss-common catches it before it ever reaches fpmsyncd — it wraps the `readData()` call in `catch (const std::runtime_error&)`, and `std::system_error` derives from `std::runtime_error`. It logs the error and returns `Select::ERROR`, which fpmsyncd's main loop doesn't check, so the loop just continues. The connection also stays up, since `io_error` isn't `FpmConnectionClosedException`, so zebra never re-drives the RIB. No crash, no replay — the update is simply lost.

Confirmed on a T2 run (build 1160993, 2026-07-13): the throw is immediately followed by `poll_descriptors: readData error ... Input/output error` on both bgp0 and bgp1, with no crash or restart markers and no core. fpmsyncd is also a critical process, so a real abort would have bounced the bgp container and flapped BGP — that didn't happen.

doc/fpmsyncd/fpmsyncd_zmq_route_delivery.md
deepak-singhal0408: Background context — no dependency. The producer serializes the same `KeyOpFieldsValuesTuple` batches to the same `(db, table)` endpoint regardless of which consumer class is on the other end, so the wire contract is unchanged.

You were right that the wiring wasn't there when you reviewed. It has since landed — [sonic-swss-common #1187](https://github.com/sonic-net/sonic-swss-common/pull/1187) added `ZmqRouteServer` / `ZmqRouteConsumerStateTable` for burst-coalesced ingress, and [sonic-swss #4564](https://github.com/sonic-net/sonic-swss/pull/4564) makes `RouteOrch` a `ZmqRouteOrch` that constructs it. I'll update the reference to point at the merged change.

PR #2489, opened 2026-08-07
CLI cpo-util HLD

Junchao-Mellanox:


doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+Compared with the `sfputil firmware` command group, the scope of subcommands is defined as follows:
+
+| Operation | Description | In scope |
+| --- | --- | --- |
+| download | Download firmware to the device bank. | Yes |
+| run | Temporarily switch to the new firmware. | Yes |
+| commit | Set the new firmware as the default. | Yes |
+| upgrade | Perform download, run, and commit. | Yes |
+| unlock | Enter module password. | No |
+| target | Select target firmware for `Y-cable`. | No (not relevant) |
```
prgeor, line 55: @Junchao-Mellanox Need an option to be able to select the target so that all subsequent CDB commands is directed to the targeted end point (Eg, MCU, OE, ELSP) on that port#

doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+- `ElsfpApi` should be updated to inherit from, or mix in, `CmisCdbFw` so that it can support CDB-based firmware operations. (Or, should each vendor provide a `VendorCdbFw`?)
+
+> Note: currently, ElsfpApi is not inherit from CmisApi, each vendor has to implement `get_fw_ops_match_attrs` by their own.
+
+
+#### 7.7 Error handling
+
+Firmware execution is delegated only after `cpoutil` completes parsing, matching, and optional checksum verification. Failures are isolated per component; a failed ELS does not roll back a successful OE. The CLI reports status per internal component identity. Fatal conditions yield a non-zero exit; per-target warnings do not mask fatal errors.
+
+**Frontend (CLI):**
```
prgeor, line 679: @Junchao-Mellanox Please add the CLI output. User should be able to see the progress of the operation (download) and if there is a error from CDB transaction, the CLI should capture the Error code mentioned in the CDB/CMIS spec

doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+# Activate staged ELS firmware
+cpoutil firmware run Ethernet0 --component els
+
+# Commit staged OE firmware
+cpoutil firmware commit Ethernet0 --component oe
+```
+
+Corresponding CLI documentation should be added to https://github.com/sonic-net/sonic-utilities/blob/master/doc/Command-Reference.md when the feature is implemented.
+
+> Note: Only one CLI instance may run at a time. This constraint should be enforced with a file lock.
```
prgeor: @Junchao-Mellanox for simplicity we can leave this to the user where multiple ports can share resources MCU/OE/ELSP. This is same as we do for 8x100G on pluggable modules.

doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+
+##### 7.6.3 ELS API
+
+- Add a new function `get_fw_ops_match_attrs` to get all the FW operation matching attributes on demand.
+- `ElsfpApi` should be updated to inherit from, or mix in, `CmisCdbFw` so that it can support CDB-based firmware operations. (Or, should each vendor provide a `VendorCdbFw`?)
+
+> Note: currently, ElsfpApi is not inherit from CmisApi, each vendor has to implement `get_fw_ops_match_attrs` by their own.
+
+
+#### 7.7 Error handling
```
prgeor, line 675: @Junchao-Mellanox @eddyk-nvidia I would wish if this HLD can be split into two phases

1. Basic CPO component firmware upgrade via CLI , download, run, commit using port#, firmware binary file
2. OIF based package which is built on top of #1 

This gives the user flexibility to build their own services around #1 (who don't want to pursue OIF package based approach). Its also easier to debug issues using #1 and much simpler to implement and on par with pluggable firmware download procedure.

doc/cpo-util/cpo-util-hld.md
Junchao-Mellanox: We can choose it from `--component` option, which supports: oe | els | all

doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+
+- a single YAML mapping describes Firmware Load, or
+- a YAML list of mappings, where each list element describes one Firmware Load
+
+Each YAML mapping that describes a Firmware Load includes the following fields:
+
+- VendorName: mandatory, exact match
+- *ComponentClass: mandatory, either ExternalLaserSource or OpticalEngine. **This field is a `cpoutil` extension to the OIF package which allows `cpoutil` to select the correct CPO component as the upgrade target**.
+- FwLoadName: mandatory, unique, identifies the filename of the Firmware Load
+- FwLoadVersion: optional, identifies the version of the Firmware Load. An informational attribute SHALL NOT be used as a firmware matching attribute and SHALL NOT be interpreted as part of CMIS-reported active or inactive firmware version information. When multiple Firmware Metadata entries match, a host may use FwLoadVersion as a tie-break policy (e.g., selecting the highest FwLoadVersion) or may prompt the operator to select among candidates.
```
bgallagher-nexthop: Is `FwLoadVersion` solely just for tie-breaking? To support packaging multiple versions of firmware with the same `FwLoadName` in a single tarball?

Also, what is the expected format of this: a string, an integer?

doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+##### 7.4.3 Stage — Matching
+
+For each `discover_entry`, select zero or one `parsing_entry` and generate a `plan_entry`.
+
+![matching-flow](./matching-flow.svg)
+
+Tie-break policy (Phase 1):
+
+1. highest `FwLoadVersion` wins;
+2. user choose;
```
bgallagher-nexthop: A third option to error out on multiple matches instead of prompting the user may be worth considering, since folks may build automation that uses the `cpoutil` tool that won't be easily able to answer prompts.

doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+
+Matching stage outputs a list of `plan_entries`:
+
+```
+plan_entry {
+  cpo_module_index
```
bgallagher-nexthop: What is the difference between `cpo_module_index` and `component_index`? A comment beside these fields may be useful to disambiguate

doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+
+Tie-break policy (Phase 1):
+
+1. highest `FwLoadVersion` wins;
+2. user choose;
+
+Matching stage outputs a list of `plan_entries`:
+
+```
+plan_entry {
```
bgallagher-nexthop, line 452: Would it make sense to rename this `match_entry` and have the `match_entries` list only contain entries for matches? I am not sure of the use-case of a `plan_entry` that has `matching_decision == None`.

doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+    Returns:
+        Runtime match attributes dictionary
+    """
+    pass
+```
+
+##### 7.6.3 ELS API
+
+- Add a new function `get_fw_ops_match_attrs` to get all the FW operation matching attributes on demand.
+- `ElsfpApi` should be updated to inherit from, or mix in, `CmisCdbFw` so that it can support CDB-based firmware operations. (Or, should each vendor provide a `VendorCdbFw`?)
```
bgallagher-nexthop: I think the former (that `ElsfpApi` should mix in `CmisCdbFw`, the same as `CmisApi`) is the better option

doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+
+> Note: currently, ElsfpApi is not inherit from CmisApi, each vendor has to implement `get_fw_ops_match_attrs` by their own.
```
bgallagher-nexthop, line 672: If `get_fw_ops_match_attrs` was moved into the CmisCdbFw mixin, would this be solved? `CmisApi` and `ElsfpApi` would then share the same definition via the mixin.

doc/cpo-util/cpo-util-hld.md
```diff
@@ -0,0 +1,742 @@
...
+
+##### 7.6.1 CPOBase
+
+Add a few new APIs and default implementations to `CpoBase`.
+
+```python
+from dataclasses import dataclass
+
+@dataclass(slots=True)
+class CpoFwOpsEntry:
```
bgallagher-nexthop, line 595: We could keep these `cpoutil` related data structures out of the platform API by instead adding the download, run, commit and upgrade methods on [`CpoDeviceBase`](https://github.com/sonic-net/sonic-platform-common/blob/bf10bc9b8e2e8e47445ed7503c9e7df882c6c42d/sonic_platform_base/sonic_xcvr/cpo/cpo_base.py#L40) instead.

Then both the `OeBase` and `ElsfpBase` classes will expose the firmware methods, but we no longer need the `CpoFwOpsEntry` to communicate the component information. Instead, `cpoutil` will just access the device it cares about based on the matching firmware operation entry it is about to perform an upgrade for:
```python3
# in cpoutil
cpo = get_cpo_object(fw_op_entry)
if fw_op_entry.component_class == "OpticalEngine":
    cpo.oe.upgrade_firmware(file_path=fw_op_entry.file_path)
elif fw_op_entry.component_class == "ExternalLaserSource":
    cpo.elsfp.upgrade_firmware(file_path=fw_op_entry.file_path)
```

doc/cpo-util/cpo-util-hld.md
Junchao-Mellanox: Will add it later.

doc/cpo-util/cpo-util-hld.md
Junchao-Mellanox: it is a string like other version, e.g: "1.2.3". Will make it clear here.

doc/cpo-util/cpo-util-hld.md
Junchao-Mellanox: good idea, will update

doc/cpo-util/cpo-util-hld.md
Junchao-Mellanox: Yes, will update with some comment.
- `cpo_module_index `: index of the CPO object
- `component_index`: index of OE or ELS

doc/cpo-util/cpo-util-hld.md
Junchao-Mellanox: `matching_decision = None` is used for logging purposes. For example, if a user wants to upgrade `Ethernet0` and `Ethernet120`, but only `Ethernet0` has a matching entry, we need to log that `Ethernet120` has no match. Maybe we don't really need this entry, we will see in the implementation and update acoordingly.

doc/cpo-util/cpo-util-hld.md
Junchao-Mellanox: Agree, I will remove the inherit option .

doc/cpo-util/cpo-util-hld.md
Junchao-Mellanox: Moving it to CmisCdbFw is an option. Another option is to put it to XcvrApi. I will compare the 2 options and update.

doc/cpo-util/cpo-util-hld.md
Junchao-Mellanox: I prefer the current way because: implementing it in CpoBase will give the max flexibility for vendor to support different hardware topology. For example, it support both joint mode and separate mode well. Implementing it in `OeBase` and `ElsfpBase` will limit it in separate mode.

bgallagher-nexthop: I am not sure I understand, both approaches should support arbitrary hardware topologies (separate/joint). There will always be OE and ELSFP objects present on the `CpoBase` object, regardless of whether the hardware operates in joint or separate mode.

The only change in my suggestion is that we don't leak `cpoutil`'s abstraction into the platform API by removing the need to add `CpoFwOpsEntry` to the platform API entirely. Instead, we just add the method definitions and let `cpoutil` do the decision making of selecting the appropriate component to operate on instead of the platform API doing that. The main motivations being:
- `cpoutil`'s data model (`CpoFwOpsEntry`) ideally should not be added to the platform API as it tightly couples `cpoutil` and the platform API
- By keeping `CpoFwOpsEntry` in `cpoutil` it makes any changes to the data model easier since you don't have to do cross-repo changes. Imagine a new field had to be added to `CpoFwOpsEntry`, then you might need two PRs across two repos -- one to add the field, one to plumb it through in `cpoutil`.
- Defining methods on `CpoBase` can be useful for backwards compatibility, so you can treat the OE/ELSFP as a single aggregated device like some form of pluggable transceiver. Since `cpoutil` is new code, this doesn't apply and we are in the convenient position to just reason about the OE and ELSFP directly.

Junchao-Mellanox: - Yes, OE and ELSFP objects will always be present on the CpoBase object. However, that does not mean each OE or ELSFP has its own FW bank. For example, a single Cpo object may contain two ELSFPs that share the same FW bank. Your approach does not support this deduplication, and it would also be difficult for the vendor platform API to override this behavior.
- Regarding `CpoFwOpsEntry`, I do not fully understand your point. If the interface between the CLI and `sonic-platform-common` needs to change, both repositories will need to be updated. To decouple them, I think a possible approach would be to define `dict`-based parameters. Would you prefer that approach?

Junchao-Mellanox: BTW, I am fine to replace `CpoFwOpsEntry ` with something else.

bgallagher-nexthop: `CpoBase` does not support multiple ELSFP or OEs per port as-is today so I think sharing of resources like CDB instances and firmware banks is something that would require changes in general to support since it is a new concept. I am not sure how firmware bank sharing would work on a single `CpoBase` object for instance, the same problems would have to be solved.

Current approach is fine to me, no complaints if you'd prefer to keep this approach 👍 Just wanted to highlight that there is an alternative that does not require adding an extra data structure to store the decision for which component should be upgraded: client code like `cpoutil` can make that decision pretty easily and the firmware methods could just take the file path as a sole arg if they were defined on each device object instead of the overall cpo object.

PR #2491, opened 2026-08-08
Add HLD for Top N Interfaces feature

Rishik-Y:


Rishik-Y:
https://github.com/sonic-net/sonic-utilities/pull/4416

Issue #138, opened 2017-12-23
Kindly please tell me the purpose of process dependent


lincy991389662:
Hi sir,
        I sincerely hope that you could support me , as I seek help many time on the sonicproject, but not response.
        My question is from source code and the running result , I find that when orchagent was exited, will not work again, not matter how restart it again?
       What's more,  from the code I know when syncd was exited will notify the orchagent exit. To my view, orchagent as a  separate component  need not to rely on the syncd. What's  the benefit of such a design？

stcheng:
we are working on separating `orchagent` with `syncd` in the future
the current design enforces that these two components need to work at the same time
one could use the command `systemctl restart swss` to restart `orchagent` and `syncd` together at the same time

Issue #137, opened 2017-12-21
How to configure ECN and PFC ?


snowzjx:
Hi,

Is there any document regarding ECN and PFC configuration ?

Thanks.

Issue #136, opened 2017-12-21
How to ping between two hosts connected to same SONiC switch.


JdotX:
Hello,

We are testing SONiC on a Mellanox SN-2100 switch. To begin with we connected two end hosts to this switch. By common sense we should be able to PING from one end host to another, but in reality what we can to is only to PING from the end host to the corresponding port on the switch it is connected to, when we try to PING the other end host however, we keep getting "Destination Host Unreachable" message. But we can get the ARP messages on the end hosts using tcpdump

***

We tried to create a VLAN and add the two interfaces to this VLAN, and 
```
brctl show
```
indicates that the two ports are indeed added to this VLAN, we also have them removed the interfaces from the other parts of configuration like "INTERFACES". But we are still unable to PING. Is there any idea what should be the following steps?

Many thanks to your help!

Jiacheng

PR #2495, opened 2026-08-12
[doc] MACsec: HLD for fallback CAK support in wpa_supplicant

liamkearney-msft:
Add a high level design for running a primary and a fallback CAK simultaneously on a MACsec port, and for failing the controlled port over between them without dropping traffic.

This delivers the Phase III functional requirement of the MACsec HLD and fills in section 3.4.1.1, Primary/Fallback decision, which was left as a TODO.

The document is scoped to the wpa_supplicant control plane. It covers hoisting the per-port SecY state out of the MKA participant, the principal CA selection rules that keep both ends of a link on the same CKN, hitless promotion via SAK migration and a deferred post-promotion rekey, and the SAK rollover defects that running two CAs on one port exposed. The CONFIG_DB MACSEC_PROFILE table already carries the optional fallback_cak and fallback_ckn fields, so no schema, MACsecMgr, plugin, orchagent or SAI change is implied.

Sonic side changes covered by this HLD : https://github.com/sonic-net/SONiC/pull/2545 

This change is done over 4 sequential wpasupplicant PRs: 
https://github.com/sonic-net/sonic-wpa-supplicant/pull/135
https://github.com/sonic-net/sonic-wpa-supplicant/pull/136
https://github.com/sonic-net/sonic-wpa-supplicant/pull/137
https://github.com/sonic-net/sonic-wpa-supplicant/pull/138

### DIFF
--- a/doc/macsec/MACsec_fallback_cak_hld.md
+++ b/doc/macsec/MACsec_fallback_cak_hld.md
@@ -76,21 +76,28 @@ the MKA control plane and reuses the existing plugin API surface.
 ## 1 Requirements
 
 1. A port may be configured with a **primary** and an optional **fallback** CAK,
-   both active simultaneously.
+   both active simultaneously. The two CKNs must differ — the CKN is what
+   identifies a CA in an MKPDU — though the CAKs may be the same.
 2. If the primary CA fails — typically a key mismatch after a one-sided key
    rotation, or the peer losing the primary CAK — the port **must fail over to
    the fallback CA without dropping traffic**.
-3. Failover must be **revertive**: once the primary CA recovers a live peer, the
-   port returns to it.
-4. Both ends must converge on the **same** CA. A configuration where each end
-   independently prefers a different CKN must not oscillate.
+3. The primary CA has **priority** over the fallback: whenever the primary has a
+   live peer and we are key server on it, that is where SAKs are distributed. An
+   in-flight rekey settles first (§3.4), so recovery never cuts one short.
+4. Both ends must end up on the **same** CA. When both CAs have live peers the
+   key server distributes on the primary and the other end follows the CKN its
+   SAKs arrive on, so the two ends cannot oscillate.
 5. The feature must be **opt-in**. With only a primary CAK configured, behaviour
    is unchanged.
 6. A CAK must be replaceable **at runtime**, without restarting the supplicant
-   or bouncing the link.
+   or bouncing the link. This is hitless only with a valid fallback to carry the
+   port across the rotation (§4.2.1).
 7. A port carries **exactly one primary CA**, enforced by the supplicant. Adding
    a second primary is rejected rather than silently changing which CA owns the
    port.
+
+A fallback-only configuration works in the supplicant, but SONiC does not expose
+it — MACsecMgr always programs a primary.
 
 ## 2 Background
 
@@ -158,33 +165,60 @@ everything that follows.
 
 ### 3.2 Principal CA selection
 
-`ieee802_1x_kay_decide_principal()` is evaluated whenever peer liveness or key
-server election changes. It answers one question: *which CA owns the Controlled
-Port?*
+`ieee802_1x_kay_select_principal()` answers one question: *which CA owns the
+Controlled Port?* It makes the whole decision and returns the CA that should own
+the port; `ieee802_1x_kay_reconcile_principal()` applies the result — migrating
+the installed-SAK bookkeeping (§3.3), moving the principal and arming the
+deferred rekey (§3.4), or tearing the port down if no CA is returned.
+
+It is re-evaluated whenever **peer liveness** changes — a CA gaining its first
+live peer, or losing its last — or when key server election changes. A peer is
+live once both ends have exchanged MKPDUs under the same CAK.
+
+Selection is four ordered rules, first match wins:
+
+1. **We are key server on a CA with live peers** — that CA, primary preferred.
+2. **We are not key server** — the CA we are receiving SAKs on, i.e. the
+   incumbent, while it still has live peers.
+3. **Otherwise** — any active CA with live peers, primary preferred.
+4. **Otherwise** — whatever is currently set, while it is still active.
 
 ```mermaid
 flowchart TD
-  A["Peer liveness / election changed"] --> B{"Are we the elected <br/>key server on a live CA?"}
-  B -- yes --> C["Choose that CA. <br/>Prefer the primary <br/>over the fallback."]
+  A["Peer liveness / election changed"] --> B{"Are we the elected key <br/>server on a CA with <br/>live peers?"}
+  B -- yes --> C["1 — prefer the primary <br/>over the fallback"]
   B -- no --> D{"Does the current <br/>principal still have <br/>a live peer?"}
-  D -- yes --> E["Keep it — the remote <br/>key server chose this CKN"]
-  D -- no --> F{"Port secured and a <br/>sibling CA is live?"}
-  F -- yes --> G["Fail over to the sibling"]
-  F -- no --> H["No CA can own the port: <br/>tear down the controlled port"]
-  C --> I["reconcile_principal()"]
-  E --> I
-  G --> I
+  D -- yes --> E["2 — keep it: the remote <br/>key server chose this CKN"]
+  D -- no --> F{"Any other active CA <br/>with live peers?"}
+  F -- yes --> G["3 — fail over to it. <br/>Prefer the primary."]
+  F -- no --> H{"Is the current principal <br/>still an active CA?"}
+  H -- yes --> I["4 — keep it as <br/>owner, peerless"]
+  H -- no --> J["No owner can be <br/>named"]
+  C --> K["reconcile_principal() <br/>applies the result"]
+  E --> K
+  G --> K
+  I --> L["No live peer anywhere: <br/>tear down the data path"]
+  J --> L
 ```
 
-Two properties fall out of this ordering and are what keep the two ends in
-step:
+Selection is not key server election: election happens inside each CA and is an
+input here. As key server we choose, and we prefer the primary; as a non key
+server we choose nothing — rule 2 keeps whichever CA the remote key server is
+distributing on. Only one end decides, so the two cannot ping-pong between CKNs.
+Election is re-run for the incoming CA on promotion, but key server priority and
+SCI are port properties, so every CA reaches the same verdict.
 
-- **Primary preference applies only to the key server.** A follower keeps the CA
-  the remote key server is actually distributing on. If both ends independently
-  preferred their own primary, they would ping-pong between CKNs; deferring to
-  the key server makes the choice unilateral and therefore stable.
-- **Selection is revertive.** As soon as the primary CA regains a live peer and
-  we are key server on it, ownership returns to the primary.
+"Active" is close to "configured": the flag is set once a CA sends or processes
+an MKPDU and cleared only when the CA is deleted or deactivated. It is not a
+liveness signal — a CA that has lost every peer is still active.
+
+Rules 3 and 4 are a floor: an owner is named whenever any CA remains, so a later
+SAK lookup or re-homing always resolves. Owning the port is not the same as
+carrying traffic, though — rules 1 to 3 all require a live peer, so a rule 4
+owner has none. When no CA on the port has a live peer the data path is torn
+down as upstream does: the SAs are deleted and the controlled port is blocked,
+while the owner pointer is retained. The returning peer re-elects and the port
+re-secures on a freshly distributed SAK.
 
 Two safety rules complete the picture:
 
@@ -192,13 +226,21 @@ Two safety rules complete the picture:
   is allowed to touch the CP, so a SAK arriving on the fallback CA cannot
   install key material for the port it does not own.
 - Only the **principal** key server distributes SAKs. A rekey requested on a
-  follower is retained (and honoured on promotion) rather than silently dropped.
+  non key server is retained (and honoured on promotion) rather than silently
+  dropped.
 
 ### 3.3 Hitless failover
 
-The promotion itself must not disturb the datapath. The incoming CA **inherits
-the installed SAK** rather than negotiating a new one, so no SA is deleted or
-created at the moment of failover — only MKA bookkeeping moves.
+The promotion itself must not disturb the datapath. Control plane and data plane
+are decoupled, so a change of principal does not require the installed SAK to be
+torn down: the SA stays programmed and keeps carrying traffic.
+
+What moves is **bookkeeping** — the KaY's record of the installed key is re-homed
+from the outgoing participant to the incoming one, so the new principal can
+account for it and retire it later. No SA is created or deleted, and no key
+material crosses between CAs. The new principal does distribute a fresh SAK under
+its own CKN, just not at the instant of promotion: that rekey is deferred until
+MKA has settled (§3.4), and the inherited SA is then rotated out normally.
 
 ```mermaid
 sequenceDiagram
@@ -212,14 +254,14 @@ sequenceDiagram
   Note over CAP,CAF: both CAs exchange MKPDUs independently <br/>only the primary owns the port
 
   CAP->>KAY: last live peer times out
-  KAY->>KAY: decide_principal() <br/>primary has no live peer, <br/>fallback is live → fallback wins
+  KAY->>KAY: select_principal() <br/>primary has no live peer, <br/>fallback is live → fallback wins
   KAY->>KAY: migrate_principal_sas() <br/>re-home installed-SAK bookkeeping
-  KAY->>CAF: set principal, elect key server
+  KAY->>CAF: set principal, re-run key server election <br/>(same verdict)
   Note over SEC: no SA delete / create — <br/>traffic keeps flowing on the inherited SAK
   KAY->>KAY: arm deferred rekey (≈3 × hello time)
 
   rect rgb(235,245,255)
-  Note over CAP,CAF: settle window — both ends converge on the new principal
+  Note over CAP,CAF: settle window — MKA hellos converge <br/>on the new principal
   end
 
   KAY->>CAF: deferred rekey fires
@@ -233,6 +275,10 @@ Rekeying *immediately* on promotion is what breaks the datapath: while the two
 ends are still converging on the new principal, the rotation races traffic and
 frames are lost. The rekey is therefore deferred by **≈3 hello times** and the
 inherited SAK carries traffic until then.
+
+The **settle window** is that deferral — the grace period between a change of
+principal and the rekey that follows it, long enough (≈3 hello times, ≈6 s at
+the default 2 s hello) for MKA hellos to converge on both ends.
 
 Arming is **non-resetting**, so a burst of ownership swaps collapses into a
 single rekey instead of each swap pushing the timer out. A `principal_generation`
@@ -284,7 +330,8 @@ sequenceDiagram
 `retire_when` is demoted to a failsafe for a peer that stays live but never
 confirms. On the key server it is lengthened to **20 s** so it can never cut off
 a slow-but-progressing peer mid-rotation — precisely the loss this change
-exists to prevent. A follower cannot observe the gate and keeps the stock timer.
+exists to prevent. A non key server cannot observe the gate and keeps the stock
+timer.
 
 #### 3.5.2 Coalesce the deferred CP step
 
@@ -357,6 +404,12 @@ When both are set, a second, standby MKA participant is created on the same
 interface at association time. Both participants exchange MKPDUs independently,
 but only the principal owns the Controlled Port and distributes SAKs.
 
+Both CAs advertise MKPDUs **continuously** while configured; the fallback is not
+held in reserve. An MKPDU carrying a CKN the peer does not hold fails ICV
+validation there and is discarded, so a mismatched CA simply never gains a live
+peer. Removing a CAK with `macsec_del_mka` stops its MKPDUs, and a SAK is only
+ever distributed on a CA that has a live peer.
+
 These map one-to-one onto the `fallback_cak` / `fallback_ckn` fields already
 defined in the CONFIG\_DB `MACSEC_PROFILE` table, so no schema change is needed.
 
@@ -371,6 +424,9 @@ the supplicant or bouncing the link.
 | `macsec_del_mka ckn=<hex>` | `MACSEC_DEL_MKA` | Remove a participant by CKN. |
 | `macsec_mka_list` | `MACSEC_MKA_LIST` | List participants with their role and peer counts. |
 | `macsec_rekey` | `MACSEC_REKEY` | Force the key server to distribute a fresh SAK. |
+
+`FAIL` is ordinary `wpa_cli` control interface semantics — the command is
+rejected and nothing changes. It is not a MACsec or MKA state.
 
 `macsec_mka_list` reports, per participant:
 
@@ -393,13 +449,27 @@ port in between:
 
 1. `macsec_del_mka` the old primary. The KaY re-selects, the fallback CA is
    promoted and inherits the installed SAK (§3.3), so the port keeps forwarding.
-2. `macsec_add_mka` the new primary. Once it has a live peer, revertive
-   selection returns the port to it.
+2. `macsec_add_mka` the new primary. Once it has a live peer, rule 1 applies —
+   the primary outranks the fallback — and selection returns the port to it.
 
 Each step costs one settle window of a few hello times, but no packet loss, and
 the port is never left without a usable CA. Note the peer does **not** have to
 have installed the new CKN before the old one is removed — the fallback covers
 the gap, which is what makes an uncoordinated, one-end-at-a-time rotation safe.
+
+These are supplicant-level primitives, not an operator workflow. MACsecMgr
+issues the sequence; an operator uses the ordinary SONiC config commands and
+never touches the control interface:
+
+| Intent | Operator command | What MACsecMgr issues |
+| ------ | ---------------- | --------------------- |
+| Rotate the primary CAK | `config macsec profile update …` | `macsec_del_mka` old primary, then `macsec_add_mka` new primary |
+| Tear the session down | `config macsec port del …` | removes the whole network block, and both CAs with it |
+
+Because the two intents map to different command streams, the supplicant never
+has to infer which was meant: a bare `macsec_del_mka` is *never* a teardown
+request. It removes one CA and the KaY keeps the port on whatever remains. The
+Controlled Port is torn down only when the port's last CA goes away.
 
 ```mermaid
 sequenceDiagram
@@ -416,7 +486,7 @@ sequenceDiagram
 
   OP->>KAY: macsec_del_mka ckn=<old>
   KAY->>CA1: destroy participant
-  KAY->>KAY: decide_principal() <br/>only the fallback is live
+  KAY->>KAY: select_principal() <br/>only the fallback is live
   KAY->>CAF: promote, migrate installed SAK
   CAF->>CP: principal (hitless, same SAK)
   Note over CAF,CP: deferred rekey after the settle window (§3.4)
@@ -424,8 +494,8 @@ sequenceDiagram
   OP->>KAY: macsec_add_mka ckn=<new> cak=<new>
   KAY->>CA2: create participant (primary)
   Note over CA2: MKA converges, peer becomes live
-  KAY->>KAY: decide_principal() <br/>primary is eligible again
-  KAY->>CA2: promote (revertive), migrate SAK
+  KAY->>KAY: select_principal() <br/>primary is eligible again
+  KAY->>CA2: promote (primary outranks fallback), migrate SAK
   CA2->>CP: principal (hitless)
   Note over CA2,CP: deferred rekey under the new CKN (§3.4)
 ```
--- a/doc/macsec/MACsec_fallback_cak_hld.md
+++ b/doc/macsec/MACsec_fallback_cak_hld.md
@@ -1,0 +1,499 @@
+<!-- omit in toc -->
+# MACsec Fallback CAK — wpa_supplicant High Level Design
+
+***Revision***
+
+|  Rev  | Date | Author       | Change Description |
+| :---: | :--: | :----------- | ------------------ |
+|  0.1  |      | Liam Kearney | Initial version    |
+
+<!-- omit in toc -->
+## Table of Contents
+
+- [About this Manual](#about-this-manual)
+- [Scope](#scope)
+- [Abbreviations](#abbreviations)
+- [1 Requirements](#1-requirements)
+- [2 Background](#2-background)
+  - [2.1 The single-CA assumption](#21-the-single-ca-assumption)
+- [3 Design](#3-design)
+  - [3.1 Per-port SecY state moves to the KaY](#31-per-port-secy-state-moves-to-the-kay)
+  - [3.2 Principal CA selection](#32-principal-ca-selection)
+  - [3.3 Hitless failover](#33-hitless-failover)
+  - [3.4 Deferred post-promotion rekey](#34-deferred-post-promotion-rekey)
+  - [3.5 SAK rollover hardening](#35-sak-rollover-hardening)
+- [4 Configuration](#4-configuration)
+  - [4.1 Network block parameters](#41-network-block-parameters)
+  - [4.2 Control interface](#42-control-interface)
+- [5 Interaction with MACsecMgr](#5-interaction-with-macsecmgr)
+- [6 Backward compatibility](#6-backward-compatibility)
+- [7 Test plan](#7-test-plan)
+- [8 Patch series](#8-patch-series)
+
+## About this Manual
+
+This document describes the `wpa_supplicant` changes required to support a
+**fallback Connectivity Association Key (CAK)** on a SONiC MACsec port, together
+with the MKA robustness fixes that make key rollover hitless in the presence of
+more than one Connectivity Association (CA).
+
+It is a companion to the existing
+[MACsec SONiC HLD](https://github.com/sonic-net/SONiC/blob/master/doc/macsec/MACsec_hld.md)
+and fills in the section marked *TODO* at
+[3.4.1.1 Primary/Fallback decision](https://github.com/sonic-net/SONiC/blob/master/doc/macsec/MACsec_hld.md#3411-primaryfallback-decision).
+It delivers **Phase III** of that document's functional requirements:
+
+> *Primary and Fallback secure Connectivity Association Key can be supported
+> simultaneously.*
+
+## Scope
+
+**In scope** — changes inside `wpa_supplicant` / the MKA `KaY` implementation
+(`src/pae/`, `wpa_supplicant/`).
+
+**Out of scope** — CONFIG\_DB schema (the `MACSEC_PROFILE` table already carries
+optional `fallback_cak` / `fallback_ckn`), MACsecMgr, the SONiC MACsec plugin,
+MACsecOrch, and SAI. None of these need to change; the feature is confined to
+the MKA control plane and reuses the existing plugin API surface.
+
+## Abbreviations
+
+| Abbreviation | Description                                              |
+| ------------ | -------------------------------------------------------- |
+| CA           | Secure Connectivity Association                          |
+| CAK / CKN    | Connectivity Association Key / CA Key Name               |
+| CP           | Controlled Port state machine (IEEE 802.1X-2010, Cl. 12) |
+| KaY          | MACsec Key Agreement Entity                              |
+| KI / KN      | Key Identifier / Key Number                              |
+| MKA          | MACsec Key Agreement protocol                            |
+| MKPDU        | MKA Protocol Data Unit                                   |
+| MI / MN      | Member Identifier / Message Number                       |
+| PN / LPN     | Packet Number / Lowest Packet Number                     |
+| SAK          | Secure Association Key                                   |
+| SC / SA      | Secure Channel / Secure Association                      |
+| SecY         | MACsec Security Entity                                   |
+
+## 1 Requirements
+
+1. A port may be configured with a **primary** and an optional **fallback** CAK,
+   both active simultaneously.
+2. If the primary CA fails — typically a key mismatch after a one-sided key
+   rotation, or the peer losing the primary CAK — the port **must fail over to
+   the fallback CA without dropping traffic**.
+3. Failover must be **revertive**: once the primary CA recovers a live peer, the
+   port returns to it.
+4. Both ends must converge on the **same** CA. A configuration where each end
+   independently prefers a different CKN must not oscillate.
+5. The feature must be **opt-in**. With only a primary CAK configured, behaviour
+   is unchanged.
+6. A CAK must be replaceable **at runtime**, without restarting the supplicant
+   or bouncing the link.
+7. A port carries **exactly one primary CA**, enforced by the supplicant. Adding
+   a second primary is rejected rather than silently changing which CA owns the
+   port.
+
+## 2 Background
+
+### 2.1 The single-CA assumption
+
+MKA permits a port to hold several CAs at once. Upstream `wpa_supplicant`
+models this — `struct ieee802_1x_kay` owns a list of participants, one per CKN —
+but only ever runs one in practice. The **first** participant created claims the
+Controlled Port and the SecY, and nothing re-evaluates that choice.
+
+The reason is that per-port state was stored **per participant**:
+
+```mermaid
+graph LR
+  subgraph before ["Before — SecY state duplicated per CA"]
+    K1["struct ieee802_1x_kay"]
+    P1["participant (CKN-A) <br/>txsc <br/>rxsc_list <br/>lki / lan / ltx / lrx <br/>oki / oan / otx / orx <br/>principal = true"]
+    P2["participant (CKN-B) <br/>txsc (second copy!) <br/>rxsc_list (second copy!) <br/>lki / lan / ... <br/>principal = false"]
+    K1 --> P1
+    K1 --> P2
+    P1 -. "both drive" .-> S1(["SecY / SAI"])
+    P2 -. "the same" .-> S1
+  end
+```
+
+```mermaid
+graph LR
+  subgraph after ["After — SecY state owned by the port"]
+    K2["struct ieee802_1x_kay <br/>txsc <br/>rxsc_list (refcounted) <br/>lki / lan / ltx / lrx <br/>oki / oan / otx / orx <br/>principal_participant"]
+    P3["participant (CKN-A) <br/>peers, MI/MN, key server"]
+    P4["participant (CKN-B) <br/>peers, MI/MN, key server"]
+    K2 --> P3
+    K2 --> P4
+    K2 --> S2(["SecY / SAI"])
+  end
+```
+
+Consequences of the old layout:
+
+- Two participants each believed they owned the transmit SC, so the second one
+  to act would delete or overwrite SAs installed by the first.
+- Receive SCs for the same peer SCI were created and destroyed independently.
+- There was no notion of *which* CA should own the port, so a dead primary CAK
+  took the link down even with a healthy fallback configured.
+
+## 3 Design
+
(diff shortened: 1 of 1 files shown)


doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+| MKPDU        | MKA Protocol Data Unit                                   |
+| MI / MN      | Member Identifier / Message Number                       |
+| PN / LPN     | Packet Number / Lowest Packet Number                     |
+| SAK          | Secure Association Key                                   |
+| SC / SA      | Secure Channel / Secure Association                      |
+| SecY         | MACsec Security Entity                                   |
+
+## 1 Requirements
+
+1. A port may be configured with a **primary** and an optional **fallback** CAK,
```
m26singhvi, line 78: So only fallback CAK/CKN configuration isn't expected to work ?

Also, can a fallback CKN be same as primary CKN ? I guess not is there way to catch that error ? This restriction need not apply to CAK though.

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+   both active simultaneously.
```
m26singhvi: So MKPDUs for fallback key will be advertised continuously if configured, right ? Can we please clear about that.

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+2. If the primary CA fails — typically a key mismatch after a one-sided key
+   rotation, or the peer losing the primary CAK — the port **must fail over to
```
m26singhvi, line 82: Just for my understanding - How does the CAK/CKN rotation work in Sonic/wpa_supplicant, do we stop MKPDUs immediately old CAK/CKN and use new CAK/CKN immediatley ?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+   the fallback CA without dropping traffic**.
+3. Failover must be **revertive**: once the primary CA recovers a live peer, the
```
m26singhvi: Instead of saying revertive - should we instead say that primary has a higher priority over fallback. If and when a higher priority CA is available, key server will choose that to generate SAKs ?

There is a case here where SAK using fallback CAK/CKN is still not completely programmed, in such a case, port should wait for that rekey to complete and then switch back to primary.

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+   port returns to it.
+4. Both ends must converge on the **same** CA. A configuration where each end
```
m26singhvi: What is the meaning of "converge" here ? What happens if both primary and fallback CAK/CKN have CA established ?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+   independently prefers a different CKN must not oscillate.
+5. The feature must be **opt-in**. With only a primary CAK configured, behaviour
+   is unchanged.
+6. A CAK must be replaceable **at runtime**, without restarting the supplicant
```
m26singhvi, line 92: Any particular reason to get this covered ?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+### 3.2 Principal CA selection
+
+`ieee802_1x_kay_decide_principal()` is evaluated whenever peer liveness or key
+server election changes. It answers one question: *which CA owns the Controlled
+Port?*
+
+```mermaid
+flowchart TD
+  A["Peer liveness / election changed"] --> B{"Are we the elected <br/>key server on a live CA?"}
+  B -- yes --> C["Choose that CA. <br/>Prefer the primary <br/>over the fallback."]
```
m26singhvi: Choose that CA isn't really required, right ?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+| `lki/lan/ltx/lrx`, `oki/oan/otx/orx` | `kay->…` | One installed-key view per port |
+| principal flag | `kay->principal_participant` | A single pointer, reachable via `get`/`is`/`set` accessors |
+
+This is a **pure refactor** — with one participant the hoisted state has exactly
+the same lifetime and values as before — but it is the precondition for
+everything that follows.
+
+### 3.2 Principal CA selection
+
+`ieee802_1x_kay_decide_principal()` is evaluated whenever peer liveness or key
```
m26singhvi: what is peer liveness ? It is liveness of the session, right ?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+server election changes. It answers one question: *which CA owns the Controlled
+Port?*
+
+```mermaid
+flowchart TD
+  A["Peer liveness / election changed"] --> B{"Are we the elected <br/>key server on a live CA?"}
+  B -- yes --> C["Choose that CA. <br/>Prefer the primary <br/>over the fallback."]
+  B -- no --> D{"Does the current <br/>principal still have <br/>a live peer?"}
```
m26singhvi, line 190: If it is not a key server, then why do we care here ? Aren't we mixing up 2 functionality/state machines ?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+  D -- yes --> E["Keep it — the remote <br/>key server chose this CKN"]
+  D -- no --> F{"Port secured and a <br/>sibling CA is live?"}
+  F -- yes --> G["Fail over to the sibling"]
```
m26singhvi: What is the meaning of this ? What extra non key server does at this point ?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+  F -- no --> H["No CA can own the port: <br/>tear down the controlled port"]
+  C --> I["reconcile_principal()"]
```
m26singhvi: Where can I read more about what all will be done as part of reconcile principal ?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+  E --> I
+  G --> I
+```
+
+Two properties fall out of this ordering and are what keep the two ends in
+step:
+
+- **Primary preference applies only to the key server.** A follower keeps the CA
```
m26singhvi: What is a follower ? Do you mean non key server ? If yes, can we use that terminology for consistency with 802.1X standard

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+
+- A `Distributed SAK` is validated **entirely within the receiving CA** before it
+  is allowed to touch the CP, so a SAK arriving on the fallback CA cannot
+  install key material for the port it does not own.
+- Only the **principal** key server distributes SAKs. A rekey requested on a
+  follower is retained (and honoured on promotion) rather than silently dropped.
+
+### 3.3 Hitless failover
+
+The promotion itself must not disturb the datapath. The incoming CA **inherits
```
m26singhvi: What is the meaning of inheritance here ? Control plane and data plane are essentially decoupled. As soon as a new CAK/CKN is chosen to be principal, it generates/distributes a new SAK keeping in mind the next association number to be used. 

Continue with SAK from an older CAK/CKN doesn't sound right to me semantically.

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+  participant CAF as CA-fallback <br/>(mka_ckn_fallback)
+  participant CP as CP state machine
+  participant SEC as SecY / SAI
+
+  Note over CAP,CAF: both CAs exchange MKPDUs independently <br/>only the primary owns the port
+
+  CAP->>KAY: last live peer times out
+  KAY->>KAY: decide_principal() <br/>primary has no live peer, <br/>fallback is live → fallback wins
+  KAY->>KAY: migrate_principal_sas() <br/>re-home installed-SAK bookkeeping
+  KAY->>CAF: set principal, elect key server
```
m26singhvi: What is the meaning of elect key server now ? key server priority is a port property and not a participant property, right ?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+  Note over SEC: no SA delete / create — <br/>traffic keeps flowing on the inherited SAK
+  KAY->>KAY: arm deferred rekey (≈3 × hello time)
+
+  rect rgb(235,245,255)
+  Note over CAP,CAF: settle window — both ends converge on the new principal
```
m26singhvi: What is the meaning of settle window ?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+defined in the CONFIG\_DB `MACSEC_PROFILE` table, so no schema change is needed.
+
+### 4.2 Control interface
+
+Four commands are added so a CAK can be rotated at runtime without restarting
+the supplicant or bouncing the link.
+
+| `wpa_cli` command | Control interface | Description |
+| ----------------- | ----------------- | ----------- |
+| `macsec_add_mka ckn=<hex> cak=<hex> [fallback=1]` | `MACSEC_ADD_MKA` | Create an MKA participant on the running KaY. Creates the **primary** CA by default; with `fallback=1` the participant only claims the port while the primary CA has no live peer. Returns `FAIL` if the port already has a primary. |
```
m26singhvi, line 423: I didn't get the `FAIL` part here. Can you please clarify

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+is_principal=Yes  is_primary=Yes
+live_peers=1      potential_peers=0
+is_key_server=Yes is_elected=Yes
+```
+
+#### 4.2.1 Rotating the primary CAK
+
+Because a port has exactly one primary CA, the new primary cannot be stacked on
+top of the old one — `macsec_add_mka` returns `FAIL` while a primary is present.
+Rotation is therefore **delete, then add**, with the fallback CA carrying the
```
m26singhvi, line 447: If user is expected to delete first, then how do we differentiate between the following two scenarios : 

1. MKA with primary working
2. User deletes primary because he/she wants to genuinely delete the session now. 

Other

1. MKA with primary working
2. User deletes primary because he/she wants to rotate.

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,499 @@
...
+port in between:
+
+1. `macsec_del_mka` the old primary. The KaY re-selects, the fallback CA is
```
m26singhvi, line 450: Why are we forcing users to take care of the deletion and not let system take care of this ?

m26singhvi (commented):
Added few comments.

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: config wise fallback only doesnt make much sense/we will probably prevent it at the sonic config layer. As far as wpa supplicant is concerned, it should support only fallback.
Primary and fallback CKN must be different, as its used in the mkpdus to identify the keys. CAK can be the same although not recommended

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: yes - can add clarification

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: for hitless runtime CAK rotation - this requires a valid fallback to achieve though. I can add that clarification

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: the MKPDUs will be advertised, but if the peers drop off (due to mismatch) it will use the fallback, since it's the only live session

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: in steady state, both should be valid CAs. The key server will choose the primary to distribute keys on if both are available

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: there is debouncing in the SAK rekeys, so that case should be covered.

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: yes, liveness of the CA / MKA / whether the session has live peers (ie. both ends agree on CAK)

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: not sure what you mean - if we have 2 CAs, then we have to choose which one to use ?

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: the system will take care of this - in sonic layer, macsecmgrd will do this sequencing.
This HLD just covers wpa supplicant - in practice the user will not be touching this control interface.

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: The user will delete the whole macsec network block / session if they want to tear it down. 
If we are deleting an MKA will a fallback available, it will always try to continue with the fallback.

Realistically, the user will not use this control interface. In sonic, it will be handled with config commands.

For key rotation, user will use "config macsec profile update ...", to tear down the macsec session, user will use "config macsec port del ..."

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: this is just wpa supplicant control interface semantics.
If we try to add multiple primaries concurrently, wpacli will reject it.

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: settle window is the time it takes for the MKA hellos to converge. Generally the "deferred rekey" grace period. 
I can reword this to be more clear

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: Yes thats correct, this is referencing the literal function call in wpa supplicant. Key server election will yield the same result for all participants, since as you mentioned, its a port/SCI property
I can reword this to be more clear

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: yep will update all references here

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: I will post the wpa_supplicant changes soon

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: yeah this is misleading - the non-key server just uses whatever mka it got its last sak on / whatever has live peers

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: non-keyserver just uses whatever it gets keys distributed on

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: yes your read is correct.
"Inherit" more means that the current SAK isnt immediately torn down, from wpasupplicant perspective, the key is still live and it "moves" (internally) from the old MKA to the newly promoted principal, and is used until the deferred rekey is invoked and the SAK under the new MKA is distributed. 
I can reword if this causes some confusion.

doc/macsec/MACsec_fallback_cak_hld.md
m26singhvi: Sorry it's not clear. MKPDUs will be advertized with new CAK/CKN only ?

doc/macsec/MACsec_fallback_cak_hld.md
m26singhvi: I meant "Prefer the primary over the fallback" is sufficient.

doc/macsec/MACsec_fallback_cak_hld.md
m26singhvi: How about using liveness to express that ?

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: mkpdus will be advertised with both configured cak/ckn, but if there is mismatch icv wont be validated and they will be ignored.
if we remove a cak/ckn (eg. key rotation scenario), then the mkpdus for that cak/ckn will stop.
SAK keys will only be distributed on a mka if the cak/ckn is valid (ie. both sides agree / icv passes)

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,569 @@
...
+  participant SEC as SecY / SAI
+
+  Note over CAP,CAF: both CAs exchange MKPDUs independently <br/>only the primary owns the port
+
+  CAP->>KAY: last live peer times out
+  KAY->>KAY: select_principal() <br/>primary has no live peer, <br/>fallback is live → fallback wins
+  KAY->>KAY: migrate_principal_sas() <br/>re-home installed-SAK bookkeeping
+  KAY->>CAF: set principal, re-run key server election <br/>(same verdict)
+  Note over SEC: no SA delete / create — <br/>traffic keeps flowing on the inherited SAK
+  KAY->>KAY: arm deferred rekey (≈3 × hello time)
```
judyjoseph, line 261: This 3 x hello-interval I assume is the max-T, and if both peers' handshake done earlier, we can switch to fallback? 

Another question if there is a slow h/w on other side which takes more than 6s ? what is the bahavior

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,569 @@
...
+Use body re-samples the shared transmit SC, so the first MKPDU of an interval
+advertises a stale PN and the rest advertise one only moments old — and which
+participant sees which depends on list order.
+
+The value to advertise is now sampled **once per interval**, when the principal
+encodes its MKPDU, and every participant reports that snapshot. Sampling cadence
+becomes a property of the port rather than of the number of CAs on it.
+
+#### 3.5.4 Two smaller fixes
+
```
judyjoseph, line 365: The Fixes may be we can share as a separate section at the end of doc ? if that is ok I beieve these were existing issues in wpa_supplicant

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,569 @@
...
+the two steps. That flow is out of scope for this document.
+
+## 6 Backward compatibility
+
+- With only `mka_cak` / `mka_ckn` configured, exactly one participant is created
+  and it is unconditionally the principal. Behaviour is byte-for-byte the
+  previous behaviour.
+- No SONiC MACsec plugin API changed, so no MACsecOrch or SAI change is implied.
+- No CONFIG\_DB / APP\_DB / STATE\_DB schema change.
+- The `MACSEC` status output gains `is_principal` and `is_primary` per
```
judyjoseph, line 530: are you referring to primary and fallback here

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,569 @@
...
+  participant and now reports secure channels once for the port rather than once
```
judyjoseph, line 531: Another comment is with the SAI counters in show command o/p in the earlier case I think was getting cummulated ( i think SAI/SDK was taking care of it ) as it was same key. Now with fallback key and we using the same SC -- when we program a new SA, it could be reset

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: "is_primary" is config convention "primary cak", as in what's preferred if we have 2 that are live.
"is_principal" is the ieee definition, which is the MKA currently driving the port (ie. generating keys)

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: SAI has no concept of the cak. Since this behaviour is identical, in terms of what the HW sees, i doubt that counters will be reset. we can test for this though

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: sure i can move this

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: realistically they should converge quickly - the MKA is already established. This is mainly because i wanted to move the timing of the rekey off the instant the mka goes down. Perhaps this time should be random with some lower/upper bound
Once the rekey is kicked, its the same process as existing sak rekey

liamkearney-msft:
@m26singhvi 
following on our conversation, I believe we just abandon the sak rekey midway through if we delete the MKA while its happening.
Then after deferred rekey period (6s) we will just rekey freshly.

karthik-nexthop:
> @m26singhvi following on our conversation, I believe we just abandon the sak rekey midway through if we delete the MKA while its happening. Then after deferred rekey period (6s) we will just rekey freshly.

hi @liamkearney-msft  ,

Thanks — that answers “we don’t finish the in-flight rekey.”

Still unclear: after abandon, which SAK remains in the SecY until the deferred rekey — the old one, or the new one if it was already installed as latest? And which AN does the fresh rekey take if the abandoned SA already consumed a slot?

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,569 @@
...
+the installed-SAK bookkeeping (§3.3), moving the principal and arming the
+deferred rekey (§3.4), or tearing the port down if no CA is returned.
+
+It is re-evaluated whenever **peer liveness** changes — a CA gaining its first
+live peer, or losing its last — or when key server election changes. A peer is
+live once both ends have exchanged MKPDUs under the same CAK.
+
+Selection is four ordered rules, first match wins:
+
+1. **We are key server on a CA with live peers** — that CA, primary preferred.
```
karthik-nexthop: If fallback gets a live peer before primary, it becomes principal. When primary later has a live peer and we are key server, rule 1 promotes primary. That is two inherit + deferred-rekey cycles at boot. Is that intended ? or should be give some time for principal to get the live peer before fallback ?

karthik-nexthop: A flaky primary (ICV fail / recover every few hellos) bounces ownership. How are we  handling this ?

liamkearney-msft:
@karthik-nexthop 
> Still unclear: after abandon, which SAK remains in the SecY until the deferred rekey — the old one, or the new one if it was already installed as latest? And which AN does the fresh rekey take if the abandoned SA already consumed a slot?

If it was installed and signaled swapped over, it will keep using that one, otherwise it will continue using the old key until the rekey is retried. Once MKA switches over it will rekey again. It will use next available AN, and will reuse the abandoned slot if we are AN constrained.

senthil-nexthop:
@liamkearney-msft , the overall design looks good and I just have a few clarifying questions.

1. Cross-vendor/Interop
- Session establishment and steady-state encryption interoperate fine, but the hitless property assumes the peer also retains the old SAK across its own failover and fails over symmetrically. That could depend on the imlementation.
- Deferred rekey is a key-server-side mechanism. When SONiC is the non-key-server and the peer is KS, SONiC cannot defer the peer's rekey. If HW has a limitation of 2 SAs (ANs) per SC, then a rekey initiated by the peer might overlap the key rotation. How is this handled?
- What if the non-SONiC peer uses a distinct SCI per CA, instead of one SCI for both CAs for the rotation? if the partner uses a second SCI for the same CA during rotation, then SONiC has to create a second SC. Will be good to verify if this is handled correctly.
- The HLD proposes an "only one side decides" coherent election outcome. Is it possible for SONiC side to be KS on primary CA and the partner wins KS on fallback CA? Might be good to call out explicitly that it is the operator's responsibility to configure the correct priority on both sides so the same side is KS for primary and fallback.
2. The HLD describes the mka primary/fallback primitives, but not the MACsecMgr workflow, as noted in section 5. Should that also be covered in this HLD?
3. "Only one end decides, the other follow" is good intent, but is it possible for fallback to be promoted on one end but not the other, resulting in split brain? How does the convergence work in such a scenario?
4. Deferring SAK rekey by 6s is fine for scheduled/operational rotation, but is that desirable? If the operator triggers a rotation because the CAK is compromised, we will be using a retired SAK for 6 more seconds.

saravanan-nexthop:
cc @senthil-nexthop for review

doc/macsec/MACsec_fallback_cak_hld.md
senthil-nexthop: Perhaps make it configurable? If the rotation is triggered by the operator in response to a compromised situation, it will be good to shorten this window.

doc/macsec/MACsec_fallback_cak_hld.md
senthil-nexthop: From the SAI point-of-view, the CAK rotation will just look like a new SAK rekey, so counters should not be affected. But good to confirm.

m26singhvi:
> > @m26singhvi following on our conversation, I believe we just abandon the sak rekey midway through if we delete the MKA while its happening. Then after deferred rekey period (6s) we will just rekey freshly.
> 
> hi @liamkearney-msft ,
> 
> Thanks — that answers “we don’t finish the in-flight rekey.”
> 
> Still unclear: after abandon, which SAK remains in the SecY until the deferred rekey — the old one, or the new one if it was already installed as latest? And which AN does the fresh rekey take if the abandoned SA already consumed a slot?

@liamkearney-msft -  Sorry for the delayed reply.  We need to make sure that this transition is hitless. The way you have described, it doesn't look like it will be hitless.  I hope you are consider all possible states of the peer as far as current SAK is concerned . New SAK will be generated only 6 seconds later ( I won't call it rekey period though ).

liamkearney-msft:
@saravanan-nexthop @m26singhvi 
Thanks for the review. There are some interoperability assumptions which i can make explicit in this doc. Ive been testing these changes against cEOS and i imagine the other vendors do something similar (all their config schemes are similar).

Regarding deletion of the active MKA during an in-flight SAK rollover: SONiC does not unconditionally discard the new SAK or choose between the old and new SAK again. It preserves the shared CP/SecY state and transfers RX SAK state to the incoming principal CA - this is to ensure the MKA is decoupled from the CP/SecY state, and old SAKs can still be retired after principal swap. Therefore:

- before `RECEIVE`, only the old SA remains installed;
- during `RECEIVE/RECEIVING`, the old transmit SA remains active while the new receive SA may already be installed;
- during `TRANSMIT/TRANSMITTING`, the new transmit SA remains active while the old receive SA is retained pending retirement.

The SA currently carrying traffic therefore remains in use across the CA ownership change. Distribution under the deleted CA stops, and the new principal key server performs a fresh rekey after the settle window.

For the next rekey, SONiC does not rely on a free-running AN counter. It starts after the most recently installed SAK, excludes ANs occupied by locally installed latest/old SAs, considers the ANs reported by live peers, and, if every local slot is occupied, selects the AN that the CP is about to release.

IEEE 802.1AE-2018 §§10.7.13 and 10.7.22 state that creating a receive or transmit SA at an occupied AN deletes the prior SA at that AN. This is the reason safe AN selection is required: reusing the wrong AN could delete the SA currently carrying traffic.

The intended transition is therefore hitless when both endpoints retain at least one common active SAK, the fallback CA is already live, and the peer/key-server SCI remains stable. I can put this in explicitly in the doc

1. Cross-vendor/interoperability

A peer does not specifically have to retain the *old* SAK if both ends have already moved to the new one; the requirement is that they retain at least one common active SAK throughout the transition.

When SONiC is key server, the rollover changes retain the old receive SA until every live peer reports that it is transmitting on the latest SAK, subject to a bounded failsafe. When SONiC is non-key-server, it cannot impose that retention behavior on the peer key server. Interoperability therefore depends on the peer implementing an equivalent make-before-break rollover. This is consistent with observed eos behavior

If an MKA participant is removed, its MKPDUs stop. The peer learns that the CA is no longer usable when its live-peer entry expires, while the continuously active fallback CA is already available for selection.

For hardware supporting fewer than four SAs per SC, SONiC raises its key-server priority to `0` so that it normally controls AN allocation. This is an implementation mechanism rather than an absolute guarantee of election: a priority tie or an obliged remote key server can still leave SONiC as non-key-server. In that case SONiC must accept the AN in the peer's Distributed SAK, so the peer must allocate an AN that permits make-before-break operation. I will call this out as an interoperability scenario requiring explicit validation.

SONiC advertises the same port-level SCI on both CAs. The implementation can create separate receive SCs for distinct peer SCIs, but a change in the elected key-server SCI signals `chgdServer` to the CP and causes the existing SAs to be replaced. The current hitless design therefore assumes that the same physical peer uses a stable SCI across the primary and fallback CAs. I will make that assumption explicit.

Key-server election is performed independently within each CA. SONiC uses the same local key-server priority and SCI for both CAs, so the elections should produce the same result when the same two actors are live and configured consistently. However, different peer membership or peer configuration can produce different results. Having a consistent key server configuration is the responsibility of the operator, but I can make that clear explicitly. 

2. MACsecMgr workflow
I was going to add a 2nd HLD for macsecmgr/ config commands, but i can add a small summary here

3. Split-brain and convergence

The intended convergence rule is:

- the key server selects the CA on which it distributes the SAK, preferring its live primary CA;
- a non-key-server follows a validated Distributed SAK on the CA selected by that key server;
- if the key server's primary peer expires, it selects the already-live fallback CA and distributes there.

This prevents oscillation while one coherent key server drives the port. It does not unconditionally prevent split brain if the two CAs elect different key servers. Consistent key-server priority/SCI configuration across both CAs is therefore a deployment requirement.

4. Deferred rekey and compromised CAKs

The delay is ~3x hellos (6s with default hello period) - Its intention was to give the mka change some time to settle before immediately issuing a rekey. I can test out and see if we can remove / move it to a config knob which we can set in macsecmgrd once validated against other vendors.

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,695 @@
...
+  Note over CA2: MKA converges, peer becomes live
+  KAY->>KAY: select_principal() <br/>primary is eligible again
+  KAY->>CA2: promote (primary outranks fallback), migrate SAK
+  CA2->>CP: principal (hitless)
+  Note over CA2,CP: deferred rekey under the new CKN (§3.4)
+```
+
+## 5 Interaction with MACsecMgr
+
+No MACsecMgr change is required for the static case: `fallback_cak` /
```
rajshekhar-nexthop, line 637: The HLD correctly states that  fallback_cak  /  fallback_ckn  already exist in  MACSEC_PROFILE  in CONFIG_DB. However, the current  MACsecMgr::configureMACsec()  only programs  mka_cak  and  mka_ckn ; it does not send the fallback parameters to wpa_supplicant.

Could we reword this to say that static fallback support requires a companion MACsecMgr change which maps the existing CONFIG_DB fields to  mka_cak_fallback  /  mka_ckn_fallback  before  enable_network ? Reusing the existing  set_network  flow is fine, but the mapping is still needed for the feature to be usable.

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,695 @@
...
+participant_idx=0
+ckn=6162636465666768...
+mi=...            mn=42
+active=Yes        participant=Yes     retain=No
+is_principal=Yes  is_primary=Yes
+live_peers=1      potential_peers=0
+is_key_server=Yes is_elected=Yes
+```
+
+#### 4.2.1 Rotating the primary CAK
```
rajshekhar-nexthop, line 569: Could this section also define the supported workflow for rotating or removing the fallback CAK/CKN on a live profile? The fallback is a standing credential, so it needs planned-rotation and revocation semantics as well as initial configuration. Replacing a standby fallback should be straightforward while the primary is live:  macsec_del_mka  the old fallback, then  macsec_add_mka  the new one. Could the HLD state the operational precondition explicitly? In particular, MACsecMgr must not remove the fallback while it is principal, because that can remove the CA currently carrying traffic.

It would also help to document that the primary continues forwarding during the update, but the fallback is temporarily not failover-ready until the new participant has a live peer.

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,695 @@
...
+
+- With only `mka_cak` / `mka_ckn` configured, exactly one participant is created
+  and it is unconditionally the principal. Behaviour is byte-for-byte the
+  previous behaviour.
+- No SONiC MACsec plugin API changed, so no MACsecOrch or SAI change is implied.
+- No CONFIG\_DB / APP\_DB / STATE\_DB schema change.
+- A principal change does not recreate an SC or SA, so it does not reset SAI
+  counters. The deferred SAK rollover has the same counter lifecycle as an
+  ordinary rekey.
+- The `MACSEC` status output gains `is_principal` and `is_primary` per
```
rajshekhar-nexthop, line 667: How do the per-participant fields reach the SONiC  MACSEC  status/monitoring path?  macsec_mka_list  exposes runtime state in wpa_supplicant, but the current  show macsec  path reads CONFIG_DB/APPL_DB and does not query wpa_supplicant.

CONFIG_DB can confirm the intended fallback CKN/CAK configuration, but it cannot show whether the fallback participant exists and has  live_peers > 0 . Please identify the proposed plumbing, or scope this status addition as companion work. A configured fallback that has no live peer should be observable before a primary failure tests it.

liamkearney-msft: Ill add some more detailed scope here for exposing MKA state into statedb, so config cli has better visibility. I agree, for operator ease/safety having some smarts there is a good idea

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: Sure thing, i can add some extra clarification here on process / expected workflow

doc/macsec/MACsec_fallback_cak_hld.md
liamkearney-msft: yes, i agree this wording is misleading - I will fix this.

bmridul:
@ChennakesavaReddy-G , Please review

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,695 @@
...
+The SAK-Use reporting role swaps before the next MKPDU, as shown above; the
+installed SAs and shared key state do not change.
+
+The SA that remains active depends on how far the CP had progressed when the
+outgoing CA was removed:
+
+| CP phase | State preserved across promotion |
+| -------- | -------------------------------- |
+| Before `CP_RECEIVE` | Only the pre-existing/current SA carries traffic. |
+| `CP_RECEIVE` / `CP_RECEIVING` | The old transmit SA remains active; a receive SA and disabled transmit SA for the new SAK may already exist. |
```
Pterosaur, line 293: Could you clarify whether packet loss caused by an inherited `transmit_when` timeout is expected during this transition?

For example, with the default 6-second transmit timeout and approximately 6-second settle window:

- **t = 0 s:** A is the key server on both CAs, and both endpoints are forwarding with K0. A starts distributing K1 on the primary CA and enters `CP_RECEIVING`, but B misses the Distributed SAK containing K1. A therefore still has `all_receiving = false` and transmits with K0.
- **t = 0.1 s:** The primary CA is deleted on A. The already-live fallback is promoted, with the same peer/key-server SCI. Per this section, the CP state is preserved, distribution on the deleted CA stops, and K1 is not redistributed under the fallback CAK. The fresh fallback rekey is deferred until approximately t = 6.1 s or later.
- **t = 6 s:** The original `transmit_when` expires. The existing `CP_RECEIVING` transition allows `CP_TRANSMIT` when `all_receiving || !controlled_port_enabled || !transmit_when`, so A can enable K1 for transmission even though `all_receiving` is still false. B does not have K1, so A-to-B traffic is dropped.
- **t >= 6.1 s:** The deferred fallback rekey can begin, but starting a fresh distribution does not itself immediately restore traffic; the endpoints still need to install and switch to the new SAK.

I can understand packet loss being an intentional consequence of the transmit-timeout failsafe. Is that the expected behavior here? If so, could this section explicitly describe this interrupted-rollover / `transmit_when`-expiry case as a possible packet-loss scenario and qualify the hitless claim accordingly? If not, what prevents the inherited timeout from enabling an unconfirmed K1 during the settle window?

liamkearney-msft: Yes, thanks that's a good edge case - we should explicitly handle it imo
If we are in receiving during a principal change, and the peer hasn't confirmed its key yet, we will cancel the transmit_when timer & just kick the CP/SM to ABANDON - we will then keep transmitting on the old keys, and a new key exchange/distsak will try again once principal change rekey fires.

eg: 
```
if ((sm->CP_state == CP_RECEIVE ||
     sm->CP_state == CP_RECEIVING) &&
    sm->lki && !sm->ltx && sm->oki && sm->otx) {
	/* Keep the old SAK active if a peer missed DistSAK on CA change. */
        ...
```

liamkearney-msft:
hey reviewers - companion HLD can be found here : https://github.com/sonic-net/SONiC/pull/2545

liamkearney-msft:
@reviewers PRs can be found here
These are sequential and need to be merged in sequence. Each PR has the previous's commits, has been broken into 4 to help assist with review
https://github.com/sonic-net/sonic-wpa-supplicant/pull/135
[MACsec: move per-port SecY state to KaY by liamkearney-msft · Pull Request #136 · sonic-net/sonic-wp…](https://github.com/sonic-net/sonic-wpa-supplicant/pull/136)
[MACsec: support fallback CAKs with automatic failover by liamkearney-msft · Pull Request #137 · soni…](https://github.com/sonic-net/sonic-wpa-supplicant/pull/137)
[MACsec: make SAK rollover safe across multiple peers by liamkearney-msft · Pull Request #138 · sonic…](https://github.com/sonic-net/sonic-wpa-supplicant/pull/138)

liamkearney-msft:
> For hardware supporting fewer than four SAs per SC, SONiC raises its key-server priority to 0 so that it normally controls AN allocation.

For clarity, this is existing sonic wpa-supplicant behavior. It is not being introduced as a part of this change. see: https://github.com/sonic-net/sonic-wpa-supplicant/pull/47

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,714 @@
...
+  PRI->>PEER: full SAK-Use<br/>KI, AN, RX/TX, LPN, DP = shared state<br/>PTx/PRx = shared status
+  FB->>PEER: full SAK-Use<br/>key-state fields = 0<br/>PTx/PRx = shared status
+
+  Note over SEC,FB: principal migrates<br/>installed SAs do not change
+  SEC-->>FB: reporting role moves
+  PRI->>PEER: full SAK-Use<br/>key-state fields = 0<br/>PTx/PRx = shared status
+  FB->>PEER: full SAK-Use<br/>KI, AN, RX/TX, LPN, DP = shared state<br/>PTx/PRx = shared status
+```
+
+SONiC accepts valid SAK-Use on either CKN and uses it for that actor's
```
ChennakesavaReddy-G: SAK-Use Processing:
What if both the Primary and Fallback actors advertise the Principal Actor's SAK-Use?
To prevent convergence on the wrong actor, the following validation checks are necessary as part of SAK-Use processing.

MKA SAK-Use Validation:
1. Receiving Participant is the Key Server
   - The received Latest_Key_Server_MI in the SAK-Use parameter set must match the local participant's my_MI.
   - In addition, the complete received latest SAK KI (Key_Server_MI + Key_Server_KN) must match the locally stored latest SAK KI.
   - If either validation fails, the received SAK-Use parameter set must be ignored.

2. Receiving Participant is a Non-Key Server
   - The received SAK-Use must reference a SAK distributed by the elected Key Server.
   - The received Latest_Key_Server_MI in the SAK-Use parameter set must match the local participant's my_elected_Key_Server_MI.
   - If the validation fails, the received SAK-Use parameter set must be ignored.

Example:
Assume the principal actor is Primary and the SAK KI carried in the received SAK-Use is:
SAK KI = 0x123456789a0001

where:
Key_Server_MI = 0x123456789a
Key_Server_KN = 0x0001

Case 1: Receiving Participant is the Key Server
Primary CA:
my_MI = 0x123456789a
received latest_Key_Server_MI = 0x123456789a
The local my_MI matches the received latest_Key_Server_MI, and the complete received SAK KI matches the locally stored latest SAK KI.
Result: Honor the SAK-Use parameter set.

Fallback CA:
my_MI = 0x9999999999
received Key_Server_MI = 0x123456789a
The local my_MI does not match the received latest_Key_Server_MI
Result: Ignore the SAK-Use parameter set.


Case 2: Receiving Participant is a Non-Key Server
Primary CA:
my_elected_Key_Server_MI = 0x123456789a
received Key_Server_MI = 0x123456789a
The locally elected Key Server MI matches the received latest_Key_Server_MI
Result: Honor the SAK-Use parameter set.

Fallback CA
my_elected_Key_Server_MI = 0x9999999999
received Key_Server_MI = 0x123456789a
The locally elected Key Server MI does not match the received Key_Server_MI.
Result: Ignore the SAK-Use parameter set.

liamkearney-msft: We will ignore sak use from non-principal, so we can handle whether fallback is sending 0s or duplicating principals sak use.
This is mainly just to specify what sonic sends, which is 0s. 

Ill update hld to specify this

doc/macsec/MACsec_fallback_cak_hld.md
```diff
@@ -0,0 +1,725 @@
...
+
+The KaY can represent more than one receive SC, keyed by peer SCI. That is not
+equivalent to hitless failover across a peer SCI change. A change in elected
+key-server SCI signals `chgdServer` to the CP, which transitions through
+`CP_CHANGE` and removes the existing SAs. The zero-loss fallback path therefore
+assumes a stable peer/key-server SCI across the primary and fallback CAs.
+
+### 3.4 Deferred post-promotion rekey
+
+Rekeying *immediately* on promotion is what breaks the datapath: while the two
```
ChennakesavaReddy-G: Is it possible to share a scenario that could potentially break the datapath?
Ideally, both parties should continue operating with the old SAK until the new SAK is successfully provisioned by the new principal actor. What specific condition could cause the datapath to break and require a 6-second delay for provisioning the new key?

liamkearney-msft: yes i agree this is strong wording. It doesn't necessarily break the datapath but for robustness, moving the rekey off the moment the CA goes away seemed prudent.
We can revisit this once the main changes go in.

Ill revise the wording to be less absolute

abdosi:
please start review of the code PR's also

PR #2500, opened 2026-08-12
[doc] HLD: BMC-initiated graceful shutdown and restart of the switch-host

william8545:
On a chassis where SONiC runs on both the BMC and the switch-host, the BMC removes switch-host power without telling the host, so the host cannot flush state, run any platform ordering it needs, or record why it went down.

This design adds a bounded handshake before the cut: the BMC asks the host over gNOI System.Reboot(HALT), waits up to graceful_shutdown_timeout, and removes power either way. The host side is the existing reboot -p path, which runs the normal reboot teardown and exits with the OS still running, so the host reports completion in band on System.RebootStatus and no new platform API or hardware readiness signal is needed. It also adds a GRACEFUL_RESTART command composed from power off, a pause and power on, so a critical leak can be re-checked before power is raised.

No SAI change, no new daemon, container, database table, CLI command or gNOI method. A platform opts in through a platform.json capability after passing the qualification checks in the document; every other platform keeps today's behaviour.

### DIFF
--- a/doc/bmc/graceful-shutdown/switch-host-graceful-shutdown-hld.md
+++ b/doc/bmc/graceful-shutdown/switch-host-graceful-shutdown-hld.md
@@ -1,0 +1,943 @@
+# Graceful Shutdown and Restart of the Switch-Host from BMC
+
+## Table of Content
+
+- [1. Revision](#1-revision)
+- [2. Scope](#2-scope)
+- [3. Definitions/Abbreviations](#3-definitionsabbreviations)
+- [4. Overview](#4-overview)
+- [5. Requirements](#5-requirements)
+- [6. Architecture Design](#6-architecture-design)
+- [7. High-Level Design](#7-high-level-design)
+  - [7.1. Shutdown flow](#71-shutdown-flow)
+  - [7.2. Host side: `reboot -p`](#72-host-side-reboot--p)
+  - [7.3. Knowing the host finished](#73-knowing-the-host-finished)
+  - [7.4. Timing](#74-timing)
+  - [7.5. Graceful restart](#75-graceful-restart)
+  - [7.6. Concurrency and preemption](#76-concurrency-and-preemption)
+  - [7.7. Reboot cause](#77-reboot-cause)
+  - [7.8. Failure handling](#78-failure-handling)
+  - [7.9. Platform requirements](#79-platform-requirements)
+  - [7.10. Security](#710-security)
+  - [7.11. Serviceability](#711-serviceability)
+  - [7.12. Compatibility](#712-compatibility)
+  - [7.13. Considered alternatives](#713-considered-alternatives)
+- [8. SAI API](#8-sai-api)
+- [9. Configuration and management](#9-configuration-and-management)
+  - [9.1. Manifest](#91-manifest)
+  - [9.2. CLI/YANG model Enhancements](#92-cliyang-model-enhancements)
+  - [9.3. Config DB Enhancements](#93-config-db-enhancements)
+  - [9.4. State DB Enhancements](#94-state-db-enhancements)
+- [10. Warmboot and Fastboot Design Impact](#10-warmboot-and-fastboot-design-impact)
+  - [Warmboot and Fastboot Performance Impact](#warmboot-and-fastboot-performance-impact)
+- [11. Memory Consumption](#11-memory-consumption)
+- [12. Restrictions/Limitations](#12-restrictionslimitations)
+- [13. Testing Requirements/Design](#13-testing-requirementsdesign)
+  - [13.1. Unit Test cases](#131-unit-test-cases)
+  - [13.2. System Test cases](#132-system-test-cases)
+- [14. Open/Action items](#14-openaction-items)
+- [15. References](#15-references)
+
+### 1. Revision
+
+| Rev | Date | Author | Change Description |
+| --- | --- | --- | --- |
+| 0.1 | 2026-08-12 | William Tsai | Initial version |
+
+### 2. Scope
+
+On a chassis where SONiC runs on both the BMC and the switch-host, the BMC owns switch-host power. It
+removes that power without telling the host, so the host cannot flush state, run any platform ordering
+it needs, or record why it went down.
+
+This design lets the BMC ask the host to prepare first, wait a bounded time, and then remove power. It
+adds:
+
+- a graceful leg to the existing `GRACEFUL_SHUT` command — ask, wait, then remove power;
+- a new `GRACEFUL_RESTART` command — the same shutdown, a short pause, then power on;
+- an extension of the existing critical-leak check to `POWER_CYCLE`, which is ungated today, and a
+  re-read of that state under the power lock. `POWER_ON` is gated on its three other entry points already;
+  the startup power-on on a non-liquid-cooled chassis is not. The re-read reaches that call site, but there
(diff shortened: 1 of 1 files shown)


yxieca:
Really solid HLD — the safety invariant (power is removed on every path; the host's confirmation only changes how the op is recorded) reads consistently throughout, and the failure model and open-items are refreshingly honest. A few clarifying questions to make sure I'm reading the design right:

1. Result attribution (§7.3, lines 277-281): the active report carries the BMC's tag, but the terminal report seems to swap it for a fixed string — so distinguishing op-N's terminal from op-(N-1)'s looks like it relies on inference (host refuses others while active + first poll lands in the active window + the counter never moves backward). Is that the intended attribution model, or would echoing a per-request nonce even on the terminal report be preferable to make it positive rather than inferred? Are there sequences (e.g. a slow first poll) where the inference could slip?

2. Stop non-propagation (§7.2): the reporting path surviving teardown appears to depend on syncd.service being ordered with only After=. Can the doc explicitly assert there's no BindsTo=/PartOf=/Requires= on syncd.service? I want to confirm a stop can't propagate and take down the database/swss/gnmi reporting path.

3. Critical-leak path (§12): during an in-flight uninterruptible POWER_CYCLE (do_power_cycle()), how is a critical leak handled? My reading is graceful-restart routes around it but legacy POWER_CYCLE does not — is a stated restriction warranted here?

4. Timeout default (§7.4 vs §14#1): bmcctld defaults to 0 while the parent design/CLI use 120s. I see this is already flagged as an open item — is the plan to freeze one value, and will the CLI upper-limit-vs-watchdog validation (§9.2) be specified before implementation?

5. gNOI authz (§7.10/§14#3): nice that the System-role-also-grants-OS/File/Containerz breadth is called out and gates enablement — is the intent that enablement stays blocked until that granularity is resolved?

oleksandrivantsiv (dismissed):

william8545:
> Really solid HLD — the safety invariant (power is removed on every path; the host's confirmation only changes how the op is recorded) reads consistently throughout, and the failure model and open-items are refreshingly honest. A few clarifying questions to make sure I'm reading the design right:
> 
> 1. Result attribution (§7.3, lines 277-281): the active report carries the BMC's tag, but the terminal report seems to swap it for a fixed string — so distinguishing op-N's terminal from op-(N-1)'s looks like it relies on inference (host refuses others while active + first poll lands in the active window + the counter never moves backward). Is that the intended attribution model, or would echoing a per-request nonce even on the terminal report be preferable to make it positive rather than inferred? Are there sequences (e.g. a slow first poll) where the inference could slip?
> 2. Stop non-propagation (§7.2): the reporting path surviving teardown appears to depend on syncd.service being ordered with only After=. Can the doc explicitly assert there's no BindsTo=/PartOf=/Requires= on syncd.service? I want to confirm a stop can't propagate and take down the database/swss/gnmi reporting path.
> 3. Critical-leak path (§12): during an in-flight uninterruptible POWER_CYCLE (do_power_cycle()), how is a critical leak handled? My reading is graceful-restart routes around it but legacy POWER_CYCLE does not — is a stated restriction warranted here?
> 4. Timeout default (§7.4 vs §14#1): bmcctld defaults to 0 while the parent design/CLI use 120s. I see this is already flagged as an open item — is the plan to freeze one value, and will the CLI upper-limit-vs-watchdog validation (§9.2) be specified before implementation?
> 5. gNOI authz (§7.10/§14#3): nice that the System-role-also-grants-OS/File/Containerz breadth is called out and gates enablement — is the intent that enablement stays blocked until that granularity is resolved?

Thanks — good questions, and a couple of them found real problems. All five are reflected in the current revision.

**1. Attribution.** You were right that the inference could slip: the acceptance the BMC holds is the *backend's*, returned as soon as its worker spawns, so an early poll can read the previous operation's report. The id is now a fresh UUID (a counter would repeat across a `bmcctld` restart), carried on every terminal report — appended, not swapped — and graceful requires all five of: our tag, terminal, `HALT`, `SUCCESS`, empty status message (§7.3).

**2. `syncd.service`.** Asserted now, but directionally rather than as that list — the set of stop-carrying directives isn't closed, and either unit can declare the edge. §7.9 asserts the property instead: on the rendered image, nothing propagates a `syncd` stop or inactivation into the reporting path. You were right to single out `swss`: it declares nothing about `syncd.service`, yet its `ExecStart` holds a container watch over `syncd` — real coupling, just not in systemd. Neither teardown branch trips it, and §13.2 injects both `syncd` endings on hardware. (`swss` isn't on the reporting path; the report lives in `sonic-hostservice`.)

**3. Critical leak.** Yes — now a stated restriction (§12): `do_power_cycle()` can't be interrupted, so a critical action arriving during it waits for the call to return, which is why graceful restart doesn't use it (§7.5). Also worse than the doc said: `rack_mgr_critical_alert_action` defaults to `syslog_only`, not `power_off`.

**4. Timeout.** One value, frozen against measurement, with an owner (open item 1) — a non-zero timeout turns a critical `graceful_shutdown` into a real delay, so that change rides the same commit. The CLI limit can only be a coarse sanity bound (§9.2); the binding watchdog check is the host's own readback (open item 2). And `graceful_shutdown_timeout` is what binds: `reboot -p` runs to completion before the host's own clock starts.

**5. gNOI authz.** Yes — enablement stays blocked until this is resolved (§7.10, open item 3), and the gap is wider than per-module: one `gnoi` target, matched by role prefix, spans all of gNOI and most of gNSI. An unmatched certificate is refused only on writes, and the read-classified set includes `File.Put`/`Remove`, `OS.Install`, `FactoryReset.Start` and the gNSI rotations — so the gap admits unauthorized mutation, not read breadth. `Healthz.Artifact` doesn't authorize at all.

On your opening: power removal covers every outcome `bmcctld` handles, not every path — a daemon death abandons the operation, now a stated restriction (§12).

yxieca (approved):
Re-reviewed at head 8ca748d1f — thanks for the thorough responses; all five clarifying questions are addressed and grounded in the current revision:

1. **Attribution (§7.3):** the fresh UUID is now carried (appended, not swapped) on every terminal report, and graceful requires all five of tag-match / terminal / HALT / SUCCESS / empty-status. That converts the previously-inferred attribution into a positive check and closes the slow-first-poll slip.
2. **`syncd` non-propagation (§7.9):** the property is asserted directionally on the rendered image, and I appreciate that the `swss` → `syncd` container-watch coupling (invisible in unit files) is explicitly called out with the teardown shown not to trip it. The report living in `sonic-hostservice` (off the `swss` path) is the right anchor.
3. **Critical leak (§12):** stated restriction — `do_power_cycle()` is uninterruptible so a leak waits for it to return, which is why graceful restart routes around it; plus the honest correction that `rack_mgr_critical_alert_action` defaults to `syslog_only`.
4. **Timeout:** `graceful_shutdown_timeout` named as the binding bound, one value to be frozen against measurement inside the <260 s bracket (open item 1), with the CLI limit as a coarse bound and the host readback as the real watchdog check.
5. **gNOI authz (§7.10):** enablement stays blocked until resolved (open item 3); the gap correctly characterized as admitting unauthorized mutation (File.Put/Remove, OS.Install, FactoryReset, gNSI rotations), not just read breadth, with Healthz.Artifact not authorizing at all.

The three remaining items (timeout owner, watchdog readback, gNOI authz granularity) are appropriately tracked as in-doc open items. No blocking design defects — the power-removal-on-every-outcome invariant and the failure model read consistently. Approving.

PR #2506, opened 2026-08-20
[vpp][acl] add vpp acl `IN_PORT` support HLD

lolyu:
Add VPP ACL `IN_PORT` support.

|Repo|PR|state|
|-|-|-|
|sonic-platform-vpp|https://github.com/sonic-net/sonic-platform-vpp/pull/278|GitHub issue/pull request detail|
|sonic-sairedis|https://github.com/sonic-net/sonic-sairedis/pull/2064|GitHub issue/pull request detail|

* Microsoft ADO (number only): 39466368

Signed-off-by: Longxiang Lyu <lolv@microsoft.com>


doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,598 @@
...
+|------|---------|---------|
+| No entry carries `IN_PORTS` | Any ordinary L3 table | No partitioning is performed at all; every port binds the default ACL, exactly as before this design |
+| Port named by no entry | Ethernet0 above | Binds the default ACL — the unscoped entries only |
+| Ports with identical applicable sets | Ethernet4 and Ethernet16 above | Equal signatures, so one shared scoped ACL rather than a copy each |
+| Ports with different applicable sets | Ethernet8 versus Ethernet4 above | Different signatures, so one scoped ACL each |
+| Distinct entries that happen to match identically | Two `DROP any` entries, one naming Ethernet4 and one naming Ethernet8 | Signatures hold entry OIDs, not rule content, so these differ and produce two ACLs. Deduplication is never inferred from rule text |
+| Every port named, with the same scope | All mux ports standby | One signature covering all of them, so one scoped ACL bound to every port |
+| Table with no unscoped entries | The mux drop table with a single scoped entry | The default ACL has no rules to hold, so it degrades to an unmatchable placeholder; ports that are not candidates bind that and match nothing |
```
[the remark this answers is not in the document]
lolyu: Correct, and thanks — "unmatchable" was wrong. The rule is `permit dst 0.0.0.0/32` with `proto = 0`, so VPP skips the L4 checks entirely and it reduces to "any packet whose destination is 0.0.0.0". Nothing legitimate carries that destination, but it is craftable, and since a match ends evaluation of the interface's whole ACL list a permit there is a bypass of every lower-priority table on the port.

This design also makes it materially worse: previously the placeholder appeared only for a table with no entries at all, whereas now every non-candidate port of an all-scoped table binds one — on a dual-ToR box that is every active mux port.

Rather than leaving it unbound (which would cost the stable default-ACL identity the tunterm and table-removal paths rely on), the placeholder rule's action is now **deny**. That closes the bypass and is the correct treatment for the destination anyway. It does not disturb the ip2me path, which derives `has_deny` from the rules generated from the SAI entries, not from the placeholder. Doc updated in 19ad39b3, plus a new entry under Restrictions.

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,598 @@
...
+
+- Toward a **more** restrictive state (forwarding → dropping), the window still behaves like the state being left, so the change lands late and packets that should have been dropped are forwarded.
+- Toward a **less** restrictive state (dropping → forwarding), the window already behaves like the state being entered, so the change merely lands a fraction of a millisecond early. No packet is treated in any way the finished transition would not also have produced.
+
+### Failure Handling and Dirty Ports
+
+Computing the changed-port set by diffing old versus new *intent* is not sufficient. If a rebind fails, the new scope is still committed; on the next reprogram old and new intent agree, the diff is empty, no rebind is attempted, and the operation reports success with the port left unbound.
+
+A `m_acl_tbl_dirty_ports` set records every port whose bind/unbind failed or whose interface could not be resolved. A dirty port is forced through unbind/rebind on the next reprogram even when its assignment appears unchanged. A port whose *unbind* failed is additionally skipped in the bind loop, since binding onto an uncleared list would append out of order.
```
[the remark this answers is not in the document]
lolyu: Confirmed, and this was the most valuable comment of the set — thank you. Traced against the implementation: on a failed unbind the port is marked dirty and skipped in the bind pass, but `m_acl_port_scope` is still overwritten with the new assignment. Since unbinding resolves its swindex through that stored scope, the next reconcile asks VPP to unbind an ACL the port never had, gets `NO_SUCH_ENTRY` (deliberately mapped to success), binds the new ACL in front of the stale one, and the stale one stays bound — permanently, because a bound ACL cannot be deleted either. For the mux table that is a port stuck on its previous verdict: a standby→active transition that keeps dropping.

Fixed as you suggested, by retaining the old assignment rather than adding a parallel stale set: a port whose unbind failed keeps its old per-port swindex, that swindex is excluded from the reap so it is not freed while still bound, and it is carried under the empty signature so it stays tracked and becomes reapable as soon as a retry unbinds it. Doc section rewritten in 19ad39b3.

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,598 @@
...
+
+A scoped ACL that VPP refuses to delete is retained under an **empty signature**, which the grouping loop can never produce. It can therefore never be reused or rebound — only retried on a later pass.
+
+### Counters
+
+An ACE may now be replicated into several VPP ACLs, so per-entry counter tracking becomes a list. `vpp_ace_cntr_info_t` holds a vector of `vpp_ace_placement_t` (`acl_index`, `vpp_rule_base_index`, `num_rules`), and `getAclEntryStats()` sums across all placements. Placements accumulate, so `acl_ace_cntr_info_clear(tbl_oid)` is invoked before reprogramming a table to prevent double counting.
```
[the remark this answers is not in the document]
lolyu: Agreed. `acl_ace_cntr_info_clear()` runs before programming and programming is not staged, so a reprogram that fails part way leaves some ACLs of the table still active with their placements already discarded, and `getAclEntryStats()` under-reports for those entries until the next successful reprogram.

I have documented rather than fixed this: counts are not corrupted or double counted, only incomplete, and it recovers on the next successful pass — whereas retaining each old placement until its ACL is confirmed inactive needs the same transactional staging that is already deferred in the first Restrictions bullet. The Counters section now states that REQ-7 holds in the steady state only, and says which restriction removing it depends on. 19ad39b3.

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,598 @@
...
+
+### Scaling
+
+| Scenario | VPP ACLs created |
+|----------|------------------|
+| No entry carries `IN_PORTS` | 1 (unchanged from today) |
+| All mux ports standby | 1 scoped ACL bound to N interfaces, plus the default-ACL placeholder |
+| Single mux port standby | 1 scoped ACL on 1 interface; all other ports sit on the placeholder |
+| K distinct `IN_PORTS` groupings | K scoped ACLs |
```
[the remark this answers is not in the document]
lolyu: Right on both counts. Reworded: the row is now "K distinct signatures among the candidate ports → K scoped ACLs, plus the default ACL", and the paragraph below spells out why signatures and `IN_PORTS` sets are not interchangeable, using your overlap case — entries scoped to `{A, B}` and `{B, C}` give three signatures from two sets, because the port both name has a signature neither set produces alone. The default ACL being additional is now stated in the row itself rather than only in the two rows above. 19ad39b3.

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,598 @@
...
+
+```
+acl-index 3 (deny)  applied inbound on sw_if_index: 2          <-- Ethernet4 only
+
+sw_if_index  2 (Ethernet4):    input acl(s): 2, 3, 1           <-- deny present, permit last
+sw_if_index  3 (Ethernet8):    input acl(s): 2, <placeholder>, 1
+sw_if_index 32 (Ethernet124):  input acl(s): <placeholder>, 1
+```
+
+| Check | Expectation |
```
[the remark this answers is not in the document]
lolyu: Correct — REQ-5 and REQ-6 had checks added earlier in this review pass but REQ-7 was left with none, which is the weakest spot given counter fan-out is entirely new behaviour. Added the two cases you describe:

- an unscoped entry replicated across several scoped ACLs reports the **sum** of the traffic hitting it on every port, not one port's share;
- after a reprogram that changes the signatures, that entry's counters are neither doubled nor reset, i.e. placements are rebuilt exactly once.

The second one is what would catch a `acl_ace_cntr_info_clear()` regression in either direction. 19ad39b3.

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,598 @@
...
+| Toggle | Repeated standby↔active transitions converge, with no ACL or swindex leak in `show acl-plugin acl` |
+| All-active | ACL set matches the known-good all-active baseline |
+| Other tables preserved | After a toggle, every other table bound to the transitioning port is still bound, with unchanged content and position — REQ-5 |
+| Scoped read failure | A forced `IN_PORTS` read failure leaves the entry out of the dataplane rather than applying it everywhere — REQ-6 |
+
+### sonic-mgmt Coverage
+
+| Test | Relevance |
+|------|-----------|
+| `tests/generic_config_updater/test_dynamic_acl.py::test_gcu_acl_drop_rule_removal` | Three `DROP` rules on three different ports in one table; removes one and asserts that port forwards while the others stay blocked — direct coverage of REQ-4 |
```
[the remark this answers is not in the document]
lolyu: Verified against the test and you are right. `test_gcu_acl_drop_rule_removal` calls `dynamic_acl_create_three_drop_rules()` — whose assertions are control-plane only, checking `show acl rule` content — then removes the third rule and makes exactly one dataplane assertion, on `scale_port_indices[2]`, the port whose rule was just removed. The first two ports are never re-tested, so a regression that dropped their scoping at the same time would pass.

Relabelled as partial coverage with the specific gap named, rather than left claiming direct coverage. I also added an *Independent scopes* dataplane check to carry REQ-4 properly — two entries of one table scoped to different ports, each dropping only on its own port, and removing one leaving the other's port still dropping — so the requirement has a real home instead of resting on a test that does not assert it. 19ad39b3.

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,610 @@
...
+| signature | The ordered vector of ACE OIDs forming an applicable entry set — **including the unscoped entries**, interleaved at their priority positions. Identifies a scoped ACL, groups interfaces that can share one, and serves as the slot-reuse key across reprogramming |
+| dirty port | A port whose VPP binding may not match the recorded scope, because a previous bind/unbind failed or its interface could not be resolved. Forced through unbind/rebind on the next reprogram |
+
+---
+
+## Background
+
+SONiC ACL entries may be scoped to a subset of ingress ports via the SAI qualifier `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS`. The ACL table is bound to a broad set of ports, and each entry may narrow itself to specific ingress ports.
+
+The VPP-based virtual switch (`saivpp`) does not honour this qualifier. Creating or updating an ACL entry's `IN_PORTS` is accepted and **reported as successful**, but the port scope is discarded and the entry is programmed as if the qualifier were absent.
```
yue-fred-gao, line 98: with the proposed approach, what if IN_PORTS is used in egress ACL?

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,610 @@
...
+|----------|-------|-------|
+| MuxOrch | `IngressTableDrop` | One rule, `IN_PORTS` = all standby mux ports, match any, `DROP` |
+| PFC watchdog | `IngressTableDrop` (same table) | One rule **per queue id** (`Rule_PfcWdAclHandler_<qid>`), `IN_PORTS` = storming ports, match `TC=<qid>`, `DROP` |
+| GCU dynamic ACL | `DYNAMIC_ACL_TABLE` | Three `DROP` rules naming three *different* ports, alongside unscoped `FORWARD` rules |
+
+The built-in `DROP` table type declares exactly two match qualifiers — `SAI_ACL_TABLE_ATTR_FIELD_TC` and `SAI_ACL_TABLE_ATTR_FIELD_IN_PORTS` (`aclorch.cpp`) — which is what allows MuxOrch and the PFC watchdog to coexist in one table. `SAI_ACL_TABLE_ATTR_FIELD_IN_PORTS` is in fact a *mandatory* ingress match for this type, so an ingress `DROP` table always carries it. Two lossless queues storming on different ports likewise produce two rules with disjoint `IN_PORTS` — reachable on a box with no dual-ToR configuration at all.
+
+---
+
+## Design
```
yue-fred-gao, line 255: I feel a much simpler design is extend vpp acl with match on in_port. With sonic-ext plugin, we already store the original ingress interface. so matching should be very simple. The only thing is the patch cannot be upstreamed due to the dependency on sonic-ext

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,618 @@
...
+
+#### Encoding and Its Limit
+
+`in_sw_if_index = 0` means **any interface**. This is safe because `sw_if_index 0` is `local0`, which is never a data-plane port, so no legitimate rule needs to name it. It follows the convention `proto = 0` already uses, and avoids spending a separate valid flag.
+
+Only the low 16 bits take part in the match, since `lsb_of_sw_if_index` is a `u16` packed into the L4 `u64`. Widening it would mean growing `fa_session_l4_key_t` past the word it fits in, which is a far larger change than this feature warrants. An index above `0xffff` would otherwise be programmed as a *different* interface than the one named, with the CLI and API dump still reporting the interface the user asked for, so `acl_add_list()` rejects it with `VNET_API_ERROR_INVALID_SW_IF_INDEX` alongside the existing prefix and port-range validation. Both the binary API and the CLI go through that function, so neither path can install a rule whose scope is not representable. SONiC interface indices stay far below the limit, so this is a guard rather than an expected condition.
```
yue-fred-gao, line 380: When does vpp recycle sw_if_index? If interfaces are deleted and created many times, do they reused freed sw_if_index or keep growing?

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,618 @@
...
+| 1 | E1 | Ethernet8 | `DROP` |
+| 2 | E1 | Ethernet16 | `DROP` |
+| 3 | E2 | 0 (any) | `FORWARD` |
+| 4 | E3 | Ethernet8 | `DROP` |
+
+The ACL stays bound to all five ports. E1's three rules occupy indices 0–2, so E1's counter sums that range; E3's single rule is index 4.
+
+Priority is preserved because the fan-out is local: an entry's copies are emitted where the entry itself sat in the order. E2 remains between E1 and E3, so a packet arriving on Ethernet8 is tried against E1's copy for that port, then E2, then E3 — the table's order, unchanged.
+
+Note that Ethernet0 and Ethernet12 are named by nothing. They match only E2, because every other rule names an interface that is not theirs. No separate ACL is needed to arrange that.
```
yue-fred-gao: What do you mean by name? it is not bind, right?

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,618 @@
...
+
+#### Cases
+
+| Case | Example | Outcome |
+|------|---------|---------|
+| No entry carries `IN_PORTS` | Any ordinary L3 table | No fan-out; one rule per entry with `in_sw_if_index = 0`, exactly as before this design |
+| Port named by no entry | Ethernet0 above | Binds the same ACL as everyone else, and matches only the unscoped rules |
```
yue-fred-gao: I don't understand this. It sounds like it is still trying to use in_ports as bind? "Binds the same ACL as everyone else": what's the ACL here?

doc/vpp/vpp_acl_in_ports_HLD.md
```diff
@@ -0,0 +1,618 @@
...
+- **Rule count grows with the scope.** An entry naming N ports costs N rules. The ACL count is unchanged, but a large fan-out of otherwise identical rules can be folded by tuplemerge onto one collision chain that `split_partition()` cannot split, since it has no interface dimension. Matching stays correct; the cost is lookup efficiency at large N.
+- **Requires a patched VPP.** The field does not exist upstream, so `sonic-sairedis` and the `sonic-platform-vpp` VPP package must stay in step. Upstreaming the plugin change would remove this coupling.
+- **ip2me state remains table-keyed.** `m_ip2me_drop_tables` is keyed by table, not by interface, so the ip2me bypass is enabled for a table if *any* of its rules contains a deny, regardless of which ports those rules name. Re-keying it per interface is out of scope.
+- **Egress scoping is not addressed.** Only `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS` is implemented; `OUT_PORTS` remains unhandled.
```
yue-fred-gao: is IN_PORTS in egress ACL supported?

doc/vpp/vpp_acl_in_ports_HLD.md
lolyu: Good question, and the answer turned out to be worse than "unsupported" — thank you for asking it twice, since the first time I only tightened the wording instead of checking.

The field this design matches on is not inherently an *ingress* interface. `dataplane_node.c` picks the interface by direction:

```c
if (is_input)
  out_sw_if_index[ii] = vnet_buffer (b[ii])->sw_if_index[VLIB_RX];
else
  out_sw_if_index[ii] = vnet_buffer (b[ii])->sw_if_index[VLIB_TX];
```

Both end up in the same `lsb_of_sw_if_index` slot. It is only an ingress interface because of where the ACL is bound.

And the combination is reachable. SAI puts no stage constraint on `SAI_ACL_TABLE_ATTR_FIELD_IN_PORTS`, and saivpp binds egress groups for real — `aclBindUnbindPort()` maps `SAI_ACL_STAGE_EGRESS` to `is_input = false`. `fill_acl_rules()` had no stage check, so an egress table with `IN_PORTS` would have been fanned out and programmed normally, then matched against the **egress** port. Not an error, not a no-op: `drop what arrives on Ethernet4` would quietly become `drop what leaves on Ethernet4`, with the dump still printing the port the operator asked for.

So it is now rejected at programming time rather than mis-programmed — the same treatment as an out-of-range index. I considered silently dropping the scope instead and rejected it, because that reproduces the exact bug this design exists to fix: an over-broad rule applying to every bound port. New `Why This Is Ingress-Only` section, two precise restrictions replacing the vague one, and a test case. d4c8d33.

doc/vpp/vpp_acl_in_ports_HLD.md
lolyu: Answered in full in [the thread on line 94](https://github.com/sonic-net/SONiC/pull/2506#discussion_r3908920231) — short version: it is not supported, and until this change it was not *safely* unsupported either.

The bullet you commented on was the problem. It conflated two different things under one sentence: `OUT_PORTS` not being implemented (true and harmless) and what happens if `IN_PORTS` lands on an egress table (unaddressed, and silently mis-programmed — the matched slot holds the TX interface when the ACL is bound outbound). It is now split into two bullets that say exactly that, and the egress case is rejected at programming time instead of installed. d4c8d33.

doc/vpp/vpp_acl_in_ports_HLD.md
lolyu: Reused, not grown — and I had asserted the 16-bit ceiling was safe without actually establishing why, so this was worth pinning down.

`sw_if_index` is a pool index, not a counter:

```c
/* vnet_create_sw_interface_no_callbacks() */
pool_get (im->sw_interfaces, sw);
sw_if_index = sw - im->sw_interfaces;
```

Deletion does `pool_put()`, and `pool_get()` drains the free list before extending the vector (`vppinfra/pool.h`):

```c
uword n_free = vec_len (ph->free_indices);
if (n_free)
  {
    uword index = ph->free_indices[n_free - 1];   /* LIFO reuse */
```

So the high-water mark is the number of interfaces alive **at once**, not the number ever created. Churn cycles the same handful of indices forever, and hitting the limit would need 65 535 interfaces present simultaneously.

Your question also exposed a second-order case I had not thought about: if indices are recycled, could a rule outlive the interface it named and then apply to whatever inherits the index? A rule stores the index resolved at programming time, and **nothing invalidates it on interface deletion** — neither the VPP patch nor saivpp. What actually makes it safe is the qualifier type, not any cleanup: `IN_PORTS` is `@objects SAI_OBJECT_TYPE_PORT`, so it can only name front-panel ports, which are created at startup and not deleted at runtime. The churn-prone objects — sub-interfaces, tunnels, LAGs — are precisely the ones that cannot appear in `IN_PORTS`.

Since that is a real dependency rather than something the design enforces, it is now stated in Restrictions with the note that extending `IN_PORTS` to LAGs would require revisiting it. d4c8d33.

doc/vpp/vpp_acl_in_ports_HLD.md
lolyu: Correct, it is not bind — and the document was sloppy enough about this that the question was unavoidable. I was using "named" for the new per-rule match and "bound" for the unchanged VPP attachment without ever saying so, and then writing sentences like "named by nothing" that read like a binding statement.

The two relations are now defined explicitly in Definitions/Abbreviations:

- **bound** (port ↔ ACL) — the VPP attachment via `vpp_acl_interface_bind()`. Decides *which ACL a packet is evaluated against*. Set by the table s port list, **unchanged by this design**.
- **named** (entry → port) — a port in an entry s `IN_PORTS`, compiled into that rule s `in_sw_if_index`. Decides *whether a rule matches*, once the ACL is already being evaluated. Never attaches or detaches anything.

The line you flagged now says Ethernet0 and Ethernet12 appear in no entry s `IN_PORTS` list, and spells out that their binding is identical to every other port s — all five ports are bound to the same single ACL. d4c8d33.

doc/vpp/vpp_acl_in_ports_HLD.md
lolyu: No — nothing here uses `IN_PORTS` as a bind, and that row was written badly enough to suggest otherwise. It was carrying over phrasing from the v0.1 design, where scope *was* expressed by binding different ACLs to different ports. In v0.2 bindings are untouched.

"The ACL" is the table s single VPP ACL — one ACL per table, bound to every port the table covers, exactly as before this feature. Ethernet0 is bound to it like everyone else; it simply matches only the unscoped rules inside it, because every scoped rule carries an `in_sw_if_index` that is not Ethernet0.

The Cases table now opens by stating that "the ACL" means the table s single VPP ACL and that **no case in the table changes any binding**, and the row reads: *Stays bound to the table s ACL exactly as every other port does, and matches only the unscoped rules within it. Being unnamed is not a weaker binding — it simply means no scoped rule s `in_sw_if_index` equals this port.* d4c8d33.

doc/vpp/vpp_acl_in_ports_HLD.md
lolyu: This is what the design does now — v0.2 replaced the binding-layer partitioning of v0.1 with exactly the match you describe, and your comment is a large part of why. Apologies for not closing the loop on this thread at the time.

One correction on the downside you noted: it does **not** depend on sonic-ext, so the upstreaming objection does not apply. The ingress interface is already in the stock 5-tuple as `lsb_of_sw_if_index` (`fa_node.h`), where the ACL plugin puts it for session handling and then masks it off during rule matching. The patch only stops the mask from discarding it — it adds no field and reads no sonic-ext state. I checked that there is no `+` line introducing that field and that `fa_node.h` is unmodified against stock VPP.

So the cost is a handful of lines in the match paths, and the patch is upstreamable as-is.

doc/vpp/vpp_acl_in_ports_HLD.md
yue-fred-gao: I think SAI should enforce the integrity. It should remove the port-to-be-deleted from ACL IN_PORTS before deleting the port.

yue-fred-gao (approved):
looks good to me

PR #2527, opened 2026-08-25
HLD for ram2disk

gauravnagesh-msft:
HLD describing the `ram2disk` service which is used for devices which have `/var/log` in `tmpfs` but since tmpfs is volatile and logs are lost when the system shuts down, reboots, or loses power and logs are required for debugging, post-incident analysis and understanding failures across boot sessions. Therefore, need a service, called `ram2disk` (also known as `LogKeeper`), that:
   - Periodically copies logs from the tmpfs-backed /var/log to persistent disk in batches and in a sequential manner by using `rsync` as the utility for copying/backing up the logs. 
   - Performs a final synchronization during an orderly shutdown or reboot
   - Restores active logs early during boot so logging can continue across sessions
   - Retains previous boot sessions for later analysis

### DIFF
--- a/doc/ram2disk/Ram2Disk_HLD.md
+++ b/doc/ram2disk/Ram2Disk_HLD.md
@@ -1,0 +1,574 @@
+# ram2disk (LogKeeper) Design
+
+## Revision
+
+| Rev | Date | Author | Change Description |
+|---|---|---|---|
+| 0.1 | 7/29/2026 | Gaurav Nagesh | Initial Version |
+
+## Glossary
+
+| Term | Meaning |
+|---|---|
+| **EROFS** | Read-only filesystem error |
+| **tmpfs** | Temporary File System that is RAM based |
+| **rsync** | Remote synchronization utility |
+
+## Scope/Purpose
+
+`/var/log` is currently disk-backed. Frequent log appends, log rotation,
+compression, and metadata updates create sustained write activity. Moving
+`/var/log` to **tmpfs** reduces disk wear and lowers the likelihood of read-only
+filesystem (`EROFS`) issues caused by storage failures or filesystem degradation.
+
+The main drawback of using **tmpfs** is that it is volatile and logs are lost when the
+system shuts down, reboots, or loses power. But logs are required for debugging,
+post-incident analysis and understanding failures across boot sessions.
+
+We therefore need a service, called **ram2disk** (also known as **LogKeeper**), that:
+
+- Periodically copies logs from the tmpfs-backed `/var/log` to persistent disk in batches
+- Performs a final synchronization during an orderly shutdown or reboot
+- Restores active logs early during boot so logging can continue across sessions
+- Retains previous boot sessions for later analysis
+- Minimizes disk writes while preserving the same logical rotation history as
+  `/var/log`
+
+Ram2Disk relies on `rsync` utility program for the purpose of backing up logs to disk in a sequential manner in batches.
+
+## Architecture / Design
+
+### Overview
+
+**ram2disk** separates active, frequently written logs from persistent historical
+storage as follows:
+
+- `/var/log` is backed by **tmpfs** and is the live source used by logging services
+- `/var/log.disk` is the disk-backed destination used to preserve logs.
+- Periodic synchronization copies changes from **tmpfs** to disk.
+
+**Example view of system when ram2disk is enabled:**
+
+```bash
+admin@sonic:~$ df -h
+Filesystem      Size  Used Avail Use% Mounted on
+tmpfs           988M  116K  988M   1% /var/log
+/dev/loop1      3.9G  3.1M  3.7G   1% /var/log.disk
+```
+
+```bash
+sudo losetup -l /dev/loop1
+NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE                   DIO LOG-SEC
+/dev/loop1         0      0         1  0 /host/disk-img/var-log.ext4   0     512
+```
+
+> **Note:** When `/var/log` is not mounted on **tmpfs**, i.e., it is already
+> disk-backed, then this is already the final copy and `/var/log.disk` will not
+> exist and work as normal.
+
+### Storage Layout on Disk
+
+```bash
+/var/log.disk/
+|-- current_session/               # Live mirror of /var/log for this boot
+|   |-- syslog
+|   |-- syslog.1
+|   |-- syslog.2.gz
+|   |-- frr/bgpd.log
+|   `-- ...
+|-- session_<timestamp1>/          # Archived previous boot1
+|-- session_<timestamp2>/          # Archived previous boot2
+`-- .boot_id
+```
+
+- **`current_session/`** is the persistent mirror of the current boot's
+  `/var/log`. Its active and rotated files are kept aligned with the tmpfs source
+  during periodic synchronization.
+- **`session_<timestampN>`** stores an immutable log snapshot from a past completed
+  boot. Multiple session directories may be retained for historical analysis.
+- **`.boot_id`** records the boot associated with **`current_session/`**. It allows
+  **ram2disk** to detect a new boot and is also used to derive the archived session
+  timestamp.
+
+### Why Each Boot Needs a Separate Session Directory
+
+Logrotate archive numbering is meaningful only within the current log history. A
+new boot starts with a fresh tmpfs-backed `/var/log` and begins creating files such
+as:
+
+```bash
+
+Before Reboot / Shutdown :
+Logs are in sync between the two
+
+tmpfs /var/log/                                 Disk /var/log.disk/
+-----------------                               ----------------
+syslog                                           syslog
+syslog.1                                         syslog.1
+syslog.2.gz                                      syslog.2.gz
+...                                              ...
+...                                              ...
+syslog.100.gz                                    syslog.100.gz
+
+
+
+After Reboot / Shutdown :
+tmpfs will be initially empty and slowly start to fill up as shown below 
+
+
+Example1 : sync here will overwrite disk files 
+tmpfs /var/log/                                  Disk /var/log.disk/
+---------------                                 --------------
+syslog                                           syslog
+                                                 syslog.1
+                                                 syslog.2.gz
+                                                 ...
+                                                 ...
+                                                 syslog.100.gz
+
+
+Example2 : sync here will overwrite disk files 
+tmpfs /var/log/                                  Disk /var/log.disk/
+---------------                                 --------------
+syslog                                           syslog
+syslog.1                                         syslog.1
+                                                 syslog.2.gz
+                                                 ...
+                                                 ...
+                                                 syslog.100.gz
+
+
+Example3 : sync here will overwrite disk files 
+tmpfs /var/log/                                  Disk /var/log.disk/
+---------------                                 --------------
+syslog                                           syslog
+syslog.1                                         syslog.1
+syslog.2.gz                                      syslog.2.gz
+                                                 ...
+                                                 ...
+                                                 syslog.100.gz
+
+                                          
+```
+
+The persistent disk may already contain archives from a much older history,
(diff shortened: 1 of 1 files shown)


doc/ram2disk/Ram2Disk_HLD.md
```diff
@@ -0,0 +1,574 @@
...
+
+## Architecture / Design
+
+### Overview
+
+**ram2disk** separates active, frequently written logs from persistent historical
+storage as follows:
+
+- `/var/log` is backed by **tmpfs** and is the live source used by logging services
+- `/var/log.disk` is the disk-backed destination used to preserve logs.
```
judyjoseph, line 47: this is a new destination DIR which we will create  ? for disk backed storage

doc/ram2disk/Ram2Disk_HLD.md
```diff
@@ -0,0 +1,574 @@
...
+/dev/loop1      3.9G  3.1M  3.7G   1% /var/log.disk
+```
+
+```bash
+sudo losetup -l /dev/loop1
+NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE                   DIO LOG-SEC
+/dev/loop1         0      0         1  0 /host/disk-img/var-log.ext4   0     512
+```
+
+> **Note:** When `/var/log` is not mounted on **tmpfs**, i.e., it is already
```
judyjoseph, line 65: Can the user control this, Is this service enabled by default if the /var/log is in tmpfs ?

doc/ram2disk/Ram2Disk_HLD.md
```diff
@@ -0,0 +1,574 @@
...
+names, and rotation replay could no longer reliably keep tmpfs and disk aligned.
+
+Copying every historical archive from disk back into tmpfs at boot does not make
+sense as it increases boot-time work and mixes historical logs with the new boot's
+active rotation sequence.
+
+For these reasons, the previous boot's **`current_session/`** directory is moved to
+a separate **`session_<timestamp>/`** directory. Only active logs are restored to
+tmpfs. The new boot then starts with a fresh **`current_session/`**, while older
+sessions remain available on disk for analysis without interfering with current
```
judyjoseph, line 167: We would need to have a separate log rotate policy for the previous boot sessions ? as that also could get added up and taking more disk space ?

gauravnagesh-msft: Yes, we will need a cleanup mechanism for these previous boot session, maybe via logrotate. Probably not mentioned in detial about this in HLD but currently the cleanup is already implemented as part of the ram2disk.sh service in `cleanup_disk` function here : [Code](https://github.com/sonic-net/sonic-buildimage/pull/29209/changes#diff-f224cf713c42421b4dac14066b220b4f7e359ad661c6c1af596f1b06f7d802c6R191) 

In this the oldest boot session directories are removed first whenever we hit a threshold of 80% consumption of disk space. Threshold is configurable.

doc/ram2disk/Ram2Disk_HLD.md
gauravnagesh-msft: Yes, `/var/log.disk` will be a new directory which will be created for disk backed storage for devices on which tmpfs is enabled for logging.

`/var/log` will not have any changes. It will continue to exist irrespective of whether the device is tmpfs enabled or not and its functionality will be the same too accordingly.

doc/ram2disk/Ram2Disk_HLD.md
gauravnagesh-msft: The `ram2disk` service will be enabled by default if the /var/log is in tmpfs. And since its a systemd service, it can be controlled using normal `systemctl` commands to enable or disable it.

doc/ram2disk/Ram2Disk_HLD.md
```diff
@@ -0,0 +1,574 @@
...
+"/var/log/frr/bgpd.log" 2026-7-31-3:0:0
+"/var/log/apt/history.log" 2026-7-31-3:0:0
+"/var/log/telemetry.log" 2026-7-31-3:0:0
+"/var/log/alternatives.log" 2026-7-31-3:0:0
+"/var/log/swss/sairedis.rec" 2026-7-31-3:0:0
+```
+
+`/dev/shm/logrotate/rotation_counts`:
+
+```bash
```
judyjoseph, line 285: rotation_counts as I understand is clculated by parsing the log. What is the above file used for "/dev/shm/logrotate/status"

gauravnagesh-msft: `/dev/shm/logrotate/status` is an existing status file used by logrotate. I’ve included it here only as an example to illustrate the existing approach. Similarly, I plan to create a new file `/dev/shm/logrotate/rotation_counts`, to keep track of the number of rotations.

doc/ram2disk/Ram2Disk_HLD.md
```diff
@@ -0,0 +1,574 @@
...
+- Run **rsync** from `/var/log/` to **`current_session/`**
+- Remove old sessions or archives when disk usage is too high.
+
+#### Boot Restore
+
+- Restore active, non-rotated logs from the previous **`current_session/`** into
+  tmpfs.
+- Rename the previous **`current_session/`** to a timestamped session directory
+  **`session_<timestamp1>/`**
+- Create a fresh **`current_session/`** for the new boot.
```
dgsudharsan, line 185: Will this duplicate the non-rotated logs in current_session and the timestamped session, since we restore it and don't delete it before renaming session_timestamp1?

dgsudharsan:
## HLD review

The problem statement and the Approach 2 decision (treat logrotate as the source of truth for rotations) are sound. The document is not yet a complete contract for the implementation ([sonic-buildimage#29209](https://github.com/sonic-net/sonic-buildimage/pull/29209)). Please address the items below.

### 1. Enablement / when the service actually runs

The HLD never states the start criteria. The implementation installs units and the logrotate divert on **every** image, but sync/restore only run when `/var/log.disk` is a mountpoint (`ConditionPathIsMountPoint=/var/log.disk`).

That mount is created in initramfs when `logs_inram` is true **and** `host/disk-img/var-log.ext4` exists.

`logs_inram` becomes true when:

- kernel cmdline has `logs_inram=on` (that spelling, not `logs_in_ram`), or
- Switch-BMC platforms force tmpfs via `switch_bmc=1` in `platform_env.conf` (even without the cmdline).

Please document this gate, the cmdline name, and the BMC exception. The current Scope section reads as if every SONiC box is moving `/var/log` to tmpfs.

The logrotate divert/wrapper is **not** gated and still runs on disk-backed `/var/log` images. Either restrict the divert to logs-in-ram platforms or justify a platform-wide ABI change.

### 2. Initramfs / loop image is missing

The `df` example shows `/dev/loop1` → `var-log.ext4` on `/var/log.disk`, but there is no section on:

- `union-mount` keeping `var-log.ext4` when `logs_inram` is set (today that file is deleted)
- `fsck` + loop mount
- image size and flash wear of the loop file

That is half of the feature and should be in the HLD.

### 3. `current_session/` vs `current/`

Layout and most prose use `current_session/`. The producer/consumer diagram, some restore text, and the implementation use `current/`. Please pick one name and use it everywhere.

### 4. Restore boot ordering is incomplete

The HLD only says restore runs before `rsyslog.service` and `syslog.socket`. Journald starts earlier.

In the implementation, `WantedBy=multi-user.target` does not pull the unit into the early boot graph, so `Before=journald` is not actually enforced. The HLD needs the exact boot edge: which target pulls the unit, and whether journald must wait.

### 5. Unclean power loss is not designed

The purpose includes “loses power”, but the design only guarantees an orderly `pre-shutdown` plus a 10-minute timer. Counters live in `/dev/shm` and die on crash.

Please state explicitly:

- the last sync interval of logs may be lost on power loss
- if logrotate ran and ram2disk did not, replay is lost and the next rsync can rewrite archives

That is acceptable only if it is written down.

### 6. Warm-reboot and fast-reboot

Warm-reboot and fast-reboot are not mentioned. They should be.

**ASIC / dataplane warm or fast path:** not impacted. ram2disk does not touch SWSS, syncd, or `WARM_RESTART_TABLE`.

**Host logging across kexec:** impacted, and incomplete.

Warm/fast use the same `fast-reboot` script and **kexec**. They do **not** go through a full systemd shutdown, so they do **not** close `syslog.socket` → `host_umount.sh` → `ram2disk.sh pre-shutdown`.

What they do today if `/var/log` is tmpfs is copy recent files to `/host/logs_before_reboot` (last 30 minutes, maxdepth 2), then kexec.

After kexec:

- tmpfs `/var/log` is empty again
- `/var/log.disk` is remounted if `logs_inram` + ext4
- `ram2disk-restore` (if it runs) archives the last disk `current/`, which may be up to ~10 minutes stale

Result: two incomplete pictures — ram2disk sessions without the last interval, and `/host/logs_before_reboot` with only recent files.

Please specify what ram2disk should do on `SONIC_BOOT_TYPE=warm` / `fast-reboot`: invoke `pre-shutdown` before kexec, or keep relying on `/host/logs_before_reboot`.

### 7. Boot-time impact

Only when `/var/log.disk` is mounted. On a normal disk `/var/log`, units skip; boot time should not change except the always-on logrotate wrapper (small).

When ram2disk is active, expect extra time:

1. **Initramfs** — today `logs_inram` deletes `var-log.ext4`. The implementation keeps it, runs `fsck.ext4`, and loop-mounts it on **every** boot, including post-kexec. A multi-GB image can add a noticeable fsck.
2. **`ram2disk-restore.service`** — rsync of active (non-rotated) logs disk → tmpfs, then `mv current/ → session_*`. If `Before=journald/rsyslog` is honored, syslog start waits on that rsync. Large `syslog` / `sairedis.rec` files can cost seconds. If the unit only joins `multi-user.target`, restore may run late and race with rsyslog (item 4).
3. **Periodic timer** — first fire is `*:05/10` after boot; not on the critical path.
4. **Cold reboot shutdown** — orderly reboot runs `pre-shutdown` rsync (no cleanup). That adds shutdown time, not next-boot time. Warm/fast skip this (item 6).

Please call this out so reviewers do not assume the feature is free on the boot path.

### 8. Other design holes to add

| Topic | Why it matters |
|---|---|
| Disk watermark | Code uses 80%; HLD says “configured limit” with no default or floor on the last session. |
| Upgrade / rollback | What happens if the wrapper is overwritten, or `logrotate.real` is missing. |
| Test plan | Unit tests for counters/replay; DUT cases for tmpfs vs disk `/var/log`; warm/fast. |
| Security | Persistent copy of syslog (secrets, TACACS). Cleanup `rm -rf session_*`. |
| CLI | `--help` example is broken (`ram2disk.shram2disk.sh --help`); the script has no `--help`. |
| CONFIG_DB / knob | Confirm there is none; enablement is cmdline + mount only. |

### 9. Consistency / editorial

- Glossary **EROFS** is unused as a requirement; the feature reduces writes, it does not prevent EROFS.
- Dual name **ram2disk / LogKeeper** is unused after the title.
- Approach 1 bullet is broken: “Does not modify… **contents from disk.**”
- Typos: `logrotate-warpper.sh`, `ram2disk.sevice`, `logroate`, `moving forword`.
- Flow diagram writes `/var/log.disk/current/` while the same page says `current_session/`.
- Restore text says previous `current/` in one place and `current_session/` in another.

### 10. Alignment with sonic-buildimage#29209

| HLD | Implementation |
|---|---|
| `current_session/` | `current/` |
| Restore before rsyslog | Also `Before=journald`; `WantedBy=multi-user.target` |
| Divert so all `logrotate` invocations count | Divert on every image, including non-tmpfs |
| Sync only if `/var/log.disk` is mounted | Same `ConditionPathIsMountPoint` |
| Cleanup when usage is high | `DISK_USED_MAX_PCT=80` |
| Initramfs / `logs_inram` | Implemented, not described |
| Warm / fast reboot | Not described; pre-shutdown skipped on kexec |

Please make the HLD the contract (directory name, `logs_inram` / mount gate, boot dependencies, crash-window and warm/fast behavior) before treating the implementation PR as complete against this document.

PR #2533, opened 2026-08-31
MAC over FPM HLD documentation

pbrisset:
# MAC Learning Synchronization over the FPM Channel HLD

## Summary

This HLD proposes moving MAC (bridge FDB) state synchronization between SONiC and FRRouting from the Linux kernel/netlink path to the existing FPM (Forwarding Plane Manager) channel. This eliminates race conditions in EVPN Multihoming scenarios where a MAC can have both local (hardware-learned) and remote (BGP EVPN) sources simultaneously.

## Motivation

Today, SONiC uses the Linux kernel bridge FDB as an inter-process message bus for MAC state between SONiC and FRR zebra. This creates problems in EVPN-MH deployments:

- **The kernel bridge FDB can only hold one effective entry per MAC**, even when zebra tracks multiple sources (local hardware + remote EVPN)
- **Race conditions occur** when one source is withdrawn but the other remains (e.g., BGP route withdrawal deletes the kernel entry, forcing SONiC to re-install the local MAC)
- **Requires an out-of-tree kernel patch** (`proto hw`/`RTPROT_HW`) for self-notification filtering that is difficult to upstream

## Proposed Solution

Synchronize MAC add/delete/update events directly over the existing FPM channel:

- **Local MACs**: `ASIC → fdbOrch → fdbsyncd → fpmsyncd =[FPM]=> zebra`
- **Remote MACs**: `zebra =[FPM]=> fpmsyncd → fdbOrch → ASIC`

**Key benefits:**
- Eliminates kernel bridge FDB as a relay for MAC synchronization
- Removes dependency on the `proto hw`/`RTPROT_HW` kernel patch
- Enables per-source MAC tracking in both zebra and fdbOrch, eliminating churn
- Aligns MAC synchronization with existing route/RMAC FPM architecture
- Backward compatible: feature can be enabled/disabled via configuration

## Scope

- EVPN-VXLAN with Multihoming deployments
- Repositories: `sonic-swss` (fdbsyncd, fpmsyncd, fdbOrch) and `FRRouting/frr` (zebra `dplane_fpm_sonic` plugin)
- No DB schema changes; existing `STATE_FDB_TABLE` and `APP_VXLAN_FDB_TABLE` contracts preserved
- Includes warm/fast reboot and FPM reconnect reconciliation design

## Related Work

- Complements existing FPM transport for routes, nexthop groups, and L3VNI RMACs
- Builds on zebra's external-learn MAC model (`--kernel-mac-ext-learn`)

### DIFF
--- a/doc/mac/mac-over-fpm.md
+++ b/doc/mac/mac-over-fpm.md
@@ -1,0 +1,477 @@
+# MAC Learning Synchronization over the FPM Channel HLD
+
+## Table of Contents
+
+- [Revision](#revision)
+- [Scope](#scope)
+- [Definitions/Abbreviations](#definitionsabbreviations)
+- [Overview](#overview)
+- [Requirements](#requirements)
+- [Architecture Design](#architecture-design)
+- [High-Level Design](#high-level-design)
+- [SAI API](#sai-api)
+- [Configuration and management](#configuration-and-management)
+- [Warmboot and Fastboot Design Impact](#warmboot-and-fastboot-design-impact)
+- [Memory Consumption](#memory-consumption)
+- [Restrictions/Limitations](#restrictionslimitations)
+- [Testing Requirements/Design](#testing-requirementsdesign)
+- [Open/Action items](#openaction-items)
+
+### Revision
+
+| Rev | Date       | Author            | Change Description |
+|:---:|:----------:|:------------------|:-------------------|
+| 1.0 | 08/31/2026 | Patrice Brissette (Cisco), Tamer Ahmed (Microsoft) | Initial version |
+
+### Scope
+
+This document describes a high-level design for synchronizing MAC (bridge FDB) state between
+SONiC and FRRouting (FRR) over the **FPM channel** instead of the Linux kernel / netlink path.
+The scope is EVPN-VXLAN with Multihoming (MH), where a MAC can be learned locally in hardware and,
+at the same time, remotely through BGP EVPN.
+
+The document covers the motivation, the target architecture, the affected SONiC and FRR components,
+and the behavioral impact (configuration, warm/fast reboot, scale). It intentionally stays at a
+high level: numeric identifiers and field-level encoding are left to the implementation PRs. The
+HLD still defines the required message **semantics** — direction, local vs remote source, replay
+generation, and end-of-replay signaling — because those are architectural contracts, not
+implementation details.
+
+### Definitions/Abbreviations
+
+| Term | Definition |
+|------|------------|
+| FDB | Forwarding Database (MAC table) |
+| FPM | Forwarding Plane Manager — netlink-over-TCP channel between FRR and SONiC |
+| fdbOrch | SONiC orchagent component that owns the ASIC FDB via SAI |
+| fdbsyncd | SONiC daemon syncing FDB state between STATE_DB/APPL_DB and the kernel |
+| fpmsyncd | SONiC daemon that terminates the FPM channel from FRR |
+| zebra | FRR routing/dataplane manager daemon |
+| dplane_fpm_sonic | SONiC-specific zebra FPM dataplane plugin (FPM client) loaded by SONiC, derived from upstream `dplane_fpm_nl` |
+| EVPN | Ethernet VPN |
+| MH | EVPN Multihoming |
+| RT-2 | EVPN Route Type 2 (MAC/IP advertisement) |
+| VTEP | VXLAN Tunnel Endpoint |
+| VNI | VXLAN Network Identifier |
+| SAI | Switch Abstraction Interface |
+| RMAC | Router MAC (L3VNI) |
+
+### Overview
+
+In SONiC today, MAC learning is hardware-driven. The ASIC reports learned/aged MACs through SAI to
+**fdbOrch**, which records them and hands local MACs to **fdbsyncd**. fdbsyncd mirrors them into the
+Linux kernel bridge FDB, and FRR **zebra** learns them from the kernel through netlink. Remote MACs
+learned from BGP EVPN travel the opposite way: zebra programs them into the kernel bridge, fdbsyncd
+picks them up from netlink, and fdbOrch installs them into the ASIC.
+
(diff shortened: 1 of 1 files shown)


doc/mac/mac-over-fpm.md
```diff
@@ -0,0 +1,477 @@
...
+A concrete example: zebra holds a MAC that is both BGP-EVPN-learned and locally learned. The BGP
+route is withdrawn, so zebra deletes the MAC through the kernel. The kernel removes the single
+entry even though the MAC is still present in hardware, and SONiC must then detect this and
+re-install the local MAC — unnecessary churn and a race window.
+
+Two distinct kernel-side mechanisms are involved, and it is important to separate them:
+
+- **External-learn MAC model (`--kernel-mac-ext-learn`).** zebra installs MACs as externally learned
+  and suppresses the kernel's own MAC learning/aging, so hardware and the EVPN control plane — not
+  the kernel — drive the MAC state machine. This model is correct and is **retained** by this design.
```
tahmed-dev, line 85: Small thing: "This model is correct" might be worth softening. It is correct in the sense that hardware and the control plane own the state machine, but it also means nothing ages a MAC out locally, which is what makes the multihomed withdrawal case tricky. A sentence noting there is no local age-out would tell readers what the model implies. Up to you.

doc/mac/mac-over-fpm.md
```diff
@@ -0,0 +1,477 @@
...
+typically lower because the duplicate kernel-echo/re-install churn is removed.
+
+### Restrictions/Limitations
+
+- The primary target — and the focus of the requirements, flows, and tests in this HLD — is
+  EVPN-VXLAN Multihoming with hardware-based MAC learning.
+- The FPM-based method is intended to generalize to **all FDB modes** (e.g. MCLAG, single-homing,
+  non-EVPN) using the same path, and the design does not preclude an all-FDB deployment. Extending
+  coverage to those modes — in particular MCLAG coexistence — still needs validation and is tracked
+  in Open/Action items, so it is a design goal here, not a validated claim.
```
tahmed-dev, line 437: This defers to Open/Action items, and the ToC links to it as well, but I do not see the section in the file. Possibly lost in a trim.

doc/mac/mac-over-fpm.md
```diff
@@ -0,0 +1,477 @@
...
+- Requires matching support on both SONiC and FRR; mixed versions fall back to the kernel path.
+- The zebra inbound-MAC-over-FPM handling is new FRR work. It lands in `dplane_fpm_sonic`, which is
+  SONiC-specific and carried in the SONiC FRR patch series rather than upstream FRR, so it does not
+  depend on upstream acceptance. Any change that reaches common zebra EVPN code outside the plugin
+  does need to be upstreamed.
```
tahmed-dev, line 442: Would it be worth capturing this as a limitation? Not caused by anything here, just inherited from keeping ext-learn.

The scenario we keep hitting:

1. A host is multihomed to PE1 and PE2 on the same ES. Both learn the MAC locally, both advertise RT-2, and each also holds it peer-synced from the other.
2. The host stops forwarding through PE1, say on a link failure or a move.
3. PE1 drops its local source and marks the MAC local-inactive. At that point it should either hold it as a peer-proxy entry and withdraw when the hold timer expires, or withdraw when PE2 does.
4. Neither happens. The MAC-only RT-2 carries the proxy indication, but the companion MAC-IP RT-2 does not, because the host's neighbour entry never ages while the peer keeps probing it. Those MAC-IP paths get counted when deciding whether the MAC is active on the peer, so there is always at least one non-proxy path and the peer looks permanently active.
5. The withdrawal hand-off is gated on that peer-active condition clearing, so it never runs. The hold timer gets armed by the proxy advertisement and then cancelled by the non-proxy one right behind it, so it never fires either.
6. PE1 keeps the MAC programmed and advertised indefinitely. Both PEs sit in the same state, so the network never converges on where the host actually is, and a later move is not reflected.

The gate is `ATTR_ES_PEER_ACTIVE` in the bgpd MH sync path and the matching check on the zebra side. Structure it however you prefer, or tell me it belongs somewhere else entirely.

doc/mac/mac-over-fpm.md
```diff
@@ -0,0 +1,477 @@
...
+- zebra: verify the new inbound FPM MAC handler updates the EVPN MAC state and triggers RT-2.
+- fdbOrch: existing FDB unit tests (local learn/age, local+remote coexistence, remote withdrawal) must
+  pass unchanged, proving the DB contract is stable. A local learn over an existing EVPN-advertised MAC must retain the control-plane contribution rather than replacing it; a local age-out must fall back to the remote contribution without deleting the ASIC entry; a remote withdraw with the MAC still local
+  must leave the local entry untouched.
+
+**Integration / system tests**
+
+- Reuse the FPM-based EVPN-MH topotest already added in the FRR work (no kernel
+  dependency) to cover hardware add, age/delete, RT-2 add/withdraw, and the local+remote combined
+  cases; SONiC VS tests for MH failover.
```
tahmed-dev, line 467: This says the topotest covers age/delete. It may be worth calling out the multihomed local-withdrawal case explicitly, and specifically with the host acting as an ARP/ND speaker. Without that condition a test can pass against a silent host and miss the case in the Restrictions comment.

tahmed-dev (commented):
Hi Patrice, thanks for writing this up. The ownership split reads well and I agree with keeping the external-learn model.

One thing I would like your take on. Because ext-learn does no age-out of its own, removing a local MAC on a multihomed ES rests entirely on the peer withdrawing it or on the MAC hold timer expiring. Where the multihomed host is also an ARP/ND speaker we have seen neither happen, and the MAC just stays. Details in a comment on the Restrictions section.

Nothing in this design causes it, but since the design keeps the model as is, it inherits the behaviour, so it might be worth a mention for readers.

Also one editorial thing: the Open/Action items section is linked from the ToC and referenced in Restrictions, but I cannot find it in the file.

Happy to be wrong on any of this, and entirely your call how you want to handle it.

PR #2534, opened 2026-09-01
Add hld for secure_upgrade_certificate_management_cli

rameshraghupathy:
#### What I did

Added the High-Level Design for a platform-agnostic SONiC Secure Boot Certificate Management CLI.

The HLD defines the architecture and management model for Secure Boot certificate and authenticated-variable lifecycle operations for the standard UEFI Secure Boot variables:

- Platform Key (`PK`)
- Key Exchange Key (`KEK`)
- allowed signature database (`db`)
- revoked/forbidden signature database (`dbx`)

The design defines:

- platform-independent SONiC Secure Boot CLI
- platform-neutral Secure Boot backend contract
- authenticated-variable update model
- Secure Boot mode and policy reporting
- vendor, customer, and unified Secure Boot variable views
- protected persistent storage abstraction
- security and authorization model
- unit and system test requirements
- UEFI-compatible lifecycle flow

The upstream SONiC CLI communicates with the platform through the platform-neutral backend entry point:

```text
/usr/sbin/secure-boot-backend
```

Related implementation:

[sonic-utilities PR #4830 - Add Secure Boot show and config CLI support](https://github.com/sonic-net/sonic-utilities/pull/4830)

#### How I did it

Defined the design in three logical layers:

1. **SONiC Secure Boot CLI**
   - Provides the platform-neutral user interface.

2. **Secure Boot backend contract**
   - Defines stable operations and structured responses between SONiC and a platform implementation.

3. **Platform implementation**
   - Implements Secure Boot variable access using the platform's native secure-variable mechanism and protected persistent storage.

The HLD preserves standard UEFI authenticated-variable semantics while allowing different platforms to use different secure storage implementations underneath the SONiC interface.

The CLI does not directly modify protected storage and does not become the cryptographic authorization boundary.

The backend remains responsible for:

- authenticated-variable validation
- ownership and mode-policy enforcement
- replay/timestamp protection where applicable
- persistent storage protection
- authorization of Secure Boot variable updates

Private signing keys remain off the target device.

The HLD also defines the corresponding CLI model:

show secure-boot status
show secure-boot mode
show secure-boot keys
show secure-boot key <PK|KEK|db|dbx> [--store vendor|customer|unified]

config secure-boot certificate update <PK|KEK|db|dbx> <auth-var-file> [--operation append|update|remove]

### DIFF
--- a/doc/secure_upgrade/secure_boot_certificate_management.md
+++ b/doc/secure_upgrade/secure_boot_certificate_management.md
@@ -133,11 +133,11 @@ The following sequence shows the end-to-end Secure Boot variable-management flow
     "primaryTextColor": "#102A43",
     "primaryBorderColor": "#2F6BBD",
 
-    "secondaryColor": "#E7F7ED",
+    "secondaryColor": "#E8F8EE",
     "secondaryTextColor": "#163A24",
     "secondaryBorderColor": "#3A8F5B",
 
-    "tertiaryColor": "#FFF2CC",
+    "tertiaryColor": "#FFF1CC",
     "tertiaryTextColor": "#5C4400",
     "tertiaryBorderColor": "#C99200",
 
@@ -145,8 +145,7 @@ The following sequence shows the end-to-end Secure Boot variable-management flow
     "actorBorder": "#2F6BBD",
     "actorTextColor": "#102A43",
 
-    "actorLineColor": "#6B7C93",
-
+    "actorLineColor": "#75869E",
     "signalColor": "#34495E",
     "signalTextColor": "#102A43",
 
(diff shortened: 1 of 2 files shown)


doc/secure_upgrade/secure_boot_certificate_management.md
```diff
@@ -0,0 +1,656 @@
...
+### 1.5. <a name='Requirements'></a>Requirements
+
+#### Functional requirements
+
+| ID | Requirement |
+| :--- | :--- |
+| SB-CERT-REQ-1 | SONiC shall provide show commands for Secure Boot backend state. |
+| SB-CERT-REQ-2 | SONiC shall provide a show command for Secure Boot mode/policy when supported. |
+| SB-CERT-REQ-3 | SONiC shall show state for `PK`, `KEK`, `db`, and `dbx`. |
+| SB-CERT-REQ-4 | SONiC may distinguish vendor/platform and customer stores when the backend exposes that distinction. |
```
saiarcot895, line 87: Is there an example of this, where a backend distinguishes between vendor and customer?

doc/secure_upgrade/secure_boot_certificate_management.md
```diff
@@ -0,0 +1,656 @@
...
+config secure-boot certificate update dbx /host/secureboot/dbx-update.auth --operation append
+```
+
+#### 1.7.6. <a name='BackendContract'></a>Backend Contract
+
+The SONiC CLI communicates with a platform-provided Secure Boot backend through the platform-neutral executable interface:
+
+```text
+/usr/sbin/secure-boot-backend
+```
```
saiarcot895: Does each platform really need to define/provide a platform-specific backend? There is a UEFI standard for this that can work for amd64; for arm64, if uboot is being used, then uboot itself supports enabling a UEFI-compatible interface for this.

I recommend having a generic file that can be used if a platform doesn't use a platform-specific interface.

doc/secure_upgrade/secure_boot_certificate_management.md
rameshraghupathy: @saiarcot895 Yes. One example is a platform secure-variable implementation that maintains
separate platform/vendor-owned and customer-owned copies of the Secure Boot
variables, for example PKVendor/PKCustomer, KEKVendor/KEKCustomer,
dbVendor/dbCustomer, and dbxVendor/dbxCustomer.  The distinction is not mandated by UEFI and SONiC does not require every platform to implement separate stores. This requirement is intentionally optional.  In our case it is: PKCisco       / PKCustomer
KEKCisco      / KEKCustomer
dbCisco       / dbCustomer
dbxCisco      / dbxCustomer

doc/secure_upgrade/secure_boot_certificate_management.md
rameshraghupathy: @saiarcot895  Agreed. The intent was not to require every platform to implement a unique backend.

I will clarify the HLD so that /usr/sbin/secure-boot-backend  (A platform supporting this CLI shall provide this entry point.)   is the stable SONiC-facing contract, while its implementation may be:

1. a generic SONiC backend using the standard UEFI variable interface, or
2. a platform-specific backend when the platform requires additional secure
   storage, policy, hardware RoT, or vendor/customer-store handling.

This allows amd64/arm64 platforms exposing standard UEFI variable services to
use a common implementation while preserving an override point for platforms
with specialized secure-variable mechanisms.

@saiarcot895  Conceptually I can do something like the following:
1. If a platform backend is explicitly registered:
       use platform backend

2. Else if standard UEFI variable services are available:
       use SONiC generic UEFI backend

3. Else:
       report Secure Boot management unsupported

Issue #289, opened 2018-11-21
ECMP hashing is not working with incremental Source / destination IPV4 Address


chaitanc:
Description:
=======
ECMP hashing is not working with incremental Source / destination IPV4 Address

Procedure:
=======
1) Configure IPV4 addresses for 3 interfaces using below command
```
root@sonic-testing:/home/admin# config interface Ethernet4 ip add 10.10.10.1/24
root@sonic-testing:/home/admin#
root@sonic-testing:/home/admin#
root@sonic-testing:/home/admin# ifconfig Ethernet4
Ethernet4: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 9100
        inet 10.10.10.1  netmask 255.255.255.0  broadcast 0.0.0.0
        inet6 fe80::eef4:bbff:fefd:47e5  prefixlen 64  scopeid 0x20<link>
        ether ec:f4:bb:fd:47:e5  txqueuelen 1000  (Ethernet)
        RX packets 11  bytes 818 (818.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 6136  bytes 1414306 (1.3 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

root@sonic-testing:/home/admin# ip route show
default via 10.130.82.1 dev eth0
2.5.6.0/24 dev Ethernet16 proto kernel scope link src 2.5.6.3 linkdown
6.6.6.0/24 dev Ethernet12 proto kernel scope link src 6.6.6.1
10.10.10.0/24 dev Ethernet4 proto kernel scope link src 10.10.10.1
10.130.82.0/25 dev eth0 proto kernel scope link src 10.130.82.30
10.150.20.0/24 dev Ethernet4 scope link
50.50.50.0/24 dev Ethernet20 proto kernel scope link src 50.50.50.1
240.127.1.0/24 dev docker0 proto kernel scope link src 240.127.1.1 linkdown
```
Note: Followed same command configuration for Ethernet12 and Ethernet20

2) Configured ECMP route for 2 next hops as below
```
root@sonic-testing:/home/admin#ip route add 9.9.0.0/24 nexthop via 10.10.10.2 nexthop via 6.6.6.4

root@sonic-testing:/home/admin# ip route show
default via 10.130.82.1 dev eth0
2.5.6.0/24 dev Ethernet16 proto kernel scope link src 2.5.6.3 linkdown
[...]

PR #2539, opened 2026-09-03
[vpp] CoPP dataplane enablement HLD

nhegde-microsoft:
## Why I did it

HLD for CoPP dataplane enablement on SONiC-VPP: [sonic-buildimage#25801](https://github.com/sonic-net/sonic-buildimage/issues/25801) asks for CoPP test coverage on `t1-lag-vpp`. SAI config-plane already worked, but `saivpp` had no dataplane mechanism to classify/rate-limit/punt control-plane traffic. Documents the design: CoPP nodes in the shared `sonic_ext` VPP plugin — `copp_ifout_node` (ethertype-keyed traps), `copp_ip2me_node` (IP2ME/SNMP/SSH/BGP/BGPV6, one slot per trap group), `copp_udld_node` (real LLC/SNAP dispatch for UDLD).

## Companion PRs

- sonic-platform-vpp: the VPP plugin nodes — https://github.com/sonic-net/sonic-platform-vpp/pull/281
- sonic-sairedis: SAI `POLICER`/`HOSTIF_TRAP`/`HOSTIF_TRAP_GROUP` wiring — https://github.com/sonic-net/sonic-sairedis/pull/2070
- sonic-mgmt: lift CoPP skip for now-passing subtests — https://github.com/sonic-net/sonic-mgmt/pull/27726

## Test result

Manually validated on `vlab-vpp-01`: all 24 applicable `tests/copp/test_copp.py` subtests pass. Details in the HLD's Status section.


doc/vpp/vpp_copp_HLD.md
```diff
@@ -0,0 +1,97 @@
...
+**What we built:** a self-contained VPP plugin, `copp_punt_policer` registering one feature node on `device-input`. It performs the following tasks:
+
+1. **Classify**: parses the raw 14-byte Ethernet header and looks up a small, bounded ethertype→policer-name table. TTL_ERROR is matched by ethertype `0x0800` (IPv4) **plus** an additional condition requiring the IP header's TTL ≤ 1.
+2. **Police**: applies VPP's existing `vnet_police_packet()` token-bucket primitive against the same VPP policer object `SwitchVppPolicer.cpp` already creates from SAI `POLICER` attributes — no new metering implementation. Uses a fixed 256-byte reference packet length since VPP's pps→token-bucket conversion assumes that fixed size internally.
```
lolyu, line 57: This is the load-bearing metering claim and it's underspecified. CoPP CIR/CBS in SONiC are in **pps**, but polishing on a *fixed* 256-byte reference length means byte-rate enforcement diverges for any packet that isn't ~256B — and the very next section runs `test_policer_mtu[BGP]` at **64/1514/4096B**. A 4096B control packet consuming one "256B" token is ~16x under-accounting. The doc says the MTU test passes (reassuring empirically), but the *design rationale* doesn't explain why fixed-256 is correct across 64–4096B rather than just "it's what VPP assumes." Please either show the token math, or state explicitly that CoPP here is pps-based so packet size is deliberately ignored (defensible for pps policing — but then say so, and reconcile it with the byte-oriented `SAI_POLICER_STAT_*_BYTES` counters in REQ-5).

doc/vpp/vpp_copp_HLD.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Design
+### `copp_punt_policer` VPP plugin
+
+**Why a new plugin, not VPP's existing classify/policer features:** VPP already ships a classify-based policer feature (`policer-classify`), but it only runs on three feature arcs — `l2-input` (bridged L2 traffic), `ip4-unicast`, `ip6-unicast`. This project's ports are `linux-cp`-paired, L3-routed ports: ARP/LACP/LLDP/UDLD traffic on them never traverses any of those three arcs — `ethernet-input` dispatches it directly to protocol-specific nodes (`arp-input`, `linux-cp-punt-xc`) that punt straight to each port's TAP.
+**What we built:** a self-contained VPP plugin, `copp_punt_policer` registering one feature node on `device-input`. It performs the following tasks:
+
+1. **Classify**: parses the raw 14-byte Ethernet header and looks up a small, bounded ethertype→policer-name table. TTL_ERROR is matched by ethertype `0x0800` (IPv4) **plus** an additional condition requiring the IP header's TTL ≤ 1.
```
lolyu, line 56: The classify parses "the raw 14-byte Ethernet header," but on linux-cp-paired L3 ports there can be **802.1Q-tagged** frames (the sub-interface / VLAN cases), where the ethertype sits after a 4-byte 802.1Q tag (0x8100 + inner type) and the IP header — hence the TTL≤1 test for TTL_ERROR — is shifted by 4 bytes. A fixed-14-byte parse would misclassify a tagged control frame (wrong ethertype offset, wrong TTL offset). Please state whether tagged frames are handled or explicitly out of scope.

doc/vpp/vpp_copp_HLD.md
```diff
@@ -0,0 +1,97 @@
...
+2. **Police**: applies VPP's existing `vnet_police_packet()` token-bucket primitive against the same VPP policer object `SwitchVppPolicer.cpp` already creates from SAI `POLICER` attributes — no new metering implementation. Uses a fixed 256-byte reference packet length since VPP's pps→token-bucket conversion assumes that fixed size internally.
+3. **Deliver**: for a conforming packet, sets the buffer's TX interface directly to the mapped linux-cp TAP and dispatches straight to `interface-output`. Resolves against the raw ingress port for every ethertype uniformly, matching what this project's PTF test harness (`ptf_nn_agent`) actually observes.
```
lolyu, line 58: Delivering by setting TX interface to the TAP and dispatching straight to `interface-output` **bypasses** the normal `arp-input`/`linux-cp-punt-xc` nodes for the trapped ethertypes. Two things to spell out: (1) does the `copp_punt_policer` node **consume** these packets (so the original path no longer sees them) or can both fire (double-punt to the TAP)? (2) Short-circuiting `arp-input` — does VPP's own `arp-input` perform any side effect (ARP learning / neighbor update) that is now lost for punted ARP? Rerouting a control-plane path means the *removed* behavior matters as much as the added policing.

doc/vpp/vpp_copp_HLD.md
```diff
@@ -0,0 +1,97 @@
...
+
+SAI wiring (`SwitchVppHostifTrap.cpp`) mirrors the existing per-trap dispatch pattern used for bookkeeping: on `createHostifTrap`/`setHostifTrap`/`setHostifTrapGroup`, resolve the trap's bound policer to its VPP policer name and call the plugin's `copp_punt_policer_bind` API — switch-wide, since the plugin auto-enables its feature on every interface as it's created (`VNET_SW_INTERFACE_ADD_DEL_FUNCTION`), no per-port bind needed.
```
lolyu, line 60: Auto-enabling the feature on **every** interface + switch-wide bind means every ingress packet on every port now hits this `device-input` classify node (14-byte Ethernet parse, plus TTL read for 0x0800) — including the overwhelming majority that is *not* control-plane traffic. REQ-7 says the feature "must not regress ACL/FDB/routing dataplane behavior," but a device-input node in front of all traffic is exactly a throughput-regression risk, and it's unquantified. Please add a line on the per-packet fast-path cost for the common (non-punted) case and confirm the ethertype→policer lookup is a bounded/hashed lookup rather than linear.

lolyu (commented):
Strong HLD overall — above the median for this repo. It does the three things design docs usually skip: (1) states *why* VPP's stock `policer-classify` can't cover this traffic with precise arc-level detail (ARP/LACP/LLDP/UDLD on linux-cp-paired L3 ports go `ethernet-input → arp-input/linux-cp-punt-xc → TAP` and never touch the `l2-input`/`ip4-unicast`/`ip6-unicast` arcs where policer-classify lives — correct, and it's the real justification for a new plugin); (2) documents alternates that were *built and disproven* (the `tc` ingress-qdisc-ordering fact — AF_PACKET copy happens before the ingress qdisc — is a real, subtle reason that approach fails); (3) carries a concrete pass/fail validation table. REQ-1..REQ-8 are well-formed and testable, and REQ-8 is unusually honest about testbed/harness blockers. Nice work.

Four technical questions I'd want answered before design-approval (inline), plus doc hygiene below. Flagging as Comment, not blocking — these are "clarify the design," not "the design is wrong."

**Doc hygiene:**
- Revisions table author is "Vesper" while the work is under nhegde-microsoft — codename/placeholder? Merged HLDs normally carry the real handle.
- Background says the ask is for `t1-lag` topology, but the testbed is `t1-lag-vpp` / `vms-kvm-vpp-t1-lag` — drop of the `-vpp` suffix.
- REQ-2 lists SNMP/SSH in its trap enumeration, but neither the in-scope table nor the Status table covers them — are they in scope for this effort or aspirational? The "etc." is doing a lot of work in a requirement.
- REQ-4 calls out runtime add/remove (`test_add_new_trap`/`test_remove_trap`) but only `test_trap_config_save_after_reboot` (REQ-6) appears in the Status table — is the runtime add/remove path passing too?
- Consider a one-line VPP node-graph fragment (`device-input → copp_punt_policer → {police → interface-output(TAP) | drop}` vs the bypassed `ethernet-input → arp-input/linux-cp-punt-xc`) — it would make the whole design legible at a glance, and HLDs here usually carry one.

doc/vpp/vpp_copp_HLD.md
nhegde-microsoft: Correct flag — this is pps-based policing (SONiC CoPP CIR/CBS on this platform are configured in pps, not bytes/sec), so the fixed 256-byte reference length is deliberately packet-size-independent — it's purely VPP's internal pps-to-token-bucket calibration constant, not a byte-accounting decision, and doesn't scale with real frame size. That's why `test_policer_mtu[BGP]` passes identically at 64/1514/4096B. Added an explicit sentence stating this and reconciling it with REQ-5's byte-oriented `SAI_POLICER_STAT_*_BYTES` counters: those still report real packet lengths for statistics; only the policing *verdict* uses the fixed 256. Pushed in eeffa10.

doc/vpp/vpp_copp_HLD.md
nhegde-microsoft: Out of scope, now stated explicitly — this project's ports are untagged L3-routed access ports; no VLAN sub-interface case exists on this testbed. Added a line: tagged/sub-interface frames are out of scope for this design, since the fixed 14-byte parse assumes untagged Ethernet (a tagged frame's ethertype, and for TTL_ERROR the IPv4 TTL, would be read from the wrong offset and misclassified). Pushed in eeffa10.

doc/vpp/vpp_copp_HLD.md
nhegde-microsoft: Yes to both, now documented explicitly. The node consumes the packet — a single `vlib_buffer_enqueue_to_next` per packet redirects a matched, conforming packet straight to `interface-output`/TAP, so `arp-input` never sees it (no double-punt). That does mean VPP's own arp-input ARP-learning side effect is lost for punted ARP; the equivalent function is expected to happen at the CPU/Linux side (kernel ARP handling on the TAP), matching the existing `linux-cp` model. Added this as an explicit design note. Pushed in eeffa10.

doc/vpp/vpp_copp_HLD.md
nhegde-microsoft: Fair, now quantified. Added a line: the per-packet cost for the non-punted majority is a 14-byte Ethernet header read plus a linear scan of a small, bounded table (`COPP_PUNT_POLICER_MAX_ENTRIES` = 16 today); no policer/counter work happens unless an entry matches. Confirmed the lookup is linear, not hashed — acceptable given the table's small, static size, but called out as a known future optimization opportunity if entry count grows materially. Pushed in eeffa10.

doc/vpp/vpp_copp_HLD.md
```diff
@@ -0,0 +1,148 @@
...
+| Rev | Date | Author(s) | Changes |
+|-----|------|-----------|---------|
+| 1.0 | 2026-09-02 | nhegde-microsoft | Initial HLD for  `copp_punt_policer` design. |
+| 1.1 | 2026-09-10 | nhegde-microsoft | Added `copp_ip2me_policer` plugin (IP2ME/SNMP/SSH enforcement on `ip4-punt`); updated status table. |
+
+---
+
+## Background
+
+[sonic-buildimage#25801](https://github.com/sonic-net/sonic-buildimage/issues/25801) asks to enable Control Plane Policing (CoPP) testing for the `t1-lag-vpp` topology (`vms-kvm-vpp-t1-lag` testbed) on the SONiC-VPP KVM testbed.
```
yue-fred-gao, line 14: I think the design should not limit the scope to t1-lag topology. We should at least include t0 to make sure it is compatible. we don't want to redesign it to work for both t1 and t0

doc/vpp/vpp_copp_HLD.md
```diff
@@ -0,0 +1,148 @@
...
+2. **Police**: applies VPP's existing `vnet_police_packet()` token-bucket primitive against the same VPP policer object `SwitchVppPolicer.cpp` already creates from SAI `POLICER` attributes — no new metering implementation. Uses a fixed 256-byte reference packet length since VPP's pps→token-bucket conversion assumes that fixed size internally.
+3. **Deliver**: for a conforming packet, sets the buffer's TX interface directly to the mapped linux-cp TAP and dispatches straight to `interface-output`. Resolves against the raw ingress port for every ethertype uniformly, matching what this project's PTF test harness (`ptf_nn_agent`) actually observes.
+
+SAI wiring (`SwitchVppHostifTrap.cpp`) mirrors the existing per-trap dispatch pattern used for bookkeeping: on `createHostifTrap`/`setHostifTrap`/`setHostifTrapGroup`, resolve the trap's bound policer to its VPP policer name and call the plugin's `copp_punt_policer_bind` API — switch-wide, since the plugin auto-enables its feature on every interface as it's created (`VNET_SW_INTERFACE_ADD_DEL_FUNCTION`), no per-port bind needed.
```
yue-fred-gao, line 60: I have concern with adding this node to device input, which will affect every packets. Most of the protocols listed (except UDLD and TTL expiry requiring investigation) are already punted to control plane today. Can we add the policer to the punt path? For example, in the interface output arc of tap interface?

doc/vpp/vpp_copp_HLD.md
nhegde-microsoft: Fair point, I did not make it generic enough to cover t0. I believe the primary difference is the VLAN/bridge setup in t0 versus the L3 routed ports in t1-lag. t0 still uses the same linux-cp tap pairing, so I _think_ this should work just fine, but I'd like to run a quick test first. I will update here and in the document when done. Thanks for the suggestion.

doc/vpp/vpp_copp_HLD.md
nhegde-microsoft: I did have the same concern but it seemed like the best option while I was trying this out. I also ran some tests and found the added delay was ~50ns per packet and the protocol match list was small (<16). The advantage was that this would work for any topology.
Having said that, your other point about policing it in the output path is interesting because that is the first option I tried and ran into issues. However, after several other changes, I suspect the issue was not related to the location of the policer because I just ran it again now with the policer in tap interface output and it seems to work, which I am surprised to see myself. I am going to do a couple of more runs to be sure, but if it is stable, I will switch the design to use this instead. Should be a small change.

doc/vpp/vpp_copp_HLD.md
```diff
@@ -0,0 +1,148 @@
+# SONiC-VPP CoPP Dataplane Enablement — HLD
+
+## Revisions
+
+| Rev | Date | Author(s) | Changes |
+|-----|------|-----------|---------|
+| 1.0 | 2026-09-02 | nhegde-microsoft | Initial HLD for  `copp_punt_policer` design. |
+| 1.1 | 2026-09-10 | nhegde-microsoft | Added `copp_ip2me_policer` plugin (IP2ME/SNMP/SSH enforcement on `ip4-punt`); updated status table. |
```
yue-fred-gao, line 8: so far sonic-ext plugin is used for multiple features for sonic. Please consider if it is better to use the same plugin. The benefit is we can avoid too many plugins. But we lose the flexibility to enable/disable a feature through plugin infra. Currently, we use API in the sonic-ext plugin to enable/disable individual feature.

nhegde-microsoft: Fair point. Let me look into feasibility of moving this into sonic-ext.

doc/vpp/vpp_copp_HLD.md
nhegde-microsoft: Checked that, in general, the design applies to t0 as well. Did some basic tests and saw a few failures with DHCP which is likely related to how it is handled in VLAN tagged versus LAG member interfaces. I propose fixing those in a follow-on ticket.

doc/vpp/vpp_copp_HLD.md
nhegde-microsoft: Point accepted and addressed in latest commit

doc/vpp/vpp_copp_HLD.md
nhegde-microsoft: Point accepted and addressed in latest commit

lolyu (approved):
The redesign resolves the core concern from the earlier review round, and cleanly. Approving.

**The central fix — device-input → output-arc/punt-path policing.** My original concern (and yue-fred-gao's follow-up) was that the `device-input` classify node ran on every packet on every interface, taxing the ~100% of ordinary forwarded traffic that never matches — a real REQ-7 throughput-regression risk. Rev 1.1 reworks this into three feature nodes that only ever see already-CPU-bound traffic: `sonic-ext-copp-ip2me` on the global `ip4-punt` arc (IP2ME/SNMP/SSH by dst-IP, BGP by TCP dport — exactly the punt-path placement suggested in review), `sonic-ext-copp-ifout` on `interface-output` of each linux-cp TAP (ethertype classify on an already-rewound intact Ethernet frame), and `sonic-ext-copp-udld` as the LLC/SNAP handler redirecting into the ifout node. This eliminates the fast-path tax structurally rather than merely quantifying it — the right call.

**All four of my original technical questions are now resolved or moot:**
1. Fixed-256B metering — addressed earlier (pps-based, reconciled with REQ-5's byte counters); policing is now consolidated through one `vnet_police_packet()` node.
2. VLAN/tagged parse — no longer a raw device-input parse; ethertype classify runs at `interface-output` on an intact frame.
3. Consume-vs-double-punt / lost `arp-input` side effects — moot: the design now polices *on* the existing linux-cp punt path instead of short-circuiting `arp-input`, so no side effects are bypassed.
4. Switch-wide fast-path cost — fully resolved; the premise (a node in front of all traffic) is gone.

**The alternatives-rejected section is genuinely excellent** — three approaches built and empirically disproven with controlled experiments: the device-input plugin (fast-path tax), Linux `tc` on the TAP (disproven via a veth-pair experiment showing `AF_PACKET`/`ptype_all` fires before the ingress qdisc, so a `tc` drop-all still let a raw-socket receiver see all frames), and VPP-native `policer_classify` on `l2-input` (disproven via `show classify tables verbose` showing `hits 0` because the L3-routed linux-cp ports never traverse `l2-input`). That's model HLD engineering. All 24 copp subtests pass, including the previously-flaky `test_remove_trap`, root-caused to a harness measurement bug (BGPTest's `RX PPS` counting real background BGP on tcp/179) rather than a product bug.

Non-blocking hygiene items (worth a final pass, none gating):
- Revisions author is still "Vesper" while the work is under nhegde-microsoft — merged HLDs normally carry the real handle.
- The doc now states the design is topology-agnostic (the output-arc/punt-path model genuinely is more topology-neutral than the old device-input parse) — a one-line explicit confirmation that t0 was actually *run*, not just expected to work, would close yue-fred-gao's t0 point cleanly.
- A one-line node-graph fragment would help the now-three-node design read at a glance (`linux-cp-punt → interface-output(TAP) → sonic-ext-copp-ifout → {police|drop}`; `ip4-punt → sonic-ext-copp-ip2me`; `llc-input → sonic-ext-copp-udld → sonic-ext-copp-ifout`).

Net: sound design, core concern resolved at the root, thoroughly validated. Approving on the design; deferring to yue-fred-gao to confirm the punt-path rework satisfies their thread since it directly answers their suggestion.

nhegde-microsoft:
> Non-blocking hygiene items (worth a final pass, none gating):
> 
> * Revisions author is still "Vesper" while the work is under nhegde-microsoft — merged HLDs normally carry the real handle.
> * The doc now states the design is topology-agnostic (the output-arc/punt-path model genuinely is more topology-neutral than the old device-input parse) — a one-line explicit confirmation that t0 was actually _run_, not just expected to work, would close yue-fred-gao's t0 point cleanly.
> * A one-line node-graph fragment would help the now-three-node design read at a glance (`linux-cp-punt → interface-output(TAP) → sonic-ext-copp-ifout → {police|drop}`; `ip4-punt → sonic-ext-copp-ip2me`; `llc-input → sonic-ext-copp-udld → sonic-ext-copp-ifout`).
> 
> Net: sound design, core concern resolved at the root, thoroughly validated. Approving on the design; deferring to yue-fred-gao to confirm the punt-path rework satisfies their thread since it directly answers their suggestion.

Thanks for the review and approval @lolyu 🙏 
I went ahead and addressed all your hygiene comments as well in latest commit. Please re-approve when you get a chance.

yue-fred-gao (approved):
lgtm

nhegde-microsoft:
Issue to test and handle T0 CoPP failing tests [#289](https://github.com/sonic-net/sonic-platform-vpp/issues/289)

PR #2543, opened 2026-09-09
[HLD] EVPN Layer-3 Multihoming (L3MH) neighbor synchronization

pbrisset:
Code PR: https://github.com/FRRouting/frr/pull/23304

### What I did

Add a High-Level Design for **EVPN Layer-3 Multihoming (L3MH) neighbor
synchronization** — keeping a multihomed host's ARP/ND (neighbor) entry in sync
across the leaves that share it, using EVPN Route Type 2 (RT-2), **even when the
host's VLAN has no L2VNI** (a pure routed fabric).

New file: `doc/vxlan/EVPN/EVPN-L3MH-Neighbor-Sync.md`.

### Why I did it

In a routed-only fabric there is no stretched Layer 2 and therefore no L2VNI, so
the Layer-2 machinery that normally carries a host's neighbor entry between the
leaves does not exist. Because of LAG hashing, a host's ARP reaches only one
member leaf; the other members have no neighbor entry and cannot reliably route
the last hop to the host (self-resolution is unreliable because the host's reply
can be hashed to a different member, and there is no flood domain to fall back
on). This feature gives the neighbor entry its own control-plane ride.

### How it works (summary)

- A leaf originates the host's neighbor as an RT-2 marked **pure Layer-3** via an
  MPLS Explicit NULL (`label[0]=0`), with `label[1]=L3VNI`, Ethernet Tag = VLAN,
  and the IP-VRF RT.
- A receiving leaf whose **ESI matches a local bond** installs only the **kernel
  neighbor** plus a **local-ES sync-MAC** pin — no bridge/FDB flood, no VXLAN
  state.
- Enabled per box with a single EVPN knob (`advertise-l3vni-neigh`) that gates
  both transmit and receive.
- Implemented in FRR (bgpd/zebra); SONiC consumes the resulting kernel
  neighbor/FDB through the existing pipeline. **No new SAI object or
  ASIC-programming change.**

### Scope

Neighbor synchronization only. Host `/32`/`/128` RIB import, backup NHGs,
remote-VTEP multipath, and the ES-Import RT optimization are called out as
follow-up work.

### References

- IETF: [draft-ietf-bess-evpn-l3mh-proto](https://datatracker.ietf.org/doc/draft-ietf-bess-evpn-l3mh-proto/) — ARP/ND synchronization via RT-2
- [RFC 7432](https://datatracker.ietf.org/doc/html/rfc7432) (base EVPN),
  [RFC 9135](https://datatracker.ietf.org/doc/html/rfc9135) (EVPN IRB)

### DIFF
--- a/doc/vxlan/EVPN/EVPN-L3MH-Neighbor-Sync.md
+++ b/doc/vxlan/EVPN/EVPN-L3MH-Neighbor-Sync.md
@@ -1,0 +1,454 @@
+# EVPN Layer-3 Multihoming Neighbor Synchronization — SONiC High-Level Design
+
+## Table of Contents
+
+- [1. Revision](#1-revision)
+- [2. Scope](#2-scope)
+- [3. Definitions and Abbreviations](#3-definitions-and-abbreviations)
+- [4. References](#4-references)
+- [5. Overview](#5-overview)
+- [6. Problem Statement](#6-problem-statement)
+- [7. Solution](#7-solution)
+  - [7.1 The idea](#71-the-idea)
+  - [7.2 The wire signal](#72-the-wire-signal)
+  - [7.3 What a receiver installs](#73-what-a-receiver-installs)
+  - [7.4 The enable knob](#74-the-enable-knob)
+- [8. Architecture](#8-architecture)
+  - [8.1 Reference topology](#81-reference-topology)
+  - [8.2 Data flow](#82-data-flow)
+  - [8.3 Ownership model](#83-ownership-model)
+  - [8.4 FRR data-model extensions](#84-frr-data-model-extensions)
+  - [8.5 ZAPI messages (bgpd ↔ zebra)](#85-zapi-messages-bgpd--zebra)
+- [9. Requirements](#9-requirements)
+- [10. Configuration and Management](#10-configuration-and-management)
+- [11. SAI / Data-Plane Impact](#11-sai--data-plane-impact)
+- [12. Warm Boot / Fast Boot](#12-warm-boot--fast-boot)
+- [13. Restrictions and Limitations](#13-restrictions-and-limitations)
+- [14. Testing](#14-testing)
+
+---
+
+## 1. Revision
+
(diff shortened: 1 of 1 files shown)


doc/vxlan/EVPN/EVPN-L3MH-Neighbor-Sync.md
```diff
@@ -0,0 +1,454 @@
...
+   straight out the correct bond.
+
+Everything Layer-2 is **suppressed**: no VXLAN/remote-MAC state, no Inclusive
+Multicast (RT-3), no flood list. If the receiver's ESI does **not** match a local
+bond, it installs **nothing** — the route is simply not relevant to it.
+
+### 7.4 The enable knob
+
+The feature is turned on per box with a single EVPN sub-command,
+**`advertise-l3vni-neigh`**, alongside the existing `advertise-all-vni`. It is a
```
mjstapp, line 191: is this a new, stand-alone cli? I think it might be clearer to add an optional token to the existing cli, and add a flag to the existing zapi message. that would avoid any ordering issues between the handling of the config commands, or between two zapi messages.

doc/vxlan/EVPN/EVPN-L3MH-Neighbor-Sync.md
```diff
@@ -0,0 +1,454 @@
...
+  still serves many SVIs.
+
+### 8.5 ZAPI messages (bgpd ↔ zebra)
+
+ZAPI is the internal channel between bgpd and zebra. One new message is added and
+the existing MAC/IP messages are extended with a flag and an Ethernet Tag.
+
+| Message | Direction | Change |
+|---|---|---|
+| **`ZEBRA_ADVERTISE_L3VNI_NEIGH`** | bgpd → zebra | **New** — carries the enable knob (1-byte flag), mirroring `ZEBRA_ADVERTISE_ALL_VNI`. |
```
mjstapp, line 334: see above - maybe just enhance the existing zapi message instead?

PR #2545, opened 2026-09-15
[doc] MACsec: add MKA operational state and key rotation HLD

liamkearney-msft:
## Summary

Define the final SONiC-side design for publishing MKA operational state, validating safe primary/fallback CAK rotation, and displaying that state through `show macsec --mka`.

The design:

- assigns sole ownership of namespace-local `MACSEC_MKA_SESSION_TABLE` and `MACSEC_MKA_PARTICIPANT_TABLE` publication to macsecmgrd;
- performs a sequential full sweep of configured MACsec ports every 20 seconds, with a hard two-second deadline per `macsec_mka_list` query and per-port failure isolation;
- stores independent `last_updated`, `query_status`, `config_status`, and optional redacted `config_error` values, while deriving freshness only at consumption time;
- maps the final WPA participant contract directly, using `active`, `retain`, `is_principal`, `is_primary`, peer counts, and key-server/election fields without synthesizing the removed `participant` management field;
- defines healthy protected Controlled Port operation as `kay_status=active`, `authenticated=false`, `secured=true`, and `failed=false`—the WPA `Authenticated` field indicates authenticated-only, unprotected mode rather than successful MKA authentication;
- requires the protected state above, query health, config consistency, age no greater than 60 seconds, correct participant roles, active state, and a live alternate CA before rollover;
- requires macsecmgrd to revalidate immediately before remove-old/add-new actioning;
- defines compact `show macsec --mka` output with natural interface sorting, a separate `Secured` column, and `Status` derived only from query/config/age health;
- derives one detailed `Controlled port mode` value from the raw KaY fields: `secured`, `authenticated-only`, `failed`, `inactive`, `inconsistent`, or `unknown`; and
- keeps existing APP_DB/ASIC programming tables, MACsecOrch, SAI, and WPA supplicant behavior unchanged.

## Companion implementation

- sonic-net/sonic-swss-common#1251
- sonic-net/sonic-swss#4827
- sonic-net/sonic-buildimage#29102
- sonic-net/sonic-wpa-supplicant#138

## Scope

Documentation only. The HLD covers SONiC-side schema ownership, status collection, safe rotation preflight/actioning, failure reconciliation, namespace behavior, CLI presentation, and validation. WPA fallback behavior and control/status interfaces are consumed as the existing companion dependency.

## Validation

- Markdown table of contents, code fences, and sequential test-plan numbering validated.
- Scheduler, deadline, freshness, failure-isolation, Controlled Port semantics, and Status/Age wording checked consistently across requirements, schemas, algorithms, examples, repository impact, and test plan.
- The removed WPA `participant` field is absent from the proposed STATE_DB schema, examples, and show/test expectations; `retain` remains unchanged.
- Safe rotation requires `authenticated=false,secured=true`; authenticated-only or contradictory Controlled Port combinations are explicitly rejected.
- Compact `Status` is limited to query health, configuration consistency, and age; `Secured` remains the separate compact protected/unprotected indication.
- Detailed output derives `Controlled port mode` through an ordered mapping while preserving the raw STATE_DB fields and rotation-safety tuple.
- No stale five-second, 15-second, concurrent scheduler, or stored-freshness wording remains.


rlhui:
can we add the PRs needed for this feature? thanks

doc/macsec/MACsec_MKA_operational_state_hld.md
```diff
@@ -0,0 +1,1135 @@
...
+runtime set, it is not a complete trustworthy observation: macsecmgrd follows
+the query-error behavior in §4.5 and does not use that response as safety
+evidence. A response that matches applied state but differs from desired
+CONFIG_DB is publishable with `query_status=ok,config_status=degraded`; this is
+how add/remove failures remain observable.
+
+### 4.4 Refresh and freshness
+
+Periodic collection is a full sweep of every configured MACsec port in the
+local namespace every 20 seconds. `macsecmgrd` executes the sweep sequentially
```
judyjoseph, line 576: Are we planning to use wpa_cli in macsec docker per namespace. To prevent overhead we should club multiple interface in that asic in a single wpa_cli call ?

liamkearney-msft: i dont think we can - each interface has its own wpa supplicant instance

PR #2547, opened 2026-09-16
Neighbor Next-hop optimization HLD

manamand2020:
Optimization of neighbor next-hops by delaying the creation until referenced.


tahmed-dev:
§6.3 enumerates every internal `NeighOrch` call site to establish the pure vs side-effecting split. The consumer side does not get the same treatment: §5.1 and §6.2 describe the `hasNextHop()` to `ensureNextHop()` change as a mechanical, compiler-checked rename with no call-site logic changes.

That does not hold for all current callers. Of the 20 `hasNextHop()` call sites outside `neighorch.cpp`, some are pure predicates whose decision inverts once the call can allocate, and some sit on teardown paths where creating a hardware object is exactly what must not happen:

1. `NextHopGroupMember::~NextHopGroupMember()` (`nhgorch.cpp`). Both branches read `hasNextHop(m_key) && getNextHopRefCount(m_key) == 0`. If that call creates the next hop, the destructor allocates a SAI object and then calls `removeMplsNextHop()` on it. This is the mid-teardown hazard §6.3 guards against for `NeighOrch`'s internal checks.

2. `RouteOrch::removeNextHopGroup()` (`routeorch.cpp`), the `isPrefixNeighborNh(nhop->first) && !hasNextHop(nhop->first)` check guarding "Skip NHG member remove: nexthop missing". If the renamed call creates the next hop, the condition flips and the other branch is taken. That is a behavior change, not a rename.

The remaining call sites outside `neighorch.cpp` may be worth studying for their patterns of use, since more than one pattern appears to be in play. A compiler-checked rename proves every caller was updated, not that allocation is correct at each one.

doc/ip/delayed-neighbor-nexthop-creation-hld.md
```diff
@@ -0,0 +1,388 @@
...
+
+<a id="73-crm-and-reference-counting"></a>
+### 7.3 CRM and Reference Counting
+- CRM next-hop counters (`CRM_IPV4_NEXTHOP` / `CRM_IPV6_NEXTHOP` / `CRM_MPLS_NEXTHOP`) are incremented/decremented exclusively inside `addNextHop()` / `removeNextHop()` / `removeMplsNextHop()` / `removeOverlayNextHop()`, unchanged. Only the *timing* of creation moves; the 1:1 accounting invariant is untouched, and counts now more accurately reflect real hardware usage than before. The debounced reclaim in [Section 7.6.3](#763-bounding-crm-staleness-debounced-reclaim-of-idle-next-hops) is a new *caller* of `removeNextHop()`, not a new decrement path, so this invariant still holds.
```
zjswhhh: `removeNextHop()` only removes software bookkeeping; SAI removal and CRM accounting happen in `removeNeighbor()`, so calling it alone would leave the hardware object allocated. Please define a shared next-hop-only teardown path that frees the hardware and updates accounting without deleting the resolved neighbor.

doc/ip/delayed-neighbor-nexthop-creation-hld.md
```diff
@@ -0,0 +1,388 @@
...
+
+1. **Add side.** `NeighOrch::addNeighbor()` explicitly calls `gFgNhgOrch->validNextHopInNextHopGroup(nhKey)` for every plain neighbor, restoring the "neighbor resolved" notification. `FgNhgOrch::validNextHopInNextHopGroup()` in turn calls `m_neighOrch->ensureNextHop(nexthop)` itself before activating a member, since being configured as an FG ECMP member is itself a legitimate reference that should trigger lazy creation (this call safely short-circuits reentrantly through `addNextHop()`'s own `validNextHopInNextHopGroup()` notification for the same key). As a direct consequence, any next hop that FG ECMP actually activates is guaranteed to exist as a SAI object from that point on, since activation is what creates it.
```
zjswhhh: FG ECMP can skip a member when `ensureNextHop()` fails and complete the route successfully, leaving no retry pending when hardware capacity becomes available. Please define how resolved-but-resource-blocked members are tracked and retried until activation succeeds.

doc/ip/delayed-neighbor-nexthop-creation-hld.md
```diff
@@ -0,0 +1,388 @@
...
+
+To bound that staleness without giving up the churn-free property above, reclaiming a next hop on reference count reaching zero is **debounced** rather than immediate or never: when a next hop's reference count drops to zero, its destruction is scheduled after a short, fixed grace period instead of happening right away. If any consumer references it again before the grace period elapses -- the ordinary churn case -- the pending destruction is simply cancelled and no SAI work happens at all, identical to today. Only a next hop that stays unreferenced for longer than the grace period is actually torn down, once, through the same teardown path `removeNeighbor()` already uses, so CRM decrement and `FgNhgOrch` invalidation ([Section 7.1](#71-fgnhgorch-fine-grained-ecmp)) stay consistent automatically, with no separate teardown logic to maintain.
```
zjswhhh, line 307: A next hop can be created with zero references and remain unused if route installation fails, so its reference count never *drops* to zero to trigger reclamation. Please cover this case by scheduling reclamation at creation and cancelling it when a reference is acquired.

doc/ip/delayed-neighbor-nexthop-creation-hld.md
```diff
@@ -0,0 +1,388 @@
...
+
+An entry in `m_bfdDownPeers` is removed in one of two ways, both required so this set cannot grow stale or unbounded: (1) `updateNextHop()` already erases it the moment the peer's BFD session reports `UP` again, unchanged from before this design; (2) `removeNeighbor()` additionally erases it once no resolved neighbor references that peer IP any more. (2) is necessary because a peer's BFD session can be torn down (e.g. the neighbor itself is removed) while its last-known state is still `DOWN` -- that session's teardown is not preceded by an `UP` transition, so (1) alone would never fire and the entry would linger indefinitely. Since `m_bfdDownPeers` is keyed purely by peer IP (matching `updateNextHop()`'s own matching semantics, which ignores alias/VRF), (2) only erases an entry once *every* neighbor sharing that IP is gone, not just the one being removed.
```
zjswhhh, line 228: BFD sessions can outlive their neighbors, so deleting cached DOWN state on neighbor removal can incorrectly enable forwarding after ARP relearning. Please tie cache cleanup to BFD session lifecycle events rather than neighbor removal, including session deletion while the neighbor remains resolved.

manamand2020:
> §6.3 enumerates every internal `NeighOrch` call site to establish the pure vs side-effecting split. The consumer side does not get the same treatment: §5.1 and §6.2 describe the `hasNextHop()` to `ensureNextHop()` change as a mechanical, compiler-checked rename with no call-site logic changes.
> 
> That does not hold for all current callers. Of the 20 `hasNextHop()` call sites outside `neighorch.cpp`, some are pure predicates whose decision inverts once the call can allocate, and some sit on teardown paths where creating a hardware object is exactly what must not happen:
> 
> 1. `NextHopGroupMember::~NextHopGroupMember()` (`nhgorch.cpp`). Both branches read `hasNextHop(m_key) && getNextHopRefCount(m_key) == 0`. If that call creates the next hop, the destructor allocates a SAI object and then calls `removeMplsNextHop()` on it. This is the mid-teardown hazard §6.3 guards against for `NeighOrch`'s internal checks.
> 2. `RouteOrch::removeNextHopGroup()` (`routeorch.cpp`), the `isPrefixNeighborNh(nhop->first) && !hasNextHop(nhop->first)` check guarding "Skip NHG member remove: nexthop missing". If the renamed call creates the next hop, the condition flips and the other branch is taken. That is a behavior change, not a rename.
> 
> The remaining call sites outside `neighorch.cpp` may be worth studying for their patterns of use, since more than one pattern appears to be in play. A compiler-checked rename proves every caller was updated, not that allocation is correct at each one.

You're right, and both examples hold. The "mechanical, compiler-checked rename with no call-site
logic changes" framing is wrong and is removed from section 5.1, 6.2 and 6.3.

New section 6.5 classifies every hasNextHop() call site outside neighorch.cpp before conversion:
  - check-before-use  -> ensureNextHop()
  - read-only check   -> hasSyncdNextHop()

Both of your examples are called out there: ~NextHopGroupMember() would create a SAI next hop and
immediately delete it, and removeNextHopGroup()'s guard would invert and run the other branch.
Two more in the same family are flagged for scrutiny during conversion: addNextHopGroup()'s MPLS
branch (would force-create the underlying IP next hop) and addRoutePost()'s post-install check.

Mux turned out to need a third case, check section 7.2 and ensureLocalNextHop().

doc/ip/delayed-neighbor-nexthop-creation-hld.md
```diff
@@ -0,0 +1,388 @@
...
+
+<a id="72-muxorch"></a>
+### 7.2 MuxOrch
+Mux neighbors in the default ("host-route") mode are plain (non `prefix_route`) neighbors and are therefore subject to delayed creation. Several mux code paths (`MuxNbrHandler::update()`/`enable()`, `MuxOrch::updateRoute()`) previously fetched a next hop's SAI id via the **pure** `getLocalNextHopId()` lookup, which would silently return `SAI_NULL_OBJECT_ID` for a deferred-but-not-yet-created next hop. Each of these call sites now calls `ensureNextHop()` first to force on-demand creation before the pure lookup. (`MuxPrefixBasedNbrHandler`, used for no-host-route mode, is unaffected -- it is always `prefix_route`, hence always eager.)
```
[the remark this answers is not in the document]
manamand2020: Agreed. `ensureNextHop()` is the wrong here: its existence check also consults
`MuxOrch::getNextHopId()`, so with a tunnel override present it returns success having created
nothing, and the following `getLocalNextHopId()` still returns SAI_NULL_OBJECT_ID.
`MuxNbrHandler::enable()` hits this on every standby->active transition, since entries in
`neighbors_` still hold the tunnel id from the standby path. The MUX_STATE_ACTIVE case in
`update()` is safe only incidentally.

Section 7.2 now adds `ensureLocalNextHop()` -- identical to `ensureNextHop()` but checking
`m_syncdNextHops` alone. The mux sites that follow with `getLocalNextHopId()` use it; every other
consumer keeps `ensureNextHop()`.

doc/ip/delayed-neighbor-nexthop-creation-hld.md
```diff
@@ -0,0 +1,388 @@
...
+- This means the one-time creation cost introduced by this feature is paid **once, at first reference**, not once per churn cycle. Steady-state route flapping is no more expensive after this change than before it; the only difference is *when* that one-time cost is paid (first route reference instead of neighbor resolution).
+- This is a genuine change in behavior only for plain ARP/ND next hops. MPLS-labeled next hops (`RouteOrch::removeMplsNextHop()`, invoked whenever `getNextHopRefCount() == 0`) and overlay/VXLAN tunnel next hops already had a "create-on-reference, destroy-on-zero-refcount" lifecycle *before* this change, so they already pay a create/destroy cost on every churn cycle, unrelated to and unaffected by this feature.
+- As a side effect, CRM `NEXTHOP` used-counters for plain neighbors stay flat across pure route churn instead of ticking down/up on every flap, which is a useful property when using those counters as a capacity signal (see [Section 7.6.1](#761-adjacency-resolution-no-longer-reserves-a-next-hop)) -- churn noise does not show up as CRM churn for the common case.
```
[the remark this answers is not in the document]
manamand2020: Correct, this conflicted with the debounced reclaim in section 7.6.3. The guarantee is now scoped to
re-references that land within the grace period. Past it the reclaim fires: the next hop is
destroyed once, CRM `used` drops, and a later SET pays the creation cost again. The
"CRM counters stay flat" bullet is qualified the same way.

doc/ip/delayed-neighbor-nexthop-creation-hld.md
```diff
@@ -0,0 +1,388 @@
...
+
+This is deliberately narrow in scope: it only bounds staleness under ordinary churn, and does not attempt to reactively reclaim capacity under table-full pressure (a different problem, not addressed here). The grace period is a fixed constant -- long enough to absorb the flap/reconverge timescales discussed in [Section 7.6.2](#762-behavior-under-frr-route-churn-withdraw--re-inject), short enough that CRM `used` staleness is now bounded rather than unbounded. A neighbor removed while one of its next hop's pending destructions hasn't yet fired tears down cleanly through the existing `removeNeighbor()` path; the now-redundant pending destruction is simply dropped.
```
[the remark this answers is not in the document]
manamand2020: Specified in section 7.6.3: 5-second grace period, swept once per second, both fixed compile-time
constants (NEXTHOP_RECLAIM_GRACE_PERIOD_SECS, NEXTHOP_RECLAIM_SWEEP_INTERVAL_SECS), no CONFIG_DB
knob. A SelectableTimer owned by NeighOrch drives the sweep. Pending reclaims sit in a queue
ordered by due time plus a set of keys still due; cancelling removes the key from the set and
leaves the queue entry to be ignored when the sweep reaches it.

Warm restart was a genuine gap. Now specified: the queue does not survive a warm boot, and since
a reclaim is scheduled when a refcount *reaches* zero, a post-reconciliation sweep schedules one
for every plain next hop already sitting at zero.

Ndancejic (commented):
Overall I think it's a good idea with practical benefits. I think the HLD could be a bit more concise and focused, it's very easy to get lost in the intricacies.

My main thoughts after reading:
1. It sounds like CRM counts will follow real nexthop creation (will not increment during initial neighbor creation) If that's the case, what's preventing us from creating more neighbors than hardware nexthop limitations will allow (assuming that's something that's possible) and then breaking when all of these nexthops are referenced and created all at once? I'm assuming neighbor/nexthop limitations are the same for most but we should definitely confirm before making that assumption.
2. I think (depending on time/complexity) implementing the bulk nexthop programming from the start is a good idea if we pursue this. It would take a bit more effort, but could save time in the future and be a big benefit if we just do things correctly from the start

Lets set up a meeting to review and align with if/when we want to do this and also make sure we're on the same page with the HLD

manamand2020:
> Overall I think it's a good idea with practical benefits. I think the HLD could be a bit more concise and focused, it's very easy to get lost in the intricacies.
> 
> My main thoughts after reading:
> 
> 1. It sounds like CRM counts will follow real nexthop creation (will not increment during initial neighbor creation) If that's the case, what's preventing us from creating more neighbors than hardware nexthop limitations will allow (assuming that's something that's possible) and then breaking when all of these nexthops are referenced and created all at once? I'm assuming neighbor/nexthop limitations are the same for most but we should definitely confirm before making that assumption.
> 2. I think (depending on time/complexity) implementing the bulk nexthop programming from the start is a good idea if we pursue this. It would take a bit more effort, but could save time in the future and be a big benefit if we just do things correctly from the start
> 
> Lets set up a meeting to review and align with if/when we want to do this and also make sure we're on the same page with the HLD

Yes you can create neighbors and it will no more be limited by neighbor nexthop creations. That's the idea. If later all these nexthops get referenced and you hit the scale limit of the hardware then you will fail those creations. Neighbor and Nexthop limitations are not same on all hardware, two different SAI object and tables. 

I agree on implementation of bulk nexthop programming from the beginning.
I will try to schedule a meeting on this and discuss further.

